PR #256, opened 2020-08-21
MM-27004: SQLStore versioning and migration

agarciamontoro:
#### Summary
This PR does a couple of things:
- it creates the needed boilerplate for versioning and migration
- it adds the first migration, that creates all the tables
- it adds the data migration from the old kv store data to the brand new database
- it modifies the checklist model to add an ID

This has been manually tested, but not at the extent that it should be. It seems to work, though. I am aiming to get this extensively tested when working on the unit test tickets, so I'm expecting the reviews here to focus on the versioning/migration structure, as well as an early look to the 0.1.0 migration itself.

The data migration is isolated in `data_migration.go`, so we will be able to easily remove when everyone is on 0.1.0 :)

I've set the base branch to the open PR #252. I'll refactor when that one's merged!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27004
https://mattermost.atlassian.net/browse/MM-27140

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/mattermost/mattermost-plugin-incident-response/server/api"
@@ -156,6 +157,7 @@ func (p *Plugin) UpgradeDatabase(sqlStore *sqlstore.SQLStore, pluginAPIClient *p
 	defer mutex.Unlock()
 
 	currentSchemaVersion, err := sqlStore.GetCurrentVersion()
+	fmt.Println("VERSION: ", currentSchemaVersion)
 	if err != nil {
 		return errors.Wrapf(err, "failed to get the current schema version")
 	}
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -143,7 +143,7 @@ func (p *Plugin) OnActivate() error {
 	}
 
 	// Cluster lock: only one plugin will perform the migration when needed
-	if err := p.DBMigration(sqlStore, pluginAPIClient, mutex); err != nil {
+	if err := p.UpgradeDatabase(sqlStore, pluginAPIClient, mutex); err != nil {
 		return errors.Wrapf(err, "failed to run migrations")
 	}
 
@@ -151,7 +151,7 @@ func (p *Plugin) OnActivate() error {
 	return nil
 }
 
-func (p *Plugin) DBMigration(sqlStore *sqlstore.SQLStore, pluginAPIClient *pluginapi.Client, mutex *cluster.Mutex) error {
+func (p *Plugin) UpgradeDatabase(sqlStore *sqlstore.SQLStore, pluginAPIClient *pluginapi.Client, mutex *cluster.Mutex) error {
 	mutex.Lock()
 	defer mutex.Unlock()
 
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -167,7 +167,7 @@ func (p *Plugin) DBMigration(pluginAPIClient *pluginapi.Client, mutex *cluster.M
 	}
 
 	if currentSchemaVersion.LT(sqlstore.LatestVersion()) {
-		if err := sqlstore.Migrate(db, currentSchemaVersion, pluginAPIClient); err != nil {
+		if err := sqlstore.Migrate(db, currentSchemaVersion, pluginAPIClient, builder); err != nil {
 			return errors.Wrapf(err, "failed to complete migrations")
 		}
 	}
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	sq "github.com/Masterminds/squirrel"
+
 	pluginapi "github.com/mattermost/mattermost-plugin-api"
 	"github.com/mattermost/mattermost-plugin-api/cluster"
 )
@@ -98,13 +99,13 @@ func (p *Plugin) OnActivate() error {
 		telemetryEnabled := diagnosticsFlag != nil && *diagnosticsFlag
 
 		if telemetryEnabled {
-			if err := telemetryClient.Enable(); err != nil {
+			if err = telemetryClient.Enable(); err != nil {
 				pluginAPIClient.Log.Warn("Telemetry could not be enabled", "Error", err)
 			}
 			return
 		}
 
-		if err := telemetryClient.Disable(); err != nil {
+		if err = telemetryClient.Disable(); err != nil {
 			pluginAPIClient.Log.Error("Telemetry could not be disabled", "Error", err)
 		}
 	}
@@ -129,7 +130,7 @@ func (p *Plugin) OnActivate() error {
 	api.NewIncidentHandler(p.handler.APIRouter, p.incidentService, p.playbookService, pluginAPIClient, p.bot, p.bot)
 	api.NewSubscriptionHandler(p.handler.APIRouter, p.subscriptionService, p.playbookService, pluginAPIClient)
 
-	if err := command.RegisterCommands(p.API.RegisterCommand); err != nil {
+	if err = command.RegisterCommands(p.API.RegisterCommand); err != nil {
 		return errors.Wrapf(err, "failed register commands")
 	}
 
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -133,13 +133,15 @@ func (p *Plugin) OnActivate() error {
 		return errors.Wrapf(err, "failed register commands")
 	}
 
-	mutex, err := cluster.NewMutex(p.API, manifest.Id+"dbMutex")
+	mutex, err := cluster.NewMutex(p.API, "IR_dbMutex")
 	if err != nil {
 		return errors.Wrapf(err, "failed creating cluster mutex")
 	}
 
 	// Cluster lock: only one plugin will perform the migration when needed
-	p.DBMigration(pluginAPIClient, mutex)
+	if err := p.DBMigration(pluginAPIClient, mutex); err != nil {
+		return errors.Wrapf(err, "failed to run migrations")
+	}
 
 	p.API.LogDebug("Incident response plugin Activated")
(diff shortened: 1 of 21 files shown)


server/sqlstore/migrate.go
```diff
@@ -0,0 +1,34 @@
...
+			return errors.Wrapf(err, "error executing migration from version %s to version %s", migration.fromVersion.String(), migration.toVersion.String())
+		}
+
+		currentSchemaVersion = migration.toVersion
+		if err := SetCurrentVersion(pluginAPIClient, currentSchemaVersion); err != nil {
+			return errors.Wrapf(err, "migration succeeded, but failed to set the current version to %s. Database is now in an inconsistent state", currentSchemaVersion.String())
+		}
+
+		// TODO: Remove when all customers are in 0.1.0
+		if currentSchemaVersion.EQ(semver.MustParse("0.1.0")) {
```
crspeller, line 26: ~~This will re-run on servers running 0.1.0. Maybe check `migration.toVersion`?~~

crspeller: Woopse. You shadowed the function param. Bit confusing. Nit: Maybe don't do that?

agarciamontoro: Good point! Changed it in https://github.com/mattermost/mattermost-plugin-incident-response/pull/256/commits/002c461d977058ef268e30b4acbcc6c68181ddac.

server/plugin.go
```diff
@@ -126,14 +128,47 @@ func (p *Plugin) OnActivate() error {
...
+		return errors.Wrapf(err, "failed creating the SQL store")
+	}
+
+	mutex, err := cluster.NewMutex(p.API, "IR_dbMutex")
+	if err != nil {
+		return errors.Wrapf(err, "failed creating cluster mutex")
+	}
+
+	// Cluster lock: only one plugin will perform the migration when needed
+	if err := p.DBMigration(sqlStore, pluginAPIClient, mutex); err != nil {
```
lieut-data, line 146: nit: can we call this using a verb, like `UpgradeDatabase`?

server/sqlstore/data_migration.go
```diff
@@ -0,0 +1,273 @@
...
+			// odds are this should not happen, so default to failing fast
+			return errors.Wrapf(err, "failed to get incident id '%s'", header.ID)
+		}
+		incidents = append(incidents, *i)
+	}
+
+	builder := sq.StatementBuilder.PlaceholderFormat(sq.Question)
+
+	playbookInsert := builder.
+		Insert("IR_Playbook").
```
lieut-data, line 132: In the event something goes wrong with the migration, we'll end up running it twice -- but the subsequent attempt will find existing records and then fail again since the rows already exist. Is there a way harden against that?

server/sqlstore/migrations.go
```diff
@@ -0,0 +1,146 @@
...
+						INDEX IR_Playbook_TeamID (TeamID),
+						INDEX IR_PlaybookMember_PlaybookID (ID)
+					);
+				`); err != nil {
+					return errors.Wrapf(err, "failed creating table IR_Playbook")
+				}
+
+				if _, err := sqlStore.db.Exec(`
+					CREATE TABLE IF NOT EXISTS IR_PlaybookMember (
+						PlaybookID VARCHAR(26) NOT NULL REFERENCES IR_Playbook(ID),
```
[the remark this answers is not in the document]
lieut-data: It's probably fine within our own schema for now, so long as we don't reference existing Mattermost tables.

server/sqlstore/versions.go
```diff
@@ -0,0 +1,43 @@
...
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
+)
+
+func LatestVersion() semver.Version {
+	return migrations[len(migrations)-1].toVersion
+}
+
+func GetCurrentVersion(pluginAPIClient *pluginapi.Client) (semver.Version, error) {
+	var versionString string
+	if err := pluginAPIClient.KV.Get("DatabaseVersion", &versionString); err != nil {
```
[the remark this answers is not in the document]
lieut-data: One advantage to using an `IR_System` table is the ability to transactionally increment the scheme version alongside the schema changes itself (at least on Postgres). Otherwise we have to jump through a few hoops like with the cluster mutex. 0/5, though, since what we have right now seems elegant.

server/sqlstore/data_migration.go
agarciamontoro: I think a transaction could fix that. What do you think @lieut-data?

server/plugin.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/256/commits/4964ccccb037ea558c78262e4bf48c1db6c021bb.

server/sqlstore/data_migration.go
agarciamontoro: We can also delete all rows in this first migration, of course.

lieut-data: Yes, a transaction would be perfect here -- I missed that we weren't doing that already!

agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/256/commits/97460524cdb260815750c4c16c1aad1c8176a412.

server/sqlstore/versions.go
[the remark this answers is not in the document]
agarciamontoro: I'd say the mutex is needed as long as we support MySQL. But I agree that if we add the `IR_System` table we can better flee mysql if we ever have that opportunity. Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/256/commits/4b7190bf31348af7e1563026b97776b295e64112.

agarciamontoro:
I'd appreciate a quick look to the latest changes, @lieut-data, I ended up changing several things for better supporting transactions and the IR_System table

server/plugin.go
```diff
@@ -155,7 +156,8 @@ func (p *Plugin) UpgradeDatabase(sqlStore *sqlstore.SQLStore, pluginAPIClient *p
 	mutex.Lock()
 	defer mutex.Unlock()
 
-	currentSchemaVersion, err := sqlstore.GetCurrentVersion(pluginAPIClient)
+	currentSchemaVersion, err := sqlStore.GetCurrentVersion()
```
lieut-data, line 159: nit: remove debugging statements?

server/sqlstore/versions.go
```diff
@@ -1,26 +1,52 @@
...
+		return semver.Version{}, errors.Wrap(err, "could not check if table IR_System exists")
+	}
+
+	if !tableExists {
+		return semver.MustParse("0.0.0"), nil
 	}
 
-	if versionString == "" {
+	builder := sq.StatementBuilder.PlaceholderFormat(sq.Question)
+	if sqlStore.db.DriverName() == model.DATABASE_DRIVER_POSTGRES {
```
lieut-data, line 30: This seems brittle -- maybe we can find a way to factor out the need to check at this level and have a function to construct a builder from a `sqlStore`?

server/sqlstore/store.go
```diff
@@ -95,3 +98,36 @@ func (sqlStore *SQLStore) execBuilder(e execer, b builder) error {
...
+
+	builder := sq.StatementBuilder.PlaceholderFormat(sq.Question)
+
+	switch sqlStore.db.DriverName() {
+	case model.DATABASE_DRIVER_MYSQL:
+		query = builder.
+			Select("count(0)").
+			From("information_schema.TABLES").
+			Where("TABLE_SCHEMA = DATABASE()").
+			Where(sq.Eq{"TABLE_NAME": tableName})
```
cpoile, line 113: @lieut-data For something like this, where we know what the driver is, can we just use straight text, instead of making it fit into the squirrel syntax? Just wondering for the future (not this particular piece, since it's already written). It would just make life easier (and as a bonus will make the sql easier to read).

server/sqlstore/versions.go
```diff
@@ -29,14 +55,41 @@ func GetCurrentVersion(pluginAPIClient *pluginapi.Client) (semver.Version, error
...
+func (sqlStore *SQLStore) SetCurrentVersion(tx *sqlx.Tx, currentVersion semver.Version) error {
+	builder := sq.StatementBuilder.PlaceholderFormat(sq.Question)
+	if sqlStore.db.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+		builder = builder.PlaceholderFormat(sq.Dollar)
+	}
+
+	versionUpdate := builder.
+		Update("IR_System").
+		Set("SValue", currentVersion.String()).
+		Where(sq.Eq{"SKey": "DatabaseVersion"})
```
cpoile, line 67: @lieut-data Would you prefer we use your `get/set SystemValue` pattern from cloud? It would make this (and future kv get/sets) simpler.

server/sqlstore/store.go
lieut-data: We definitely don't /have/ to use squirrel to write static queries. Though I admit it's somewhat nice here given the interpolation of `tableName` and the ability have confidence that there's no injection possibility.

server/sqlstore/versions.go
lieut-data: I confess I can't remember the pattern -- are you asking about the column names? Or the way the store is structured?

server/sqlstore/store.go
cpoile: Sounds good. Keep it in mind for the future (if there's no interpolation we might just use a string if it's clearer).

server/sqlstore/versions.go
cpoile: Good point. I've removed all these PlaceholderFormat calls to a single place: the sqlStore itself, and added a `builder` field to the sqlStore struct. Now you can get a builder that's always been set up.

server/sqlstore/versions.go
cpoile: I've added it in the latest commit -- I think it's worth it.

PR #261, opened 2020-08-22
MM-27002: Playbook sql store

cpoile:
#### Summary
- Based on #252 , so will need to rebase off that once it's merged (sorry @agarciamontoro)
- note for #252 : I think I was using transactions wrong, and needed to add the execer and queryer back to the sqlstore/store.go
- In order to move the sorting logic in `GetPlaybooksForTeam` completely into a SQL query, I'm adding two columns to the schema `Stages` and `Steps`. They won't be pulled out into the model, they're just there for sorting in SQL queries. I think this is pretty cool. This is why /not/ using an ORM was the best decision!
- Like #252, I'm eyeballing the SQL, so I'm sure I've made a few mistakes, sorry Alejandro
- Alejandro will take this over for the reviews and revisions, thanks!

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27002

### DIFF
--- a/server/api/playbooks.go
+++ b/server/api/playbooks.go
@@ -185,7 +185,7 @@ func (h *PlaybookHandler) getPlaybooks(w http.ResponseWriter, r *http.Request) {
 	userID := r.Header.Get("Mattermost-User-ID")
 	opts, page, perPage, err := parseGetPlaybooksOptions(r.URL)
 	if err != nil {
-		HandleError(w, errors.Wrap(err, "failed to parse playbook options"))
+		HandleError(w, err)
 		return
 	}
 
@@ -262,24 +262,24 @@ func parseGetPlaybooksOptions(u *url.URL) (opts playbook.Options, page, perPage 
 	param := strings.ToLower(params.Get("sort"))
 	switch param {
 	case "title", "":
-		sortField = playbook.Title
+		sortField = playbook.SortByTitle
 	case "stages":
-		sortField = playbook.Stages
+		sortField = playbook.SortByStages
 	case "steps":
-		sortField = playbook.Steps
+		sortField = playbook.SortBySteps
 	default:
-		sortField = playbook.SortField(param)
+		return playbook.Options{}, 0, 0, errors.Errorf("bad parameter 'sort' (%s): it should be empty or one of 'title', 'stages' or 'steps'", param)
 	}
 
 	var sortDirection playbook.SortDirection
 	param = strings.ToLower(params.Get("direction"))
 	switch param {
 	case "asc", "":
-		sortDirection = playbook.Asc
+		sortDirection = playbook.OrderAsc
 	case "desc":
-		sortDirection = playbook.Desc
+		sortDirection = playbook.OrderDesc
 	default:
-		sortDirection = playbook.SortDirection(param)
+		return playbook.Options{}, 0, 0, errors.Errorf("bad parameter 'direction' (%s): it should be empty or one of 'asc' or 'desc'", param)
 	}
 
 	pageParam := params.Get("page")
--- a/server/api/playbooks.go
+++ b/server/api/playbooks.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/gorilla/mux"
 	"github.com/mattermost/mattermost-plugin-incident-response/server/bot"
@@ -257,14 +258,28 @@ func (h *PlaybookHandler) hasPermissionsToPlaybook(thePlaybook playbook.Playbook
 func parseGetPlaybooksOptions(u *url.URL) (opts playbook.Options, page, perPage int, err error) {
 	params := u.Query()
 
-	sortField := playbook.SortField(params.Get("sort"))
-	if sortField == "" {
+	var sortField playbook.SortField
+	param := strings.ToLower(params.Get("sort"))
+	switch param {
+	case "title", "":
 		sortField = playbook.Title
+	case "stages":
+		sortField = playbook.Stages
+	case "steps":
+		sortField = playbook.Steps
+	default:
+		sortField = playbook.SortField(param)
 	}
 
-	sortDirection := playbook.SortDirection(params.Get("direction"))
-	if sortDirection == "" {
+	var sortDirection playbook.SortDirection
+	param = strings.ToLower(params.Get("direction"))
+	switch param {
+	case "asc", "":
 		sortDirection = playbook.Asc
+	case "desc":
+		sortDirection = playbook.Desc
+	default:
+		sortDirection = playbook.SortDirection(param)
 	}
 
 	pageParam := params.Get("page")
--- a/server/sqlstore/playbook.go
+++ b/server/sqlstore/playbook.go
@@ -39,7 +39,7 @@ func NewPlaybookStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 
 	playbookSelect := builder.
 		Select("ID", "Title", "TeamID", "CreatePublicIncident", "CreateAt",
-			"DeleteAt", "ChecklistJSON").
+			"DeleteAt", "ChecklistJSON", "Stages", "Steps").
 		From("IR_Playbook")
 
 	memberIDsSelect := builder.
@@ -88,6 +88,8 @@ func (p *playbookStore) Create(pbook playbook.Playbook) (id string, err error) {
 			"CreateAt":             model.GetMillis(),
 			"DeleteAt":             0,
 			"ChecklistsJSON":       checklistsJSON,
+			"Stages":               len(pbook.Checklists),
+			"Steps":                getSteps(pbook),
 		}))
 	if err != nil {
 		return "", errors.Wrap(err, "failed to store new playbook")
@@ -264,6 +266,8 @@ func (p *playbookStore) Update(updated playbook.Playbook) (err error) {
 			"CreatePublicIncident": updated.CreatePublicIncident,
 			"DeleteAt":             updated.DeleteAt,
 			"ChecklistsJSON":       checklistsJSON,
+			"Stages":               len(updated.Checklists),
+			"Steps":                getSteps(updated),
 		}).
 		Where(sq.Eq{"ID": updated.ID}))
 
@@ -335,4 +339,12 @@ func addMembersToPlaybooks(memberIDs playbookMembers, out []playbook.Playbook) {
 	for _, p := range out {
 		p.MemberIDs = pToM[p.ID]
 	}
+}
+
+func getSteps(pbook playbook.Playbook) int {
+	steps := 0
+	for _, p := range pbook.Checklists {
+		steps += len(p.Items)
+	}
+	return steps
 }
--- a/server/sqlstore/playbook.go
+++ b/server/sqlstore/playbook.go
@@ -3,7 +3,6 @@ package sqlstore
 import (
 	"database/sql"
 	"encoding/json"
-	"fmt"
 
 	sq "github.com/Masterminds/squirrel"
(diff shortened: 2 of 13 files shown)


server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+
+	err = p.store.getBuilder(tx, &out, p.playbookSelect.Where(sq.Eq{"ID": id}))
+	if err == sql.ErrNoRows {
+		return out, errors.Wrapf(playbook.ErrNotFound, "playbook with does not exist for id '%s'", id)
+	} else if err != nil {
+		return out, errors.Wrapf(err, "failed to get playbook by id '%s'", id)
+	}
+
+	var memberIDs playbookMembers
+	err = p.store.selectBuilder(tx, &memberIDs, p.memberIDsSelect.Where(sq.Eq{"PlaybookID": id}))
```
crspeller, line 135: Would it be possible to use a subquery instead of having two separate queries here? Would that be better?

crspeller: Also some of the other queries below.

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+
+	for _, m := range memberIDs {
+		out.MemberIDs = append(out.MemberIDs, m.MemberID)
+	}
+
+	return out, nil
+}
+
+// GetPlaybooks retrieves all playbooks that are not deleted.
+func (p *playbookStore) GetPlaybooks() (out []playbook.Playbook, err error) {
```
crspeller, line 152: I would be amiss to not point out that this one is quite inefficient. But it exists for compatibility to be improved later right?

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+
+	if err = tx.Commit(); err != nil {
+		return errors.Wrap(err, "could not commit transaction")
+	}
+
+	return nil
+}
+
+// Delete deletes a playbook.
+// TODO: is this what we want to do now? (Never delete, just set deleteAt?)
```
crspeller, line 290: Not sure if mimicking server behavior here is necessary.

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	defer func() {
+		cerr := tx.Rollback()
+		if err == nil && cerr != sql.ErrTxDone {
+			err = cerr
+		}
+	}()
+
+	err = p.store.getBuilder(tx, &out, p.playbookSelect.Where(sq.Eq{"ID": id}))
+	if err == sql.ErrNoRows {
```
crspeller, line 128: How do these work with the transaction?

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+// Ensure playbookStore implements the playbook.Store interface.
+var _ playbook.Store = (*playbookStore)(nil)
+
+type playbookMembers []struct {
+	PlaybookID string
+	MemberID   string
+}
+
+// NewPlaybookStore creates a new store for playbook service.
+func NewPlaybookStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLStore) playbook.Store {
```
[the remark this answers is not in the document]
agarciamontoro: I agree, but we should export the type so the linter does not complain:
```
exported func NewPlaybookStore returns unexported type *sqlstore.playbookStore, which can be annoying to use (golint)
```
I'd like to keep it private unless we need to export it, so I'd keep the interface as return type here.

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+		return "", err
+	}
+
+	// Beginning a transaction because we're doing multiple statements and need a consistent view of the db.
+	tx, err := p.store.db.Beginx()
+	if err != nil {
+		return "", errors.Wrap(err, "could not begin transaction")
+	}
+	defer func() {
+		cerr := tx.Rollback()
```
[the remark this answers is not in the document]
agarciamontoro: I think it's a way to simplify the management of the original error and the possible Rollback error. It's a bit convoluted, but it does make sense. We're updating the named output parameter `err` only if it was originally nil (that is, no other error occurred before the rollback) and the error from the rollback is different from the expected error that will happen when a commit was successful and we try to roll it back. If the rollback error is nil, nothing changes.
The issue here is that we have to manage two errors. This is an alternative to keep both errors logged:
```go
defer func() {
	if err != nil {
		if cerr := tx.Rollback(); cerr != nil {
				err = errors.Wrapf(cerr, "failed to roll back after error (%s)", err.Error())
		}
		err = errors.Wrapf(err, "failed to create playbook, but transaction successfully rolled back")
	}
}()
```
I don't like wrapping the rollback error and printing the other one as a simple string, but I don't thing there's a way to wrap more than one error.
Any suggestion here, @lieut-data?

server/sqlstore/playbook.go
agarciamontoro: In postgres we could use the `array` function, and we have `group_concat` in mysql, but I don't know whether the deseralizing would work with that. I propose we tackle this outside of this PR.

server/sqlstore/playbook.go
agarciamontoro: Same as before, I propose to investigate and address efficiency problems when we encounter them.

server/sqlstore/playbook.go
agarciamontoro: I do think it's better to never delete. We can always delete very old items if needed.

server/api/playbooks.go
```diff
@@ -257,14 +258,28 @@ func (h *PlaybookHandler) hasPermissionsToPlaybook(thePlaybook playbook.Playbook
...
+
+	var sortDirection playbook.SortDirection
+	param = strings.ToLower(params.Get("direction"))
+	switch param {
+	case "asc", "":
 		sortDirection = playbook.Asc
+	case "desc":
+		sortDirection = playbook.Desc
+	default:
+		sortDirection = playbook.SortDirection(param)
```
lieut-data, line 282: Why do we take the raw input from the user and cast it to a `SortDirection`? Would this allow for a SQL injection?

server/sqlstore/playbook.go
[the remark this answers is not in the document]
lieut-data: Yeah, this was a pattern I introduced in the server (IIRC) -- maybe we adopt the `finalizeTransaction` language there to help clarify what's going on?

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+		queryBuilder:    builder,
+		playbookSelect:  playbookSelect,
+		memberIDsSelect: memberIDsSelect,
+	}
+	return newStore
+}
+
+// Create creates a new playbook
+func (p *playbookStore) Create(pbook playbook.Playbook) (id string, err error) {
+	pbook.ID = model.NewId()
```
lieut-data, line 62: Should we assert that this was empty to start?

server/sqlstore/playbook.go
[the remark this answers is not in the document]
lieut-data: (I note too that we don't care about the error from the `Rollback` in that case, since it's not really material to the caller. Fine to just log it only, IMO.)

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+	for _, m := range memberIDs {
+		out.MemberIDs = append(out.MemberIDs, m.MemberID)
+	}
+
+	return out, nil
+}
+
+// GetPlaybooks retrieves all playbooks that are not deleted.
+func (p *playbookStore) GetPlaybooks() (out []playbook.Playbook, err error) {
+	// Beginning a transaction because we're doing multiple selects and need a consistent view of the db.
```
lieut-data, line 153: It's not clear that we /actually/ need this -- and it's weird to have a transaction commit without writing anything.

server/sqlstore/playbook.go
lieut-data: Yeah, having a log of events to trace back old playbooks would be very valuable. 👍 for keeping them around forever and just marking as deleted.

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,350 @@
...
+		cerr := tx.Rollback()
+		if err == nil && cerr != sql.ErrTxDone {
+			err = cerr
+		}
+	}()
+
+	query := p.playbookSelect.
+		Where(sq.Eq{"DeleteAt": 0}).
+		Where(sq.Eq{"TeamID": teamID}).
+		OrderBy(fmt.Sprintf("%s %s", opts.Sort, opts.Direction))
```
lieut-data, line 204: See injection concern above.

server/api/playbooks.go
agarciamontoro: I've reworked all the option management in https://github.com/mattermost/mattermost-plugin-incident-response/pull/261/commits/334ebeac648acb7e00e900442adaf0692dea166d. This is now more consistent with how the incident manage the options.

server/sqlstore/playbook.go
agarciamontoro: Good catch! Fixed this in https://github.com/mattermost/mattermost-plugin-incident-response/pull/261/commits/047257daceb84ae8549f1aa5aa09dd494b85a984 (even if the commit message only talks about the injection, it's there!)

server/sqlstore/playbook.go
agarciamontoro: But what if we read first from IR_Playbook and then from IR_PlaybookMember and there is a change in the middle?

server/sqlstore/playbook.go
agarciamontoro: Fixed in https://github.com/mattermost/mattermost-plugin-incident-response/pull/261/commits/047257daceb84ae8549f1aa5aa09dd494b85a984.

server/sqlstore/playbook.go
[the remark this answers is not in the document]
agarciamontoro: @lieut-data Sorry, but what's the `finalizeTransaction` language?

server/sqlstore/playbook.go
agarciamontoro: Ticket created: https://mattermost.atlassian.net/browse/MM-28149

server/sqlstore/playbook.go
agarciamontoro: Ticket: https://mattermost.atlassian.net/browse/MM-28149

server/sqlstore/playbook.go
agarciamontoro: Propose we tackle this as part of https://mattermost.atlassian.net/browse/MM-28149

server/sqlstore/playbook.go
```diff
@@ -200,8 +203,15 @@ func (p *playbookStore) GetPlaybooksForTeam(teamID string, opts playbook.Options
...
 		Where(sq.Eq{"DeleteAt": 0}).
-		Where(sq.Eq{"TeamID": teamID}).
-		OrderBy(fmt.Sprintf("%s %s", opts.Sort, opts.Direction))
+		Where(sq.Eq{"TeamID": teamID})
+
+	if playbook.IsValidSortBy(opts.Sort) {
+		query = query.OrderBy(string(opts.Sort))
+	}
+
+	if playbook.IsValidOrderBy(opts.Direction) {
```
lieut-data, line 212: I don't /think/ this is quite right -- the former would generate `ORDER BY Sort Direction`, while this will generate `ORDER BY Sort, Direction`, treating the second call to `OrderBy` as a tie-breaking column.

agarciamontoro: Thank you! We'll need a test to check this, so adding it to the unit tests.

PR #267, opened 2020-09-02
Unify backstage styling

lieut-data:
#### Summary
A handful of changes to align with mocks:
* Update the incident details header to feature the same styling as the playbook edit screen's header, notably without showing the top-level "Playbooks | Incidents" navigation.
* Remove the "back arrow" from the main list view, since we are expecting the user to use the `X` on the right instead.
* Remove the `X` from the playbook edit screen.
* Update to `Export Channel Log` nomenclature.

Playbook list:
image

Playbook details:
image

Incident list:
image

Incident details:
image


webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
```diff
@@ -14,46 +15,246 @@ import {getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
...
+    .profile {
+        font-size: 14px;
+    }
+`;
+
+const NavbarPadding = styled.div`
+    flex-grow: 1;
+`;
+
+const BackstageIncidentDetailsContainer = styled.div`
```
lieut-data, line 72: I didn't break this up into discrete components -- for now.

webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
```diff
@@ -14,46 +15,246 @@ import {getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
...
-        try {
-            setIncident(await fetchIncidentWithDetails(incidentId));
-            setFetchingState(FetchingStateType.fetched);
-        } catch {
-            setFetchingState(FetchingStateType.notFound);
-        }
-    };
-
     useEffect(() => {
+        const fetchIncident = async (incidentId: string) => {
```
lieut-data, line 249: Tidying up some tslint warnings.

webapp/src/components/backstage/incidents/status_badge.tsx
```diff
@@ -3,16 +3,31 @@
...
+    background-color: var(--center-channel-color-16);
+    font-weight: 600;
 
-    return (
-        <span className={badgeClass}>
-            {badgeText}
-        </span>
-    );
-}
+    &.ongoing {
```
lieut-data, line 21: I think I'm supposed to use `props` here, but then I'd have to define a type... so I cheated.

asaadmahmood (approved):
Looks great!

webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
cpoile: The only thing I'm annoyed with (with styled components) is that my IDE doesn't highlight it and autocomplete it, like it does for css or sass files. :(

cpoile (approved):

PR #268, opened 2020-09-02
MM-27005 - Unit Test SQL Incident Store

cpoile:
#### Summary
- Pretty self explanatory. 
- Many thanks to @agarciamontoro, much of this is his hard work (especially getting the framework set up)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27005


server/incident/filter_options.go
```diff
@@ -100,21 +99,19 @@ func ValidateOptions(options *HeaderFilterOptions) error {
 	sort := strings.ToLower(options.Sort)
 	switch sort {
 	case SortByCreateAt, "": // default
-		options.Sort = SortByCreateAt
+		options.Sort = "CreateAt"
```
lieut-data, line 102: No need to block here, but I think this "mapping" from the API sort values to the database sort values doesn't belong here. It's an internal detail of the SQLStore, and shouldn't really be exposed outside it.

server/sqlstore/incident.go
```diff
@@ -91,16 +101,16 @@ func (s *incidentStore) GetIncidents(options incident.HeaderFilterOptions) (*inc
...
-			}
-			incidents = append(incidents, *k)
+			j.Checklists = []playbook.Checklist{}
+			incidents = append(incidents, j)
 		}
 	}
 
 	totalCount := len(incidents)
 	incidents = pageIncidents(incidents, options.Page, options.PerPage)
+	if len(incidents) == 0 {
```
lieut-data, line 111: An empty list and a `nil` list are semantically equivallent: except when we serialize to JSON. As such, this feels like a concern outside of the scope of the `sqlstore` package. Should we "deal" with this in the api package instead?

server/sqlstore/incident.go
```diff
@@ -120,9 +130,10 @@ func (s *incidentStore) CreateIncident(newIncident *incident.Incident) (*inciden
 	if newIncident.ID != "" {
 		return nil, errors.New("ID should not be set")
 	}
-	newIncident.ID = model.NewId()
+	incidentCopy := *newIncident
```
lieut-data, line 133: This is only a shallow copy -- are there fields remaining in the `Incident` that will still be shared with the copied object? Should we formalize this by adding a `Clone` method?

server/sqlstore/incident.go
```diff
@@ -259,13 +278,15 @@ func (s *incidentStore) GetCommanders(options incident.HeaderFilterOptions) ([]i
 	}
 
 	var ret []incident.CommanderInfo
+	added := make(map[string]bool)
 	for _, c := range commanders {
 		// TODO: move to permission-checking in the sql call (MM-28008)
-		if options.HasPermissionsTo(c.ChannelID) {
+		if !added[c.CommanderUserID] && options.HasPermissionsTo != nil && options.HasPermissionsTo(c.ChannelID) {
```
lieut-data, line 284: Maybe just add a `DISTINCT` to the query instead?

server/sqlstore/migrations.go
```diff
@@ -43,12 +45,12 @@ var migrations = []Migration{
 						ActiveStage BIGINT NOT NULL,
 						PostID VARCHAR(26) NOT NULL DEFAULT '',
 						PlaybookID VARCHAR(26) NOT NULL DEFAULT '',
-						ChecklistsJSON VARCHAR(60000) NOT NULL,
+						ChecklistsJSON TEXT NOT NULL,
```
lieut-data, line 48: With this change, do we still need to have two `CREATE TABLE` invocations? (Fine if we do, just checking :))

server/sqlstore/migrations.go
```diff
@@ -76,11 +78,17 @@ var migrations = []Migration{
...
 						INDEX IR_PlaybookMember_MemberID (MemberID)
-					);
-				`); err != nil {
+					)
+				` + charset); err != nil {
 					return errors.Wrapf(err, "failed creating table IR_PlaybookMember")
 				}
 			} else {
+				if _, err := e.Exec(`
+					CREATE EXTENSION IF NOT EXISTS unaccent;
```
lieut-data, line 87: This may not be a permission we can rely upon -- is there something built-in that we could do instead?

lieut-data (changes requested):
A handful of comments -- thanks, @cpoile!

server/incident/filter_options.go
cpoile: Agreed, small ticket for that: https://mattermost.atlassian.net/browse/MM-28318

server/pluginkvstore/incident_store.go
```diff
@@ -332,19 +329,19 @@ func sortHeaders(headers []incident.Header, sortField, order string) {
 	// sort by CreateAt, unless we're told otherwise
 	var sortFn = func(i, j int) bool { return orderFn(headers[i].CreateAt > headers[j].CreateAt) }
 	switch sortField {
-	case incident.SortByID:
+	case "ID":
```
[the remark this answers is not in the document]
cpoile: 👍 Added to ticket above.

server/sqlstore/incident.go
cpoile: It is handled in the API... after looking at this, I realize what's happening. The original kvstore tests are expecting a `[]incident.Incident(nil)`, but this is returning a `[]incident.Incident{}`, and that's why Alejandro added this check here. But it's only returning an `[]incident.Incident{}` because we do paging after, and that will return a `[]incident.Incident{}` if there are no incidents in the page. Anyway, no big deal, I just changed the tests.

server/sqlstore/incident.go
cpoile: Yep, you're right (bc of the slices). Done!

server/sqlstore/incident.go
cpoile: I'm not sure if you can, unless you can think of a way? What if you have CommanderA on channel1, and CommanderA for channel2, but the person doing this request only has permissions to channel2?

server/sqlstore/migrations.go
cpoile: I think we do want separate invocations, unless we want to move the Postgres "ideal" `TEXT` columns into the MySQL `VARCHAR(26/1024/etc)` columns. If we're fine with that, we could make it a single `create table` -- up to you. I'm 1/5 to keep them separate, for no reason other than I like doing 'the right thing' for Postgres.

server/sqlstore/migrations.go
cpoile: Sorry, not sure what you mean here. Do you mean the unaccent extension? I'm not exactly sure why it is added here (that's Alejandro), but it looks like it's included in Postgres since at least 9.1, so we should always have it. Did you mean something else?

server/sqlstore/incident.go
lieut-data: Ah, yes -- this would be blocked on the permission being inlined. Once that's done, should tidy up nicely!

server/sqlstore/migrations.go
lieut-data: I mean that invoking `CREATE EXTENSION` may not be a permission granted to the `mattermost` user. Just wondering how dependent we are on this.

cpoile: Is it okay if we wait for Alejandro to answer that -- I'll create a ticket for him to look into it. https://mattermost.atlassian.net/browse/MM-28332

cpoile:
Changes are up, still not sure about the CircleCI tests -- I'll work on them this afternoon I guess (would be nice to have them working)

lieut-data (approved):

cpoile:
@lieut-data Got the sql testing to work with circleci.

cpoile:
Will merge now so I can rebase and work on the playbook store tests -- all that changed since last review was the circleci config, we can iterate on that later. Thanks!

PR #269, opened 2020-09-03
prompt for incident name and description

lieut-data:
#### Summary
When starting an incident, prompt for an incident name (vs. the `channel name` currently provided), and then add a prompt for an optional description. The description ends up as the channel header.

When starting an incident from a post, this description will be pre-filled with the markdown we currently use to record a link to the post, but can be modified freely. We only show `The channel was created by the Incident Response plugin.` in the channel header if no description is provided.

image

image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-28311

### DIFF
--- a/server/api/incidents_test.go
+++ b/server/api/incidents_test.go
@@ -69,8 +69,8 @@ func TestIncidents(t *testing.T) {
 			UserId: "testUserID",
 			State:  "{}",
 			Submission: map[string]interface{}{
-				incident.DialogFieldNameKey:       "incidentName",
 				incident.DialogFieldPlaybookIDKey: "playbookid1",
+				incident.DialogFieldNameKey:       "incidentName",
 			},
 		}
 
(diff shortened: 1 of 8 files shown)


server/api/incidents_test.go
```diff
@@ -69,8 +69,58 @@ func TestIncidents(t *testing.T) {
...
+				TeamID:          dialogRequest.TeamId,
+				Name:            "incidentName",
+			},
+			Playbook: &withid,
+		}
+		retI := i
+		retI.PrimaryChannelID = "channelID"
+		pluginAPI.On("GetChannel", mock.Anything).Return(&model.Channel{}, nil)
+		pluginAPI.On("HasPermissionToTeam", mock.Anything, mock.Anything, model.PERMISSION_CREATE_PUBLIC_CHANNEL).Return(true)
+		pluginAPI.On("HasPermissionToTeam", mock.Anything, mock.Anything, model.PERMISSION_LIST_TEAM_CHANNELS).Return(true)
```
cpoile, line 90: Not a big deal, but you /could/ use the expected channelID and teamIDs here, instead of `mock.Anything` (I can't imagine why they wouldn't be the right ones, but you had mentioned in our call that `mock.Anything`s felt like you weren't testing anything useful?

server/incident/incident.go
```diff
@@ -18,6 +18,7 @@ type Incident struct {
 type Header struct {
 	ID               string `json:"id"`
 	Name             string `json:"name"`
+	Description      string `json:"description"`
```
cpoile, line 21: added ticket to migrate and add to SQL: https://mattermost.atlassian.net/browse/MM-28316

cpoile (approved):

PR #270, opened 2020-09-04
MM-27006 - Unit test the playbook store

cpoile:
#### Summary
- tests the playbook store
- Adds descriptions to incident and migrates them (https://mattermost.atlassian.net/browse/MM-28316)
- Add descriptions to playbooks and migrates them (https://mattermost.atlassian.net/browse/MM-28292)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-27006
- https://mattermost.atlassian.net/browse/MM-28316
- https://mattermost.atlassian.net/browse/MM-28292

### DIFF
--- a/server/sqlstore/incident.go
+++ b/server/sqlstore/incident.go
@@ -39,7 +39,7 @@ var _ incident.Store = (*incidentStore)(nil)
 // NewIncidentStore creates a new store for incident ServiceImpl.
 func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLStore) incident.Store {
 	incidentSelect := sqlStore.builder.
-		Select("ID", "Name", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
+		Select("ID", "Name", "Description", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
 			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "PostID", "PlaybookID").
 		From("IR_Incident")
 
@@ -140,6 +140,7 @@ func (s *incidentStore) CreateIncident(newIncident *incident.Incident) (*inciden
 		SetMap(map[string]interface{}{
 			"ID":              rawIncident.ID,
 			"Name":            rawIncident.Name,
+			"Description":     rawIncident.Description,
 			"IsActive":        rawIncident.IsActive,
 			"CommanderUserID": rawIncident.CommanderUserID,
 			"TeamID":          rawIncident.TeamID,
@@ -178,6 +179,7 @@ func (s *incidentStore) UpdateIncident(newIncident *incident.Incident) error {
 		Update("IR_Incident").
 		SetMap(map[string]interface{}{
 			"Name":            rawIncident.Name,
+			"Description":     rawIncident.Description,
 			"IsActive":        rawIncident.IsActive,
 			"CommanderUserID": rawIncident.CommanderUserID,
 			"EndAt":           rawIncident.EndAt,
@@ -200,9 +202,8 @@ func (s *incidentStore) GetIncident(incidentID string) (*incident.Incident, erro
 		return nil, errors.New("ID cannot be empty")
 	}
 
-	withChecklistsSelect := s.store.builder.
-		Select("ID", "Name", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
-			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "PostID", "PlaybookID", "ChecklistsJSON").
+	withChecklistsSelect := s.incidentSelect.
+		Columns("ChecklistsJSON").
 		From("IR_Incident")
 
 	var rawIncident sqlIncident
--- a/server/sqlstore/incident.go
+++ b/server/sqlstore/incident.go
@@ -130,7 +130,7 @@ func (s *incidentStore) CreateIncident(newIncident *incident.Incident) (*inciden
 	incidentCopy := newIncident.Clone()
 	incidentCopy.ID = model.NewId()
 
-	rawIncident, err := toSQLIncident(incidentCopy)
+	rawIncident, err := toSQLIncident(*incidentCopy)
 	if err != nil {
 		return nil, err
 	}
@@ -170,7 +170,7 @@ func (s *incidentStore) UpdateIncident(newIncident *incident.Incident) error {
 		return errors.New("ID should not be empty")
 	}
 
-	rawIncident, err := toSQLIncident(newIncident)
+	rawIncident, err := toSQLIncident(*newIncident)
 	if err != nil {
 		return err
 	}
@@ -302,15 +302,7 @@ func (s *incidentStore) NukeDB() (err error) {
 	if err != nil {
 		return errors.Wrap(err, "could not begin transaction")
 	}
-
-	defer func() {
-		if err != nil {
-			cerr := tx.Rollback()
-			if cerr != nil && cerr != sql.ErrTxDone {
-				err = errors.Wrap(err, "because of error, tried to rollback. Rollback returned: "+cerr.Error())
-			}
-		}
-	}()
+	defer s.store.finalizeTransaction(tx)
 
 	if _, err := tx.Exec("DELETE FROM IR_Incident"); err != nil {
 		return errors.Wrap(err, "could not delete IR_Incident")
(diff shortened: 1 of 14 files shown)


server/sqlstore/playbook.go
```diff
@@ -304,36 +325,27 @@ func (p *playbookStore) Delete(id string) error {
...
-			c.ID = model.NewId()
-		}
-		for _, i := range c.Items {
-			if i.ID == "" {
-				i.ID = model.NewId()
-			}
-		}
+// replacePlaybookMembers replaces the members of a playbook
+func (p *playbookStore) replacePlaybookMembers(e execer, pbook playbook.Playbook) error {
+	delBuilder := sq.Delete("IR_PlaybookMember").
```
lieut-data, line 330: It's probably not terribly important right now, but it's generally quite "thrashy" to delete all and re-insert every time, given the tombstoning of records and such. Would suggest something more like `DELETE FROM IR_PlaybookMember WHERE PlaybookId = :ID and UserID NOT IN (:member_ids)`, syntax issues aside.

server/sqlstore/playbook.go
```diff
@@ -352,5 +364,33 @@ func getSteps(pbook playbook.Playbook) int {
...
 		steps += len(p.Items)
 	}
+
 	return steps
 }
+
+func toSQLPlaybook(origPlaybook playbook.Playbook) (*sqlPlaybook, error) {
+	for _, checklist := range origPlaybook.Checklists {
+		if len(checklist.Items) == 0 {
+			return nil, errors.New("checklists with no items are not allowed")
```
lieut-data, line 374: As a random aside, I think it would be as simpler UI experience if we just allowed empty checklists, since the frontend doesn't make it super clear "what's wrong" when this happens anyway.

lieut-data (changes requested):
Super cool! A few comments, mostly around the supporting changes rather than the tests themselves.

server/sqlstore/incident.go
```diff
@@ -298,9 +304,11 @@ func (s *incidentStore) NukeDB() (err error) {
 	}
 
 	defer func() {
-		cerr := tx.Rollback()
-		if err == nil && cerr != sql.ErrTxDone {
-			err = cerr
+		if err != nil {
```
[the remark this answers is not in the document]
cpoile: Ok, sounds good. Done!

server/sqlstore/incident.go
```diff
@@ -353,6 +361,27 @@ func toSQLIncident(origIncident *incident.Incident) (*sqlIncident, error) {
 	}, nil
 }
 
+// An incident needs to assign unique ids to its checklist items
+func checklistsToJSON(checklists []playbook.Checklist) (json.RawMessage, error) {
```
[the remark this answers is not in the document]
cpoile: Yep, makes sense. That's a bit of a bug on my part, although you'd never know with the way we use pointers in parameters...
We're now passing in an `incident.Incident` (not a pointer). I've never liked how we pass around pointers in the incident package. But I haven't made any changes to the interface; I would want everyone's input on that to see if other people agree with me or not.

server/sqlstore/incident_test.go
```diff
@@ -508,14 +504,14 @@ func TestGetIncidents(t *testing.T) {
 
 		createIncidents(incidentStore)
 
-		for _, test := range testData {
-			t.Run(driverName+" - "+test.Name, func(t *testing.T) {
-				result, err := incidentStore.GetIncidents(test.Options)
+		for _, tt := range testData {
```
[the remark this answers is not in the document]
cpoile: Either way is fine with me, made the change. :)

server/sqlstore/playbook.go
```diff
@@ -70,38 +75,41 @@ func (p *playbookStore) Create(pbook playbook.Playbook) (id string, err error) {
...
-			"Steps":                getSteps(pbook),
+			"ID":                   rawPlaybook.ID,
+			"Title":                rawPlaybook.Title,
+			"Description":          rawPlaybook.Description,
+			"TeamID":               rawPlaybook.TeamID,
+			"CreatePublicIncident": rawPlaybook.CreatePublicIncident,
+			"CreateAt":             rawPlaybook.CreateAt,
+			"DeleteAt":             rawPlaybook.DeleteAt,
+			"ChecklistsJSON":       rawPlaybook.ChecklistsJSON,
+			"Stages":               len(rawPlaybook.Checklists),
```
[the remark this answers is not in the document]
cpoile: Yep we need it for query time to sort by Stages or Steps -- this will require a change to our validation (to turn `Stages` into `NumStages`. Will do.

server/sqlstore/playbook.go
```diff
@@ -146,17 +165,19 @@ func (p *playbookStore) Get(id string) (out playbook.Playbook, err error) {
 	return out, nil
 }
 
-// GetPlaybooks retrieves all playbooks that are not deleted.
+// GetPlaybooks retrieves all playbooks that are not deleted. Does not return Checklists
```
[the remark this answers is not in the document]
cpoile: I'm not sure, because the only method that does return the checklist is the straight `Get` -- we'd have to change the name of all the `GetPlaybooks*` methods (granted, there's only two...). Maybe just remove that comment?

server/sqlstore/playbook.go
```diff
@@ -190,26 +211,26 @@ func (p *playbookStore) GetPlaybooksForTeam(teamID string, opts playbook.Options
...
 	}()
 
 	query := p.playbookSelect.
 		Where(sq.Eq{"DeleteAt": 0}).
 		Where(sq.Eq{"TeamID": teamID})
 
-	if playbook.IsValidSortBy(opts.Sort) {
+	if playbook.IsValidSortBy(opts.Sort) && playbook.IsValidOrderBy(opts.Direction) {
+		query = query.OrderBy(fmt.Sprintf("%s %s", opts.Sort, opts.Direction))
+	} else if playbook.IsValidSortBy(opts.Sort) {
```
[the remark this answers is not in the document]
cpoile: We do a validation in the API layer in `parseGetPlaybooksOptions`. Are you sure we should double that here as well? Or can we assume that if the request made it through the API layer (or was called directly), we don't have to return errors on invalid options?

server/sqlstore/playbook.go
```diff
@@ -190,26 +211,26 @@ func (p *playbookStore) GetPlaybooksForTeam(teamID string, opts playbook.Options
...
 		query = query.OrderBy(string(opts.Sort))
 	}
 
-	if playbook.IsValidOrderBy(opts.Direction) {
-		query = query.OrderBy(string(opts.Direction))
-	}
-
 	err = p.store.selectBuilder(tx, &out, query)
-	if err == sql.ErrNoRows {
+	if err == sql.ErrNoRows || len(out) == 0 {
```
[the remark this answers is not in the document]
cpoile: Right, good catch. That is how it should work.

server/sqlstore/playbook.go
cpoile: Great, the delete was doable. Sorry it took so long, but I found that the "insert if not exists" is tricky for Postgres pre 9.5. So I did it manually for now. 
As an aside, I'm a little frustrated using squirrel for anything other than what it's built to accommodate. Anything unusual and you're trying to both get the SQL right and find out how to mangle it in the way that squirrel expects.

server/sqlstore/playbook.go
cpoile: Should we make a ticket for Asaad?

server/sqlstore/playbook.go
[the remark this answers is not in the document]
lieut-data: I don't think we should too much about the origin of the data being passed in here. That being said, discarding is a valid and safe approach for now, which we can revisit if we ever have clients other than the api package.

server/sqlstore/playbook.go
lieut-data: Hmm, can you elaborate on the Squirrel challenges? Generally speaking, you can drop into SQL at any time (including for subqueries), so perhaps we should just do that more often?

server/sqlstore/incident.go
```diff
@@ -170,7 +170,7 @@ func (s *incidentStore) UpdateIncident(newIncident *incident.Incident) error {
 		return errors.New("ID should not be empty")
 	}
 
-	rawIncident, err := toSQLIncident(newIncident)
+	rawIncident, err := toSQLIncident(*newIncident)
```
lieut-data, line 173: I'm not sure passing the value here is any "safer" -- the internal `Checklists` is still a pointer, and still subject to potential mutation. It now seems hard to realize this than when it was unambiguously a pointer. Thoughts?

(I realize we appear to copy it correctly down the stack, so I don't think there's an actual bug here.)

server/sqlstore/playbook.go
```diff
@@ -326,9 +291,11 @@ func (p *playbookStore) Delete(id string) error {
 }
 
 // replacePlaybookMembers replaces the members of a playbook
-func (p *playbookStore) replacePlaybookMembers(e execer, pbook playbook.Playbook) error {
+func (p *playbookStore) replacePlaybookMembers(q queryer, e execer, pbook playbook.Playbook) error {
```
lieut-data, line 294: Squirrel defines a `BaseRunner` that embeds both `Execer` and `Queryer` -- should we define a variant for ourselves and accept that here instead to avoid the illusion that we could pass distinct transactions (to separate databases, even)?

server/sqlstore/playbook.go
```diff
@@ -337,16 +304,39 @@ func (p *playbookStore) replacePlaybookMembers(e execer, pbook playbook.Playbook
 		return nil
 	}
 
-	insBuilder := sq.Insert("IR_PlaybookMember").
+	// Now, only add new members
```
lieut-data, line 307: Would something like this help, assuming we just ran a prepared query multiple times vs. binding multiple input parameters?
```sql
INSERT INTO IR_PlaybookMember(PlaybookID, MemberID)
    SELECT :PlaybookId, :MemberId
    WHERE NOT EXISTS (
        SELECT 1 FROM IR_PlaybookMember WHERE PlaybookId = :PlaybookID AND MemberID = :MemberId
    );
```

lieut-data (approved):
A few thoughts, but not strictly blocking I think.

server/sqlstore/incident.go
cpoile: Sorry, I should have been clearer -- I meant: to me if a parameter is a pointer it implies that the parameter can be changed within the function (maybe I'm thinking C, but that's what I always think when I see that). So yah, it's not any "safer", but it does read more clearer (at least for me) -- as long as the function obeys that convention. But I'm not sure what the Go standards are... Are there any conventions about that?

server/sqlstore/playbook.go
cpoile: Yep, that's much better, thank you!

server/sqlstore/playbook.go
cpoile: I was looking at that too, but what stopped me was: if we're thinking of doing this to prevent thrashing (in other words, for performance) then wouldn't it be orders of magnitude more performant to prepare all the values ahead of time and make one call to the db?

lieut-data: My concern with thrashing in his case was relative to writes -- deleting all the rows and inserting them again is effectively `2N` writes in terms of the number of members. Writes are orders of magnitude more expensive than selects, given the eventual write to disk, cache invalidations, etc. So a handful of `SELECT`s would still be way better than writes.

In terms of the number of calls -- a prepared query that's executed multiple times inside a transaction is roughly equivalent to a single call binding a ton of parameters. That is, the actual number of `Execute()` calls isn't necessarily something to optimize. In this case, pulling the existing members down in order to decide which ones to send back might be as expensive as sending every member back for the database to ignore some. (That is, the number of bits across the wire is about the same -- when statements are prepared.)

server/sqlstore/incident.go
lieut-data: I think Go's default position is pass-by-value, unless you need to mutate or you've measured performance issues that passing by pointer helps address. But it's not as "idiomatic" as some of the other guidance I've seen. 0/5 on either direction here.

lieut-data (approved):

server/sqlstore/playbook.go
cpoile: Okay, I tried to make it work with Squirrel, but I can't seem to find a way to build the `SELECT ?, ?` statement. I looked at all the squirrel tests, and they don't test for it, so I'm guessing they can't do it. And since it doesn't have a Values clause, it fails to build. So I just went to raw SQL. /shrug

lieut-data (approved):

PR #274, opened 2020-09-08
The Big sql Merge

cpoile:
#### Summary
- Also contains #270 (which is still under review)
- Please ask away about anything.
- Don't run it yet, I still have to do the stress testing and migration testing.


server/sqlstore/store.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, err
+	}
+	db = sqlx.NewDb(origDB, pluginAPI.Store.DriverName())
+
+	builder := sq.StatementBuilder.PlaceholderFormat(sq.Question)
+	if pluginAPI.Store.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+		builder = builder.PlaceholderFormat(sq.Dollar)
+	}
+
+	// TODO: Leave the default mapper as strings.ToLower?
```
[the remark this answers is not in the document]
cpoile: Nvm, left over from earlier. Removed, thanks.

server/sqlstore/incident.go
```diff
@@ -0,0 +1,404 @@
...
+		searchString := options.SearchTerm
+
+		// Postgres performs a case-sensitive search, so we need to lowercase
+		// both the column contents and the search string
+		if s.store.db.DriverName() == model.DATABASE_DRIVER_POSTGRES {
+			column = "LOWER(UNACCENT(Name))"
+			searchString = normalize(options.SearchTerm)
+		}
+
+		builder = builder.Where(sq.Like{column: fmt.Sprint("%", searchString, "%")})
```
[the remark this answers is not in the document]
cpoile: Honestly I don't mind either way between like or `match against` -- up to you guys.

About injection and squirrel, it's tough for me to find a definitive answer to this. Some people say it does it, but the documentation is so poor I can't find an official statement on it. @lieut-data do you know?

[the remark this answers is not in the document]
cpoile: Update from Jesse: it does indeed (when you construct the query using the builder and the struct map). So no worries there. And the `LIKE` is good for now (keeping in mind for later if we think there's a performance hit, as usual).

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
+)
+
+type sqlPlaybook struct {
+	playbook.Playbook
+	ChecklistsJSON json.RawMessage
+}
+
+// playbookStore is a sql store for playbooks. DO NO USE DIRECTLY Use NewPlaybookStore
```
lieut-data, line 20: Is the warning here necessary? `playbookStore` is not exported

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,404 @@
...
+		return "", errors.New("ID should be empty")
+	}
+	pbook.ID = model.NewId()
+
+	rawPlaybook, err := toSQLPlaybook(pbook)
+	if err != nil {
+		return "", err
+	}
+
+	// Beginning a transaction because we're doing multiple statements and need a consistent view of the db.
```
lieut-data, line 72: nit: I don't think it's necessary to justify starting a transaction with a comment.

server/sqlstore/playbook.go
```diff
@@ -0,0 +1,404 @@
...
+// Delete deletes a playbook.
+func (p *playbookStore) Delete(id string) error {
+	pbook, err := p.Get(id)
+	if err != nil {
+		return err
+	}
+
+	pbook.DeleteAt = model.GetMillis()
+
+	return p.Update(pbook)
```
lieut-data, line 290: nit: this is "racey" with respect to other updates. Would probably recommend a direct `UPDATE` that simply sets `DeleteAt` discretely.

lieut-data (approved):

server/sqlstore/playbook.go
cpoile: I removed the caps, but I left the instruction, because I think it doesn't hurt to document its usage (in case someone isn't familiar with the pattern).

server/sqlstore/playbook.go
cpoile: Yes, should have done that in the first place. 🤦  :)

crspeller (approved):

server/sqlstore/playbook.go
lieut-data: It's not a pattern, per se -- it's literally impossible to use `playbookStore` outside of this package, since it's not exported. (Or maybe I misunderstand what's being communicated?)

cpoile: To me it means: even if you're inside the package don't create this directly, use the `NewPlaybookStore` method, since it does other important things.

server/incident/service.go
```diff
@@ -73,18 +71,15 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, public bool) (*Incident,
 
 	// New incidents are always active
 	incdnt.IsActive = true
-	incdnt.PrimaryChannelID = channel.Id
-	incdnt.CreatedAt = time.Now().Unix()
+	incdnt.ChannelID = channel.Id
+	incdnt.CreateAt = model.GetMillis()
```
agarciamontoro, line 75: Little bug I found while working on something else:
duration


I've filed a ticket and I'll open a PR agains this branch: https://mattermost.atlassian.net/browse/MM-28701

cpoile: Thanks @agarciamontoro

agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/274/commits/210e0d6a4c87adfb65fb62d4d1a9becde352af2f.

PR #275, opened 2020-09-09
MM-27983: use defaultChecked when no onChange

lieut-data:
#### Summary
React throws a warning when using `checked` without `onChange`. Since we use `onClick` with this component, use `defaultChecked` instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27983


crspeller (approved):
I would say that we should make this a controlled component. But I think that the checkbox CSS breaks that?

cpoile (approved):

lieut-data:
Seems to work as a controlled component -- updated! Thanks, @crspeller :)

crspeller (approved):

PR #276, opened 2020-09-10
Playbook edit polish

lieut-data:
#### Summary
A grabbag of changes from a recent design review and some opinionated changes along the way.

* When sharing a playbook, show profiles on first click even before searching.
* Show "Add Slash Command" button until a slash command is added, and then automatically show slash command auto-complete on focus. (Requires parallel webapp changes.)
* Tweak step description adding and removing.
* Style placeholder text with italics:

Finally, allow `Save` at all times, and not just when there are changes. This eliminates a number of weird edge cases where the `Save` button sometimes shows up enabled and sometimes not, and always provides a safe "exit path" vs. having to figure out how to get out of editing without losing potential changes. As part of this, allow empty stage and step names, and just replace with default text for consistency in saving.

### Ticket
Fixes: https://mattermost.atlassian.net/browse/MM-28009

### DIFF
--- a/webapp/src/components/backstage/editable_text.tsx
+++ b/webapp/src/components/backstage/editable_text.tsx
@@ -62,6 +62,9 @@ const EditableText: FC<EditableTextProps> = (props: EditableTextProps) => {
     };
 
     const enterEditMode = () => {
+        // When editing, copy the props value at the instant editing starts.
+        setText(props.text);
+
         // Start the input size at least as wide as the span, itself with a max width.
         setInputWidth(textElement.current?.offsetWidth || 0);
         setEditMode(true);
@@ -93,17 +96,18 @@ const EditableText: FC<EditableTextProps> = (props: EditableTextProps) => {
             </Container>
         );
     }
+
     return (
         <Container
             id={props.id}
             onClick={enterEditMode}
         >
-            {text && text.trim().length > 0 &&
+            {props.text && props.text.trim().length > 0 &&
                 <Text ref={textElement}>
-                    {text}
+                    {props.text}
                 </Text>
             }
-            {(!text || text.length === 0) && props.placeholder && props.placeholder.length > 0 &&
+            {(!props.text || props.text.length === 0) && props.placeholder && props.placeholder.length > 0 &&
                 <Text>
                     {props.placeholder}
                 </Text>
--- a/webapp/src/components/backstage/editable_text.tsx
+++ b/webapp/src/components/backstage/editable_text.tsx
@@ -46,6 +46,14 @@ const Text = styled.span`
     white-space: nowrap;
 `;
 
+const Placeholder = styled.span`
+    overflow: hidden;
+    text-overflow: ellipsis;
+    max-width: 450px;
+    white-space: nowrap;
+    font-style: italic;
+`;
+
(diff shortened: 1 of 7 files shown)


webapp/src/components/backstage/editable_text.tsx
```diff
@@ -93,20 +104,21 @@ const EditableText: FC<EditableTextProps> = (props: EditableTextProps) => {
...
         );
     }
+
     return (
         <Container
             id={props.id}
             onClick={enterEditMode}
         >
-            {text && text.trim().length > 0 &&
+            {props.text && props.text.trim().length > 0 &&
```
lieut-data, line 113: We render the props unless in edit mode so that we're always in sync, i.e. after initializing asynchronously from a template.

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -246,34 +243,20 @@ const PlaybookEdit: FC<Props> = (props: Props) => {
...
-
-        // It's possible there was actually nothing there.
-        if (playbookExcludingEmpty.checklists.length === 0) {
-            return;
-        }
-
-        if (playbookExcludingEmpty.title.trim().length === 0) {
-            playbookExcludingEmpty.title = 'Untitled Playbook';
-        }
+            title: playbook.title.trim() || 'Untitled Playbook',
```
lieut-data, line 249: Filling in any "missing" data everywhere instead of rejecting the save (and not closing).

cpoile (approved):

webapp/src/components/backstage/playbook_edit.tsx
crspeller, line 249: Nit: Since this needed a comment in code review, maybe pull this out into a named function as docs?

webapp/src/components/backstage/step_edit.tsx
```diff
@@ -99,118 +106,261 @@ const OverrideWebappStyle = createGlobalStyle`
...
+};
+
+interface StepDescriptionProps {
+    description: string;
+    setDescription: (description: string) => void;
+}
+
+const StepDescription: FC<StepDescriptionProps> = (props: StepDescriptionProps) => {
+    const [description, setDescription] = useState(props.description);
+    const [hasValue, setHasValue] = useState(props.description.length > 0);
```
crspeller, line 179: Nit: Confusing naming. Maybe `desciptionOpen` and `setDescriptionOpen`? Because by name it's seems like useless state completely dependent on description above.

webapp/src/components/backstage/step_edit.tsx
```diff
@@ -99,118 +106,261 @@ const OverrideWebappStyle = createGlobalStyle`
...
+
+interface StepCommandProps {
+    command: string;
+    setCommand: (command: string) => void;
+    autocompleteOnBottom: boolean;
+}
+
+const StepCommand: FC<StepCommandProps> = (props: StepCommandProps) => {
+    const [command, setCommand] = useState(props.command);
+    const [hasValue, setHasValue] = useState(props.command.length > 0);
```
crspeller, line 251: Nit: Same naming comment as above.

crspeller (approved):

PR #277, opened 2020-09-13
MM-28008 - Permissions checking in SQL; use subqueries

cpoile:
#### Summary
- This is why we moved to SQL
- moves permission checking into the sql calls for playbooks and incidents. 
- huge speedup; for playbooks we went from 8.6s -> 49ms, for incidents we go from 1 minute -> 84ms (on a stress test of ~4400 incidents and 1000 playbooks).
- had to use a string for the nested exists, because assembling it with Squirrel worked for MySQL but not for Postgres

NOTE:
- Have asked for reviews from 3 devs, first 2 who want to can review it (I wasn't sure how overloaded Jesse is this week)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-28008
- https://mattermost.atlassian.net/browse/MM-28149
- https://mattermost.atlassian.net/browse/MM-26698


server/playbook/filter_options.go
```diff
@@ -29,6 +29,13 @@ const (
 type Options struct {
 	Sort      SortField
 	Direction SortDirection
+
+	// Pagination options.
+	Page    int
+	PerPage int
+
+	// Permissions Check
+	HasPermissionsTo func(pb Playbook, userID string) bool
```
crspeller, line 38: This is the kind of weird stuff you get when we do backwards compatibility. It confused me a bit why we needed this if we had switched to the queries... but then I remembered.

server/sqlstore/incident.go
```diff
@@ -33,6 +33,30 @@ type incidentStore struct {
 	incidentSelect sq.SelectBuilder
 }
 
+const (
+	isAdminOrMemberOrPublicChannelSQLText = `
```
crspeller, line 37: This query isn't perfect. It gives broader permissions then previously granted. For example, a guest user is a member of the team and there might be open incident channels, but you might not want them to see them. 

I think we might need to actually copy the permissions queries from the server.

crspeller (changes requested):

server/sqlstore/incident.go
cpoile: I agree, we should define the exact permissions -- @lieut-data can you help with this?
For guests, we could a `not guest` check, if we do want to limit guests to not seeing public incidents. Is that what we want?

What we actually do is:
1. check if you're a system admin; if not try:
1. check if you have permission to read channel; if not try:
1. check if the channel is public and if you can "LIST_TEAM_CHANNELS". 

It's the `LIST_TEAM_CHANNELS` that I'm somewhat guessing on.

But I'm unclear on what the exact permissions we were using. Because I tried tracing the permissions in the server code, and it's not clear what's happening when you get into `mergeChannelHigherScopedPermissions`.

server/playbook/filter_options.go
cpoile: Yah. I'm not sure when we'll pull out all that code, so I had to make sure it still worked even with the refactoring.

server/playbook/playbook.go
```diff
@@ -69,7 +76,7 @@ type Service interface {
 	// GetPlaybooks retrieves all playbooks
 	GetPlaybooks() ([]Playbook, error)
 	// GetPlaybooksForTeam retrieves all playbooks on the specified team given the provided options
-	GetPlaybooksForTeam(teamID string, opts Options) ([]Playbook, error)
+	GetPlaybooksForTeam(requester, teamID string, opts Options) (GetPlaybooksResults, error)
```
agarciamontoro, line 79: Nit: can this be `requesterID`, just for consistency?

server/sqlstore/playbook.go
```diff
@@ -180,51 +182,101 @@ func (p *playbookStore) GetPlaybooks() (out []playbook.Playbook, err error) {
...
+	pageCount := int(math.Ceil(float64(total) / float64(opts.PerPage)))
+	hasMore := opts.Page+1 < pageCount
 
-	if err = tx.Commit(); err != nil {
-		return out, errors.Wrap(err, "could not commit transaction")
+	ret := make([]playbook.Playbook, 0, len(playbookWithMembers))
+	for _, p := range playbookWithMembers {
+		p.MemberIDs = strings.Fields(p.Members)
+		ret = append(ret, p.Playbook)
 	}
```
agarciamontoro: This is perfect, but I had to double check what was happening with `p.MemberIDs`, because we are only appending `p.Playbook` to`ret`. Of course, the `p.MemberIDs` is part of `p.Playbook`, but it is a bit confusing (this is Go's fault, not yours).

server/sqlstore/playbook.go
```diff
@@ -403,3 +455,12 @@ func directionOptionToSQL(direction playbook.SortDirection) string {
 		return ""
 	}
 }
+
+func correctPaginationOpts(opts *playbook.Options) {
+	if opts.PerPage <= 0 {
+		opts.PerPage = 1000
+	}
+	if opts.Page <= 0 {
```
agarciamontoro, line 462: The nittiest nit you can imagine:
```suggestion
	if opts.Page < 0 {
```

server/sqlstore/support_for_test.go
```diff
@@ -94,3 +94,204 @@ func setupSQLStore(t *testing.T, db *sqlx.DB) (PluginAPIClient, bot.Logger, *SQL
...
+
+		_, err = db.Exec(`
+			CREATE TABLE IF NOT EXISTS public.channelmemberhistory (
+				channelid character varying(26) NOT NULL,
+				userid character varying(26) NOT NULL,
+				jointime bigint NOT NULL,
+				leavetime bigint
+			);
+		`)
+		require.NoError(t, err)
```
agarciamontoro, line 142: This was left here after the refactor, but I think it only belongs in `setupChannelMemberHistoryTable`.

server/sqlstore/incident.go
agarciamontoro: Also, can we change this into a function, so we can call something like `buildPermissionsWhereExpr(requesterID, opts.TeamID)`? This would prevent the need to look into the position of the parameters in this string, and we can rely on the function parameteres.

agarciamontoro (changes requested):
The speedups are amazing, the change was indeed worth it!

I left some comments below, mostly nits.

server/sqlstore/playbook.go
agarciamontoro: (I wrote this in an earlier commit, but the comment still applies after the latest changes)

cpoile: Ok, makes sense -- code for clarity while reading, not for terseness. Fixed. :)

server/sqlstore/playbook.go
cpoile: That's pretty nitty, but I'll own it. Fixed. 😂

server/sqlstore/support_for_test.go
cpoile: Ah yes, missed that. Thanks.

server/sqlstore/incident.go
cpoile: That's a great idea! Done.

agarciamontoro (approved):
Great, thank you! :tada:

lieut-data (commented):
Apologies, I had typed up a bunch of comments yesterday but didn't have time to finish a detailed SQL review. I'm posting them now, knowing that some have already been raised, and then will do a second pass to get through the SQL and respond to the "queries" above (pun intended ;P).

server/sqlstore/incident.go
lieut-data: On the issue of building the query, I think we can pre-flight / use-the-api to check the admin and guest bits and make decisions accordingly. (If guest, then only check channel membership, not `LIST_TEAM_CHANNELS`).

Would that simplify things?

cpoile: @lieut-data It's an interesting solution, but I think it might end up causing more complication---if we use the api to check for system admin and guest, then we're sending those bits and pushing the logic into the sqlstore, which has to build a custom query based on those bits. I think it might be simpler to add the guest check in the SQL. Then we use the same permissions logic for every query.

cpoile: Or I could be wrong, and it's simpler to build it up... How about I'll try both and we'll see?

lieut-data: Yeah, that's fair. It will be especially good if we can replicate precisely what the server is doing here.

cpoile: From our discussion in sprint planning: do the tests first, then send the permissions over to the sqlstore.

server/api/playbooks.go
```diff
@@ -183,12 +177,14 @@ func (h *PlaybookHandler) getPlaybooks(w http.ResponseWriter, r *http.Request) {
...
 	teamID := params.Get("team_id")
 	userID := r.Header.Get("Mattermost-User-ID")
-	opts, page, perPage, err := parseGetPlaybooksOptions(r.URL)
+	opts, err := parseGetPlaybooksOptions(r.URL)
 	if err != nil {
 		HandleError(w, err)
 		return
 	}
 
+	opts.HasPermissionsTo = h.hasPermissionsToPlaybook
```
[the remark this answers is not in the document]
cpoile: Yep, this'll be removed in the-big-kvstore-removal PR. :)

server/incident/filter_options.go
```diff
@@ -118,9 +118,9 @@ func ValidateOptions(options *HeaderFilterOptions) error {
 
 	order := strings.ToLower(options.Order)
 	switch order {
-	case OrderDesc, "": // default
+	case OrderDesc:
 		options.Order = OrderDesc
-	case OrderAsc:
+	case OrderAsc, "": // default (like in SQL)
```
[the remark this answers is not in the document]
cpoile: It mattered because while I was writing the tests I had to keep remembering what the default ordering was, and it was different from the playbookStore's. So I wanted to make them consistent, and chose asc because I bet it's what people would expect as the default. I will remove the comment though, since you're right it doesn't really matter that SQL does it like this.

server/pluginkvstore/incident_store.go
```diff
@@ -93,7 +93,7 @@ func (s *incidentStore) GetIncidents(options incident.HeaderFilterOptions) (*inc
 	totalCount := len(filtered)
 	filtered = pageHeaders(filtered, options.Page, options.PerPage)
 
-	var result []incident.Incident
+	result := []incident.Incident{}
```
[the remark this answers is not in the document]
cpoile: This is just for the kvstore. I just made a change to return nil and do the empty slice conversion in the api layer. :)

server/sqlstore/incident.go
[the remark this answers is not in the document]
cpoile: Yep, done!

server/sqlstore/incident.go
```diff
@@ -33,6 +33,30 @@ type incidentStore struct {
...
+		(
+              EXISTS(SELECT 1
+                         FROM Users AS u
+                         WHERE u.Id = ?
```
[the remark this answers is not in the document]
cpoile: No need now, there's only one parameter. But I checked, and it doesn't look like it lets you--it's not in the docs (such as they are) and I couldn't find an example of named parameters in the tests.

server/sqlstore/incident.go
```diff
@@ -33,6 +33,30 @@ type incidentStore struct {
...
+                           AND u.Roles LIKE '%system_admin%')
+              OR EXISTS(SELECT 1
+                            FROM ChannelMembers as cm
+                            WHERE cm.ChannelId = inc.ChannelID
+                              AND cm.UserId = ?)
```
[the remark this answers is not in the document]
cpoile: Nope, I traced the code -- the row is deleted. (As you know, we keep track of past membership in the history table.)

server/sqlstore/incident.go
```diff
@@ -33,6 +33,30 @@ type incidentStore struct {
...
+              OR (
+                      EXISTS(SELECT 1
```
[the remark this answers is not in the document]
cpoile: Not needed now -- we test for CanViewTeamChannels, and if they can then all we need to test for is whether the channel is open.

server/sqlstore/incident.go
```diff
@@ -33,6 +33,30 @@ type incidentStore struct {
...
+                                 FROM Channels as c
+                                 WHERE c.Id = inc.ChannelID
+                                   AND c.Type = 'O')
```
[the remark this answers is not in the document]
cpoile: At the moment we don't have that restriction -- you can still list and view details of and even visit incident channels that have been archived. Is there another level of Delete that I haven't encountered yet?

server/sqlstore/incident.go
```diff
@@ -33,6 +33,30 @@ type incidentStore struct {
...
+                      AND EXISTS(SELECT 1
+                                     FROM TeamMembers AS t
+                                     WHERE t.TeamId = ?
+                                       AND t.UserId = ?)
```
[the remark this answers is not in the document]
cpoile: Removed now, so not needed.

server/sqlstore/incident.go
```diff
@@ -41,7 +65,7 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 	incidentSelect := sqlStore.builder.
 		Select("ID", "Name", "Description", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
 			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "PostID", "PlaybookID").
-		From("IR_Incident")
+		From("IR_Incident AS inc")
```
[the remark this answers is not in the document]
cpoile: Yep, I didn't like that either -- I kept thinking it was "increment" which is what Clojure uses all the time.

server/sqlstore/incident.go
```diff
@@ -53,66 +77,80 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
...
-	if err := incident.ValidateOptions(&options); err != nil {
+func (s *incidentStore) GetIncidents(requesterID string, opts incident.HeaderFilterOptions) (*incident.GetIncidentsResults, error) {
+	if err := incident.ValidateOptions(&opts); err != nil {
 		return nil, err
 	}
 
-	builder := s.incidentSelect.
-		Where(sq.Eq{"DeleteAt": 0})
+	isAdminOrMemberOrPublicChannel := sq.Expr(isAdminOrMemberOrPublicChannelSQLText,
+		requesterID, requesterID, opts.TeamID, requesterID)
```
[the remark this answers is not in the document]
cpoile: Yep, not needed anymore. :)

server/api/incidents.go
```diff
@@ -309,7 +304,13 @@ func (h *IncidentHandler) getIncidents(w http.ResponseWriter, r *http.Request) {
 		return err2 == nil
 	}
 
-	results, err := h.incidentService.GetIncidents(*filterOptions)
+	requesterInfo := incident.RequesterInfo{
+		UserID:              userID,
+		IsAdmin:             permissions.IsAdmin(userID, h.pluginAPI),
+		CanViewTeamChannels: permissions.CanViewTeam(userID, filterOptions.TeamID, h.pluginAPI),
```
lieut-data, line 310: nit: we don't "remember" the `TeamID` here, so `CanViewTeamChannels` can be a but worrisome as a general flag. Maybe a `map[string]bool` to make this explicit and safe by default?

server/sqlstore/support_for_test.go
```diff
@@ -94,3 +94,332 @@ func setupSQLStore(t *testing.T, db *sqlx.DB) (PluginAPIClient, bot.Logger, *SQL
 
 	return pluginAPIClient, logger, sqlStore
 }
+
+func setupUsersTable(t *testing.T, db *sqlx.DB) {
+	t.Helper()
+
+	// Statements copied from mattermost-server/scripts/mattermost-postgresql-5.0.sql
```
lieut-data, line 101: If we copied from here, this schema is "out of date" relative to v5.28 (by over two years). Possibly not an issue for now.

server/sqlstore/playbook.go
```diff
@@ -180,51 +182,100 @@ func (p *playbookStore) GetPlaybooks() (out []playbook.Playbook, err error) {
...
+				ON p.ID = pmember.PlaybookID
+			`
+	}
+
+	isAdminOrTeamAndPlaybookMember := sq.Expr(`
+		(
+            EXISTS(SELECT 1
+                       FROM Users AS u
+                       WHERE u.Id = ?
+                         AND u.Roles LIKE '%system_admin%')
```
lieut-data, line 209: Should we have a requester info that passes the `isAdmin` as well?

server/sqlstore/playbook.go
```diff
@@ -180,51 +182,100 @@ func (p *playbookStore) GetPlaybooks() (out []playbook.Playbook, err error) {
...
+		return playbook.GetPlaybooksResults{}, errors.Wrap(err, "failed to get playbooks")
 	}
 
-	nestedQuery := p.queryBuilder.
-		Select("ID").
-		From("IR_Playbook").
+	queryForTotal := p.store.builder.
+		Select("COUNT(*)").
+		From("IR_Playbook AS p").
+		Join(getMembers).
```
lieut-data, line 255: nit: I find this embedded `getMembers` to be much harder to parse than the original two queries. It seems like a modest performance difference between the two implementations?

server/sqlstore/playbook.go
```diff
@@ -180,51 +182,100 @@ func (p *playbookStore) GetPlaybooks() (out []playbook.Playbook, err error) {
...
+				(SELECT PlaybookID, GROUP_CONCAT(MemberID SEPARATOR ' ') members
+				   FROM IR_PlaybookMember
+				   GROUP BY PlaybookID
+				) pmember
+				ON p.ID = pmember.PlaybookID
+			`
+	}
+
+	isAdminOrTeamAndPlaybookMember := sq.Expr(`
+		(
```
lieut-data, line 205: Is this parenthesis necessary? Wondering if `WHERE` from the builder will do the right thing by itself?

server/sqlstore/playbook.go
```diff
@@ -180,51 +182,100 @@ func (p *playbookStore) GetPlaybooks() (out []playbook.Playbook, err error) {
...
+            EXISTS(SELECT 1
+                       FROM Users AS u
+                       WHERE u.Id = ?
+                         AND u.Roles LIKE '%system_admin%')
+            OR (
+                    EXISTS(SELECT 1
+                               FROM TeamMembers AS t
+                               WHERE t.TeamId = ?
+                                 AND t.UserId = ?)
```
lieut-data, line 214: Check `DeleteAt`?

server/sqlstore/incident.go
```diff
@@ -53,66 +53,76 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 }
 
 // GetIncidents returns filtered incidents and the total count before paging.
-func (s *incidentStore) GetIncidents(options incident.HeaderFilterOptions) (*incident.GetIncidentsResults, error) {
-	if err := incident.ValidateOptions(&options); err != nil {
+func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, opts incident.HeaderFilterOptions) (*incident.GetIncidentsResults, error) {
```
lieut-data, line 56: nit: why `opts` from `options`?

server/sqlstore/incident.go
```diff
@@ -53,66 +53,76 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
...
 
-	var incidents []incident.Incident
-	for _, j := range rawIncidents {
-		// TODO: move to permission-checking in the sql call (MM-28008)
-		if options.HasPermissionsTo == nil || options.HasPermissionsTo(j.ChannelID) {
-			j.Checklists = []playbook.Checklist{}
-			incidents = append(incidents, j)
-		}
+	for i := range incidents {
+		incidents[i].Checklists = []playbook.Checklist{}
```
lieut-data, line 114: Why is this necessary?

server/sqlstore/incident.go
```diff
@@ -254,46 +264,32 @@ func (s *incidentStore) GetAllIncidentMembersCount(channelID string) (int64, err
...
-	// At the moment, the options only includes teamID and the HasPermissionsTo
+	isAdminOrMemberOrPublicChannel := s.buildPermissionsExpr(requesterInfo)
+
+	// At the moment, the opts only includes teamID
 	query := s.queryBuilder.
-		Select("CommanderUserID", "ChannelID", "Username").
-		From("IR_Incident AS i").
-		Join("Users AS u ON i.CommanderUserID = u.Id").
-		Where(sq.Eq{"TeamID": options.TeamID})
+		Select("DISTINCT CommanderUserID AS UserID", "Username").
```
lieut-data, line 276: nit: propose always namespacing the columns when there are multiple being selected, so `incident.CommandUserID` and `u.Username` (or just `u.Id` and `u.Username`).

server/sqlstore/incident.go
```diff
@@ -254,46 +264,32 @@ func (s *incidentStore) GetAllIncidentMembersCount(channelID string) (int64, err
...
-		// TODO: move to permission-checking in the sql call (MM-28008)
-		if !added[c.CommanderUserID] && options.HasPermissionsTo != nil && options.HasPermissionsTo(c.ChannelID) {
-			ret = append(ret, incident.CommanderInfo{
-				UserID:   c.CommanderUserID,
-				Username: c.Username,
-			})
-			added[c.CommanderUserID] = true
-		}
+	if commanders == nil {
+		commanders = []incident.CommanderInfo{}
```
lieut-data, line 289: nit: this is similar to a comment of my above, but ideally shouldn't be necessary at this layer.

server/sqlstore/incident.go
```diff
@@ -322,18 +318,36 @@ func (s *incidentStore) NukeDB() (err error) {
...
 	}
-	return b
+
+	if info.CanViewTeamChannels {
+		checkMembershipOrPublicChannel := sq.Expr(`
+		  (
+			  -- If requester is a channel member
+			  EXISTS(SELECT 1
+						 FROM ChannelMembers as cm
+						 WHERE cm.ChannelId = incident.ChannelID
```
lieut-data, line 332: Check for channel membership deletion?

lieut-data (changes requested):
Nice work, @cpoile! A few follow up questions.

server/api/incidents.go
cpoile: Not sure what you mean here by remembering the `TeamID`, and why that would affect the `CanViewTeamChannels` flag...?

server/sqlstore/support_for_test.go
cpoile: Right -- at this point all we need are the few fields that we check in the tests, so none of the changes up to 5.28 will affect it. I left Alejandro's comment here because it's good to remind people that it's an old schema, and if they run into a mismatch and can't run the tests, that's why. I'll add a comment to that effect.

server/sqlstore/playbook.go
cpoile: Yep, I made playbooks consistent with how we do it in incident.

server/sqlstore/playbook.go
cpoile: Sorry, I'm not following -- we use `Join(getMembers)` the same way in both `queryForResults` and `queryForTotal`, what original two queries do you mean?

> It seems like a modest performance difference between the two implementations?

Also not sure what you mean here.

server/sqlstore/playbook.go
cpoile: It didn't because Postgres doesn't like the way Squirrel `or`s `exists` statements. But either way, it doesn't matter -- with the `requesterInfo` these checks reduce to one `exists` statement, which is fine for both dbs, so I converted it to a builder statement. :)

server/sqlstore/playbook.go
cpoile: I thought I commented about that before? (maybe it got lost by Github?) Anyway, it's moot since we're using the `requesterInfo` and we don't need this check anymore.

server/sqlstore/incident.go
cpoile: To be consistent with the `GetCommanders` parameter name -- I'm 0/5, so I changed them both to `options`.

server/sqlstore/incident.go
cpoile: It's the frontend expecting an array problem. I moved this to the api. :)

server/sqlstore/incident.go
cpoile: (also pretty sure I commented on this before) -- there's no DeleteAt column for these rows (we have ChannelMemberHistory)

server/api/incidents.go
lieut-data: The `RequesterInfo` isn't self-contained. We compute a bit that is `CanViewTeamChannels` and assume that somewhere, downstream, someone is using that bit against the same `TeamID` that we used. It works that way right now, but lends itself to misuse IMO.

server/sqlstore/playbook.go
lieut-data: Before, we appeared to collect all the playbook IDs, then query for the members discretely. Now we're using the aggregate functions to do that in "one query", but we have to deal with MySQL vs Postgres... and it seems complicated :)

server/api/incidents.go
cpoile: Ah, I see what you mean. Hah, I never even thought about that. Sure, I'll make that change!

server/sqlstore/playbook.go
lieut-data: Sorry, it's not super obvious why requester info "guards" us against this -- I think the SQL queries should be reasonably "self-contained", and assuming that the user won't be deleted because we checked it upstream seems brittle.

server/sqlstore/incident.go
lieut-data: Sounds good! Sorry, I definitely missed these :)

server/api/playbooks.go
```diff
@@ -199,7 +199,13 @@ func (h *PlaybookHandler) getPlaybooks(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	playbookResults, err := h.playbookService.GetPlaybooksForTeam(userID, teamID, opts)
+	requesterInfo := playbook.RequesterInfo{
+		UserID:              userID,
+		IsAdmin:             permissions.CanViewTeam(userID, teamID, h.pluginAPI),
```
lieut-data, line 204: `IsAdmin` !== `CanViewTeam`?

server/command/command.go
```diff
@@ -127,7 +127,13 @@ func (r *Runner) actionStart(args []string) {
 		postID = args[1]
 	}
 
-	playbooksResults, err := r.playbookService.GetPlaybooksForTeam(r.args.UserId, r.args.TeamId,
+	requesterInfo := playbook.RequesterInfo{
+		UserID:              r.args.UserId,
+		IsAdmin:             permissions.CanViewTeam(r.args.UserId, r.args.TeamId, r.pluginAPI),
```
lieut-data, line 132: See above.

lieut-data (changes requested):
Great, just spotted a permissions error.

server/api/playbooks.go
cpoile: Shoot, I was too quick typing. Sorry, will fix.

server/command/command.go
cpoile: Was a cut and paste from above. 🤕

server/sqlstore/playbook.go
cpoile: Yah, I guess It's true, we're moving the complexity into the sql store, and there probably isn't a huge speedup... But it's so coooool. :P

Okay, I'll change it back -- priority should be readability and maintainability, not technical cleverness.  I'm going to save it though, it makes me happy.

server/sqlstore/playbook.go
cpoile: You're right, I guess we are assuming that if the user isAdmin or if the user CanViewTeam, then they aren't deleted and they should be given those permissions in the store layer.  But that's the same permissions sequence we used before. We never checked if the user was not deleted before, do we want to do that for every query now?  

I'm okay to do it, I just wonder what situation we're guarding against, which the previous kvstore implementation was not guarded against?

server/api/incidents.go
```diff
@@ -306,8 +306,9 @@ func (h *IncidentHandler) getIncidents(w http.ResponseWriter, r *http.Request) {
 
 	requesterInfo := incident.RequesterInfo{
 		UserID:              userID,
-		IsAdmin:             permissions.IsAdmin(userID, h.pluginAPI),
-		CanViewTeamChannels: permissions.CanViewTeam(userID, filterOptions.TeamID, h.pluginAPI),
+		TeamID:              filterOptions.TeamID,
+		UserIDtoIsAdmin:     map[string]bool{userID: permissions.IsAdmin(userID, h.pluginAPI)},
```
lieut-data, line 310: I apologize for all the thrashing on this -- and please ignore this comment for the purpose of merging this PR.

I really like your solution of just adding `TeamID` to the struct. Then it's all grouped together -- `UserID`, `TeamID`, `IsAdmin`, `CanViewTeam` now all make sense in context. Hard to use "incorrectly."

But doing /both/ `TeamID` and a `TeamIDtoCanViewTeam` feels weird, since it's not a set of `TeamIDs` for which the request is being made, so I really don't need "support for one."

lieut-data (approved):

server/api/incidents.go
cpoile: Ok, makes sense! I'll merge this and then fix this a bit later, but I understand now. :)

PR #279, opened 2020-09-15
MM-28701: Fix timestamp handling

agarciamontoro:
#### Summary
The database tables are storing the timestamps as milliseconds, but the KV store was storing them as seconds. This PR fixes a conflict caused by this difference with the Duration component:

1. Parse timestamps with [`moment`](https://momentjs.com/docs/#/parsing/unix-timestamp-milliseconds/) instead of [`moment.unix`](https://momentjs.com/docs/#/parsing/unix-timestamp/) in the Duration component.
2. Fix the migration code to convert the old timestamps to milliseconds.

This PR assumes that we'll get rid of the SQL feature flag. If that flag was in place, it was disabled, and we would decide to migrate the old KV data to the new data model (right now this is not working properly), the changes in the Duration component would make the duration behave incorrectly for the old incidents.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28701


cpoile:
>  If that flag was in place, it was disabled, and we would decide to migrate the old KV data to the new data model (right now this is not working properly)

Not sure what you mean by this, can you explain?

> , the changes in the Duration component would make the duration behav incorrectly for the old incidents.

Right, that's correct -- I forgot to feature flag the front end duration changes. We will want to do that if we plan on using the feature flag on community.

cpoile (approved):

agarciamontoro:
>> If that flag was in place, it was disabled, and we would decide to migrate the old KV data to the new data model (right now this is not working properly)

> Not sure what you mean by this, can you explain?

Right now, as far as I know, we cannot decode old incidents (the ones in the KV store written by the code in `master`) with the kv store code in `the-big-sql-merge` branch, because of [the change in the data model](https://github.com/mattermost/mattermost-plugin-incident-response/pull/274/files#diff-de035e092cd77feecef9a95700023bdfR17-R25). I understand that this makes the SQL feature flag ineffective, as it is only guarding the usage of the SQL store, but the data model is still changed.
If we decide to fix this in order to make the feature flag effective (we would need to convert the data that we have in the KV store to use the new data model), then a disabled flag would make the Duration component behave incorrectly, because it is not checking that flag, and parsing the timestamps as milliseconds.

We can discuss this in the meeting today!

lieut-data (approved):

PR #280, opened 2020-09-15
MM-28332: Add a fallback for when unaccent is not available

agarciamontoro:
#### Summary

- I've changed the migration so we don't finish it when we are unable to create the extension, but only log the error. Also, in case the extension fails to be created, I've added a fallback so we don't use it (I considered implementing a noop SQL function so we don't need to check whether the real function exists, but we would have the same problem: does the user has permissions to add a function?)
- I've smoke-tested that the general Mattermost search is not affected by this change. Also, I'm 99% sure it won't affect anything because all the extension does is adding a dictionary and a function, which you need to explicitly call to get the behavior.
- Also, I was testing this mostly with the ñ letter, and after all was implemented, I tried to create an incident with letters with accents (áéíóú) and it turns out we cannot create incidents with those letters :sob: (my gut feeling is because of channel names restrictions). We may want to look into that, of course, but as this work was already done, I think it's better to merge it and tackle that in the future.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28332


crspeller (approved):

cpoile (approved):

cpoile:
I'd recommend to hold off merging this with the sql-merge branch, and use "rebase --onto" once we have the sql-merge branch merged with master.

agarciamontoro:
> I'd recommend to hold off merging this with the sql-merge branch, and use "rebase --onto" once we have the sql-merge branch merged with master.

I'm curious why. Is there any conflict you want to avoid?

agarciamontoro:
Oops, messed up the rebase (and that's why you should never force push anything). I'll see if I can open a new one, I still have the commits.

PR #281, opened 2020-09-16
MM-28683: Make the SQL store the default, remove all usage of incident/playbook KV stores

agarciamontoro:
#### Summary
The change was a bit more involved than expected, as I had to adapt the API tests, which were mocking the incident/playbook KV store calls directly. I changed them so they only mock the incident/playbook service, which is what they really use.

We cannot get rid of all the pluginkvstore directory because the subscriptions service is still using a KV store as backend. I filed a ticket to change this: https://mattermost.atlassian.net/browse/MM-28732

Also, I propose to remove the incident and playbook KV store files *after* we make The Big SQL Merge. If we ever want to get that code back, it will be much easier to revert an isolated commit that does not change, for example, the tests, as this PR does.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28683


cpoile (approved):

cpoile:
> I propose to remove the incident and playbook KV store files after we make The Big SQL Merge.
That sounds like a good plan.

crspeller (approved):

cpoile:
@agarciamontoro  
After working on the other PRs that just got merged -- I think there are a few other changes this PR needs:
1. All references to `HasPermissionsTo` in the options are not needed anymore, so that can go, and anything that touches that.
1. If we need the kvstore for the event subscription, we should still be able to remove all the code from all the other kv stores
1. All the tests too. (I think there were some still left -- but apologies if there aren't, I can't tell anymore since the branch this was based on was deleted -- sorry!)

agarciamontoro:
@cpoile 
> 1. All references to `HasPermissionsTo` in the options are not needed anymore, so that can go, and anything that touches that.

Agree!

> 2. If we need the kvstore for the event subscription, we should still be able to remove all the code from all the other kv stores
> 3. All the tests too. (I think there were some still left -- but apologies if there aren't, I can't tell anymore since the branch this was based on was deleted -- sorry!)

Yes, we *can* get rid of it. But I was hoping to have that in a different PR, so we can easily revert the commit that removes all the code (in case we want it back).

agarciamontoro:
Sorry for the force-pushed, but this needed a `rebase` onto the new master

agarciamontoro:
Merging this now and tackling the permissions and removal of the KV code as part of https://mattermost.atlassian.net/browse/MM-28816. Removing the permissions here would be impossible without removing or modifying the legacy code.

PR #282, opened 2020-09-16
Turn on the sql

cpoile:
#### Summary
- Flip the switch!  (community will start to use sql when this gets merged to master and we cut 0.7.0 alpha)


cpoile:
Adding this directly to #278 -- looks like the `IR_USE_SQL=false` option stopped working after the merge, and since we're going to remove all that code anyway there's no point in trying to fix it.

PR #284, opened 2020-09-17
MM-28332: Add a fallback for when unaccent is not available (now ready to merge)

agarciamontoro:
#### Summary

A rebased copy of #280. The original summary:

- I've changed the migration so we don't finish it when we are unable to create the extension, but only log the error. Also, in case the extension fails to be created, I've added a fallback so we don't use it (I considered implementing a noop SQL function so we don't need to check whether the real function exists, but we would have the same problem: does the user has permissions to add a function?)
- I've smoke-tested that the general Mattermost search is not affected by this change. Also, I'm 99% sure it won't affect anything because all the extension does is adding a dictionary and a function, which you need to explicitly call to get the behavior.
- Also, I was testing this mostly with the ñ letter, and after all was implemented, I tried to create an incident with letters with accents (áéíóú) and it turns out we cannot create incidents with those letters :sob: (my gut feeling is because of channel names restrictions). We may want to look into that, of course, but as this work was already done, I think it's better to merge it and tackle that in the future.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-28332

### DIFF
--- a/server/sqlstore/incident.go
+++ b/server/sqlstore/incident.go
@@ -95,12 +95,9 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
 		// Postgres performs a case-sensitive search, so we need to lowercase
 		// both the column contents and the search string
 		if s.store.db.DriverName() == model.DATABASE_DRIVER_POSTGRES {
-			var unaccentExists bool
-			err := s.store.getBuilder(s.store.db, &unaccentExists,
-				sq.Expr("SELECT EXISTS(SELECT * FROM pg_proc WHERE proname = 'unaccent')"))
+			unaccentExists, err := s.store.isUnaccentAvailable()
 			if err != nil {
 				s.store.log.Errorf("failed to check if the unaccent function exists: %w", err)
-				unaccentExists = false
 			}
 
 			if unaccentExists {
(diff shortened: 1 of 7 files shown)


agarciamontoro:
Sorry for the re-request of the review, @crspeller and @cpoile. But I messed up the rebase in the other PR and, although the changes are minimal, wanted your re-approval :)

server/sqlstore/incident.go
```diff
@@ -95,8 +95,21 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
 		// Postgres performs a case-sensitive search, so we need to lowercase
 		// both the column contents and the search string
 		if s.store.db.DriverName() == model.DATABASE_DRIVER_POSTGRES {
-			column = "LOWER(UNACCENT(Name))"
-			searchString = normalize(options.SearchTerm)
+			var unaccentExists bool
+			err := s.store.getBuilder(s.store.db, &unaccentExists,
+				sq.Expr("SELECT EXISTS(SELECT * FROM pg_proc WHERE proname = 'unaccent')"))
```
lieut-data, line 100: Should we cache this?

cpoile: +1

cpoile (approved):
👍 to Jesse's comment, otherwise looks great!

crspeller (approved):

agarciamontoro:
@lieut-data, not sure if using the `sqlStore` struct is the best place to put the cached value. Any other ideas?

lieut-data (approved):
A work of beauty!

PR #286, opened 2020-09-17
currentChannelId vs currentChannel.Id

lieut-data:
#### Summary
`currentChannel` from the Redux store is changing unexpectedly, even if the channel itself is not changing -- depend on `currentChannel.Id` only instead to avoid flashing the RHS until the webapp is fixed.

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-28825


lieut-data:
I'll plan to release this as an updated alpha today.

crspeller (approved):

cpoile (approved):

PR #287, opened 2020-09-17
MM-28038: reset pagination on filtering

lieut-data:
#### Summary
When changing the filter settings on the incident list, reset pagination to avoid being stuck on a second page without any results.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-28038


crspeller (approved):

agarciamontoro (approved):
Neat :tada:

PR #291, opened 2020-09-21
MM-28285: /incident list

agarciamontoro:
#### Summary

This PR adds a new slash command: `/incident list`. It lists the ongoing incidents of the current team, ordered by creation date (most recent first).

(I also needed to do a teeny tiny refactor in the `Ephemeral` function from bot.Poster, which is isolated in the first commit)

As predicted on the ticket, the final style (see the screenshots below) is quite different from the original designs:
- I used [Message Attachments](https://docs.mattermost.com/developer/message-attachments.html) to differentiate each incident. Styling with raw markdown was far from perfect, and the rendering of the message attachments (different in mobile and in the webapp) is more similar to the designs. The biggest difference is that each incident is enclosed within a box whose left side is coloured. But I do like it.
- As each incident is enclosed in a box, I did not add the horizontal lines separating them.
- The interactive button cannot be used, as it is designed to send data to the plugin, not to go to another location. Instead, the link to the channel is embedded in the incident name, which I styled as a title to make it easier to tap into it on mobile. I thought about adding a separate link with the text `Go To Channel`, but I guess we don't want to waste vertical spacing here.
- The commander field does not contain the avatar image.

A few more decisions worth pointing out:
- I am listing 10 incidents max. I did not add any way of requesting more, but I don't see the use case for being on your phone and browsing more than 10 incidents in this UI. If you think we should add that (or increase the 10 limit), please say so!
- I am rendering the same thing in mobile and webapp. The original goal was to list the incidents as an ephemeral message for mobile and let the webapp open the RHS. When we implement that feature, we can make the distinction.

| Mobile 	| Web 	|
|-:	|-:	|
|image|image|

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28285


asaadmahmood (approved):

server/bot/poster.go
```diff
@@ -36,12 +36,10 @@ func (b *Bot) DMWithAttachments(userID string, attachments ...*model.SlackAttach
 }
 
 // Ephemeral sends an ephemeral message to a user
-func (b *Bot) Ephemeral(userID, channelID, format string, args ...interface{}) {
-	post := &model.Post{
-		UserId:    b.botUserID,
-		ChannelId: channelID,
-		Message:   fmt.Sprintf(format, args...),
-	}
+func (b *Bot) EphemeralPost(userID, channelID string, post *model.Post) {
```
lieut-data, line 39: 0/5, but at this point I'd be inclined to just use the API directly and abandon the bot interface.

server/command/command.go
```diff
@@ -225,6 +233,90 @@ func (r *Runner) actionAnnounce(args []string) {
...
+	}
+
+	requesterInfo := incident.RequesterInfo{
+		UserID:              r.args.UserId,
+		TeamID:              r.args.TeamId,
+		UserIDtoIsAdmin:     map[string]bool{r.args.UserId: permissions.IsAdmin(r.args.UserId, r.pluginAPI)},
+		TeamIDtoCanViewTeam: map[string]bool{r.args.TeamId: permissions.CanViewTeam(r.args.UserId, r.args.TeamId, r.pluginAPI)},
+	}
+
+	options := incident.HeaderFilterOptions{
```
lieut-data, line 250: Will this list all incidents, or only incidents of which the current user is a part? @asaadmahmood, can you also confirm intent, noting that the only material difference would be public incidents.

server/command/command.go
```diff
@@ -225,6 +233,90 @@ func (r *Runner) actionAnnounce(args []string) {
...
+		TeamID:  r.args.TeamId,
+		PerPage: 10,
+		Sort:    incident.SortByCreateAt,
+		Order:   incident.OrderDesc,
+		Status:  incident.Ongoing,
+	}
+
+	result, err := r.incidentService.GetIncidents(requesterInfo, options)
+	if err != nil {
+		r.postCommandResponse(fmt.Sprintf("Error retrieving the incidents: %v", err))
```
lieut-data, line 260: nit: This pattern seems to come up often. What about a `postCommandResponseError` to make things a bit more explicit?

lieut-data: Actually, should we necessarily be posting the error details directly as a command response? For a non system-admin, this could theoretically leak too much info (i.e. "Postgres database down...")

server/command/command.go
```diff
@@ -225,6 +233,90 @@ func (r *Runner) actionAnnounce(args []string) {
...
+			return
+		}
+
+		channel, err := r.pluginAPI.Channel.Get(theIncident.ChannelID)
+		if err != nil {
+			r.postCommandResponse(fmt.Sprintf("Error retrieving channel of incident '%s': %v", theIncident.Name, err))
+			return
+		}
+
+		duration := now.Sub(time.Unix(0, theIncident.CreateAt*1000000)).Round(time.Second)
```
lieut-data, line 290: Could we factor this out into a helper method that's possibly unit tested separately?

server/command/command.go
```diff
@@ -225,6 +233,90 @@ func (r *Runner) actionAnnounce(args []string) {
...
+
+		durationStr := duration.String()
+		if duration.Hours() > 23 {
+			days := duration / (24 * time.Hour)
+			duration %= 24 * time.Hour
+
+			durationStr = fmt.Sprintf("%dd%s", days, duration.String())
+		}
+
+		attachments[i] = &model.SlackAttachment{
```
lieut-data, line 300: Thoughts on separate posts vs. separate attachments? It might avoid the "long message" having to be expanded.

server/command/command.go
```diff
@@ -225,6 +233,90 @@ func (r *Runner) actionAnnounce(args []string) {
...
+			Text: fmt.Sprintf("### [%s](/%s/channels/%s)", channel.DisplayName, team.Name, channel.Name),
```
lieut-data, line 301: Is it possible to use a `~` reference instead? I think this might force a page refresh (not sure), but also has the problem of not handling subpaths (i.e. `/mattermost/<team_name>/channels/<channel name>`).

lieut-data (changes requested):
Nice, pragmatic approach! A handful of comments, but not all blocking.

server/command/command.go
cpoile: Maybe: "Your request could not be completed. Check the system logs for more information." And then log the error with details.

If you made this a function, like `r.ReturnMsgAndLogError(err)`, then you can use it for all the errors below, too.

server/command/command.go
```diff
@@ -225,6 +233,90 @@ func (r *Runner) actionAnnounce(args []string) {
...
+	requesterInfo := incident.RequesterInfo{
+		UserID:              r.args.UserId,
+		TeamID:              r.args.TeamId,
+		UserIDtoIsAdmin:     map[string]bool{r.args.UserId: permissions.IsAdmin(r.args.UserId, r.pluginAPI)},
+		TeamIDtoCanViewTeam: map[string]bool{r.args.TeamId: permissions.CanViewTeam(r.args.UserId, r.args.TeamId, r.pluginAPI)},
+	}
+
+	options := incident.HeaderFilterOptions{
+		TeamID:  r.args.TeamId,
+		PerPage: 10,
```
cpoile, line 252: Re: your question about max 10 incidents in the response. It might be useful to offer a paged response, with a next and previous page link, like in the backstage incident list. What do you think @asaadmahmood, @itao ?

server/command/command.go
```diff
@@ -225,6 +233,90 @@ func (r *Runner) actionAnnounce(args []string) {
...
+	if len(result.Items) == 0 {
+		message = "There are no ongoing incidents in **" + team.DisplayName + "** team."
+	}
+
+	now := time.Now()
+	attachments := make([]*model.SlackAttachment, len(result.Items))
+	for i, theIncident := range result.Items {
+		thePlaybook, err := r.playbookService.Get(theIncident.PlaybookID)
+		if err != nil {
+			r.postCommandResponse(fmt.Sprintf("Error retrieving playbook of incident '%s': %v", theIncident.Name, err))
```
cpoile, line 274: Here for example: `r.ReturnMsgAndLogError(err)`

cpoile (changes requested):
Nice work Alejandro! Some suggestions for the error reporting. Not sure Asaad and Tao want for the interface, but you could do paging if we need that.

server/bot/poster.go
agarciamontoro: I think encapsulating the bot details is good, so I'd say we keep the interface. We can abandon it later if we see there's no need for it anymore.

server/command/command.go
agarciamontoro: You're right! I added a `MembersIDs` field to the `HeaderFilterOptions` so we can filter by channel membership. I'd like you to take a look at the query here: https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/0a1b4c94837f9e821080d2918c873c524c792dfd

server/command/command.go
agarciamontoro: I followed your suggestion, Christopher: https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/de7ddecdad787535f76a356ea4796db5710dcd2b

server/command/command.go
agarciamontoro: I have a concern with the separate posts: as they are ephemeral, the user can dismiss them, so they may end up with what looks like a complete list that is, in fact, incomplete.

server/command/command.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/de7ddecdad787535f76a356ea4796db5710dcd2b

server/command/command.go
lieut-data: Ah, good point!

agarciamontoro:
> I realize we validate the incoming member list as IDs, but this is only loosely connected to protecting ourselves against a SQL injection attack. In general, I'm extremely hesitant anytime we find ourselves needing to fmt.Sprintf and run SQL, and recommend always finding a way to avoid same.
In this particular case, I feel like filtering by MemberIDs in the plural might be overkil -- can we just filter by a particular MemberID instead and trivially bind that parameter? It seems unlikely that we need to support queries for incidents of which a set of users are members.

Makes sense! (It was fun to build that query, though)

server/command/command.go
agarciamontoro: Ended up allowing filter by only one member. Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/87317454e93487727be365779cc28208b7d62166.

server/command/command.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/529f972d6dc1985caa88e95c2440aa7b08d40547.

server/command/command.go
agarciamontoro: - `Title: fmt.Sprintf("~%s", channel.Name)` does not render as a link in mobile, but works on webapp.
- `Text: fmt.Sprintf("~%s", channel.Name)` ends up in Not Found in webapp, but works on mobile.
- `Pretext: fmt.Sprintf("~%s, channel.Name)` works on both, but does force a page refresh on webapp. This changes the styling a bit, but I do like the change.

I ended up using the latter. I took a real quick look at the webapp code parsing the attachments but saw nothing completely obvious. We can use the `Pretext` field and investigate the bug later.

Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/1b25f2c639a3aa32c27324fbe92b02ee1a8b68d5.

server/command/command.go
agarciamontoro: We decided to keep the maximum of 10 incidents in the meeting.

agarciamontoro:
Last commit changed a bit the styling. The options boxes are now preceded by the incident name with the link (that is now rendered with a tilde, so it may be more obvious that it is indeed a link to a channel):

image

lieut-data (approved):

asaadmahmood (approved):
I think the link in the header is fine for now, let us conduct some tests and see how users are interacting with this.

agarciamontoro:
I've added the `ActiveStageTitle` to the incident struct in https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/6ee4ea2ae151618cc34c7ae101a2d5b93099f0d9. I ended up needing to unmarshal the checklists for each incident in the `sqlStore.GetIncidents` method. But the return value is unchanged (except for the ActiveStageTitle, of course), as I am not adding the checklists to each individual incident (see `toIncidentWithoutChecklists`)
Please take a look, @lieut-data and @cpoile, thanks!

server/sqlstore/incident.go
```diff
@@ -397,6 +418,33 @@ func toIncident(rawIncident sqlIncident) (*incident.Incident, error) {
 	if err := json.Unmarshal(rawIncident.ChecklistsJSON, &i.Checklists); err != nil {
 		return nil, errors.Wrapf(err, "failed to unmarshal checklists json for incident id: '%s'", rawIncident.ID)
 	}
+
+	numChecklists := len(i.Checklists)
+	if numChecklists != 0 {
+		if i.ActiveStage < 0 || i.ActiveStage >= len(i.Checklists) {
+			return nil, errors.Errorf("invalid ActiveStage in incident '%s': index is %d, but incident has %d checklists", rawIncident.ID, i.ActiveStage, len(i.Checklists))
```
lieut-data, line 425: Should we permanently fail to return the incident when this happens? Wondering if we just log a warning instead, and leave the title blank?

lieut-data (approved):
Nice!

server/sqlstore/incident.go
agarciamontoro: If this happens there's something completely wrong, so I'd say it's better to error as soon as possible.

lieut-data (approved):

agarciamontoro:
Waiting for v0.7.0 to be released before we merge this PR.

agarciamontoro:
Feel free to re-review when you have time, @cpoile, 0.7.0 is already tagged.

server/command/command.go
```diff
@@ -124,6 +124,13 @@ func (r *Runner) postCommandResponse(text string) {
 	r.poster.EphemeralPost(r.args.UserId, r.args.ChannelId, post)
 }
 
+func (r *Runner) warnUserAndLogErrorf(format string, args ...interface{}) {
+	r.logger.Errorf(format, args...)
+	r.poster.EphemeralPost(r.args.UserId, r.args.ChannelId, &model.Post{
+		Message: "Your request could not be completed. Check the system logs for more information.",
+	})
+}
```
cpoile: Nice!

server/command/command.go
cpoile: This is good -- it's similar to something I did for the new RHS list, unfortunately. :) I'll have to deal with the merge later.

server/command/command_test.go
```diff
@@ -0,0 +1,165 @@
...
+		expected, err := time.Parse(layout, layout)
+		require.NoError(t, err)
+
+		actual := getTimeForMillis(unixTime * 1000)
+
+		require.Equal(t, expected.Unix(), actual.Unix())
+	})
+}
+
+func TestDurationString(t *testing.T) {
```
cpoile, line 46: Nice tests!

server/sqlstore/incident.go
```diff
@@ -36,7 +36,7 @@ var _ incident.Store = (*incidentStore)(nil)
 func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLStore) incident.Store {
 	incidentSelect := sqlStore.builder.
 		Select("ID", "Name", "Description", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
-			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "PostID", "PlaybookID").
+			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "PostID", "PlaybookID", "ChecklistsJSON").
```
cpoile, line 39: Is this supposed to be in here? I don't think we should retrieve `ChecklistsJSON` in the standard `incidentSelect`

server/sqlstore/incident.go
```diff
@@ -101,11 +114,20 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
 
 	queryForResults = queryForResults.OrderBy(fmt.Sprintf("%s %s", options.Sort, options.Order))
 
-	var incidents []incident.Incident
-	if err := s.store.selectBuilder(s.store.db, &incidents, queryForResults); err != nil {
+	var rawIncidents []sqlIncident
+	if err := s.store.selectBuilder(s.store.db, &rawIncidents, queryForResults); err != nil {
```
cpoile, line 118: Why do we need to include checklists in `GetIncidents`?

server/sqlstore/incident.go
```diff
@@ -383,5 +401,32 @@ func toIncident(rawIncident sqlIncident) (*incident.Incident, error) {
...
+
+	numChecklists := len(i.Checklists)
+	if numChecklists != 0 {
+		if i.ActiveStage < 0 || i.ActiveStage >= len(i.Checklists) {
+			return nil, errors.Errorf("invalid ActiveStage in incident '%s': index is %d, but incident has %d checklists", rawIncident.ID, i.ActiveStage, len(i.Checklists))
+		}
+		i.ActiveStageTitle = i.Checklists[i.ActiveStage].Title
+	}
+
+	return &i, nil
```
cpoile, line 413: I think we might be filling in the `ActiveStageTitle` in the wrong place. If we keep it updated whenever we create and update an incident, then it will be current. And we can go back to the way `GetIncidents` was before (with no unmarshaling of `ChecklistsJSON`). 

That will save a lot of processing Checklists (since we call `GetIncidents` often, but we only need the active stage for the Incident List command).

cpoile (changes requested):
Nice work Alejandro! I have a concern about where we update the ActiveStage title. Otherwise, nicely done.

server/command/command.go
agarciamontoro: Thanks!

server/command/command_test.go
agarciamontoro: :tada:

server/sqlstore/incident.go
agarciamontoro: @cpoile How else could we retrieve the active stage title if we don't have the checklists? Or was your suggestion adding a new column to the incident table?

server/sqlstore/migrations.go
```diff
@@ -180,9 +183,65 @@ var migrations = []Migration{
...
+
+				incidentUpdate := sqlStore.builder.
+					Update("IR_Incident").
+					Set("ActiveStageTitle", checklists[theIncident.ActiveStage].Title).
+					Where(sq.Eq{"ID": theIncident.ID})
+
+				if _, err := sqlStore.execBuilder(e, incidentUpdate); err != nil {
+					return errors.Errorf("failed updating the ActiveStageTitle field of incident '%s'", theIncident.ID)
+				}
+			}
```
agarciamontoro, line 242: I decided to do an update per incident for two reasons:
- building an update for multiple rows it's different for postgresql and mysql, and I did not find it as straightforward as it should be in any of them
- the number of incidents any server can have at this point is still low, so I don't see there's any performance concern.

agarciamontoro:
After an offline conversation with @cpoile I solved my misunderstanding, as I was not aware that the proposal was to add a new column to the database! My last commit changes the implementation of the `ActiveStageTitle` field in order to move it inside the `IR_Incident` table, as it was first intended. Christopher, please, take a look into it and see if I missed something :)

server/sqlstore/migrations.go
```diff
@@ -180,9 +183,65 @@ var migrations = []Migration{
...
+					return errors.Wrapf(err, "failed to unmarshal checklists json for incident id: '%s'", theIncident.ID)
+				}
+
+				numChecklists := len(checklists)
+				if numChecklists == 0 {
+					continue
+				}
+
+				if theIncident.ActiveStage < 0 || theIncident.ActiveStage >= numChecklists {
+					return errors.Errorf("index %d out of bounds: incident %s has %d stages", theIncident.ActiveStage, theIncident.ID, numChecklists)
```
lieut-data, line 231: nit: /if/ this every fails, the plugin is dead in the water, requiring an upgrade to run the new version. Should we just fall back to an empty string here, with a warning?

lieut-data (approved):

server/sqlstore/migrations.go
agarciamontoro: Good catch! Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/cd28c42487068430ced220c8f5c80444412a99ac.

lieut-data (approved):

server/sqlstore/incident.go
```diff
@@ -226,8 +227,12 @@ func (s *incidentStore) GetIncident(incidentID string) (*incident.Incident, erro
 		return nil, errors.New("ID cannot be empty")
 	}
 
+	withChecklistsSelect := s.incidentSelect.
+		Columns("ChecklistsJSON").
+		From("IR_Incident")
```
cpoile: I don't think you need to add the `From` clause (it's in the `s.incidentSelect`), and if we do this I'm worried we'll lose the `AS incident` naming (if it's needed -- I can't remember).

server/sqlstore/incident.go
```diff
@@ -194,6 +190,10 @@ func (s *incidentStore) UpdateIncident(newIncident *incident.Incident) error {
 		return errors.New("ID should not be empty")
 	}
 
+	if err := syncActiveStageTitle(newIncident); err != nil {
+		return err
+	}
+
```
cpoile, line 196: I /think/ we might be able to do this simpler in `incident/service.go` line 515 -- set the active stage title when we change the active stage. Then we don't have to have the `syncActiveStageTitle` method, and we don't have to do bounds checking (it's done already in the `ChangeActiveStage` method). I'm prettttttty sure that's the only place it's possible to change the active stage, can you double check? (If you can set it directly through the API, then maybe this is better.) Let me know what you think is better and we'll go with that.

cpoile (changes requested):
Nice!

server/sqlstore/incident.go
agarciamontoro: Hm, I just brought back what was there, I'll test whether we need it or not, but it looks like we don't.

server/sqlstore/incident.go
agarciamontoro: I spent a bit of time trying to decide where to put this, and decided to do it here to be as close to the real data as possible. I'm afraid we call this method directly, wanting it to sync, and later discovering it does not. But I like the fact that we don't need to do bounds checking (it's a bit annoying) so I'll move it to the service and be explicit on the documentation. I'm also pretty sure it's the only place, I double checked, but I'll triple check!
We can do it directly through the API, but we would need to repeat the bounds checking, then. I think the service is the least evil option.

server/sqlstore/incident.go
agarciamontoro: You were right, if we leave the `From` clause, it removes the `AS incident` part. The alias is not needed here, but neither is this additional line. Removed in https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/9b69d5881c74bdee8614908098d369776c0de28b.

server/sqlstore/incident.go
agarciamontoro: Moved all syncing to the service (it turns out I had already added it to `ChangeActiveStage`, too). See https://github.com/mattermost/mattermost-plugin-incident-response/pull/291/commits/b63da87497657f664cb83d47ad2c55b1f89adeb5.

lieut-data (approved):

cpoile (approved):
Awesome! Thanks @agarciamontoro, apologies for the churn!

agarciamontoro:
> Apologies for the churn!

No problem at all! Thank you for a thorough awesome review :)

PR #293, opened 2020-09-23
/incident commander

lieut-data:
#### Summary
Support `/incident commander` to show the current commander:
image

and `/incident commander @username` to change it:
image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-28287
Relates-to: https://github.com/mattermost/mattermost-plugin-api/pull/75


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/google/go-querystring v1.0.0
 	github.com/gorilla/mux v1.7.4
 	github.com/jmoiron/sqlx v1.2.0
-	github.com/mattermost/mattermost-plugin-api v0.0.12-0.20200908141642-fbaca40730c4
+	github.com/mattermost/mattermost-plugin-api v0.0.12-0.20200923151715-5a2508c31e48
```
lieut-data, line 12: This is related to https://github.com/mattermost/mattermost-plugin-api/pull/75.

server/command/command.go
```diff
@@ -184,6 +190,91 @@ func (r *Runner) actionCheck(args []string) {
...
+	}
+}
+
+func (r *Runner) actionShowCommander([]string) {
+	incidentID, err := r.incidentService.GetIncidentIDForChannel(r.args.ChannelId)
+	if errors.Is(err, incident.ErrNotFound) {
+		r.postCommandResponse("This is not an incident channel.")
+		return
+	} else if err != nil {
+		r.postCommandResponse(fmt.Sprintf("Error retrieving incident: %v", err))
```
lieut-data, line 210: I'll probably leverage your changes, @agarciamontoro, to not expose errors directly, resolving merge conflicts along the way.

tests-e2e/cypress/fixtures/theme.json
```diff
@@ -0,0 +1,28 @@
+{
```
lieut-data, line 1: Copying stuff from the webapp.

server/command/command.go
```diff
@@ -184,6 +190,91 @@ func (r *Runner) actionCheck(args []string) {
...
+	if err != nil {
+		r.postCommandResponse(fmt.Sprintf("Error retrieving commander user: %v", err))
+		return
+	}
+
+	r.postCommandResponse(fmt.Sprintf("**@%s** is the current commander for **%s**", commanderUser.Username, currentIncident.Name))
+}
+
+func (r *Runner) actionChangeCommander(args []string) {
+	targetCommanderUsername := strings.TrimLeft(args[0], "@")
```
cpoile, line 230: Are all leading and trailing whitespace removed before sending arguments?

server/command/command.go
```diff
@@ -184,6 +190,91 @@ func (r *Runner) actionCheck(args []string) {
...
+	_, err = r.pluginAPI.Channel.GetMember(r.args.ChannelId, targetCommanderUser.Id)
+	if errors.Is(err, pluginapi.ErrNotFound) {
+		r.postCommandResponse(fmt.Sprintf("User @%s must be part of this channel to make them commander.", targetCommanderUsername))
+		return
+	} else if err != nil {
+		r.postCommandResponse(fmt.Sprintf("Failed to find user @%s as channel member: %v", targetCommanderUsername, err))
+		return
+	}
+
+	currentIncident.CommanderUserID = targetCommanderUser.Id
```
cpoile, line 270: Is this needed?

tests-e2e/cypress/integration/frontstage/slash_command_spec.js
```diff
@@ -0,0 +1,210 @@
...
+        cy.viewport('macbook-13');
+
+        // # Login as user-1
+        cy.apiLogin('user-1');
+
+        // # Reset the commander to test-1 as necessary.
+        cy.apiChangeIncidentCommander(incidentId, userId);
+    });
+
+    describe('/incident commander', () => {
```
cpoile, line 65: awesome tests!

cpoile (approved):

server/command/command.go
```diff
@@ -71,6 +72,11 @@ func getAutocompleteData() *model.AutocompleteData {
 		"api/v1/incidents/checklist-autocomplete", true)
 	slashIncident.AddCommand(checklist)
 
+	commander := model.NewAutocompleteData("commander", "[@username]",
+		"Show or change the current commander")
+	commander.AddTextArgument("The desired commander.", "[@username]", "")
```
agarciamontoro, line 77: Can we be super explicit here? Having the same command to show current commander and to set a new one can be a bit dangerous. Maybe we can just say "The desired new commander." Otherwise, I can think of the command `/incident commander @user` as a way of showing info of the commander `@user`

server/command/command.go
```diff
@@ -184,6 +190,91 @@ func (r *Runner) actionCheck(args []string) {
...
 }
 
+func (r *Runner) actionCommander(args []string) {
+	switch len(args) {
+	case 0:
+		r.actionShowCommander(args)
+	case 1:
+		r.actionChangeCommander(args)
+	default:
+		r.postCommandResponse(helpText)
```
agarciamontoro, line 200: nit: is it possible to prepend a specific error here for informing the user what happened? Something like "`/incident commander` expects one argument at most`? Maybe this is not how it is handled elsewhere, but I think this command UI should be polished for being useful.

server/command/command.go
```diff
@@ -184,6 +190,91 @@ func (r *Runner) actionCheck(args []string) {
...
+		return
+	}
+
+	commanderUser, err := r.pluginAPI.User.Get(currentIncident.CommanderUserID)
+	if err != nil {
+		r.postCommandResponse(fmt.Sprintf("Error retrieving commander user: %v", err))
+		return
+	}
+
+	r.postCommandResponse(fmt.Sprintf("**@%s** is the current commander for **%s**", commanderUser.Username, currentIncident.Name))
```
agarciamontoro, line 226: I've seen this formatting misbehave when implementing `/incident list`: as the incident name may have trailing spaces, we may end up with the asterisks in the text itself, which will be no longer bold. Can we trim `currentIncident.Name` here to make sure the format is ok?

server/command/command.go
```diff
@@ -184,6 +190,91 @@ func (r *Runner) actionCheck(args []string) {
...
+		r.actionChangeCommander(args)
+	default:
+		r.postCommandResponse(helpText)
+	}
+}
+
+func (r *Runner) actionShowCommander([]string) {
+	incidentID, err := r.incidentService.GetIncidentIDForChannel(r.args.ChannelId)
+	if errors.Is(err, incident.ErrNotFound) {
+		r.postCommandResponse("This is not an incident channel.")
```
agarciamontoro, line 207: Nit: this kind of error (also repeated in `actionChangeCommander`) is shown elsewhere as:
- `You can only check/uncheck an item from within the incident's channel`
- `You can only announce from within the incident's channel.`
- `You can only end an incident from within the incident's channel.`
- `You can only restart an incident from within the incident's channel.`

Maybe we should stick to that format (or change all others to this one) for the sake of consistency.

tests-e2e/cypress/integration/frontstage/slash_command_spec.js
agarciamontoro: agree!

tests-e2e/cypress/integration/frontstage/slash_command_spec.js
```diff
@@ -0,0 +1,210 @@
...
+            it('when the username has an @-prefix', () => {
+                // # Run a slash command to change the current commander
+                cy.executeSlashCommand('/incident commander @sysadmin');
+
+                // * Verify the expected error message.
+                cy.verifyEphemeralMessage('User @sysadmin must be part of this channel to make them commander.');
+            });
+        });
+
+        describe('should show an message when the user is already the commander', () => {
```
agarciamontoro, line 150: Typo:
```suggestion
        describe('should show a message when the user is already the commander', () => {
```

agarciamontoro (changes requested):
Nice!!! I'll copy these tests to `/incident info`!

I left some suggestions below

server/command/command.go
lieut-data: Nope! Vestigial from when I thought I had to `UpdateIncident` to effect a commander change. Thanks :)
```suggestion
```

server/command/command.go
lieut-data: Indeed -- `args` here effectively comes from `strings.Fields` on the incoming slash command string, which handles this natively.

server/command/command.go
lieut-data: Good catch! Stripping markdown here.

lieut-data: Actually, it's probably easier just to say `this incident` here like the other error messages.

server/command/command.go
lieut-data: Excellent!

server/command/command.go
lieut-data: I like this much better -- but going to just show that message vs. prepending to the whole help text. Hopefully that's enough?

agarciamontoro: Yeah, I think the valuable part here is the specific message.

agarciamontoro (approved):
:tada:

lieut-data:
I'm going to force merge this on the belief that the linting error above is bogus. If not, I'll keep iterating.

PR #294, opened 2020-09-24
MM-29002: Rollback unaccent creation if it fails

agarciamontoro:
#### Summary
A failure in the creation of the unaccent extension should not abort the transaction. This commit creates a savepoint just before the attempt of creating it and, if that fails, it rolls back to it.
I also changed the `log.Errorf` for a `log.Warnf`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29002

### DIFF
--- a/server/sqlstore/migrations.go
+++ b/server/sqlstore/migrations.go
@@ -87,9 +87,21 @@ var migrations = []Migration{
 				}
 			} else {
 				if _, err := e.Exec(`
+					SAVEPOINT before_unaccent;
+				`); err != nil {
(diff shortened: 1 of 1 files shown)


lieut-data (approved):
This makes sense! I'm still worried that we even saw a permission error re: `CREATE EXTENSION`, which makes me think it's likely this will fail anyway for customers. Should we consider just ripping it out altogether?

agarciamontoro:
After an offline discussion with @lieut-data, we decided to close this PR in favour of a new one completely removing the extension.

PR #295, opened 2020-09-24
MM-29002: Remove unaccent extension

agarciamontoro:
#### Summary
This PR completely removes the support for the `unaccent` extension. I removed all the related code, which is always nice. Only weird thing here: as PostgreSQL and MySQL are now handling the search differently, I needed to reduce the tests to the lowest common denominator, which is no support for accent-insensitive searches.

We can split the tests in the future to actually test that MySQL has this feature, but I'm leaving in a bit and we need this merged today, so I decided to adapt the tests for what we really support in every case (mysql addition being a happy coincidence).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29002


lieut-data (approved):
Thanks, @agarciamontoro!

agarciamontoro:
Feel free to merge as soon as this is approved by everybody, I may not be around when that happens!

cpoile (approved):
I still think that if we need a case-insensitive search (@itao should consider it) then we should keep an `UnaccentedTitle` column -- super easy and not really resource intensive.

prapti (approved):
The issue is fixed in this PR: there is no `Unable to activate plugin` in the cloud server anymore (tested in the cloud server spun up from this PR since testing with the one generated by the cloud plugin was giving a lot of issues).
Approving.

PR #296, opened 2020-09-24
MM-28949 - Fix the GetPlaybooksForTeam when used for starting an incident

cpoile:
#### Summary
- Simplify the GetPlaybooksForTeams. We could have put the MemberOnly flag in the options, but I figured that should probably be kept for paging and searching options, the kind that are sent from the client, whereas RequesterInfo is more of the internal information. But I'm open to other suggestions. 
- Had a lot of trouble with the tests, that's what took half the day. Turns out I was modifying the original slice of playbooks, and then inserting them again into the next db, and that's why things were failing. It was a mess.
- Alternatively, we could get all memberIDs like in `GetPlaybooks` and keep the check in the Go code, but keeping it all in the store seems cleaner to me.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-28994

### DIFF
--- a/server/playbook/playbook.go
+++ b/server/playbook/playbook.go
@@ -73,7 +73,7 @@ type RequesterInfo struct {
 	TeamID          string
 	UserIDtoIsAdmin map[string]bool
 
- 	// MemberOnly filters playbooks to those for which UserId is a member
+	// MemberOnly filters playbooks to those for which UserId is a member
 	MemberOnly bool
 }
 
--- a/server/playbook/playbook.go
+++ b/server/playbook/playbook.go
@@ -73,7 +73,7 @@ type RequesterInfo struct {
 	TeamID          string
 	UserIDtoIsAdmin map[string]bool
 
-	// MemberOnly when true will return playbooks only if the requester is a member of those playbooks.
+ 	// MemberOnly filters playbooks to those for which UserId is a member
 	MemberOnly bool
 }
 
--- a/server/playbook/playbook.go
+++ b/server/playbook/playbook.go
@@ -72,6 +72,9 @@ type RequesterInfo struct {
 	UserID          string
 	TeamID          string
 	UserIDtoIsAdmin map[string]bool
+
+	// MemberOnly when true will return playbooks only if the requester is a member of those playbooks.
+	MemberOnly bool
 }
 
 // Service is the playbook service for managing playbooks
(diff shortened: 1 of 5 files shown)


server/playbook/playbook.go
```diff
@@ -72,6 +72,9 @@ type RequesterInfo struct {
 	UserID          string
 	TeamID          string
 	UserIDtoIsAdmin map[string]bool
+
+	// MemberOnly when true will return playbooks only if the requester is a member of those playbooks.
```
lieut-data, line 76: nit: comment suggestion
```suggestion
 	// MemberOnly filters playbooks to those for which UserId is a member
```

server/sqlstore/playbook.go
```diff
@@ -152,28 +151,34 @@ func (p *playbookStore) Get(id string) (out playbook.Playbook, err error) {
...
+	var out []playbook.Playbook
+	err = p.store.selectBuilder(tx, &out, p.store.builder.
+		Select("ID", "Title", "Description", "TeamID", "CreatePublicIncident", "CreateAt",
+			"DeleteAt", "NumStages", "NumSteps").
+		From("IR_Playbook AS p").
+		Where(sq.Eq{"DeleteAt": 0}))
+
 	if err == sql.ErrNoRows {
-		return out, errors.Wrap(playbook.ErrNotFound, "no playbooks found")
+		return nil, errors.Wrap(playbook.ErrNotFound, "no playbooks found")
```
lieut-data, line 169: Apologies if I've asked this in a previous PR and lost the context, but why is this an error?
```suggestion
return nil, nil
```

lieut-data (approved):
Nice simplification!

lieut-data:
(Sending this to @crspeller on the chance we can merge today.)

server/sqlstore/playbook.go
cpoile: It was for code that was doing logic when it received ErrNotFound. But really, we don't even use this function, except in the self-test and the testing code, so we can basically remove it if you want. 
I prefer to keep it -- it caught a nasty (self-inflicted) bug today, since it's the only place where multiple playbooks are returned with memberIDs.

server/sqlstore/playbook.go
```diff
@@ -263,7 +239,7 @@ func (p *playbookStore) GetPlaybooksForTeam(requesterInfo playbook.RequesterInfo
 }
 
 func (p *playbookStore) buildPermissionsExpr(info playbook.RequesterInfo) sq.Sqlizer {
-	if info.UserIDtoIsAdmin[info.UserID] {
+	if info.UserIDtoIsAdmin[info.UserID] && !info.MemberOnly {
```
crspeller, line 242: Nit: This changes the what this function does. it does multiple things now, permissions and filtering. Maybe refactor or change name?

crspeller (approved):

server/sqlstore/playbook.go
cpoile: Yep, makes sense. Done.

PR #297, opened 2020-09-24
MM-29009 - Allow playbooks and incident stages with no checklists

cpoile:
#### Summary
- This was a bit more tricky than just removing the error. We are expecting there to always be checklist items for a checklist in the frontend. I cleared up the errors, so things work, but the backstage incident details checklist graph doesn't render if there are no items in the first checklist. I recommend splitting that off into another ticket. There might be other effects of this change. 
- I'm not sure this should go in last minute for 0.7.0, considering the possible unanticipated effects...
- I also fixed a bug where if the incident being sent over websockets didn't fit the typescript definition of an incident (e.g., there was a null instead of an array --- that again) the websocket event wouldn't be passed on and the frontend wouldn't update. So you can click a checkbox and nothing in the UI would change (but you would see the 'checklist item has been checked' message in the centre channel).

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29009

### DIFF
--- a/server/api/incidents.go
+++ b/server/api/incidents.go
@@ -357,6 +357,8 @@ func (h *IncidentHandler) getIncident(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	removeNulls(incidentToGet)
+
 	jsonBytes, err := json.Marshal(incidentToGet)
 	if err != nil {
 		HandleError(w, err)
@@ -387,6 +389,8 @@ func (h *IncidentHandler) getIncidentWithDetails(w http.ResponseWriter, r *http.
 		HandleError(w, err)
 		return
 	}
+
+	removeNulls(&incidentToGet.Incident)
 
 	jsonBytes, err := json.Marshal(incidentToGet)
 	if err != nil {
@@ -909,4 +913,16 @@ func parseIncidentsFilterOptions(u *url.URL) (*incident.HeaderFilterOptions, err
 		CommanderID: commanderID,
 		SearchTerm:  searchTerm,
 	}, nil
+}
+
+func removeNulls(theIncident *incident.Incident) {
+	// replace nils with empty slices for the frontend
+	if theIncident.Checklists == nil {
+		theIncident.Checklists = []playbook.Checklist{}
+	}
+	for i, cl := range theIncident.Checklists {
+		if cl.Items == nil {
+			theIncident.Checklists[i].Items = []playbook.ChecklistItem{}
+		}
+	}
 }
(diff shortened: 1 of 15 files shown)


webapp/src/websocket_events.ts
```diff
@@ -29,9 +29,9 @@ export function handleWebsocketIncidentUpdate() {
         if (!msg.data.payload) {
             return;
         }
-        const incident = JSON.parse(msg.data.payload);
+        let incident = JSON.parse(msg.data.payload);
         if (!isIncident(incident)) {
-            return;
+            incident = removeNulls(incident);
```
crspeller, line 34: What if it's still not an incident?

server/api/incidents.go
```diff
@@ -910,3 +914,15 @@ func parseIncidentsFilterOptions(u *url.URL) (*incident.HeaderFilterOptions, err
 		SearchTerm:  searchTerm,
 	}, nil
 }
+
+func removeNulls(theIncident *incident.Incident) {
```
crspeller, line 918: I wonder if it would be cleaner, instead of doing this here, to use a custom `MarshalJSON` for the incident and playbook types that does this.

crspeller (changes requested):

webapp/src/websocket_events.ts
cpoile: I'm not sure when that would happen -- unless we're sending bad data to the client? Either way, I don't think this `isIncident` thing does what it's supposed to do. It's supposed to help Typescript know what the object really is, so it can use that later. But it doesn't seem to do that for me. Does it do anything for you? For now I'm using it as `check that the arrays aren't null`.

server/api/incidents.go
cpoile: Maybe, but would that effort be worth it?

cpoile: I thought about it overnight, and you're right -- I'll wright a MarshalJSON and that will eliminate the need for the screwy checks and changes in the frontend. Thanks!

webapp/src/websocket_events.ts
cpoile: I decided to log an error to the console. This shouldn't happen, so if it does we should know about it and fix it (it probably means we're sending bad data from the backend).

cpoile:
Fixed the checklist timeline in the backstage. I've poked around, and I think everything works as expected now. 
Will need Prapti to do another smoke test to make sure there aren't any other side effects from allowing this change.

server/api/incidents.go
crspeller: Awesome 😄

crspeller (approved):
Looks good!

lieut-data (approved):
Really slick!

PR #298, opened 2020-09-25
MM-29021: Fix timestamp bug in data migration

agarciamontoro:
#### Summary

Before #274, the `ChecklistItem` structs had two `time.Time` fields: `StateModified` and `AssigneeModified`. Now, those timestamps are stored as `int64` fields that contain the number of milliseconds since the unix epoch.

When doing the data migration, we used [`model.GetMillisForTime`](https://github.com/mattermost/mattermost-server/blob/30bd506e7676033bdfb632f566e538d6cd3d8be0/model/utils.go#L179-L182) directly on the timestamps, but we never checked whether they had a zero value. It turns out that converting the zero value of a `time.Time` using that function returns a negative number. Specifically, the value `-6795364578871`. This value is then stored in all items that were never checked (in which case, `StateModified` has this value) or that never got an assignee (in which case, `AssigneeModified` has this value). Parsing that value in the client with `moment` returns an invalid date, but `moment('-6795364578871').isSameOrAfter('2020-01-01')` (a common check in the webapp) is always `false`, which is nice.

This PR fixes the data migration, just making sure that the zero value is never converted to milliseconds using the function, using an `int64` 0 instead. I refactored the loop in `checklistsToJSON` into a new function `oldChecklistToNewChecklist`, where the fix is applied. I added a test to make sure that this new function behaves as expected.

This bug is present only in the servers that have already migrated to the SQL store (probably, only ours). However, we need to make sure that we're not using these buggy timestamps anywhere. In the server code, we only use `StateModified` and `AssigneeModified` to change its value, never to read it (except for when we marshal it into JSON and send it to the client). In the client code, `assignee_modified` is never used (interesting) and we use the `state_modified` field in these cases:
- in [components/checklist_item.tsx](https://github.com/mattermost/mattermost-plugin-incident-response/blob/2049861337258fc77a9aaa42233653a389342d10/webapp/src/components/checklist_item.tsx#L203): inside the `ChecklistItemDetails` component, used to render the timestamp of checked items; it is guarded by a check that makes sure that the timestamp is after 2020-01-01, so we're safe.
- in [components/backstage/incidents/incident-detials/checklist_timeline.tsx](https://github.com/mattermost/mattermost-plugin-incident-response/blob/ee86e9eaede65f5da5acc13cebe17c3989717d11/webapp/src/components/backstage/incidents/incident_details/checklist_timeline.tsx#L240): inside the `initData` function; only parsed timestamps that are after 2020-01-01 are handled, so we're safe there.
- in [components/backstage/incidents/incident-detials/checklist_timeline.tsx](https://github.com/mattermost/mattermost-plugin-incident-response/blob/ee86e9eaede65f5da5acc13cebe17c3989717d11/webapp/src/components/backstage/incidents/incident_details/checklist_timeline.tsx#L177): inside the graph callback `tooltipLabel`, it is directly parsed with moment. This is the only one that is never checked, but I *believe* (I'm not 100% sure) that this label is only used for checked items, in which case the timestamp will be correct. The worst-case scenario here is that the graph with the details contain a label with 'Invalid date', in which case we'll get a bug report immediately (however, I have checked and saw no such timestamps in the graph).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29021

### DIFF
--- a/server/sqlstore/data_migration.go
+++ b/server/sqlstore/data_migration.go
@@ -253,31 +253,44 @@ func DataMigration(store *SQLStore, tx *sqlx.Tx, kvAPI KVAPI) error {
 	return nil
 }
 
+func oldChecklistToNewChecklist(oldChecklist oldChecklist) playbook.Checklist {
+	newItems := make([]playbook.ChecklistItem, len(oldChecklist.Items))
+
+	for j, oldItem := range oldChecklist.Items {
+		var newStateModified int64
+		if !oldItem.StateModified.IsZero() {
+			newStateModified = model.GetMillisForTime(oldItem.StateModified)
+		}
+		var newAssigneeModified int64
+		if !oldItem.AssigneeModified.IsZero() {
+			newAssigneeModified = model.GetMillisForTime(oldItem.AssigneeModified)
+		}
+
+		newItems[j] = playbook.ChecklistItem{
+			ID:                     model.NewId(),
+			Title:                  oldItem.Title,
+			State:                  oldItem.State,
+			StateModified:          newStateModified,
+			StateModifiedPostID:    oldItem.StateModifiedPostID,
+			AssigneeID:             oldItem.AssigneeID,
+			AssigneeModified:       newAssigneeModified,
+			AssigneeModifiedPostID: oldItem.AssigneeModifiedPostID,
+			Command:                oldItem.Command,
+			Description:            oldItem.Description,
+		}
+	}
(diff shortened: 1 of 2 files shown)


crspeller (approved):

lieut-data (approved):

lieut-data:
@prapti, I'm going to merge this one and we can finalize any required testing from `master`.

PR #300, opened 2020-09-29
Fix link in slash commands help text

agarciamontoro:
#### Summary
This PR fixes a dead link in the help text from the slash commands.

It was linking to https://mattermost.com/pl/default-incident-response-app-documentation as our documentation, but that resolves to a weirdly unstyled frontpage of the docs site. I changed it to what I *think* is the canonical location of the user-facing documentation, https://docs.mattermost.com/administration/devops-command-center.html, but I'm not entirely sure I am right.

#### Ticket Link
--

### DIFF
--- a/server/command/command.go
+++ b/server/command/command.go
@@ -24,7 +24,7 @@ const helpText = "###### Mattermost Incident Response Plugin - Slash Command Hel
 	"* `/incident check [checklist #] [item #]` - check/uncheck the checklist item. \n" +
 	"* `/incident announce ~[channels]` - Announce the currrent incident in other channels. \n" +
 	"\n" +
-	"Learn more [in our documentation](https://mattermost.com/pl/default-incident-response-app-documentation). \n" +
+	"Learn more [in our documentation](https://docs.mattermost.com/administration/devops-command-center.html). \n" +
 	""
 
 const confirmPrompt = "CONFIRM"


lieut-data:
@justinegeffen, I suspect the problem here is the permalink and not necessarily something we should fix in-product. Are you able to correct where that permalink lands?

justinegeffen:
Hmmm, that is interesting. That *should* work - the only issue I can think of is that it should end in ``html`` as per: https://handbook.mattermost.com/operations/research-and-development/product/product-management-team-handbook#frequently-asked-questions-faq. 

In the interim, I know @jasonblais has access to the admin panel so tagging him for confirmation.

jasonblais:
This is now fixed.

https://mattermost.com/pl/default-incident-response-app-documentation redirected to https://docs.mattermost.com/administration/incident-response-application.html, which no longer exists. I fixed this by changing the redirect to https://docs.mattermost.com/administration/devops-command-center.html in Wordpress.

@agarciamontoro for in-product documentation links, the practice we do is use a default permalink to a mattermost.com which then redirects to the appropriate page in the docs. This ensures that if the page moves (like in this case), we only need to fix the redirect and not require in-product changes. More about it in our [handbook page](https://handbook.mattermost.com/operations/research-and-development/product/product-management-team-handbook#how-do-i-create-redirects-for-in-product-documentation).

@justinegeffen I think the handbook page has an error, typically a website page URL doesn't end with `.html`. I've submitted a correction to it: https://github.com/mattermost/mattermost-handbook/pull/481

agarciamontoro:
Thank you everyone! And thank you for the links to the handbook :) Closing this PR.

PR #306, opened 2020-09-30
Add /incident info

agarciamontoro:
#### Summary

Unsurprisingly, this adds an `/incident info` slash command. I do have an E2E test, but it is based on the changes introduced on #293; as I want these changes to be merged before deploying to community tomorrow, I'm opening this PR without the test. I'll open a new one with the test tomorrow, when the commander slash command is merged.

- If not on an incident channel, it shows an ephemeral message `You can only show the details of an incident from within the incident's channel.`
- If on an incident channel:
  - If command executed from webapp, it just opens the RHS. If it was already open, it shows an ephemeral message `Your incident details are already open in the right hand side of the channel.`
  - If command executed from mobile, it shows the following, non-interactive, info:

image

### DIFF
--- a/server/command/command.go
+++ b/server/command/command.go
@@ -23,8 +23,9 @@ const helpText = "###### Mattermost Incident Response Plugin - Slash Command Hel
 	"* `/incident start` - Start a new incident. \n" +
 	"* `/incident end` - Close the incident of that channel. \n" +
 	"* `/incident check [checklist #] [item #]` - check/uncheck the checklist item. \n" +
-	"* `/incident announce ~[channels]` - Announce the currrent incident in other channels. \n" +
+	"* `/incident announce ~[channels]` - Announce the current incident in other channels. \n" +
 	"* `/incident list` - List all your incidents. \n" +
+	"* `/incident info` - Show a summary of the current incident. \n" +
 	"\n" +
 	"Learn more [in our documentation](https://mattermost.com/pl/default-incident-response-app-documentation). \n" +
 	""
@@ -45,7 +46,7 @@ func getCommand() *model.Command {
 		DisplayName:      "Incident",
 		Description:      "Incident Response Plugin",
 		AutoComplete:     true,
-		AutoCompleteDesc: "Available commands: start, end, restart, check, announce, list",
+		AutoCompleteDesc: "Available commands: start, end, restart, check, announce, list, info",
 		AutoCompleteHint: "[command]",
 		AutocompleteData: getAutocompleteData(),
 	}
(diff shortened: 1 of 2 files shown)


lieut-data (approved):

webapp/src/slash_command.ts
```diff
@@ -20,6 +22,17 @@ export function makeSlashCommandHook(store: Store<GlobalState>) {
             return Promise.resolve({message: messageTrimmed, args});
         }
 
+        if (messageTrimmed && messageTrimmed.startsWith('/incident info')) {
+            const state = store.getState();
+            const currentChannel = getCurrentChannel(state);
+
+            if (isIncidentChannel(state, currentChannel.id) && !isIncidentRHSOpen(state)) {
```
crspeller, line 29: I think you are missing the already open detection you mention in the PR description. 
But I also think this behavior is kinda strange. The RHS in an incident channel is always opened automatically. Why would you open it with a slash command? Maybe just show the info as text just like on mobile?

webapp/src/slash_command.ts
```diff
@@ -20,6 +22,17 @@ export function makeSlashCommandHook(store: Store<GlobalState>) {
...
+                store.dispatch(toggleRHS());
+            }
+
+            return Promise.resolve({message: messageTrimmed, args});
```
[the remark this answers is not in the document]
agarciamontoro: @lieut-data we still need to get to the server to send the message with our bot, right? I did not find any way to do so from the webapp.

webapp/src/slash_command.ts
agarciamontoro: > I think you are missing the already open detection you mention in the PR description.

The bot is sending the ephemeral message from the server. But it's true that it is posting it in any case that the user is in an incident channel, regardless of the previous state of the RHS.

> But I also think this behavior is kinda strange. The RHS in an incident channel is always opened automatically. Why would you open it with a slash command? Maybe just show the info as text just like on mobile?

The main motivation behind `/incident info` was being able to have a quick summary in mobile. The behaviour of the RHS is just a fallback in webapp. I prefer the current design, which shows the RHS, instead of showing a downgraded version of that information in a text message.

server/command/command.go
```diff
@@ -309,7 +313,77 @@ func (r *Runner) actionList() {
 			"attachments": attachments,
 		},
 	}
+	r.poster.EphemeralPost(r.args.UserId, r.args.ChannelId, post)
+}
+
+func (r *Runner) actionInfo() {
+	incidentID, err := r.incidentService.GetIncidentIDForChannel(r.args.ChannelId)
+	if err != nil {
+		if errors.Is(err, incident.ErrNotFound) {
```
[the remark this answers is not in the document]
agarciamontoro: Thanks! Done in https://github.com/mattermost/mattermost-plugin-incident-response/pull/306/commits/819154ec3367b0121d680c86a1a72e3258cda89d.

webapp/src/slash_command.ts
agarciamontoro: Removed the ephemeral message when the RHS is opened as a result of executing the command: https://github.com/mattermost/mattermost-plugin-incident-response/pull/306/commits/45c5afbd01ff027ac5a21298988476a93cf842d0.

lieut-data (approved):

crspeller (approved):

PR #307, opened 2020-09-30
MM-29136 - Fix Jump to Channel in backstage incident details

cpoile:
-- sorry, not working, one sec --


webapp/src/types/incident.ts
```diff
@@ -31,20 +37,48 @@ export interface FetchIncidentsReturn {
...
-        typeof arg.channel_id === 'string' &&
-        typeof arg.end_at === 'number' &&
+        arg.channel_id && typeof arg.channel_id === 'string' &&
         typeof arg.create_at === 'number' &&
-        optional &&
+        typeof arg.end_at === 'number' &&
+        typeof arg.delete_at === 'number' &&
+        typeof arg.active_stage === 'number' &&
+        typeof arg.active_stage_title === 'string' &&
+        typeof arg.post_id === 'string' &&
```
cpoile, line 54: Fun gotcha: since a blank string is falsey, writing `arg.post_id && typeof arg.post_id === 'string' &&` turned `post_id` into a field that needed to have text, and therefore the incident would fail this test. So we need to be careful with these.

PR #308, opened 2020-10-01
MM-29136 - Fix backstage jump to channel

cpoile:
#### Summary
- The bug was caused by our new MarshalJSON being called on an `incident.Details` struct, but only the embedded `incident.Incident` was being marhaled. I tried creating a MarshalJSON for incident.Details, but it kept calling the embedded incident.Incident's marshaller. Looks like this might be a bug in Go itself: https://github.com/golang/go/issues/31167 
- I worked around it by separating the embedded structs, calling each and splicing them together to create the full json. I think this is what we have to do, or else make a whole new struct for the details (not do the embedding) as Christopher suggested. In the end, this is what a marshaler does (byte slice manipulation), so it doesn't feel too hacky. Depends on what you guys think after seeing the code.
- This was not found by the api tests for get details... because the tests were wrong. Sorry about that. Tests are fixed now.

#### Retrospective
- This would have been caught if we were doing type guarding and surfacing the mismatches correct in the front end, but that had gone out of sync, and we were just casting raw json into the types we /thought/ we were getting from the backend.
- And in the backstage details we were casting an empty object into an `Incident`, so I don't know how any of that passed the type checker. We should probably go through and make sure types really are that type, and use the union null  to be explicit that the object might not be there.
- This PR fixes the problem for incident details, but it doesn't fix all the other similar problems we have in the frontend.
- I think we should chat about this in the Monday meeting, because it will be a bit of work, and will be work to maintain equivalence between the backend/frontend. I'd love opinions on cost/benefit of doing this for every type coming from the backend. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29136


webapp/src/client.ts
```diff
@@ -37,12 +44,26 @@ export async function fetchIncidents(params: FetchIncidentsParams) {
...
+
+    // Prefer to continue to run even though we may be missing a field.
+    return data as Incident;
 }
 
-export function fetchIncidentWithDetails(id: string) {
-    return doGet(`${apiUrl}/incidents/${id}/details`);
+export async function fetchIncidentWithDetails(id: string) {
+    const data = await doGet(`${apiUrl}/incidents/${id}/details`);
+    if (!isIncidentWithDetails(data)) {
```
lieut-data, line 60: I'm not convinced we want this in production code. Can we find a way to run this only for development?

webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
```diff
@@ -269,7 +268,7 @@ const BackstageIncidentDetails: FC = () => {
         );
     }
 
-    if (fetchingState === FetchingStateType.notFound) {
+    if (fetchingState === FetchingStateType.notFound || incident === null) {
```
lieut-data, line 271: Won't `incident` always be `null` when the component first renders, resulting in an unconditional `Redirect` before we have a chance to load the incident?

webapp/src/types/incident.ts
```diff
@@ -6,21 +6,27 @@ import {Checklist, isChecklist} from './playbook';
...
-    post_id?: string;
+    delete_at: number;
+    active_stage: number;
+    active_stage_title: string;
+    post_id: string;
     playbook_id: string;
     checklists: Checklist[];
+}
+
+export interface IncidentWithDetails extends Incident {
```
lieut-data, line 24: Hmm, I was expecting to see an object with fields `incident` and `details`, thus matching the backend type. Actually, this would eliminate the need to do the special `MarshalJSON` handling altogether. Thoughts?

lieut-data (changes requested):
Thanks, @cpoile! A couple of comments below (which might be stale, since I think I'm reviewing a slightly older set of changes with this review).

webapp/src/types/incident.ts
cpoile: Sorry, I don't know how it would prevent the need for the MarshalJSON, can you clarify?

webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
cpoile: That will be caught by the conditional above `if (fetchingState === FetchingStateType.loading)`.

webapp/src/types/incident.ts
lieut-data: I'd presume we'd just send down `{incident: {...}, details: {...}}`.

cpoile: Oh, the structure of what's sent down is fine--it's flat. We're doing the custom MarshalJSON to convert nil slices into empty slices.

server/incident/incident.go
```diff
@@ -64,16 +64,40 @@ type UpdateOptions struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	detailsJSON, err := json.Marshal(d.Details)
+	if err != nil {
+		return nil, err
+	}
+
+	combinedJSON := make([]byte, 0, len(incidentJSON)+len(detailsJSON)-1)
```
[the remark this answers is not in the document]
cpoile: Yep, it's nicer. Been awhile since I've used it, forgot about it. Thanks!

webapp/src/client.ts
cpoile: I'm using the client config from redux, but calling it the SystemSettings (since that's what it is--Harrison says it's misnamed anyway). I'm putting it in its own module so that we can get that info from non-react components (like the functions in client).

webapp/src/types/incident.ts
cpoile: Oh, no I see what you mean now...  Let me check that.

cpoile: Yep, it works. It leads to stuttering in the backstage details, but we can fix that if we want by calling `incident` something else.

webapp/src/client.ts
```diff
@@ -46,7 +47,7 @@ export async function fetchIncidents(params: FetchIncidentsParams) {
 
 export async function fetchIncident(id: string) {
     const data = await doGet(`${apiUrl}/incidents/${id}`);
-    if (!isIncident(data)) {
+    if (SystemSettings.EnableDeveloperMode && !isIncident(data)) {
```
lieut-data, line 50: Actually, instead of tying this to a server setting, can we just have webpack compile this out for production builds, that way everything is tied together with the same "concept" vs. having something else to enable to get testing? See the following for an example:

https://github.com/mattermost/mattermost-webapp/blob/f8e7fc90a138c1c7ea6e018fb7426de4104273d8/stores/redux_store.jsx#L17-L19

webapp/package.json
```diff
@@ -2,8 +2,8 @@
   "private": true,
   "scripts": {
     "build": "webpack --mode=production",
-    "build:watch": "webpack --mode=production --watch",
-    "debug": "webpack --mode=none",
+    "build:watch": "webpack --mode=development --watch",
```
lieut-data, line 5: Can we avoid changing this, and just rely on the developer running with `MM_DEBUG` instead? Otherwise we'll want to surface this back to the starter-template, etc.

webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
```diff
@@ -273,7 +273,7 @@ const BackstageIncidentDetails: FC = () => {
     }
 
     const goToChannel = () => {
-        navigateToUrl(`/${incident.team_name}/channels/${incident.channel_name}`);
+        navigateToUrl(`/${incident.details.team_name}/channels/${incident.details.channel_name}`);
```
lieut-data, line 276: Maybe we call it `incidentWithDetails.details` to make `incident.incident.*` a bit less of a mouthful?

lieut-data (changes requested):
A bit of a mouthful, but seems slightly cleaner. A few thoughts on the developer mode below.

webapp/src/client.ts
cpoile: I thought about that, except that a lot of developers use `make deploy` to deploy the plugin while developing, right? And then the environment would be production... What do you think?

webapp/package.json
cpoile: It's so weird that `build:watch` is production mode, isn't it? That's what I use for development all the time (along with `make deploy`, actually). Is there a reason this is set to production?

webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
cpoile: It's the stutter vs more typing, I guess.  🤷

webapp/package.json
cpoile: I reverted these changes -- we can discuss and do a wholesale change to the starter template later

webapp/src/components/backstage/incidents/incident_details/incident_details.tsx
cpoile: I made the change, I do think it's better without the stutter: less confusing

webapp/src/client.ts
cpoile: After thinking about it, I vote for keeping it pegged to the server's mode. Basically my reasoning is: we don't know what the npm mode is because of the weird way we set production and developer flag below, and the many different ways a developer could be sending a plugin to the server. So, we don't know if a developer will have the npm developer flag on. But these errors need attention, so we want to make sure every developer sees them.

lieut-data: If `MM_DEBUG` is set (or `DEBUG` as currently in incident response until we update from the starter-template), then `make deploy ` and `make watch` will both generate debug builds suitable for developers, including unmified code, debugging symbols, etc. 

Is there a use-case for getting debugging statements like this from a minified production build being used for development?

cpoile: To be honest, I didn't know about `MM_DEBUG` or `DEBUG`... But, I just tried it, and setting it style compiles the webapp with `mode=none`.  So those errors would not be surfaced if you run `DEBUG=true make deploy` or `make debug`.  Is that okay? (I feel like I might be missing something, since it seems like we should be erring (hah) on the side of outputting these errors on everything except a production server, no?)

webapp/package.json
lieut-data: I admit to having done this intentionally, primarily so that the `MM_DEBUG` semantics were consistently applied, vs. being "special" depending on some contexts. But I do think there is a rationale for `build:watch` in production mode, simply because there is a non-trivial performance difference, and it can be useful to develop in production mode and only jump into development mode if issues arise that require further debugging. At least, that's how I usually develop.

We could certainly raise the visibility of `MM_DEBUG` when doing development, though!

webapp/src/client.ts
lieut-data: Ah, interesting. `mode=none` does indeed sound like the wrong choice, and maybe we need to explicitly go for `mode=development` to get the desired behaviour.

agarciamontoro (commented):
Great catch, Chris, this was not a simple bug! I like the solution for the specific bug, but I agree that we should talk on Monday about how to handle this holistically from now on.

webapp/src/client.ts
```diff
@@ -37,12 +45,26 @@ export async function fetchIncidents(params: FetchIncidentsParams) {
...
-    return doGet(`${apiUrl}/incidents/${id}`);
+export async function fetchIncident(id: string) {
+    const data = await doGet(`${apiUrl}/incidents/${id}`);
+    if (SystemSettings.EnableDeveloperMode && !isIncident(data)) {
+        // eslint-disable-next-line no-console
+        console.error('expected an Incident in fetchIncident, received:', data);
+    }
+
+    // Prefer to continue to run even though we may be missing a field.
+    return data as Incident;
```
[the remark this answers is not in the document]
cpoile: @agarciamontoro That's an interesting solution. I was about to say: Surface an error but return a <Incident | null>. But then I realize that we don't actually want to set a non-conforming JSON object from the server to `null`, do we? That would break everything if just one field was missing. Or is that what we want to do?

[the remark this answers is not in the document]
cpoile: After the discussion in the channel, it makes more sense to cast and pass it along, after doing a development-only type guard.

webapp/src/index.tsx
```diff
@@ -42,6 +44,9 @@ export default class Plugin {
     public initialize(registry: PluginRegistry, store: Store<GlobalState>): void {
         registry.registerReducer(reducer);
 
+        const config = getConfig(store.getState());
+        SystemSettings.EnableDeveloperMode = Boolean(config.EnableDeveloper);
+
```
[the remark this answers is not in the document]
cpoile: not needed now, but yes that would have been necessary, thanks!

webapp/package.json
cpoile: After discussion offline, making one small change here.

webapp/src/client.ts
cpoile: As discussed offline, going to `mode=development` makes the most sense, and using `process.env.NODE_ENV` is most desirable because webpack will tree-shake it out.

lieut-data (approved):
:tada:

PR #310, opened 2020-10-01
MM-28607: run slash command polish

lieut-data:
#### Summary
Track when the last slash command was successfully run, and use that to show `Rerun` instead of just `Run`.

A bit of visual polish:
image

Oh, and I eliminated the spinners in most of the weird places we used them before because they just glitch right now. We can revisit that holistically later.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28607


webapp/src/components/checklist_item.tsx
```diff
@@ -153,6 +158,13 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
     };
 
     const [running, setRunning] = useState(false);
+    const [lastRun, setLastRun] = useState(props.checklistItem.command_last_run);
+
+    // Immediately stop the running indicator when we get notified of a more recent execution.
+    if (props.checklistItem.command_last_run && lastRun && props.checklistItem.command_last_run > lastRun) {
```
agarciamontoro, line 164: nit: do we need the first two conditionals here? The last condition will fail in case they are both 0, which I understand is their default value.

webapp/src/components/checklist_item.tsx
```diff
@@ -265,12 +277,12 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
                             }
                         }}
                     >
-                        {'(Run)'}
+                        {props.checklistItem.command_last_run ? 'Rerun' : 'Run'}
```
agarciamontoro, line 280: Actually, do we need the state of `lastRun, setLastRun`? Can't we just use this field of the prop? If it changes, it will re-render, isn't it?

webapp/src/types/playbook.ts
```diff
@@ -55,6 +55,7 @@ export interface ChecklistItem {
     assignee_modified?: number;
     assignee_modified_post_id?: string;
     command: string;
+    command_last_run?: number;
```
agarciamontoro, line 58: Does this need to be optional? We can check for its zero value.

agarciamontoro: Also, we may need to add this new field to the `isChecklistItem` function at the end of this file.

agarciamontoro (changes requested):
A much nicer look. And thank you for removing the spinners, agree that the blinking was not a good experience.

webapp/src/components/checklist_item.tsx
lieut-data: Indeed, I [wrestled](https://community-daily.mattermost.com/core/pl/3u3xwpkhzbnypcw5g56kw97p9a) with this for a bit before giving up and letting TypeScript win. Since both are optional, they might be null.

webapp/src/components/checklist_item.tsx
lieut-data: The `lastRun`/`setLastRun` is mostly to force the spinner off as soon as we hear back from the server (allowing me to sense the delta, as opposed to just using the current value). Otherwise, the text re-renders to `Rerun` while the spinner is still showing, which felt weird somehow. But maybe I'm overthinking this, and that behaviour is just fine?

webapp/src/types/playbook.ts
lieut-data: It's only optional because otherwise I have to litter parts of the code with `command_last_run: 0` when creating a playbook before sending it up to the server. Good catch on `isChecklistItem` -- I didn't realize this existed :)

webapp/src/components/checklist_item.tsx
agarciamontoro: Ah ok ok, it makes sense. Agree to have the state, then, I had not thought about that case!

webapp/src/types/playbook.ts
lieut-data: Moved to making `command_last_run` not optional. Thoughts?

agarciamontoro (approved):
Yay! I think that's much better, thank you! :tada:

webapp/src/components/checklist_item.tsx
cpoile: This looks great, thanks Jesse!

cpoile (approved):

PR #315, opened 2020-10-05
MM-29158: fix custom components in react select

lieut-data:
#### Summary
Fix an issue with the `ProfileSelector` being given a `CustomComponent` that is redefined on every render. This breaks the internal logic of react-select, forcing it to render new DOM elements instead of reusing them, and ultimately losing focus. On an active site like community-daily, the re-renders almost most the component unusable.

The fix here was applied both to the backstage and the incident RHS for assignees, making it technically the requisite follow-up to https://github.com/mattermost/mattermost-plugin-incident-response/pull/260. I'm still not sure why the webapp insists on re-rendering practically everything so often.

While in here, I removed one thing tsc no longer complained about, and fixed another.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29158

### DIFF
--- a/webapp/src/components/profile/profile_selector.tsx
+++ b/webapp/src/components/profile/profile_selector.tsx
@@ -38,6 +38,7 @@ interface Props {
     selfIsFirstOption?: boolean;
     getUsers: () => Promise<UserProfile[]>;
     onSelectedChange: (userId?: string) => void;
+    customControlProps?: any;
 }
 
 export default function ProfileSelector(props: Props) {
@@ -195,6 +196,7 @@ export default function ProfileSelector(props: Props) {
                 onChange={(option, action) => onSelectedChange(option as Option, action as ActionObj)}
                 classNamePrefix='incident-user-select'
                 className='incident-user-select'
+                {...props.customControlProps}
             />
         </Dropdown>
     );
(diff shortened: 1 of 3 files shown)


agarciamontoro (approved):
The code looks good, but I don't exactly understand why this avoids the re-render. Was the problem that we were using the `fetchParams.commander_user_id` directly on the definition of `ControlComponent`? Or that it was defined inside the `BackstageIncidentList component?

lieut-data:
> Or that it was defined inside the `BackstageIncidentList component?

This -- I forgot to link my react-select issues dive, but the comment at https://github.com/JedWatson/react-select/issues/2702#issuecomment-441970157 basically summarizes the issue:

> Every time the render function is fired ... it will create a new instance of the Input function, which triggers a re-render and replaces the previous Input component.

> This example would work if you remove the props input Value and onInputChange, but **defining components in the render function is discouraged because of unnessecary rerendering and remounting.**

webapp/src/components/profile/profile_selector.tsx
```diff
@@ -166,12 +167,7 @@ export default function ProfileSelector(props: Props) {
         );
     }
 
-    // The following is awkward, but makes TS happy.
-    const baseComponents = {DropdownIndicator: null, IndicatorSeparator: null};
-    const components = props.customControl ? {
-        ...baseComponents,
-        Control: props.customControl,
-    } : baseComponents;
```
cpoile, line 174: :tada:

cpoile (approved):
Nice, looks cleaner now too. Thanks!

cpoile:
> I'm still not sure why the webapp insists on re-rendering practically everything so often.

This is becoming an "age-old problem."

PR #316, opened 2020-10-06
[MM-28729] Fix days calculation for duration

jmakhack:
#### Summary

Fixes the days calculation for duration by using the `asDays()` function as opposed to `days()` to get the accurate count of days when the duration has exceeded a month's length of time. `asDays()` is needed since days is the largest unit of time being displayed here. The value of days is rounded down to the nearest integer since the remainder is displayed as hours and minutes.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/15785


jasonblais:
@lieut-data @crspeller @itao  I don't have permissions to add reviewers, but this would be for your team :)

Please also add `hacktoberfest-accepted` label, thanks!

agarciamontoro (commented):
Thank you @jmakhack, this is perfect! :tada:

Would you be open to add a new test for this? You can add it in `webapp/src/components/duration.test.tsx`, verifying that durations of 30 or more days are correctly rendered.

crspeller (approved):

agarciamontoro (approved):
Thank you, @jmakhack, this is great!

agarciamontoro:
@jmakhack, just a heads up that we will wait after we release v1.0.0 (which we are testing now) to merge this PR. I'll add the Do Not Merge label and we can remove it when v1.0.0 is tagged and released.

agarciamontoro:
Thank you again for your contribution, @jmakhack!

PR #318, opened 2020-10-06
MM-28604 - RHS List View

cpoile:
#### Summary
- Implements the list view as described in the ticket.
- The cypress tests were tricky; I couldn't figure out a way to log in as a sysadmin, do some work, log out, log back in as user-1, and continue on with the test--whenever I loaded the town-square, the app would send me to the login screen (even after I verified that current user was user-1). So I had to write some of the tests assuming previous state within their `describe` group. I think that should be okay.
- Based on #307, so only commits from 6f5c06a on are related to this PR (that prev code doesn't show up in the Files Changed, at least). 

#### UX Review
- Looks like:
image

- After clicking the three-dot menu: 
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-28604


server/api/incidents.go
```diff
@@ -308,7 +308,6 @@ func (h *IncidentHandler) getIncidents(w http.ResponseWriter, r *http.Request) {
 
 	requesterInfo := incident.RequesterInfo{
 		UserID:          userID,
-		TeamID:          filterOptions.TeamID,
```
cpoile, line 311: no longer needed (prev PR)

webapp/src/reducer.ts
```diff
@@ -43,29 +51,46 @@ function clientId(state = '', action: SetClientId) {
...
+        return action.nextState;
+    default:
+        return state;
+    }
+}
+
+// myChannelIdToIncidents is a set of incidents for which the current user is an incident member. Note
+// that it is lazy loaded on team change, but will also track incremental updates as provided by
+// websocket events.
+const myChannelIdToIncidents = (state: Record<string, Incident> = {}, action: IncidentCreated | IncidentUpdated | ReceivedTeamIncidents | RemovedFromIncidentChannel) => {
```
cpoile, line 66: Turns out, it's simpler to turn the "known incident channels ids" into the user's "list of incidents they are a member of", then use the selectors to pull out the information we need.

webapp/src/rhs_opener.ts
```diff
@@ -43,6 +47,16 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
         currentChannelId = currentChannel.id;
         currentChannelIsIncident = isIncidentChannel(state, currentChannelId);
 
+        // Decide whether to show the Incident Details or Incident List view when we change channels
```
cpoile, line 50: The main logic for when to show the incident list vs incident view.

webapp/src/selectors.ts
```diff
@@ -15,10 +17,28 @@ export const isIncidentRHSOpen = (state: GlobalState): boolean => pluginState(st
...
+    const incidents = [] as Incident[];
+    for (const incident of Object.values<Incident>(pluginState(state).myChannelIdToIncidents)) {
+        if (incident.is_active) {
+            incidents.push(incident);
+        }
+    }
+    return sortIncidentsDescByCreateAt(incidents);
+};
+
+function sortIncidentsDescByCreateAt(incidents: Incident[]) {
```
cpoile, line 34: Not sure if this needs to be its own function?

tests-e2e/cypress/integration/frontstage/incident_rhs_spec.js
```diff
@@ -140,6 +140,7 @@ describe('incident rhs', () => {
         it('for a new, ongoing incident channel opened from the lhs', () => {
             // # Navigate to the application.
             cy.visit('/');
+            cy.get('#centerChannelFooter').findByTestId('post_textbox').should('be.visible');
```
cpoile, line 143: These (and the same ones in the `incident_rhs_list_spec`) are to ensure the page is loaded and redux has been updated before clicking on the rhs. If we don't, cypress will click before we have the incident list in redux.

webapp/src/hooks.ts
```diff
@@ -73,22 +72,22 @@ export function useCurrentIncident(): [Incident | null, CurrentIncidentState] {
...
+        }
+
         window.addEventListener('keydown', downHandler);
 
         // Remove event listeners on cleanup
         return () => {
             window.removeEventListener('keydown', downHandler);
         };
-    }, []); // Empty array ensures that effect is only run on mount and unmount
+    }, [handler, targetKey]); // Empty array ensures that effect is only run on mount and unmount
```
lieut-data, line 89: Comment here is out-of-date?

webapp/src/index.tsx
```diff
@@ -63,7 +65,8 @@ export default class Plugin {
         // Would rather use a saga and listen for ActionTypes.UPDATE_MOBILE_VIEW.
         window.addEventListener('resize', debounce(updateMainMenuAction, 300));
 
-        const {toggleRHSPlugin} = registry.registerRightHandSidebarComponent(RightHandSidebar, <RHSTitle/>);
+        const {toggleRHSPlugin} = registry.registerRightHandSidebarComponent(RightHandSidebar,
```
lieut-data, line 68: nit: if we're going to line-wrap, can we pull down all the parameters to separate lines, vs. splitting on one only?

webapp/src/rhs_opener.ts
```diff
@@ -29,10 +31,14 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
             return;
         }
 
-        // Update the known set of incident channels whenever the team changes.
+        // Update the known set of incidents whenever the team changes.
         if (currentTeamId !== currentTeam.id) {
             currentTeamId = currentTeam.id;
-            store.dispatch(receivedTeamIncidentChannels(await fetchIncidentChannels(currentTeam.id)));
+            store.dispatch(fetchingIncidents(true));
```
lieut-data, line 37: nit: I wonder if we could have `store.dispatch(startingToFetchIncidents)` and `store.dispatch(finishedFetchingIncidents)` vs. the opaque `true` and` false`?

webapp/src/rhs_opener.ts
```diff
@@ -43,6 +49,16 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
         currentChannelId = currentChannel.id;
         currentChannelIsIncident = isIncidentChannel(state, currentChannelId);
 
+        // Decide whether to show the Incident Details or Incident List view when we change channels
+        if (sentRHSStateForChannelId !== currentChannelId) {
+            sentRHSStateForChannelId = currentChannelId;
+            if (currentChannelIsIncident || currentlyFetchingIncidents(state)) {
+                store.dispatch(setRHSState(RHSState.ViewingIncident));
```
lieut-data, line 62: This state /seems/ a bit weird -- if we don't know we're in an incident channel yet, we track our state as `ViewingIncident`. I wonder if we should instead model the `IncidentWelcomeScreen` or some such?

webapp/src/selectors.ts
```diff
@@ -15,10 +17,30 @@ export const isIncidentRHSOpen = (state: GlobalState): boolean => pluginState(st
 
 export const clientId = (state: GlobalState): string => pluginState(state).clientId;
 
-export const isIncidentChannel = (state: GlobalState, channelId: string): boolean => pluginState(state).myIncidentChannelIds.has(channelId);
+export const isIncidentChannel = (state: GlobalState, channelId: string): boolean => {
+    return Boolean(pluginState(state).myChannelIdToIncidents[channelId]);
+};
+
+export const activeIncidentList = (state: GlobalState): Incident[] => {
```
lieut-data, line 24: Do we need to do some memoization here? Otherwise, it seems like this selector always returns a new object regardless of whether the underlying values have changed.

webapp/src/websocket_events.ts
```diff
@@ -39,11 +44,13 @@ export function handleWebsocketIncidentUpdate() {
         }
         const incident = data as Incident;
 
-        websocketSubscribers.forEach((fn) => fn(incident));
+        dispatch(incidentUpdated(incident));
+
+        websocketSubscribersToIncidentUpdate.forEach((fn) => fn(incident));
```
lieut-data, line 49: If we have the `dispatch` built-in, do we need to still support subscribers?

lieut-data (commented):
I haven't finished the review, but a handful of early feedback for discussion.

tests-e2e/cypress/support/plugin_api_commands.js
```diff
@@ -19,6 +19,20 @@ Cypress.Commands.add('apiGetAllIncidents', (teamId) => {
     });
 });
 
+/**
+* Get all incidents directly via API
+*/
+Cypress.Commands.add('apiGetAllActiveIncidents', (teamId) => {
```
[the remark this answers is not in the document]
cpoile: I changed it to 10000 just in case. But I also noticed that most of the tests that used it were searching to make sure an incident was created, so I changed those to use a search term instead.

tests-e2e/cypress/support/index.js
```diff
@@ -35,14 +35,31 @@ Cypress.Commands.add('requireIncidentResponsePlugin', (version) => {
 });
 
 /**
- * Delete all incidents directly from API
+ * End all active incidents directly from API with sysadmin. Need to login after this.
  */
-Cypress.Commands.add('deleteAllIncidents', (teamId) => {
-    cy.apiGetAllIncidents(teamId).then((response) => {
-        const incidents = JSON.parse(response.body);
+Cypress.Commands.add('endAllActiveIncidents', (teamId) => {
```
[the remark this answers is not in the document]
cpoile: That was the comment about logging in and out with sysadmin. It was causing cypress to load the login screen, and I spent hours trying to figure out why. I couldn't figure it out, so I made the `endAllMyActiveIncidents`. For some reason `endAllActiveIncidents` works if it's used in the first `it` test within a `describe` block, so I kept that one. But when used in any other `it` test, it would randomly show the login screen for any subsequent `cy.visit` call.

webapp/src/components/backstage/incidents/status_badge.tsx
```diff
@@ -30,7 +30,6 @@ const Badge = styled.div<BadgeProps>`
     line-height: 24px;
 
     ${(props) => props.compact && css`
-        top: -1px;
```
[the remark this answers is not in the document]
cpoile: That now pushes it too far up and clips the badge icon, but probably it's a change I made in the `rhs_title`.

webapp/src/index.tsx
cpoile: Whoops, that was an auto-format change I forgot to fix. Reverted.

webapp/src/rhs_opener.ts
cpoile: Yah, I felt bad about that too. I was able to simplify this in ab1bc2f.

webapp/src/websocket_events.ts
cpoile: This change is only renaming the existing subscriber set that Christopher added for the `useCurrentIncident` hook. It's still needed for that hook, unless we want to replace that hook with the state (which would be in a new PR, in my opinion).

cpoile:
@lieut-data Ready for re-review.  (Sorry, couldn't leave it alone.) I think this is much cleaner now, thanks for thinking it through with me.

webapp/src/components/assets/icons/dot_menu_icon.tsx
```diff
@@ -6,7 +6,7 @@ import React from 'react';
 import styled from 'styled-components';
 
 const DotMenuIcon = (props: React.PropsWithoutRef<JSX.IntrinsicElements['i']>): JSX.Element => (
-    <i className={`icon icon-dots-horizontal icon-16 ${props.className}`}/>
+    <i className={`icon icon-16 ${props.className}`}/>
```
agarciamontoro, line 9: Doesn't this remove the icon itself?

webapp/src/components/dot_menu.tsx
```diff
@@ -11,13 +11,17 @@ const DropdownMenuWrapper = styled.div`
     position: relative;
 `;
 
-const DropdownMenu = styled.div`
+interface DropdownMenuProps {
+    openLeft?: boolean;
+}
+
+const DropdownMenu = styled.div<DropdownMenuProps>`
```
agarciamontoro, line 18: I also changed this in https://github.com/mattermost/mattermost-plugin-incident-response/pull/319/files#diff-5a75622b83cc511c04fc73cb85ef95d5, so there may be some conflicts when merging master. Ping me if you run into any problems!

webapp/src/components/dot_menu.tsx
```diff
@@ -71,12 +78,12 @@ const DotMenu: FC<DotMenuProps> = (props: DotMenuProps) => {
             }}
         >
             <IconWrapper>
-                <DotMenuIcon/>
+                <DotMenuIcon className={props.vertical ? 'icon-dots-vertical' : 'icon-dots-horizontal'}/>
```
agarciamontoro, line 81: Aaaah ok I see. I solved this by adding an icon prop https://github.com/mattermost/mattermost-plugin-incident-response/pull/319/files#diff-5a75622b83cc511c04fc73cb85ef95d5R45 and an explicit vertical dots icon https://github.com/mattermost/mattermost-plugin-incident-response/pull/319/files#diff-040192edb708c970ae3babd409c7b6dd. Pick whichever you prefer, I'm 0/5 on any of the solutions.

webapp/src/components/rhs/rhs_list_view.tsx
```diff
@@ -0,0 +1,253 @@
...
+    border-radius: 4px;
+    background: transparent;
+    font-size: 12px;
+    font-weight: 600;
+    line-height: 9.5px;
+    color: var(--button-bg);
+    text-align: center;
+    padding: 10px 0;
+    margin-top: 10px;
+`;
```
agarciamontoro, line 128: I also added this (without the `margin-top`, apparently)  in https://github.com/mattermost/mattermost-plugin-incident-response/pull/319/files#diff-f69bd71cf9a9f067cc3968522f01f83aR115-R126. Feel free to refactor if needed, or we can keep them separated.

webapp/src/components/rhs/rhs_list_view.tsx
```diff
@@ -0,0 +1,253 @@
...
+                                            ended_at={incident.end_at}
+                                        />
+                                    </Col2>
+                                </Row>
+                                <Row>
+                                    <Col1>{'Commander:'}</Col1>
+                                    <Col2>
+                                        <SmallerProfile userId={incident.commander_user_id}/>
+                                    </Col2>
+                                </Row>
```
agarciamontoro, line 210: I really like what you did here with the rows and cols :tada:

webapp/src/components/rhs/rhs_title.tsx
```diff
@@ -22,19 +27,43 @@ const RHSTitleText = styled.div`
...
+    const theme = useSelector<GlobalState, Record<string, string>>(getTheme);
+
+    if (rhsState === RHSState.ViewingIncident) {
+        return (
+            <RHSTitleContainer>
+                <Button
+                    onClick={() => dispatch(setRHSViewingList())}
+                    data-testid='back-button'
+                >
+                    <LeftChevron theme={theme}/>
```
agarciamontoro, line 51: Why do we need the theme here? Isn't it possible to solve it with CSS vars?

agarciamontoro (commented):
Incomplete review, but publishing it in case you can read it. I'll continue after lunch from `webapp/src/reducer.ts` on (and the tests).

webapp/src/components/assets/icons/dot_menu_icon.tsx
agarciamontoro: Nvm.

webapp/src/reducer.ts
```diff
@@ -43,29 +52,46 @@ function clientId(state = '', action: SetClientId) {
...
```
agarciamontoro, line 67: Nit: can we name this something like `myIncidentsSet`? Or maybe `myIncidentsMap`? I did not understand what it was at first sight, but maybe it's just me.

webapp/src/selectors.ts
```diff
@@ -15,10 +19,31 @@ export const isIncidentRHSOpen = (state: GlobalState): boolean => pluginState(st
...
+    (channelIdToIncidents) => {
+        const incidents = [] as Incident[];
+        for (const incident of Object.values(channelIdToIncidents)) {
+            if (incident.is_active) {
+                incidents.push(incident);
+            }
+        }
+
+        // return descending by create_at
+        return incidents.sort((a, b) => b.create_at - a.create_at);
```
agarciamontoro, line 39: Nitting all day long: what about something like this (which I have not tested)?
```js
return channelIdToIncidents.values.
    filter(i => i.is_active).
    sort((a, b) => b.create_at - a.create_at);
```

webapp/src/types/rhs.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export enum RHSState {
+    ViewingList,
+    ViewingIncident,
+}
```
agarciamontoro, line 7: The whole management of the RHS state is really nice :tada:

agarciamontoro (approved):
This is great, and a ton of work :tada: Thank you, @cpoile!

I left some nits below, but nothing blocking, so approved!

webapp/src/components/dot_menu.tsx
lieut-data: Let's go with the prop, vs. more class names, so we can more fully embrace styled components.

webapp/src/components/rhs/rhs_list_view.tsx
cpoile: I'll keep them separate for now. We might be due for a pretty major css simplification/cleanup anyway, so let's worry about it then.

webapp/src/reducer.ts
cpoile: Good idea, changed to `myIncidentsMap`

webapp/src/selectors.ts
cpoile: It would have to be `Object.values(...)`, but I think there's something wrong with this, so I'll leave that for another commit. Thanks though!

lieut-data (approved):
Really solid work, @cpoile! Loved where we ended up :)

PR #319, opened 2020-10-07
MM-28674: Stage simplification

agarciamontoro:
#### Summary
This PR simplifies the stage with exactly the behaviour explained in the ticket. A couple of more additions while I was here:
- I added a couple of slash commands to be consistent with the UI: `/incident stage next` and `/incident stage prev`.
  - When the stage is the last one, running `/incident stage next` posts an ephemeral message: `The active stage is the last one. If you want to end the incident, run /incident end`.
  - When the stage is the first one, running `/incident stage prev` posts an ephemeral message: `The active stage is the first one.`
  - When running `/incident stage next` with unfinished tasks in the stage, an interactive dialog appears (the same one that appears when clicking Next Stage button, see screenshot below)
- I made the checkboxes theme-compatible. 

##### Common cases (incidents with more than one stage, currently in a middle stage)

| With tasks unfinished 	| With all tasks finished 	| Clicking Next Stage with tasks unfinished |
|:-:	|:-:	|:-:|
|image|image|image|

##### Special cases

|  (themed) An incident with one stage has only the `End Incident` button: 	| First stage contains only `End Incident` in the dropup menu 	| Last stage contains only `Previous Stage` in the dropup menu |
|:-:	|:-:	|:-: |
|image|image|image|

##### Dark theme

| Middle stage 	| Middle stage with all tasks finished 	| Ended incident|
|:-:	|:-:	|:-:	|
|  	image|image|  	image|

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28674


server/api/incidents.go
```diff
@@ -630,6 +631,34 @@ func (h *IncidentHandler) changeCommander(w http.ResponseWriter, r *http.Request
 	_, _ = w.Write([]byte(`{"status": "OK"}`))
 }
 
+// nextStageDialog handles the interactive dialog submission when a user confirms they
+// want to go to the next stage.
+func (h *IncidentHandler) nextStageDialog(w http.ResponseWriter, r *http.Request) {
```
lieut-data, line 636: Instead of an incident-dialog-specific endpoint, would it make sense to extend the existing `updateIncident` endpoint to a) support `POST` (because of the interactive dialog) and b) support query string parameters, so we can just encode a final URL of `POST /incidents/<id>?stage=1`?

tests-e2e/cypress/integration/frontstage/incident_rhs_spec.js
```diff
@@ -357,4 +417,420 @@ describe('incident rhs', () => {
             });
         });
     });
+
+    describe('has stages information', () => {
```
lieut-data, line 421: :tada:

tests-e2e/cypress/integration/frontstage/slash_command_spec.js
```diff
@@ -208,6 +224,208 @@ describe('slash command', () => {
         });
     });
 
+    describe('/incident stage', () => {
```
lieut-data, line 227: :tada: :tada:

webapp/src/components/assets/icons/three_dots_icon.tsx
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {FC} from 'react';
+
+const ThreeDotsIcon = (props: React.PropsWithoutRef<JSX.IntrinsicElements['i']>): JSX.Element => (
```
lieut-data, line 6: nit: this type definition seems... heavy. Does it suffice just to accept `props: {className: string}`?

webapp/src/components/checklist.scss
```diff
@@ -69,7 +69,7 @@ $font-family: Open Sans;
         min-width: 20px;
         height: 20px;
         background: #ffffff;
-        border: 1px solid #dddddd;
+        border: 2px solid var(--center-channel-color-24);
```
lieut-data, line 72: 👍

lieut-data (approved):
Really solid work, @agarciamontoro -- I love the slash command extension to round it out!

lieut-data:
This feels pretty self-contained that I'd be comfortable merging after all the tests pass, and letting things soak on daily vs. requiring a second review.

agarciamontoro:
Phew, that was quite a lot of changes, thank you, @lieut-data!

PR #325, opened 2020-10-08
MM-29286: Change API version to v0

agarciamontoro:
#### Summary

Just that, replacing all occurrences of `v1` with `v0`.

I considered coding a simple function that builds the API url prefix and call that from all the places that are using a path, but it's too late to fight paths and double slashes and stuff that will break before code freeze. So good old replace. We can do the right thing when we move to v1.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29286

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	apiVersion = "v1"
+	apiVersion = "v0"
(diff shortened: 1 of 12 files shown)


agarciamontoro:
Feel free to merge as soon as reviews are done and tests pass :)

PR #326, opened 2020-10-08
MM-29485: explicitly support "no stages"

lieut-data:
#### Summary
This handles "old incidents" that were started as such, but also hardens the code against any unexpectedness without just hacking in required changes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29485


server/incident/incident.go
```diff
@@ -41,6 +44,13 @@ func (i *Incident) MarshalJSON() ([]byte, error) {
 		}
 	}
 
+	// Define consistent semantics for empty checklists and out-of-range active stages.
+	if len(old.Checklists) == 0 {
+		old.Header.ActiveStage = NoActiveStage
+	} else if old.Header.ActiveStage < 0 || old.Header.ActiveStage >= len(old.Checklists) {
+		old.Header.ActiveStage = 0
+	}
```
cpoile, line 52: Ah, okay, this took me a second. You're saying an `ActiveStage` of -1 is fine only when we have no stages. Otherwise, force the active stage to be 0 if it's out-of-range. And we don't use ActiveStage below anymore for the tests (that's what was crashing).

cpoile (approved):

crspeller (approved):
Bit of a mind bender where cpoile pointed out, but it makes sense.

PR #328, opened 2020-10-08
MM-29481, MM-29483, MM-29484 - Commander dropdown, /incident list fixes

cpoile:
#### Summary
- Commander dropdown has a max-width to keep it within the RHS size, and the chevron is removed (in the backstage too):
image

- The incident time list is the same as the frontend now:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29481
- https://mattermost.atlassian.net/browse/MM-29483
- https://mattermost.atlassian.net/browse/MM-29484


crspeller (approved):

webapp/src/components/profile/profile_selector.tsx
```diff
@@ -167,7 +167,8 @@ export default function ProfileSelector(props: Props) {
         );
     }
 
-    const components = props.customControl ? {Control: props.customControl} : {};
+    const noDropdown = {DropdownIndicator: null, IndicatorSeparator: null};
```
lieut-data, line 170: Ah, that's why this was there! Apologies for breaking it :D

lieut-data (approved):

PR #336, opened 2020-10-14
client: go.mod

lieut-data:
#### Summary
Configure the `client/` package as a separate go module (a "submodule" in our context), complete with a separate license that satisfies pkg.go.dev. Conveniently, we can also version this module separately using tags such as `client/v0.0.1`.

https://pkg.go.dev/github.com/mattermost/mattermost-plugin-incident-response/client@v0.0.1 now shows:

image

At the moment, the vanilla URL of https://pkg.go.dev/github.com/mattermost/mattermost-plugin-incident-response/client still shows the license restrictions of the parent repository, but I think that's because the corresponding `v0.7.0` tag didn't identify `client` as a separate module. This should get fixed with the release of `v1.0.0`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29150


go.mod
```diff
@@ -6,11 +6,10 @@ require (
 	github.com/Masterminds/squirrel v1.4.0
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/golang/mock v1.4.3
-	github.com/google/go-querystring v1.0.0
 	github.com/gorilla/mux v1.7.4
 	github.com/jmoiron/sqlx v1.2.0
 	github.com/mattermost/mattermost-plugin-api v0.0.12-0.20200925172129-8ea2e7c0b42b
-	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200731154015-c5c6a5ce5399
+	github.com/mattermost/mattermost-server/v5 v5.27.0
```
agarciamontoro, line 12: Oh, we are now ok with 5.27?

agarciamontoro: I think we're not. CI s complaining about the build, probably because of this change?

lieut-data:
Oops, you're right! `go mod tidy` pulled in `v5.27.0` because our branching strategy in the server repository isn't "normal", and the commits from that branch appear to be "newer" than where we had previously pinned in `master`. I've updated to `v5.28.0-rc2` to address this.

client/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/mattermost/mattermost-plugin-incident-response/client
+
+go 1.15
+
+require (
+	github.com/google/go-querystring v1.0.0
+	github.com/mattermost/mattermost-plugin-incident-response v1.0.0-alpha.3.0.20201014200312-0376ccebc7c3
```
lieut-data, line 7: For the purposes of this branch, I'm depending on a commit /in/ this branch, where the `go.mod` has spliced off `client/` as not part of the `incident-response` module. Otherwise, `go mod tidy` in this module complains about ambiguous imports (rightly so!).

crspeller (approved):

PR #337, opened 2020-10-15
MM-29699: Fix crash after /incident stage with no arguments

agarciamontoro:
#### Summary
There was a missing return after the check for no arguments, so although the help text informing the user was posted, the function kept running. As the code following that check assumed that there was at least one argument, it panicked when trying to retrieve it.

I've also added an E2E test covering this case. I check that no error is posted in the footer but, if the plugin crashes, all remaining tests will fail. That is, we will know that something is wrong one way or another.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29699


lieut-data (approved):
Nice!

cpoile (approved):
Nice!

lieut-data:
Proposing we validate after merge here, given the existing e2e coverage.

PR #340, opened 2020-10-15
MM-26182: Standardize filtering parameters

agarciamontoro:
#### Summary
All collection endpoints that support `sort`, `direction` and `search_term` (namely, `GET /incidents` and `GET /incidents/channels`) are now consistent. However, there are collections endpoints that only support a subset of those parameters. I did not add any code to support the missing parameters because we don't need it right now (and because adding new code at this stage does not seem like a good idea). The endpoints that don't support everything are:
  - `GET /playbooks` does support `sort` and `direction`, but does not support `search_term`.
  - `GET /incidents/commanders` does not support `sort`, `order` nor `search_term`.

I also removed the `status` value for the `sort` parameter because it feels redundant:
- status filtering is usually done through the `status` parameter
- we also have the `is_active` value for the `sort` parameter, in case we wanna sort it this way

Also: all tests are passing, and the renaming was automatic, but I feel like it's a big change to push after E2E testing. What do you think (cc/ @prapti)?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26182

### DIFF
--- a/client/incidents.go
+++ b/client/incidents.go
@@ -69,9 +69,6 @@ const (
 
 	// EndAt sorts by the "end_at" field.
 	EndAt IncidentSort = "end_at"
-
-	// ByStatus sorts by the "status" field.
-	ByStatus IncidentSort = "by_status"
 )
 
 // IncidentList contains the paginated result.
--- a/client/incidents.go
+++ b/client/incidents.go
@@ -45,7 +45,7 @@ type IncidentListOptions struct {
 	TeamID string `url:"team_id,omitempty"`
 
 	Sort      IncidentSort  `url:"sort,omitempty"`
-	Direction SortDirection `url:"order,omitempty"`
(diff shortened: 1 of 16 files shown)


lieut-data (approved):

client/incidents.go
```diff
@@ -69,9 +69,6 @@ const (
 
 	// EndAt sorts by the "end_at" field.
 	EndAt IncidentSort = "end_at"
-
-	// ByStatus sorts by the "status" field.
-	ByStatus IncidentSort = "by_status"
```
cpoile, line 74: Does this affect the "Status" column in the backstage incidents list?

cpoile (changes requested):
Nice, just one question!

cpoile (approved):
nvm, looks good! :)

PR #345, opened 2020-10-16
set RUDDER env for build artifacts only

lieut-data:
#### Summary
Avoid spamming telemetry from e2e tests by only setting the required RUDDER environment variables during the build step.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-29753

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -292,6 +292,7 @@ workflows:
             tags:
               only: /^v.*/
       - plugin-ci/build:
+          context: mattermost-plugin-incident-response-production
           filters:
             tags:
               only: /^v.*/


.circleci/config.yml
```diff
@@ -292,6 +292,7 @@ workflows:
             tags:
               only: /^v.*/
       - plugin-ci/build:
+          context: mattermost-plugin-incident-response-production
```
lieut-data, line 295: For context (hah!), I've removed the default CircleCI environment variables, and am now using this to selectively re-apply them. I've extracted the strings from the generated binary, and can confirm that the production key is still present in this artifact.

crspeller (approved):

agarciamontoro (approved):
:tada:

PR #347, opened 2020-10-17
MM-26194: Standardize HTTP status codes

agarciamontoro:
#### Summary

This PR standardizes the HTTP status codes returned by the API. The changes affect only the API layer (I tried at first to modify all the layers: API, service and store; but I'd like us to decide together how to approach this). There are a few things more that we can do to standardize, as checking empty IDs as the first step of the data validation in the API, but I also left those for the data validation approach we follow.

The main changes are:
- Returning 400 (Bad Request) instead of 500 (Internal Server Error) when it is clear in the API layer that the data is not valid. This happens when the query parameters or the body values are not valid.
- Returning 403 (Forbidden) instead of 500 (Internal Server Error) when there is a permissions error.
- Returning 201 (Created) instead of 200 (OK) when creating an entity. In these cases, I've also added a Location header with the URL of the created entity, as per the documentation.
- Returning 204 (No Content) instead of 200 (OK) when deleting an entity.

I've made a few more changes:
- I've removed the returned body ``w.Write([]byte(`{"status": "OK"}`))``, which was used inconsistently, and which was completely ignored in the client side.
- I've changed a bunch of repeated code just by using the `ReturnJSON` function.

One thing I was not able to do was to add `w.Header().Set("Content-Type", "application/json")` to the `ReturnJSON` function. This header was used only in a couple of places, but I thought it would be nice to have it in all JSON-encoded responses. However, the E2E tests failed when trying to parse the response when this header was added, as the body was encoded as a quoted string (?). Surprisingly, this happened in CI, but not locally. I think I must be missing something very basic here, but I decided to go without it for now, as it does not change current behaviour. We can discuss it later.

##### Detailed changes
<details>
<summary>Click on the arrow to read the detailed notes I took when doing this. They are unprocessed, so maybe not very useful for anybody else, but I'll leave them here for the sake of self-containment.</summary>

- GET /incidents: getIncidents
  + Nothing done
- POST /incidents: createIncidentFromPost
  + CHANGED: Return 400 instead of 500 when unable to decode incident
  + CHANGED: Return 400 instead of 500 when any of the params is not valid
  + CHANGED: Return 403 instead of 500 when the commander does not have permissions for the team
  + ADDED: Return location
  + CHANGED: Return 201 instead of 200 when the incident is created
- POST /incidents/dialog: createIncidentFromDialog
  + CHANGED: Return 400 instead of 500 when unable to decode dialog request
  + CHANGED: Return 400 instead of 500 when unable to unmarshal dialog state
  + CHANGED: Return 400 instead of 500 when any of the params is not valid
  + NOT CHANGED: the permission error (that should return a 404) is handled with a submitDialogResponse. Did not know what to do here.
  + ADDED: Return location
  + CHANGED: Return 201 instead of 200 when the incident is created
- GET /incidents/commanders: getCommanders
  - Already has data validation and permission
- GET /incidents/channels: getChannels
  - Already has data validation and permission
- GET /incidents/checklist-autocomplete: getChecklistAutocomplete
  + Nothing done
- GET /incidents/[id]: getIncident
  + Nothing done
- GET /incidents/[id]/metadata: getIncidentMetadata
  + Nothing done
- PATCH /incidents/[id]: updateIncident
  + CHANGED: Return 400 when the options cannot be decoded
- PUT POST /incidents/[id]/end: endIncident
  + Nothing done
- PUT /incidents/[id]/restart: restartIncident
  + Nothing done
- POST /incidents/[id]/commander: changeCommander
  + CHANGED: Return 400 instead of 500 when parsing of params fails
- POST /incidents/[id]/next-stage-dialog: nextStageDialog
  + CHANGED: Return 400 instead of 500 when dialog request cannot be decoded
  + CHANGED: Return 400 instead of 500 when state cannot be parsed
- GET /incidents/channel/[id]: getIncidentByChannel
  + Already: If no permissions, return 404
  + Already: if not found by the service, return 404
- PUT /incidents/checklists/[id]/add: addChecklistItem
  + CHANGED: Return 400 instead of 500 when query param cannot be parsed
  + CHANGED: Return 400 instead of 500 when body cannot be parsed
  + CHANGED: Return 201 instead of 200 when the item is added
- PUT /incidents/checklists/[id]/reorder: reorderChecklist
  + CHANGED: Return 400 instead of 500 when query param cannot be parsed
  + CHANGED: Return 400 instead of 500 when body cannot be parsed
- DELETE /incidents/checklists/[id]/item/[id]: itemDelete
  + CHANGED: Return 400 instead of 500 when any of the query params cannot be parsed
  + CHANGED: Return 204 instead of 200 when item is succesfully deleted
- PUT /incidents/checklists/[id]/item/[id]: itemRename
  + CHANGED: Return 400 instead of 500 when any of the query params cannot be parsed
  + CHANGED: Return 400 instead of 500 when the body cannot be decoded
- PUT /incidents/checklists/[id]/item/[id]/state: itemSetState
  + CHANGED: Return 400 instead of 500 when any of the query params cannot be parsed
  + CHANGED: Return 400 instead of 500 when the body cannot be decoded
  + CHANGED: Return 400 instead of 500 when the new checklist item state is not valid
- PUT /incidents/checklists/[id]/item/[id]/assignee: itemSetAssignee
  + CHANGED: Return 400 instead of 500 when any of the query params cannot be parsed
  + CHANGED: Return 400 instead of 500 when the body cannot be decoded
- POST /incidents/checklists/[id]/item/[id]/run: itemRun
  + CHANGED: Return 400 instead of 500 when any of the query params cannot be parsed
- POST /playbooks: createPlaybook
  + CHANGED: Return 400 instead of 500 when the body cannot be decoded
  + It already returns 403 when there is a permissions error
  + CHANGED: Return 201 instead of 200 when the playbook is succesfully created
  + ADDED: Return Location with the URL of the new entity
- GET /playbooks: getPlaybooks
  + CHANGED: Return 400 instead of 500 when the options are not valid
  + It already returns 403 when there is a permissions error
- GET /playbooks/[id]: getPlaybook
  + It already returns 403 when there is a permissions error.
- PUT /playbooks/[id]: updatePlaybook
  + CHANGED: Return 400 instead of 500 when the body cannot be decoded
  + It already returns 403 when there is a permissions error.
- DELETE /playbooks/[id]: deletePlaybook
  + It already returns 403 when there is a permissions error
  + CHANGED: Return 204 instead of 200 when the playbook is successfully deleted

</details>

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-26194

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -42,14 +42,14 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, sourcePlugin
 }
 
 // ReturnJSON writes the given pointer to object as json with a success response
-func ReturnJSON(w http.ResponseWriter, pointerToObject interface{}) {
+func ReturnJSON(w http.ResponseWriter, pointerToObject interface{}, httpStatus int) {
 	jsonBytes, err := json.Marshal(pointerToObject)
 	if err != nil {
 		HandleError(w, errors.Wrapf(err, "unable to marshal json"))
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
+	w.WriteHeader(httpStatus)
 	_, _ = w.Write(jsonBytes)
 }
 
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -50,7 +50,10 @@ func ReturnJSON(w http.ResponseWriter, pointerToObject interface{}, httpStatus i
 	}
 
 	w.WriteHeader(httpStatus)
-	_, _ = w.Write(jsonBytes)
+	if _, err = w.Write(jsonBytes); err != nil {
+		HandleError(w, err)
+		return
+	}
 }
 
 // HandleError writes err as json into the response.
--- a/server/api/incidents.go
+++ b/server/api/incidents.go
@@ -103,11 +103,22 @@ func (h *IncidentHandler) createIncidentFromPost(w http.ResponseWriter, r *http.
 
 	var payloadIncident incident.Incident
 	if err := json.NewDecoder(r.Body).Decode(&payloadIncident); err != nil {
-		HandleError(w, errors.Wrapf(err, "unable to decode incident"))
+		HandleErrorWithCode(w, http.StatusBadRequest, "unable to decode incident", err)
 		return
 	}
 
 	newIncident, err := h.createIncident(payloadIncident, userID)
+
+	if errors.Is(err, incident.ErrPermission) {
+		HandleErrorWithCode(w, http.StatusForbidden, "unable to create incident", err)
+		return
+	}
+
+	if errors.Is(err, incident.ErrMalformedIncident) {
+		HandleErrorWithCode(w, http.StatusBadRequest, "unable to create incident", err)
+		return
+	}
+
 	if err != nil {
 		HandleError(w, errors.Wrapf(err, "unable to create incident"))
 		return
@@ -117,7 +128,8 @@ func (h *IncidentHandler) createIncidentFromPost(w http.ResponseWriter, r *http.
 		"incident": newIncident,
 	}, userID)
 
-	ReturnJSON(w, &newIncident)
+	w.Header().Add("Location", fmt.Sprintf("/api/v0/incidents/%s", newIncident.ID))
+	ReturnJSON(w, &newIncident, http.StatusCreated)
 }
 
 func (h *IncidentHandler) updateIncident(w http.ResponseWriter, r *http.Request) {
@@ -133,7 +145,7 @@ func (h *IncidentHandler) updateIncident(w http.ResponseWriter, r *http.Request)
 
 	var updates incident.UpdateOptions
 	if err = json.NewDecoder(r.Body).Decode(&updates); err != nil {
-		HandleError(w, errors.Wrapf(err, "unable to decode payload"))
+		HandleErrorWithCode(w, http.StatusBadRequest, "unable to decode payload", err)
 		return
 	}
 
@@ -165,14 +177,14 @@ func (h *IncidentHandler) updateIncident(w http.ResponseWriter, r *http.Request)
 func (h *IncidentHandler) createIncidentFromDialog(w http.ResponseWriter, r *http.Request) {
 	request := model.SubmitDialogRequestFromJson(r.Body)
 	if request == nil {
-		HandleError(w, errors.New("failed to decode SubmitDialogRequest"))
+		HandleErrorWithCode(w, http.StatusBadRequest, "failed to decode SubmitDialogRequest", nil)
 		return
 	}
 
 	var state incident.DialogState
 	err := json.Unmarshal([]byte(request.State), &state)
 	if err != nil {
-		HandleError(w, errors.Wrapf(err, "failed to unmarshal dialog state"))
+		HandleErrorWithCode(w, http.StatusBadRequest, "failed to unmarshal dialog state", err)
 		return
 	}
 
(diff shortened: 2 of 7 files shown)


server/api/playbooks.go
```diff
@@ -78,7 +79,8 @@ func (h *PlaybookHandler) createPlaybook(w http.ResponseWriter, r *http.Request)
 	}{
 		ID: id,
 	}
-	ReturnJSON(w, &result)
+	w.Header().Add("Location", fmt.Sprintf("/api/v0/playbooks/%s", pbook.ID))
```
crspeller, line 82: Does this actually get used? What's the point? If it's not actually used it's just going to get out of date.

server/api/api.go
```diff
@@ -42,15 +42,18 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, sourcePlugin
...
 	if err != nil {
 		HandleError(w, errors.Wrapf(err, "unable to marshal json"))
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
-	_, _ = w.Write(jsonBytes)
+	w.WriteHeader(httpStatus)
+	if _, err = w.Write(jsonBytes); err != nil {
+		HandleError(w, err)
```
crspeller, line 54: If this where to actually fail this would probably end up just failing again trying to do the w.Write inside HandleError.
I think we usually just ignore this error because if the write fails there isn't a whole lot we can do. 
We could just log an error here though.

server/api/playbooks.go
agarciamontoro: We are not using this right now, but the API is intended for public consumption, so this is just to adhere to the standards: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

server/api/api.go
agarciamontoro: You're right! Let me fix it.

agarciamontoro: I took a look and I think it's ok for use to `HandleError` here, as that function is already logging the error.

server/api/incidents.go
```diff
@@ -103,11 +103,22 @@ func (h *IncidentHandler) createIncidentFromPost(w http.ResponseWriter, r *http.
...
+		return
+	}
+
+	if errors.Is(err, incident.ErrMalformedIncident) {
+		HandleErrorWithCode(w, http.StatusBadRequest, "unable to create incident", err)
+		return
+	}
+
 	if err != nil {
 		HandleError(w, errors.Wrapf(err, "unable to create incident"))
```
cpoile, line 123: (As an aside) I thought we chose not to return internal errors? Maybe we haven't converted all our code to that yet, but the reason was: we didn't want to accidentally leak internal error details. I think we should probably move over to that way of doing things, or else it's pretty tough to make sure we never leak anything important.

We could make that change within the `HandleErrorWithCode` function, by not sending the error back over the response. Instead, log the error for the administrator. Unless we can think of a way to easily distinguish between errors that are okay for the user to see, vs errors that we should not leak. (but I can't think of an easy way to do that for now and into the future)

cpoile (approved):
Looks great! Just one broader question, which could be handled in this PR or we can make a new PR -- if you agree with the question.

server/api/incidents.go
agarciamontoro: > We could make that change within the HandleErrorWithCode function, by not sending the error back over the response. Instead, log the error for the administrator.

I like this option. I think we should handle this separately, though, as there are quite a few places to review. I created a ticket: https://mattermost.atlassian.net/browse/MM-29868

> Unless we can think of a way to easily distinguish between errors that are okay for the user to see, vs errors that we should not leak. (but I can't think of an easy way to do that for now and into the future)

I have not thought this through a lot, but: maybe we can have some basic `ErrPublic` and `ErrPrivate` empty errors and wrap the appropriate one with the errors we generate to decide if we want to leak or not. Or maybe we can just have our own type implementing `Error` that contains fields for public and admin-only errors.

server/api/api.go
crspeller: Ya. It's just a bit strange to call it here if that's the only thing we are getting from it. 0/5

crspeller (approved):

PR #290, opened 2020-09-19
MM-26944 Implement PlaybooksService

JeremyShih:
#### Summary

Implement PlaybooksService

#### Ticket Link

mattermost/mattermost-server#15412
JIRA: [MM-26944](https://mattermost.atlassian.net/browse/MM-26944)

I am not sure which fields are required and how to verify it so I decided to create the pull request with current progress.
Please help check with it and tell me what you think about the ticket.

PR #349, opened 2020-10-19
Use replace for local modules.

crspeller:
#### Summary
Uses replace directives for the new local module. Should fix the issues in #290


cpoile (approved):

agarciamontoro (approved):
Oh nice!

PR #353, opened 2020-10-21
Update telemetry formatting

justinegeffen:
Fixed telemetry table formatting which was off.


agarciamontoro (commented):
Looks like the asterisks render as asterisks, not as a list. I'm not sure if that's what's intended:
image

justinegeffen:
I'm seeing some inconsistencies in the GitHub preview vs. what renders in GitBook. So I'm just testing this out for now. It's a bit of a pain to do it this way but it's easy to change if it doesn't render as expected.

agarciamontoro:
Oh ok, it makes sense!

agarciamontoro (approved):

PR #355, opened 2020-10-23
[MM-29944] Add plugin icon and metadata

hanzei:
#### Summary
This PR adds a plugin icon that is shown in the Marketplace modal

### Bright theme
Screenshot from 2020-10-23 13-07-13

### Dark theme
Screenshot from 2020-10-23 11-29-47

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29944


abhijit-singh (approved):
Looks good for now.

plugin.json
```diff
@@ -2,6 +2,10 @@
     "id": "com.mattermost.plugin-incident-response",
     "name": "Incident Response",
     "description": "This plugin allows users to coordinate and manage incidents within Mattermost.",
+    "homepage_url": "https://github.com/mattermost/mattermost-plugin-incident-response/",
+    "support_url": "https://github.com/mattermost/mattermost-plugin-incident-response//issues",
```
crspeller, line 6:
```suggestion
    "support_url": "https://github.com/mattermost/mattermost-plugin-incident-response/issues",
```
(typo on the last slash)

crspeller (approved):
Pre-approved, one typo.

plugin.json
hanzei: Good catch :+1:

agarciamontoro (approved):
We still need to review the icon in the circle, but that can wait. Looks good to me!

hanzei:
@itao What do you think about this change?

itao:
Hmm I would've expected the dark theme icon to have inverse colours rather than a white circle. @hanzei Is showing a different icon file in dark theme supported? And if so, @abhijit-singh could you help us make an inverse version?

hanzei:
> Hmm I would've expected the dark theme icon to have inverse colours rather than a white circle. @hanzei Is showing a different icon file in dark theme supported? And if so, @abhijit-singh could you help us make an inverse version?

@itao Good point. Using a white circle vs inverting the colours was discussed when the Marketplace was introduced and the decision was to use the white circle for every icon as for themes that are between bright and dark it was hard to properly invert the colors. Therefore showing a different icon for different themes is currently not supported.

Please note that the "white circle" approach is used in other areas as well like the slash command icon.

itao:
@hanzei Ah gotcha, thanks for explaining that to me.

itao (approved):

PR #357, opened 2020-10-24
MM-26942 Incidents Service

christian-lim:
#### Summary
This pull requests implements the rest of the incidents service. I've been struggling finding some time to finish the test updates, but I wanted to get the PR going for the incidents service itself. Will be posting a question on an issue I've been having with replacing the HTTP calls in the tests.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/15786


agarciamontoro:
I'm gonna review this PR tomorrow! Sorry for the delay, @christian-lim :pray:

christian-lim:
Addressed PR comments. Thanks for the in-depth explanations!
Seems like the failing tests have something to do with playbooks, which I don't think my changes would've impacted.

jasonblais:
Thanks @christian-lim for the updates! @agarciamontoro FYI on above, ready for a review :)

agarciamontoro:
Thank you @christian-lim! I only see this one commit (https://github.com/mattermost/mattermost-plugin-incident-management/pull/357/commits/e9d575a8520dbf7a9d1835c45a7d4ed0e38bb22f). Is there maybe something else you did not push?

christian-lim:
Hello, thanks for being so patient! I apologize, I didn't actually push the changes. I just pushed the commit that addresses all the PR comments.

client/incidents.go
```diff
@@ -112,41 +106,70 @@ func (s *IncidentsService) Create(ctx context.Context, opts IncidentCreateOption
...
 // Update an incident.
-func (s *IncidentsService) Update(ctx context.Context, incidentID string, opts IncidentUpdateOptions) (*Incident, error) {
-	return nil, errors.New("not implemented")
+func (s *IncidentsService) Update(ctx context.Context, incidentID string, opts incident.UpdateOptions) (*Incident, error) {
+	incidentURL := fmt.Sprintf("incidents/%s", incidentID)
+	incidentUpdateRequest := Incident{
+		CommanderUserID: *opts.CommanderUserID,
+	}
+
+	req, err := s.client.NewRequest(http.MethodPatch, incidentURL, incidentUpdateRequest)
```
agarciamontoro, line 150: We can directly pass the `opts` param as the last parameter in `NewRequest` (so we don't need to build an Incident with that info). We need to keep the pointers as received, even if they are nil.
```suggestion
	req, err := s.client.NewRequest(http.MethodPatch, incidentURL, opts)
```

client/incidents.go
```diff
@@ -112,41 +106,70 @@ func (s *IncidentsService) Create(ctx context.Context, opts IncidentCreateOption
...
+	req.Header.Add("Mattermost-User-ID", *opts.CommanderUserID)
```
agarciamontoro, line 151: We can remove this line
```suggestion
```

client/incidents.go
```diff
@@ -115,41 +114,70 @@ func (s *IncidentsService) Create(ctx context.Context, opts IncidentCreateOption
...
+	_, err = s.client.Do(ctx, req, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return nil, nil
 }
 
 // List the incidents.
 func (s *IncidentsService) List(ctx context.Context, opts IncidentListOptions) (*IncidentList, error) {
```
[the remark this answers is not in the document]
agarciamontoro: The comment above still applies. We need to add the `url` tags to `incident.GetIncidentsResults` so the `addOptions` function works as expected: see [the comment in that function](https://github.com/mattermost/mattermost-plugin-incident-management/blob/master/client/client.go#L208-L228).

server/incident/incident.go
```diff
@@ -102,7 +102,8 @@ type StatusPost struct {
 }
 
 type UpdateOptions struct {
-	ActiveStage *int `json:"active_stage"`
+	CommanderUserID *string `json:"commander_user_id"`
+	ActiveStage     *int    `json:"active_stage"`
```
agarciamontoro, line 106: We can revert this change, as [the current implementation](https://github.com/mattermost/mattermost-plugin-incident-management/blob/master/server/api/incidents.go#L140-L168) only changes the active stage.
```suggestion
	ActiveStage *int `json:"active_stage"`
```

agarciamontoro (changes requested):
Cool, this is looking much better now! :tada:

I noticed that there are still some unresolved comments. I re-wrote the comments with some more information below.

Thank you for your work on this, @christian-lim!

jfrerich:
Thanks for the PR, @christian-lim!  Do you need any assistance or clarification from the PR reviews?

christian-lim:
> Thanks for the PR, @christian-lim! Do you need any assistance or clarification from the PR reviews?

Nope, addressed comments, but I do need to remove a gmail filter...

server/incident/incident.go
```diff
@@ -123,6 +124,7 @@ type Metadata struct {
 // GetIncidentsResults collects the results of the GetIncidents call: the list of Incidents matching
 // the HeaderFilterOptions, and the TotalCount of the matching incidents before paging was applied.
 type GetIncidentsResults struct {
+	URL        string     `json:"url"`
```
agarciamontoro, line 127: Sorry for the confusion in my previous comment. When I said "url tags" I meant adding tags to the existing fields in the struct using the `url` key. I also meant to do it in the `HeaderFilterOptions` struct (my bad!). We need that information in the struct because the `addOptions` function uses it to create the url with the specified query parameters in the struct (you can read more about this in https://pkg.go.dev/github.com/google/go-querystring/query#Values).

In short, we need to remove this new `URL` field from the `GetIncidentsResults` struct, and then change the `HeaderFilterOptions` struct ([in the incident/filter_options.go file](https://github.com/mattermost/mattermost-plugin-incident-management/blob/master/server/incident/filter_options.go#L26-L55)) into this:
```go
// HeaderFilterOptions specifies the optional parameters when getting headers.
type HeaderFilterOptions struct {
	// Gets all the headers with this TeamID.
	TeamID string `url:"team_id,omitempty"`

	// Pagination options.
	Page    int `url:"page,omitempty"`
	PerPage int `url:"per_page,omitempty"`

	// Sort sorts by this header field in json format (eg, "create_at", "end_at", "name", etc.);
	// defaults to "create_at".
	Sort string `url:"sort,omitempty"`

	// Direction orders by Asc (ascending), or Desc (descending); defaults to desc.
	Direction string `url:"direction,omitempty"`

	// Status filters by All, Ongoing, or Ended; defaults to All.
	Status Status `url:"status,omitempty"`

	// CommanderID filters by commander's Mattermost user ID. Defaults to blank (no filter).
	CommanderID string `url:"commander_user_id,omitempty"`

	// MemberID filters incidents that have this member. Defaults to blank (no filter).
	MemberID string `url:"member_id,omitempty"`

	// SearchTerm returns results of the search term and respecting the other header filter options.
	// The search term acts as a filter and respects the Sort and Direction fields (i.e., results are
	// not returned in relevance order).
	SearchTerm string `url:"search_term,omitempty"`
}
```

The values of the `url` tags come from the parameters retrieved in [this function](https://github.com/mattermost/mattermost-plugin-incident-management/blob/master/server/api/incidents.go#L948-L1005).

agarciamontoro (changes requested):
Thank you for the changes, @christian-lim! :tada:

One last round of changes and I think we're ready to go!

server/incident/incident.go
christian-lim: Pushed, thanks for the docs!

agarciamontoro (commented):
Great, thank you @christian-lim :rocket: 

I think the code looks good now! But we need to test it first, of course. I re-read your PR description and you said that you would "be posting a question on an issue with replacing the HTTP calls in the tests".

What was the issue? We can sort it out together!

lieut-data:
@agarciamontoro & @crspeller, I've made some proposed changes to this PR that dramatically reworks the `client` package. A few highlights:
* To focus the effort, I'm removing all "unsupported" functionality from the `client` package: we'll add it back later.
* The client package no longer tests itself: instead, the `server/api` package tests /using/ the client.
* I'm /duplicating/ the model structures into the `client` package: this allows the use of same from a single package (vs. having to pull in a different package defining them), /and/ exposes them for documentation purposes, something our enterprise license restricts outside of the special `client` module.
* I've streamlined constructing the client on the assumption it's almost always going to be used with a `Client4` instance.

There are probably other improvements to make, but since we're `< 1.0` for this module, I'd like to get alignment on the proposed changes and then plow more changes into subsequent PRs.

server/api/api.go
```diff
@@ -10,11 +9,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type contextKey string
-
-// PluginIDContextKey Key used to store the sourcePluginID for http requests.
-const PluginIDContextKey = "plugin_id"
-
```
agarciamontoro, line 17: These changes were already done in another PR, right?

server/api/api.yaml
```diff
@@ -549,7 +549,7 @@ paths:
       x-codeSamples:
         - lang: curl
           source: |
-            curl -X PATCH 'http://localhost:8065/plugins/com.mattermost.plugin-incident-response/api/v0/incidents/mx3xyzdojfgyfdx8sc8of1gdme' \
+            curl -X PATCH 'http://localhost:8065/plugins/com.mattermost.plugin-incident-management/api/v0/incidents/mx3xyzdojfgyfdx8sc8of1gdme' \
```
agarciamontoro, line 552: Good catches!

client/unexport_test.go
```diff
@@ -0,0 +1,4 @@
+package client
+
+var BuildAPIURL = buildAPIURL
+var NewClient = newClient
```
agarciamontoro, line 4: What is this for?

server/api/api_test.go
```diff
@@ -30,3 +35,49 @@ func TestAPI(t *testing.T) {
...
+	var apiIncident icClient.Incident
+
+	incidentBytes, _ := json.Marshal(internalIncident)
+	err := json.Unmarshal(incidentBytes, &apiIncident)
+	if err != nil {
+		panic(err)
+	}
+
+	return apiIncident
+}
```
agarciamontoro, line 59: Why panic here (and in the following ones) instead of making this a `t.Helper()` and calling `require.NoError(t, err)`?

server/api/incidents_test.go
```diff
@@ -583,44 +577,12 @@ func TestIncidents(t *testing.T) {
...
-		testreq.Header.Add("Mattermost-User-ID", "testUserID")
-		require.NoError(t, err)
-		handler.ServeHTTP(testrecorder, testreq, "testpluginid")
-
-		resp := testrecorder.Result()
-		defer resp.Body.Close()
-		assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
-	})
-
-	t.Run("create invalid incident - channel id already set", func(t *testing.T) {
```
agarciamontoro, line 600: Is this test gone?

agarciamontoro (changes requested):
Nice! :tada: I really like how readable the tests are now :pray:

Apart from the individual comments below, I have a couple of general questions:

1. - `Incident.Update` is not yet implemented. Is that intended?
2. - I guess that the panic in the `to{API,Internal}Incident` functions is what makes the struct duplication safe, right?
3. - There are a couple of things here from other PRs, if I'm not mistaken. Namely, the removal of the subscriptions and the context stuff in the API. Is that intended?
4. - Not for this PR, but worth discussing: `assert.Equal(t, testIncident, toInternalIncident(*resultIncident))` will fail if we ever decide to make the two structs (the one in the API and the normal one) diverge. Any plans for that?

client/doc.go
```diff
@@ -1,104 +1,5 @@
...
-		}
-
-		allIncidents = append(allIncidents, list.Items...)
-		if !list.HasMore {
-			break
-		}
-	}
-
-*/
+// Package client provides an HTTP client for using the Incident Management API.
```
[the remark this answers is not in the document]
lieut-data: I /think/ `pkg.go.dev` will do this automatically for us, pulling up examples that aren't bound to a particular service directly into the root. (If it doesn't, I'm happy to revisit this!)

client/error_response.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// Unwrap exposes the underlying error of an ErrorResponse.
+func (r *ErrorResponse) Unwrap() error {
+	return r.Err
+}
+
+// Error describes the error from the API request.
+func (r *ErrorResponse) Error() string {
+	return fmt.Sprintf("%v %v [%d]: %v", r.Method, r.URL, r.StatusCode, r.Err)
```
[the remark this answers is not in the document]
lieut-data: Good question: fixed!

client/go.mod
```diff
@@ -4,10 +4,11 @@ go 1.15
 
 require (
 	github.com/google/go-querystring v1.0.0
-	github.com/mattermost/mattermost-plugin-incident-collaboration v1.0.0
-	github.com/mattermost/mattermost-server/v5 v5.28.0
+	github.com/mattermost/mattermost-plugin-incident-collaboration v1.4.0
+	github.com/mattermost/mattermost-server/v5 v5.31.0
```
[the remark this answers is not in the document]
lieut-data: Ah, good catch -- I think this got upgraded in one of my many attempts at consolidation, but I didn't think to revisit the requirement. (It doesn't strictly hurt: in particular, we're still backwards compatible with v5.28.0 at an API level, but I'm also not opposed to just leaving the existing version intact.)

client/incidents.go
```diff
@@ -7,166 +7,109 @@ import (
...
 }
 
-// Delete an incident.
-func (s *IncidentsService) Delete(ctx context.Context, incidentID string) (*Incident, error) {
-	return nil, errors.New("not implemented")
-}
+// Create an incident.
+func (s *IncidentsService) Create(ctx context.Context, opts IncidentCreateOptions) (*Incident, error) {
+	incidentURL := "incidents"
+	req, err := s.client.newRequest(http.MethodPost, incidentURL, opts)
```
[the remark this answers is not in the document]
lieut-data: I'm inclined to keep the client as thin as possible so as to give the server the maximum flexibility on required options, etc.

server/api/api.go
lieut-data: Indeed, sorry for the distraction here: I tried to pull out some of the potentially contentious changes into discrete PRs, but didn't want to rebase this one.

client/unexport_test.go
lieut-data: This is "Go magic" -- all `_test.go` files are unconditionally excluded from the exposed API, even if they otherwise define what would be public methods and such. But they *are* used when running tests. By exploiting this, and putting this particular file into the `client` package, I can access the internals that are `newClient` and `buildAPIURL` from the `client_test` package without exposing those to the rest of the world.

server/api/api_test.go
lieut-data: Honestly, I wanted to just ignore the err altogether, given it's not strictly relevant to the scope of the tests. But the linter wouldn't let me, so I just panicked instead. I'd prefer not to have to also pass around `* testing.T`, but I'm only 0/5 here.

server/api/incidents_test.go
lieut-data: It is: the payload no longer accepts a `ChannelID` (since we no longer accept a "raw" incident).

lieut-data:
>     1. `Incident.Update` is not yet implemented. Is that intended?

Yep: mostly wanted to align on the structure before I spent too much time fleshing out the specifics.

>     2. I guess that the panic in the `to{API,Internal}Incident` functions is what makes the struct duplication safe, right?

By "safe", I'm assuming you mean as a defence against diverging? Now that you mention it, I recall our explicitly discussing a mechanism to ensure this but I forgot altogether to add that here. I think it's still viable to do that explicitly, but it's possible this code also guards against that by luck. I'd love to make this explicit in a subsequent PR either way.

>     3. There are a couple of things here from other PRs, if I'm not mistaken. Namely, the removal of the subscriptions and the context stuff in the API. Is that intended?

Indeed, apologies for the distraction: I didn't want to rebase this particular PR, but did want to get the meat of the changes out for consideration early.

>     4. Not for this PR, but worth discussing: `assert.Equal(t, testIncident, toInternalIncident(*resultIncident))` will fail if we ever decide to make the two structs (the one in the API and the normal one) diverge. Any plans for that?

Good point. I can see how we might want to explicitly diverge the public definition of an incident from the private one. Ok to cross that bridge if/when we hit it?

client/incidents.go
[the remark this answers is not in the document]
agarciamontoro: It makes sense!

client/unexport_test.go
agarciamontoro: Oooh, I see, that's clever! Thanks for the clarification!

server/api/api_test.go
agarciamontoro: I think not ignoring it is a good thing, as we would otherwise let potential syncing problems between the types slip through. And I think t]at passing `*testing.T` is not too bad if that let us mark this semantically as a helper.

server/api/incidents_test.go
agarciamontoro: Oooh, I see, thanks for the clarification!

agarciamontoro:
> By "safe", I'm assuming you mean as a defence against diverging? Now that you mention it, I recall our explicitly discussing a mechanism to ensure this but I forgot altogether to add that here. I think it's still viable to do that explicitly, but it's possible this code also guards against that by luck. I'd love to make this explicit in a subsequent PR either way.

Yes, that's what I meant! I think the key for that is in the `to{API,Internal}Incident` functions, which will fail when marshalling and unmarshalling the type if they are not the same. That's the only comment I left unresolved, as I think we can convert them into `t.Helper()`s so we can fail with a good error message explaining what happened. 0/5 doing that here or in subsequent PRs.

> Good point. I can see how we might want to explicitly diverge the public definition of an incident from the private one. Ok to cross that bridge if/when we hit it?

Absolutely!

agarciamontoro (approved):
Yay, this looks amazing :tada:

The comment on converting the `to{API,Internal}Incident` functions to helpers is non-blocking, so approving now!

crspeller (approved):

lieut-data:
Prapti, the bulk of this PR affects not the core product, but the updated Go client interface to streamline using the REST API. I'm not expecting that to be qualified, per se -- it's still under development and will continue to be versioned as "unstable" separate from the core product. Also, the unit tests already cover much of the requisite functionality. Nevertheless, I'd love a quick smoketest before merging to master.

prapti (approved):
Thank you for the context @lieut-data! That helped with understanding the testing expectations.
Smoke tested and verified that the plugin still works as intended and didn't come across anything that is 'broken'. 🎉 
Thank you for your contribution and patience on this @christian-lim!
Approving for merge!

lieut-data:
Thanks again, @christian-lim!

PR #358, opened 2020-10-27
MM-29867: Commands to generate test data

agarciamontoro:
#### Summary
This PR adds two new slash commands:
- `/incident test create-incident`, which accepts a playbook ID (that can be chosen from the playbooks the user is a member of, using the autocomplete system), a timestamp and an incident name. It creates an ongoing incident whose creation date is set to the specified timestamp.
- `/incident test bulk-data`, which accepts a number of ongoing incidents, a number of ended incidents, a begin and an end date, and an optional seed. It creates as many ongoing and ended incidents as specified, all of them with their creation date randomly picked between the begin and end dates. The seed, if available, is used to get reproducible results. The names of the incidents are randomly chosen from a list of incident names and a list of fake company names I defined in the code.

Other things I did while I was here:
- I fixed the `sqlstore.NukeDB` function, which failed with a foreign key error when trying to delete `IR_Playbook` before
`IR_PlaybookMember`.
- I moved the `st` command as a subcommand of `/incident test`, for the sake of consistency.
- I excluded `G404: Insecure random number source (rand)` rule from [gosec](https://github.com/securego/gosec): it was complaining because I am using `math/rand` instead of `crpyto/rand` to generate the random numbers needed in the bulk data generation. Of course, we do not need `crypto/rand` for such a task. And, if we ever need to do something crypto-related, I think we will know by ourselves that we need to use `crypto/rand`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29867


agarciamontoro:
E2E tests are failing in CI but not locally. I'll investigate.

server/api/playbooks.go
```diff
@@ -208,6 +209,36 @@ func (h *PlaybookHandler) getPlaybooks(w http.ResponseWriter, r *http.Request) {
 	ReturnJSON(w, playbookResults, http.StatusOK)
 }
 
+func (h *PlaybookHandler) getPlaybooksAutoComplete(w http.ResponseWriter, r *http.Request) {
```
crspeller, line 212: I think this needs a permission check.

server/incident/incident.go
```diff
@@ -234,6 +235,9 @@ type Service interface {
 
 	// NukeDB removes all incident related data.
 	NukeDB() error
+
+	// ChangeCreationDate changes the creation date of the specified incident.
+	ChangeCreationDate(incidentID string, creationTimestamp time.Time) error
```
crspeller, line 240: Seems like a strange function to have. Maybe a function to create with a specific timestamp would make more sense? 0/5 probably fine as is.

crspeller (changes requested):

itao:
@agarciamontoro Could you request review from @muratbayan and @sadohert as they'll be using it in the field?

itao (approved):

server/incident/incident.go
agarciamontoro: I'd argue that they are equally weird, actually. The only reason I did it this way was to separate both calls, so there is only one way of creating an incident. As we're both 0/5 on this, I'll leave it as it is just not to add any more work (even if it's negligible).

server/api/playbooks.go
agarciamontoro: You're right! Fixed in https://github.com/mattermost/mattermost-plugin-incident-management/pull/358/commits/bc7d032c8d4ea75ae8551aebf22ef19ebbcc95d4

agarciamontoro:
I know now why the E2E tests are failing only in CI (we have an environment variable overriding the server config). I'm waiting on some configuration we need to do in CircleCI itself, but consider it fixed!

crspeller (approved):

server/command/command.go
```diff
@@ -746,7 +770,7 @@ func (r *Runner) actionSelftest(args []string) {
 
 	if len(args) != 2 || args[0] != confirmPrompt || args[1] != "SELF-TEST" {
 		r.postCommandResponse("Are you sure you want to self-test (which will nuke the database and delete all data -- instances, configuration)? " +
-			"All incident data will be lost. To self-test, type `/incident st CONFIRM SELF-TEST`")
+			"All incident data will be lost. To self-test, type `/incident test self CONFIRM SELF-TEST`")
```
cpoile, line 773: maybe:

```suggestion
			"All incident data will be lost. To self-test, type `/incident test self CONFIRM TEST SELF`")
```

along with the change the check would require above

server/command/command.go
```diff
@@ -960,6 +984,348 @@ And... yes, of course, we have emojis
...
 }
 
+func (r *Runner) actionTest(args []string) {
+	if r.pluginAPI.Configuration.GetConfig().ServiceSettings.EnableTesting == nil ||
+		!*r.pluginAPI.Configuration.GetConfig().ServiceSettings.EnableTesting {
+		r.postCommandResponse("Setting `EnableTesting` must be set to `true` to run the test command.")
+		return
+	}
+
+	if !r.pluginAPI.User.HasPermissionTo(r.args.UserId, model.PERMISSION_MANAGE_SYSTEM) {
```
cpoile, line 994: Will Murat, Stuart, Paul, etc have system admin privileges when they run the test command?

server/command/command.go
```diff
@@ -960,6 +984,348 @@ And... yes, of course, we have emojis
...
+			return
+		}
+
+		seed = parsedSeed
+	}
+
+	r.generateTestData(ongoing, ended, begin, end, seed)
+}
+
+var fakeCompanyNames = []string{
```
cpoile, line 1123: nice. :)

server/command/command.go
```diff
@@ -960,6 +984,348 @@ And... yes, of course, we have emojis
...
+	"Mertz Group",
+	"Welch LLC",
+	"Baumbach Group",
+	"Ward - Schmitt",
+	"Romaguera Group",
+	"Hickle - Kemmer",
+	"Stewart Corp",
+}
+
+var incidentNames = []string{
```
cpoile, line 1166: Did you write these? Nice job

cpoile (approved):
Nice work Alejandro -- this is fiddly stuff to code, but super useful, thanks!

server/command/command.go
agarciamontoro: I was assuming that, but maybe it's not true. @muratbayan @sadohert, will you have system admin privileges when running this command to generate the data?

server/command/command.go
agarciamontoro: Yes! It was fun :joy:

server/command/command.go
agarciamontoro: It makes it more consistent, thanks for the suggestion!

agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-management/pull/358/commits/8e7adeb91db64c6c7da27e29f94ff891bf5bddaf

server/command/command.go
agarciamontoro: Confirmed offline that they will have system admin privileges when they run the test command.

muratbayan (approved):
Reviewed today with Alejandro as I don't have a local setup, the behaviour is great.
Approved

agarciamontoro:
@sadohert are you ok with merging this?

sadohert:
I'm good to merge this.  Functionality looks great!

Probably doesn't need to block this PR, but as discussed in the channel, I was having an issue with a simple `make` of the plugin.  `make deploy` got me up and running in my local dev environment (after make noticed I didn't have my environment details setup correctly and notified me).
image

sadohert (approved):
I didn't review code.  Only tested the functionality with:
```
/incident test bulk-data 10 3 2020-01-31 2020-11-22 2
```

AND
```
/incident test create-incident 6utgh6qg7p8ndeef9edc583cpc 2020-11-23 PR-Testing
```

And I surfed around to check out what was produced.  Nice job!

agarciamontoro:
@justinegeffen is it ok if I merge this PR? Or do you want to complete the docs before?

justinegeffen:
@agarciamontoro, I am working on the docs at the moment so you can go ahead and merge. :)

PR #359, opened 2020-10-27
MM-29754 - Revisit telemetry event structure

cpoile:
#### Summary
- Based on the discussion with Eric N. the new telemetry looks like this:
  - we'll use a `type` for each major category of events that we're tracking. Within each `type` we have a `properties` bag that we put as much info into as seems useful, such as:
    - the `action` which is the event "subtype" to narrow down what this event is reporting
- So, we now have the following event `type`s and `action`s:
  - `type`: `incident`
    - `action`s: `create`, `end`, `restart`, `change_stage`, ` change_commander`
  - `type`: `tasks`
    - `action`s: `add_task`, `remove_task`, `rename_task`, `modify_task_state`, `move_task`, `set_assignee_for_task`, `run_task_slash_command`
      - note that the actions include the term `task`--this is because if the action were `add`, then the action would be "adding a set of tasks" -- which we are not doing. Instead, we are "adding a task within a set of tasks".
  - `type`: `playbook`
    - `action`s: `create`, `update`, `delete`

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29754

### DIFF
--- a/server/incident/incident.go
+++ b/server/incident/incident.go
@@ -159,8 +159,8 @@ type Service interface {
 	// GetIncidents returns filtered incidents and the total count before paging.
 	GetIncidents(requesterInfo RequesterInfo, options HeaderFilterOptions) (*GetIncidentsResults, error)
 
-	// CreateIncident creates a new incident.
-	CreateIncident(incdnt *Incident, public bool) (*Incident, error)
+	// CreateIncident creates a new incident. userID is the user who initiated the CreateIncident.
+	CreateIncident(incdnt *Incident, userID string, public bool) (*Incident, error)
 
 	// OpenCreateIncidentDialog opens an interactive dialog to start a new incident.
 	OpenCreateIncidentDialog(teamID, commanderID, triggerID, postID, clientID string, playbooks []playbook.Playbook, isMobileApp bool) error
@@ -266,15 +266,16 @@ type Store interface {
 }
 
 // Telemetry defines the methods that the ServiceImpl needs from the RudderTelemetry.
+// Unless otherwise noted, userID is the user initiating the event.
 type Telemetry interface {
-	// CreateIncidenttracks the creation of a new incident.
-	CreateIncident(incident *Incident, public bool)
+	// CreateIncident tracks the creation of a new incident.
+	CreateIncident(incident *Incident, userID string, public bool)
 
 	// EndIncident tracks the end of an incident.
-	EndIncident(incident *Incident)
+	EndIncident(incident *Incident, userID string)
 
 	// RestartIncident tracks the restart of an incident.
-	RestartIncident(incident *Incident)
+	RestartIncident(incident *Incident, userID string)
 
 	// ModifyCheckedState tracks the checking and unchecking of items.
 	ModifyCheckedState(incidentID, userID, newState string, wasCommander, wasAssignee bool)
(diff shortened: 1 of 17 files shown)


server/incident/incident.go
```diff
@@ -277,31 +277,31 @@ type Telemetry interface {
 	// RestartIncident tracks the restart of an incident.
 	RestartIncident(incident *Incident, userID string)
 
+	// ChangeCommander tracks changes in commander.
+	ChangeCommander(incident *Incident, userID string)
+
+	// ChangeCommander tracks changes in stage
```
agarciamontoro, line 283:
```suggestion
	// ChangeStage tracks changes in stage
```

server/telemetry/rudder.go
```diff
@@ -23,21 +23,25 @@ type RudderTelemetry struct {
...
+	actionChangeCommander = "change_commander"
+
+	eventTasks                = "tasks"
+	actionAddTask             = "add_task"
+	actionRemoveTask          = "remove_task"
+	actionRenameTask          = "rename_task"
+	actionModifyTaskState     = "modify_task_state"
+	actionMoveTask            = "move_task"
+	actionSetAssigneeForTask  = "set_assignee_for_task"
+	actionRunTaskSlashCommand = "run_task_slash_command"
```
agarciamontoro, line 40: This is not important, actually, because we only define it here and never write that string again, but: does it make sense to remove the `_task` part of the name?

server/telemetry/rudder.go
```diff
@@ -23,21 +23,25 @@ type RudderTelemetry struct {
...
+
+	eventPlaybook = "playbook"
+	actionUpdate  = "update"
+	actionDelete  = "delete"
```
agarciamontoro, line 44: A maybe more important thing: do you think naming the actions here as `action<EventName><Action>` as in `actionPlaybookUpdate`, so it's easier to skim through them and now what event they're referring to?

crspeller (approved):

server/telemetry/rudder.go
agarciamontoro: Oh wait, after reading more code it looks like we're using the same action, as `create`, for different events. I guess that's intended?

agarciamontoro (approved):
Looks good to me! I left a couple of nits below, but non-blocking at all, so feel free to dismiss them.

server/telemetry/rudder.go
cpoile: I wrote about that in the PR notes about -- see that and then ask again if I wasn't clear (or if you disagree :) )

justinegeffen:
Thanks, @cpoile. I've started updating https://github.com/mattermost/mattermost-plugin-incident-response/pull/360.

server/telemetry/rudder.go
cpoile: Reusing create is on purpose, and we don't have to be more specific because the action is always a part of a particular event.

itao (commented):
@cpoile Does `modify_task_state` differentiate between checking and unchecking the task? If not, it'd be worth using two different events.

cpoile:
@itao yes, it will be an empty string for an unchecked state, and a "closed" for checked state.

itao (approved):

PR #371, opened 2020-10-30
MM-30135 - Fix: Playbook update in 1.1 is a bit broken

cpoile:
#### Summary
- We were trying to fetch json, but the update endpoint is now returning nothing.
- Fixing that, and a bunch of other little mistakes we were making in the client file. We probably need to clean `client.ts` up a bit, but leaving that for next version.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-30135


crspeller (approved):

agarciamontoro (approved):
Looks good, thanks!

PR #374, opened 2020-11-02
MM-29706 - update status dialog; recent update post

cpoile:
#### Summary
- Next piece of the status RHS epic
- `/incident status`, or clicking on the footer button, opens the dialog:
image
- filling out the message and pressing `Update Status` will post it to the channel:
image
- The status posts are kept in a list in the incident. Not in a separate table, that's overkill since we never have to query based on postID -- just joined together and pulled apart with `strings.Fields` when saving/loading from the store.
- When we're handling displaying the most recent non-deleted status in the RHS, and the finding the "last updated status" time, we'll go through the list of status postIDs backwards (most recent first) and use the first non-deleted post (if any).
- I've only done the basic tests, I'm waiting for the whole new RHS to be ready before doing the e2e tests that this will need. I think it's a waste of effort now to do mocked api tests, imo, unless there's some logic in the api layer that we need to test (which isn't the case here)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29706


server/incident/service.go
```diff
@@ -269,6 +278,63 @@ func (s *ServiceImpl) OpenEndIncidentDialog(incidentID, triggerID string) error
...
+	incidentToModify.StatusPostsIDs = append(incidentToModify.StatusPostsIDs, postID)
+	if err = s.store.UpdateIncident(incidentToModify); err != nil {
+		return errors.Wrapf(err, "failed to update incident")
+	}
+
+	s.poster.PublishWebsocketEventToChannel(incidentUpdatedWSEvent, incidentToModify, incidentToModify.ChannelID)
+	s.telemetry.UpdateStatus(incidentToModify, userID)
+
+	return nil
+}
```
agarciamontoro, line 336: I think we should check the permissions here: if I'm logged in as a valid Mattermost user and I send a POST request to the `/incidents/{id}/update-status-dialog` endpoint, this function will be directly called. As I can send the request to any incident ID, this would let me update the status in any incident, regardless of my membership to the channel. (If I'm not missing something here)

server/sqlstore/migrations.go
```diff
@@ -242,6 +242,23 @@ var migrations = []Migration{
...
 
+			return nil
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.3.0"),
+		toVersion:   semver.MustParse("0.4.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DATABASE_DRIVER_MYSQL {
+				if _, err := e.Exec("ALTER TABLE IR_Incident ADD StatusPostsIDsString VARCHAR(4096) DEFAULT ''"); err != nil {
```
agarciamontoro, line 253: This would fit 151 post IDs, right? I'd say that this is enough for normal incidents, but I don't know if we should increase it for incidents that have a high frequency of status updates, and for those more rare incidents that last more time (or that are restarted).

server/sqlstore/incident.go
```diff
@@ -357,8 +361,9 @@ func toSQLIncident(origIncident incident.Incident) (*sqlIncident, error) {
 	}
 
 	return &sqlIncident{
-		Incident:       origIncident,
-		ChecklistsJSON: checklistsJSON,
+		Incident:             origIncident,
+		ChecklistsJSON:       checklistsJSON,
+		StatusPostsIDsString: strings.Join(origIncident.StatusPostsIDs, " "),
```
agarciamontoro, line 366: Nit: our IDs are always 26 characters long. Maybe we can concatenate them all together?

agarciamontoro (changes requested):
This is really cool, thank you! :tada: I left some suggestions below, only one thing blocking for the permissions.
I agree with your opinion on the tests, btw, let's wait for having the whole thing to add proper E2E tests.

server/api/incidents.go
```diff
@@ -60,6 +60,7 @@ func NewIncidentHandler(router *mux.Router, incidentService incident.Service, pl
 	incidentRouterAuthorized.HandleFunc("/restart", handler.restartIncident).Methods(http.MethodPut)
 	incidentRouterAuthorized.HandleFunc("/commander", handler.changeCommander).Methods(http.MethodPost)
 	incidentRouterAuthorized.HandleFunc("/next-stage-dialog", handler.nextStageDialog).Methods(http.MethodPost)
+	incidentRouterAuthorized.HandleFunc("/update-status-dialog", handler.updateStatusDialog).Methods(http.MethodPost)
```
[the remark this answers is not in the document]
cpoile: Yah, personally I think we don't need to document them --- not because of the work (as you say, I can cut and paste most of this) but because we're advertising something that isn't officially part of the public API. That's my 1/5 on it. :)

server/command/command.go
```diff
@@ -586,6 +591,41 @@ func (r *Runner) actionEnd() {
...
+		}
+		r.warnUserAndLogErrorf("Error retrieving incident: %v", err)
+		return
+	}
+
+	if err = permissions.EditIncident(r.args.UserId, incidentID, r.pluginAPI, r.incidentService); err != nil {
+		if errors.Is(err, permissions.ErrNoPermissions) {
+			r.postCommandResponse(fmt.Sprintf("userID `%s` is not an admin or channel member", r.args.UserId))
+			return
+		}
```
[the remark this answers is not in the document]
cpoile: You could also make an execute command call to the API -- which, come to think of it, means we have to double check that we're doing permissions checks on all of our command functions...

server/command/command.go
```diff
@@ -586,6 +591,41 @@ func (r *Runner) actionEnd() {
...
+		r.warnUserAndLogErrorf("Error retrieving incident: %v", err)
+		return
+	}
+
+	if !permissions.CanPostToChannel(r.args.UserId, r.args.ChannelId, r.pluginAPI) {
+		r.postCommandResponse("user does not have permissions to post in this channel")
+	}
```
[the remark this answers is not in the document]
cpoile: Same answer as above :)

server/incident/incident.go
```diff
@@ -15,9 +15,10 @@ const NoActiveStage = -1
 // Incident holds the detailed information of an incident.
 type Incident struct {
 	Header
-	PostID     string               `json:"post_id"`
-	PlaybookID string               `json:"playbook_id"`
-	Checklists []playbook.Checklist `json:"checklists"`
+	PostID         string               `json:"post_id"`
+	PlaybookID     string               `json:"playbook_id"`
+	Checklists     []playbook.Checklist `json:"checklists"`
+	StatusPostsIDs []string             `json:"status_posts_ids"`
```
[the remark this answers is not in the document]
cpoile: Nice catch! Thanks!

server/incident/service.go
cpoile: Ahhh, now I remember why I didn't check edit permissions here -- they're checked in the router. :)
I should be checking post permissions though, thanks for reminding me about that.

server/sqlstore/incident.go
cpoile: Yah, screw it, I'm just changing it to being text for both dbs. I'm tired of MySQL's "optimizations" :)

cpoile: nvm, mysql can't have default values for text columns.

server/sqlstore/migrations.go
cpoile: Expanded to 8192 -- 303 updates should be enough... :)

server/sqlstore/incident.go
cpoile: About concatenating -- I think using the join with a space and splitting with fields is less error-prone.

itao (commented):
Let's use `/incident update` instead of `/incident status`. It's a verb that implies user is going to do a thing, whereas status reads kind of ambiguous whether it's fetching info or posting.

This may not be part of this PR, but just in case: the `status`, `message`, and `reminder` fields in the modal should default to the previous value rather than blank values.

server/api/incidents.go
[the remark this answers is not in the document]
agarciamontoro: Sounds good!

server/sqlstore/incident.go
agarciamontoro: Sounds good!

server/incident/service.go
agarciamontoro: Nice, thanks!

agarciamontoro (approved):
Great, thanks for addressing my comments! :tada:

server/command/command.go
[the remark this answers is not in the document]
crspeller: Actually are these checks not redundant since this path just opens the dialog? It doesn't actually perform any action. Maybe leave these off and just check when the action is performed?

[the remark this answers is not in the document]
crspeller: Also even if we want these, I think the "edit incident" check is redundant given the can post to channel below. It's not clear, but edit incident checks read and can post checks write.

crspeller (changes requested):

server/incident/service.go
```diff
@@ -269,6 +278,63 @@ func (s *ServiceImpl) OpenEndIncidentDialog(incidentID, triggerID string) error
...
+	incidentToModify, err := s.store.GetIncident(incidentID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to retrieve incident")
+	}
+
+	if !incidentToModify.IsActive {
+		return ErrIncidentNotActive
+	}
+
+	postID, err := s.poster.PostMessage(incidentToModify.ChannelID, options.Message)
```
agarciamontoro, line 322: @christopher.poile: any reason why it is the bot who posts the update, instead of the user that created it? I just realized that this is in conflict with the mocks, that show the user being the author of the update posts.

agarciamontoro: While doing the E2E tests for the update status feature, I realized that this is probably a must. Otherwise, regular users won't be able to edit/delete updates. (Up until now, I was always logged in as sysadmin, so I could edit/delete any post)

server/command/command.go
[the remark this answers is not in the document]
cpoile: Good point, I removed the can post check -- I think we should still have the edit incident/incident member check (the update dialog will have the previous update text pre-filled in, and we don't want to send that to just anyone).

server/incident/service.go
cpoile: Good question -- given the recent meeting we had about the status update, this will be posted as the user in the incident channel.

cpoile:
I'm leaving the status update post styling for another ticket (https://mattermost.atlassian.net/browse/MM-30551), since that scope has changed and the rest of our PRs are based on this one.

crspeller (approved):

PR #376, opened 2020-11-11
[MM-15772] Showing more user details in the share playbook auto-complete

BenCookie95:
### Summary
This adds more info to the auto-complete dropdown. I found that the profile component has exactly what we need.

**Where to find the change:**
1. Click Create a Playbook
2. On the RHS type in a name

**Expected Result:** We see the profile picture and name (display configuration dependent) in the dropdown suggestions

### Ticket Link
 https://github.com/mattermost/mattermost-server/issues/15772

### DIFF
--- a/tests-e2e/cypress/support/plugin_ui_commands.js
+++ b/tests-e2e/cypress/support/plugin_ui_commands.js
@@ -31,7 +31,7 @@ Cypress.Commands.add('executeSlashCommand', (command) => {
     cy.findByTestId('post_textbox').clear().type(command);
 
     // Using esc to make sure we exit out of slash command autocomplete
-    cy.findByTestId('post_textbox').type('{esc}{esc}{esc}{esc}', {delay: 100}).type('{enter}');
+    cy.findByTestId('post_textbox').type('{esc}{esc}{esc}{esc}', {delay: 120}).type('{enter}');
 });
 
 // Opens incident dialog using the `/incident start` slash command
--- a/tests-e2e/cypress/support/plugin_ui_commands.js
+++ b/tests-e2e/cypress/support/plugin_ui_commands.js
@@ -31,7 +31,7 @@ Cypress.Commands.add('executeSlashCommand', (command) => {
     cy.findByTestId('post_textbox').clear().type(command);
 
     // Using esc to make sure we exit out of slash command autocomplete
-    cy.findByTestId('post_textbox').type('{esc}{esc}{esc}{esc}', {delay: 100}).type('{enter}');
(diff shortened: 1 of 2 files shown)


BenCookie95:
Looking into resolving the failing end-to-end test

lieut-data (approved):
Wow, I didn't realize we had all the parts already in place! Nice investigation work, @BenCookie95!

If you need help with the e2e tests, don't hesitate to ask :)

BenCookie95:
I'm getting unpredictable results with the end-to-end tests. When I run these locally sometimes they all pass, sometimes the `frontstage/incident_creation_spec.js` test fails. On my pipeline the `frontstage/slash_command/commander_spec.js` test fails but it has never failed locally. 

@lieut-data Is this a known issue with the cypress tests for this plugin? I've read a little online and it looks like a timeout issue, I can slightly increase some of the timeouts and see if it helps?

lieut-data:
It's not /supposed/ to be flaky, but sometimes the webapp version into which we're deploying the e2e test changes out from underneath us, and things regress like you're unfortunately running into. For the particular failures you're seeing locally/on CI, does anything stand out in the video recordings of the failures?

lieut-data:
Note that you can see the video recordings on CI from the Artifacts tab:

image

BenCookie95:
So I checked out the videos and compared them to the ones that were generated locally. On circle ci the tests were much quicker and the `Enter` button was being clicked too quickly. It looked like to me that it was clicking `Enter` on the user list selection instead of sending the message, if you slow down the [video](https://8905-243358763-gh.circle-artifacts.com/0/tests-e2e/cypress/videos/frontstage/slash_command/commander_spec.js.mp4) you might see it. 

I increased the slash command delay to 200ms instead of 100ms, this allowed the user selection list to disappear before it sent the message. 

As you can see from the commit history I tried a bunch of different values until I found the lowest delay that would work.

@lieut-data Let me know if this fix is satisfactory!

lieut-data:
That works for me, @BenCookie95! Thanks :)

cpoile (approved):
Awesome work @BenCookie95!

agarciamontoro (approved):
Great work here, that one-line change will have a great impact! :tada:

Also, I was running into the same issue with that E2E test in a couple of other PRs and it was becoming a time sink for me! Thank you for fixing it, @BenCookie95, it saved me a lot of debugging time :open_hands:

lieut-data:
@prapti, heads up that I'm removing `3: QA Review` given the blocking of other PRs. Let's plan to test this post-merge instead.

PR #377, opened 2020-11-13
MM-29673: Recent Update view

agarciamontoro:
#### Summary

latest_update_demo

This adds two components:
- `LatestUpdate`, which implements the recent update view in the RHS.
- `ShowMore`, a refactored copy of the webapp component, now using hooks and styled components. This is used by the `LatestUpdate` component to add a `Show More` button in case the post is too long.

The behaviour of the recent update view is as follows:
- A 'No recent updates' message is shown if there are no updates or all of the previous updates have been deleted.
- The most recent update that is not deleted is otherwise shown, adding an (edited) indicator in case the post was edited.
- Everything happens in real time, edited and deleted posts included.
- The behaviour of the `ShowMore` component is responsive, so if the post fits when the RHS is expanded, it is no longer shown.
- The post markdown is rendered using `formatText` and `messageHtmlToComponent`, already exported by the webapp. Please note that I tried to be as faithful to the mocks as possible. This includes that all markdown headers in the original post are rendered in the RHS with a `font-weight: 600`, but the font size is the same. We can change this easily in the CSS if needed.
- The author's avatar and name are not interactive, but the update timestamp is: it links to the post itself.

A couple of more notes:
- I think it will be easier for the reviewers to go commit by commit. I tried to make them as isolated as possible.
- The base branch is set to the one in #374, because this depends on that code. We can change it as soon as that PR is merged.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29673


agarciamontoro:
Of course, an E2E tests that was passing locally is failing on CircleCI :sob:

itao (approved):
Nice looks sweet! Good job!

server/plugin.go
```diff
@@ -206,3 +207,20 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+			p.API.LogWarn(fmt.Sprintf("unable to get incident from edited post: %v", err))
+		}
+
+		return
+	}
+
+	p.bot.PublishWebsocketEventToChannel(incident.IncidentUpdatedWSEvent, theIncident, theIncident.ChannelID)
+}
+
+func (p *Plugin) MessageHasBeenUpdated(c *plugin.Context, newPost, oldPost *model.Post) {
```
crspeller, line 224: Is this necessary? Shouldn't the regular websocket update events be enough to update the post in the RHS?

crspeller (changes requested):
Looks good! Just one question about how edits are handled.

server/plugin.go
agarciamontoro: @crspeller this is to catch the editing action, that can happen in any post. Here we listen to all those events, filter out the ones that are not updating a status update post and send a websocket update event for the ones that do. Is there a way to catch this editing actions in other way?

crspeller: We already send update websocket events to the client for edits. Can't we just listen for those on the client?

agarciamontoro: @crspeller Oh, and move the filtering logic there? That makes sense. But my gut feeling says that it's less performant to do it in the client. We need to filter somewhere, and I think it's better to do it in the server and let the client listen only to our plugin-specific events. But I'm 2/5 on that. What do you think?

crspeller: I think it would be more performant to do it on the client. Doing it on the server we have to listen to every modified event. On the client we only have to listen to relevant ones.

agarciamontoro: I'm not sure I understand how to listen only to the relevant events. We'll need to listen to `WEBSOCKET_EVENT_POST_EDITED` and filter in the same way, isn't it. Maybe I'm missing something obvious :/

crspeller: Ya that's all I am saying. Just do the filtering there on the client.

webapp/src/components/rhs/latest_update.tsx
```diff
@@ -0,0 +1,174 @@
...
+                        }}
+                    >
+                        {updateTimestamp}
+                    </UpdateTimeLink>
+                </UpdateHeader>
+                <ShowMore
+                    text={latestUpdate.message}
+                >
+                    <UpdateBody>
+                        {messageHtmlToComponent(formatText(latestUpdate.message, markdownOptions), true, {})}
```
cpoile, line 165: Nice!

server/plugin.go
cpoile: Do I have this right: I think the main reason is to offload the processing to the client (which already does this kind of thing all the time), because the server can get bogged down if there's many message hooks?

cpoile (approved):
Very nice, good work! I owe you a 🍺 !

server/plugin.go
agarciamontoro: >  Ya that's all I am saying. Just do the filtering there on the client.

Oh, ok! I'm always inclined to err on the side of adding load to the server instead of adding it to the client, but I'll try and do this on the client side, see how it works.

agarciamontoro: Reverted the commit (https://github.com/mattermost/mattermost-plugin-incident-management/pull/377/commits/33b3e9daf431e8db76ecaf1fecb1f5e27abd31d4) and implemented the filter in client side: https://github.com/mattermost/mattermost-plugin-incident-management/pull/377/commits/bc0897b88016866aa3219fffe9f65f019da37778
It's true that we were already doing it for deleted posts, so it's more consistent now. Thank you for the conversation, Christophers!

crspeller (approved):

agarciamontoro:
I did not merge the latest commits from the branch this PR was based on, and when merging from master I needed to fix some conflicts. And now Github shows commits from @cpoile here, but I reviewed the files changed and everything is ok.
@cpoile can you re-review this PR, please? :pray:

webapp/src/components/rhs/latest_update.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    return [profileUrl, preferredName];
+}
+
+function usePostFromState(statusPosts: StatusPost[]) : [string, Post | null] {
+    const sortedPosts = [...statusPosts]
+        .filter((a) => a.delete_at === 0)
+        .sort((a, b) => b.create_at - a.create_at);
+
+    const postID = sortedPosts[0]?.id;
```
cpoile, line 105: Not sure if you want to do this in the current PR, or leave it for later, but I was thinking we could change the `lastUpdatedByIncidentId` selector into a `newestStatusPostByIncidentID`, which would be a `incidentID -> {postId, create_at}` map, and then we don't have to do the same filter/sort in multiple places.

tests-e2e/cypress/integration/frontstage/rhs/latest_update_spec.js
```diff
@@ -0,0 +1,233 @@
...
+            // # Create a second status update
+            cy.updateStatus(secondMessage);
+
+            // * Verify that the RHS shows the second status update
+            cy.get('div[class^=UpdateSection-]').within(() => {
+                cy.findByText(secondMessage).should('exist');
+            });
+        });
+
+        it('when the last one was deleted', () => {
```
cpoile, line 95: This relates to a comment in the websocket code below.

Can we have a test for the last one was deleted in a channel we're not currently viewing, and then we view that channel? We should see the first status update.

tests-e2e/cypress/integration/frontstage/rhs/latest_update_spec.js
```diff
@@ -0,0 +1,233 @@
...
+                cy.deletePost(postId);
+            });
+
+            // # Verify that the RHS shows the first status update
+            cy.get('div[class^=UpdateSection-]').within(() => {
+                cy.findByText(firstMessage).should('exist');
+            });
+        });
+
+        it('when the last one was edited', () => {
```
cpoile, line 121: Same as above -- this test, but for a channel we're not currently viewing?

tests-e2e/cypress/integration/frontstage/rhs/latest_update_spec.js
```diff
@@ -0,0 +1,233 @@
...
+        });
+    });
+
+    describe('shows no updates', () => {
+        it('in a brand new incident', () => {
+            // * Verify that the RHS shows that there are no updates.
+            cy.get('#incidentRHSUpdates').contains('No recent updates. Click here to update status.');
+        });
+
+        it('when the only update is deleted', () => {
```
cpoile, line 177: Same -- in a channel we're not viewing

tests-e2e/cypress/integration/frontstage/rhs/latest_update_spec.js
```diff
@@ -0,0 +1,233 @@
...
+
+                // # Delete the status update
+                cy.deletePost(postId);
+
+                // * Verify that the RHS shows that there are no updates.
+                cy.get('#incidentRHSUpdates').contains('No recent updates. Click here to update status.');
+            });
+        });
+
+        it('when all updates are deleted', () => {
```
cpoile, line 196: Same -- in a channel we're not currently viewing

webapp/src/websocket_events.ts
```diff
@@ -100,3 +102,44 @@ export function handleWebsocketUserRemoved(getState: GetStateFunc, dispatch: Dis
...
+    if (incident.status_post_ids.includes(post.id)) {
+        return incident;
+    }
+
+    return null;
+}
+
+export function handleWebsocketPostDeleted(getState: GetStateFunc, dispatch: Dispatch) {
+    return async (msg: WebSocketMessage) => {
+        if (getCurrentChannelId(getState()) === msg.broadcast.channel_id) {
```
cpoile, line 123: This is what the req for more tests is about.

I think this might be a problem -- if we're not currently in that channel, we'll miss updating the store, correct? We'd want to check if the `msg.broadcast.channel_id` is one of the incident channels, something like `Boolean(incidentMapByTeam[teamId]?.[channelId])` from `selectors.ts`.

webapp/src/websocket_events.ts
```diff
@@ -100,3 +102,44 @@ export function handleWebsocketUserRemoved(getState: GetStateFunc, dispatch: Dis
...
+                    websocketSubscribersToIncidentUpdate.forEach((fn) => fn(incident));
+                }
+            });
+        }
+    };
+}
+
+export function handleWebsocketPostEdited(getState: GetStateFunc, dispatch: Dispatch) {
+    return async (msg: WebSocketMessage) => {
+        if (getCurrentChannelId(getState()) === msg.broadcast.channel_id) {
```
cpoile, line 136: Same problem?

webapp/src/websocket_events.ts
```diff
@@ -100,3 +102,44 @@ export function handleWebsocketUserRemoved(getState: GetStateFunc, dispatch: Dis
...
+            getIncidentFromStatusUpdate(JSON.parse(msg.data.post)).then((incident) => {
+                if (incident !== null) {
+                    dispatch(incidentUpdated(incident));
```
cpoile, line 139: I'm wondering, how does this update the store? Since what would change in what we track about the incident if a status post is only edited? We don't track the `edit_at` field.

cpoile (changes requested):
Nice work!

webapp/src/websocket_events.ts
agarciamontoro: I first thought about adding the `edit_at` field to what we send from the server, but this works even without it.
I understood that the `dispatch(incidentUpdated(incident))`, that triggers the `useCurrentIncident` to re-retrieve the incident, would cause the re-render of `LatestUpdate`. But I think now that the re-render happens because I'm first of all retrieving the post from the store, and as it has changed, its content is different and is re-drawn.

*But* this happens in the cases I have tested, and probably not in the ones you pointed out (channels I'm not currently viewing). Let me add those tests and see what happens. It will most likely fail, so I'll try to add the latest update to the store (`incidentID -> Post` instead of `incidentID -> {postId, create_at}`) so we can simply get it from there (and update it in these handlers). What do you think?

agarciamontoro:
After an offline conversation with @cpoile, we saw that the behaviour of the component is actually as expected. I added the required tests to verify it, but no other changes in the component were needed.

webapp/src/components/rhs/latest_update.tsx
agarciamontoro: Btw, I did not answered this, but I propose changing it in a future PR.

cpoile (approved):
Thanks @agarciamontoro !!

PR #378, opened 2020-11-16
MM-30249: Avoid sending internal errors to the client

agarciamontoro:
#### Summary

Changed `HandleErrorWithCode` to only log the internal error, while it sends the public message to the client. I renamed the parameters in the function to make the distinction between both arguments clearer.

I also needed to change a couple of tests that were expecting the error where it no longer is. I reviewed all the places that were using this function and all of them look to be using it correctly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30249


crspeller (approved):

cpoile (approved):
👍

PR #379, opened 2020-11-16
MM-29771 - Reminder for next update

cpoile:
#### Summary
- Depends on: https://github.com/mattermost/mattermost-plugin-api/pull/79 for the reminder scheduling.
- The reminder looks like (ignore the bot badge):
image
- Clicking `Dismiss` will delete the post. Clicking `Update Status` will bring up the update status dialog (mobile and desktop)
- Posting an update (no matter how it's done--clicking the button, clicking Update Status in the RHS, typing the `/incident update` slash command) will delete the post.
- If a new update is posted before the pending reminder has expired, the pending reminder is removed. There will never be two reminders active for the same incident.
- For now, there’s no way to remove a pending reminder, other than making a new update. If a "none" reminder is selected, there will be no more reminders (obviously).
- Removed the status field in the update dialog (https://community-daily.mattermost.com/core/pl/ksungmfn4pg5bk8afimt3qgfpo)
- @crspeller I had to remove the golint for goimports localprefixes -- it kept giving me the "not imported with --local" error, even though I did. Why did we have that check? Is it okay to remove it?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29711
- https://mattermost.atlassian.net/browse/MM-30544


crspeller:
@cpoile No idea why we added that check. From the goimports docs: `put imports beginning with this string after 3rd-party packages; comma-separated list`. So it just changes the order of the imports. 
Not sure why it was erroring, but removing it is fine by me.

server/incident/reminder.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	key := model.NewId()
+	reminder := Reminder{IncidentID: incidentID}
+
+	if _, err = s.pluginAPI.KV.Set(reminderPrefix+key, reminder); err != nil {
+		return errors.Wrap(err, "unable to set reminder data in kv store")
+	}
+
+	if _, err = s.scheduler.ScheduleOnce(key, time.Now().Add(timeInMinutes*time.Minute)); err != nil {
```
lieut-data, line 99: Should we only `ScheduleOnce` after we've saved the `ReminderID`?

server/incident/reminder.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+func (s *ServiceImpl) SetReminder(incidentID string, timeInMinutes time.Duration) error {
+	incidentToModify, err := s.GetIncident(incidentID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve incident from store")
+	}
+
+	key := model.NewId()
```
lieut-data, line 92: Do we need a unique `key` all the time? Could we just use the `IncidentID`, given there's only one reminder at play per incident?

server/incident/reminder.go
```diff
@@ -0,0 +1,164 @@
...
+	reminderPrefix = "reminder_"
+)
+
+type Reminder struct {
+	IncidentID string `json:"incident_id"`
+}
+
+func (s *ServiceImpl) HandleReminder(key string) {
+	var reminder Reminder
+	err := s.pluginAPI.KV.Get(reminderPrefix+key, &reminder)
```
lieut-data, line 24: It's slightly unfortunate that we have to maintain two KV store entries here: one managed by the new `cluster` code, and another to keep the `IncidentID` around. See below for question on just using the `IncidentID` as the key, but also wondering why we aren't just querying the database for the incident with `ReminderId == key`? Is that because it's in the props bag?

lieut-data (changes requested):
Neat stuff! A few comments below.

webapp/src/components/rhs/rhs_list_view.tsx
```diff
@@ -254,4 +270,17 @@ const ThreeDotMenu = (props: ThreeDotMenuProps) => (
     </DotMenu>
 );
 
+const findLastUpdated = (getPostFromState: (postId: string) => Post, updatePostIDs: string[]): number => {
+    const postIDs = [...updatePostIDs].reverse();
+
+    for (const postId of postIDs) {
+        const post = getPostFromState(postId);
```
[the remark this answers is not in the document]
agarciamontoro: Hm that's a good point. I tested it with recent posts and it worked, but I did not tests with old posts outside of the loaded ones. I bet those are not in the store :/

agarciamontoro (changes requested):
This is great, Christopher! Thank you :tada: 

I left some minor comments below.

server/api/incidents.go
```diff
@@ -628,6 +629,56 @@ func (h *IncidentHandler) updateStatusDialog(w http.ResponseWriter, r *http.Requ
...
+// reminderButtonUpdate handles the POST /incidents/{id}/reminder/button-update endpoint, called when a
+// user clicks on the reminder interactive button
+func (h *IncidentHandler) reminderButtonUpdate(w http.ResponseWriter, r *http.Request) {
+	requestData := model.PostActionIntegrationRequestFromJson(r.Body)
+	if requestData == nil {
+		HandleErrorWithCode(w, http.StatusBadRequest, "missing request data", nil)
+		return
+	}
+
+	incidentID, err := h.incidentService.GetIncidentIDForChannel(requestData.ChannelId)
```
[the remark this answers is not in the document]
cpoile: Thanks, fixed.

server/api/incidents.go
```diff
@@ -736,12 +787,13 @@ func (h *IncidentHandler) itemRun(w http.ResponseWriter, r *http.Request) {
...
 
-	if err := h.incidentService.RunChecklistItemSlashCommand(incidentID, userID, checklistNum, itemNum); err != nil {
+	triggerID, err := h.incidentService.RunChecklistItemSlashCommand(incidentID, userID, checklistNum, itemNum)
+	if err != nil {
 		HandleError(w, err)
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
+	ReturnJSON(w, map[string]interface{}{"trigger_id": triggerID}, http.StatusOK)
```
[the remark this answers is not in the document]
cpoile: This fixes https://mattermost.atlassian.net/browse/MM-30544 (see: 7d88b34) - it passes the slash command's trigger_id to the client so we can set it.
I updated the api.yaml (though I'm not totally sure this needs to be public? It's a response only needed by internal code.) @agarciamontoro can you take a look to see that I made the yaml change correctly? Thanks!

server/incident/incident.go
```diff
@@ -15,10 +16,22 @@ const NoActiveStage = -1
...
 type Incident struct {
 	Header
-	PostID         string               `json:"post_id"`
-	PlaybookID     string               `json:"playbook_id"`
-	Checklists     []playbook.Checklist `json:"checklists"`
-	StatusPostsIDs []string             `json:"status_posts_ids"`
+	PostID     string               `json:"post_id"`
+	PlaybookID string               `json:"playbook_id"`
+	Checklists []playbook.Checklist `json:"checklists"`
+	JSONBag
```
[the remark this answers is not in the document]
cpoile: Yep, that's a good idea. Done.
--update: moot with the new SQL

server/incident/incident.go
```diff
@@ -15,10 +16,22 @@ const NoActiveStage = -1
...
+}
+
+// JSONBag is a place to put info that we don't think needs to be its own SQL column. This includes
+// info that doesn't need to be searched for or ordered by when selecting incidents. Putting a new
+// field here is all that needs to be done; it will be saved and retrieved from the db
+// automatically. If we need it to be its own column in the db, "graduate" it by moving it to the
+// Incident struct, add it to the migrations and the store select/insert/update statements, and the
```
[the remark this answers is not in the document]
cpoile: I think we would have to go through each incident, unmarshal the PropsJSON, set the new column, and set the updated PropsJSON. Like we did with the data migration code. Not trivial, but doable. I added that description to the comments. 

IMO, I think this would be rare--we would have to have originally thought that the new field would never need to be searched on/ordered by, and then realized that it does. I think the ease of using Props outweighs the rare case of needing to graduate a field.

-- edit after offline discussion: looks like we should move it all into SQL. :)

server/incident/incident.go
```diff
@@ -84,9 +97,8 @@ type UpdateOptions struct {
 
 // StatusUpdateOptions encapsulates the fields that can be set when updating an incident's status
 type StatusUpdateOptions struct {
-	Status   string
-	Message  string
-	Reminder int
+	Message           string
+	ReminderInMinutes int
```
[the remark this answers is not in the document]
cpoile: Yah, that makes sense (it will also fix the comments below). Reminder will be a `time.Duration`, rather than an `int64`, just to make it clearer what it is.

server/incident/service.go
```diff
@@ -336,6 +338,21 @@ func (s *ServiceImpl) UpdateStatus(incidentID, userID string, options StatusUpda
 	s.poster.PublishWebsocketEventToChannel(incidentUpdatedWSEvent, incidentToModify, incidentToModify.ChannelID)
 	s.telemetry.UpdateStatus(incidentToModify, userID)
 
+	// Remove pending reminder (if any), even if current reminder was set to "none" (0 minutes)
+	if err = s.RemoveReminder(incidentID); err != nil {
+		return errors.Wrap(err, "failed to remove reminder")
+	}
+
+	if options.ReminderInMinutes != 0 {
+		if err = s.SetReminder(incidentID, time.Duration(options.ReminderInMinutes)); err != nil {
```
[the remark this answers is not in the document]
cpoile: It's fixed now to be clearer.

server/sqlstore/migrations.go
```diff
@@ -250,15 +250,23 @@ var migrations = []Migration{
...
-					return errors.Wrapf(err, "failed adding column StatusPostsIDs to table IR_Incident")
+				if _, err := e.Exec("ALTER TABLE IR_Incident ADD JSONBag json"); err != nil {
+					return errors.Wrapf(err, "failed adding column JSONBag to table IR_Incident")
 				}
 			}
 
+			incidentUpdate := sqlStore.builder.
+				Update("IR_Incident").
+				Set("JSONBag", "{}").
+				Where(sq.Eq{"JSONBag": nil})
```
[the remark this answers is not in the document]
cpoile: With a complete nil, squirrel fails reading into `*json.RawMessage`, so we need to have something in there. I chose `"{}"` but I think `""` or `"null"` would probably work, since you can unmarshal that into the Props struct.

-- edit: moot after changing to full-on SQL. Except there's a similar migration I had to do for the new column. nil leads to a sql error.

webapp/src/components/rhs/rhs_list_view.tsx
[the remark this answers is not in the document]
cpoile: I made the change; had to make a `RHSListIncident` component (because of the async call to get the post from the server). I think there are ways to improve this if we're concerned about efficiency (fetch all posts at once), but that's probably premature.

There is one issue though -- it won't update when a post is deleted. It's because this branch doesn't have the message update hook that alejandro's building in another PR. I'll have to fix this later this week: https://mattermost.atlassian.net/browse/MM-30787

-- update: there's a new issue, the entire webapp (and especially our components in the RHS) update on every key press in the center channel's post input. It's brutal. Not sure if it's something I've done, but it's realllllly bad. Will have to look into it tomorrow.

webapp/src/components/duration.tsx
```diff
@@ -43,12 +44,16 @@ const Duration: FC<DurationProps> = (props: DurationProps) => {
...
     }, []);
 
-    const start = moment(props.created_at);
-    const end = (props.ended_at && moment(props.ended_at)) || now;
-    const duration = moment.duration(end.diff(start));
+    if (!props.from) {
+        return <div className='time'>{'-'}</div>;
+    }
 
+    const start = moment(props.from);
```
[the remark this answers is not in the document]
cpoile: Done!

server/incident/reminder.go
cpoile: Right, in case the save fails or the job fires immediately (I don't think that's possible since it spawns a go routine, and that won't preempt the current goroutine... but just in case)

server/incident/reminder.go
cpoile: Right, that makes sense -- didn't think of that. Nice.

server/incident/reminder.go
cpoile: Yep, you're both right on the money. Good idea, making that change.

server/api/incidents.go
```diff
@@ -628,6 +629,56 @@ func (h *IncidentHandler) updateStatusDialog(w http.ResponseWriter, r *http.Requ
...
+	requestData := model.PostActionIntegrationRequestFromJson(r.Body)
+	if requestData == nil {
+		HandleErrorWithCode(w, http.StatusBadRequest, "missing request data", nil)
+		return
+	}
+
+	incidentID, err := h.incidentService.GetIncidentIDForChannel(requestData.ChannelId)
+	if err != nil {
+		h.log.Errorf("reminderButtonUpdate failed to find incidentID for channelID: %s", requestData.ChannelId)
+		HandleError(w, errors.New("reminderButtonUpdate failed to find incidentID"))
```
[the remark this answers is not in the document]
cpoile: Right, good catch, removed. (I hadn't updated to the new error handlers, but it's there now.)

server/api/incidents.go
```diff
@@ -628,6 +629,56 @@ func (h *IncidentHandler) updateStatusDialog(w http.ResponseWriter, r *http.Requ
...
+		return
+	}
+
+	if err = h.incidentService.OpenUpdateStatusDialog(incidentID, requestData.TriggerId); err != nil {
+		h.log.Errorf("reminderButtonUpdate failed to open update status dialog")
+		HandleError(w, errors.New("reminderButtonUpdate failed to open update status dialog"))
+		return
+	}
+
+	ReturnJSON(w, nil, http.StatusOK)
```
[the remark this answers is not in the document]
cpoile: Yes, the internal do plugin request expects a response, not just the code, it will error without it.

server/bot/poster.go
```diff
@@ -20,6 +20,21 @@ func (b *Bot) PostMessage(channelID, format string, args ...interface{}) (string
...
 }
 
+// PostMessage posts a message with slack attachments to channelID. Returns the post id if
+// posting was successful. Often used to include post actions.
+func (b *Bot) PostMessageWithAttachments(channelID, message string, attachments []*model.SlackAttachment) (createdPostID string, err error) {
+	post := &model.Post{
+		Message:   message,
+		UserId:    b.botUserID,
+		ChannelId: channelID,
+	}
```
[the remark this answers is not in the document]
cpoile: Yes, good call. I think I meant to do that then forgot. Thanks!

server/incident/reminder.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	if incidentToModify.ReminderID == "" {
+		return nil
+	}
+
+	s.scheduler.Close(incidentToModify.ReminderID)
+	if err = s.pluginAPI.KV.Delete(reminderPrefix + incidentToModify.ReminderID); err != nil {
+		return errors.Wrap(err, "unable to delete reminder data from kv store")
+	}
```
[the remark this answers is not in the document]
cpoile: It doesn't matter now, because this code is gone, but: if this key was never deleted then yah, I think it would live forever.

server/incident/reminder.go
lieut-data: Yeah, my primary concern was about the save failing. Though your second case is, I think, still possible: there are multiple underlying threads and it's perfectly possible for another goroutine to start and run to completion while this one is still active and finishing up. No pre-emption required.

webapp/src/components/rhs/rhs_list_view.tsx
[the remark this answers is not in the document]
cpoile: fixed

cpoile:
Everything is good to go now -- the new StatusPosts is in there, and.

@agarciamontoro see the `lastUpdatedByIncidentId` in the selectors for an example of how we could make a selector that returns a map of the current "most recent status update postIDs". (for #377) 
It's returning a map because then we only have to compute that map once (when the `myIncidentsByTeam` changes). Nice and simple.

server/sqlstore/incident.go
```diff
@@ -358,29 +441,62 @@ func (s *incidentStore) toIncident(rawIncident sqlIncident) (*incident.Incident,
...
 }
 
+func (s *incidentStore) replaceStatusPosts(q queryExecer, incidentToSave incident.Incident) error {
+	// Delete existing posts that are not in the new incidentToSave.StatusPostsIDs list
+	delBuilder := sq.Delete("IR_StatusPosts").
+		Where(sq.Eq{"IncidentID": incidentToSave.ID}).
+		Where(sq.NotEq{"PostID": incidentToSave.StatusPostsIDs})
+	if _, err := s.store.execBuilder(q, delBuilder); err != nil {
+		return err
+	}
```
agarciamontoro: I'm not entirely sure I understand why we need to delete any rows in this table. When creating a new incident, the incident ID should not exist in this table, and when updating it, we should only add posts, not delete them (they are soft deleted with delete_at != 0). But maybe I'm missing something here?

server/sqlstore/migrations.go
```diff
@@ -249,22 +250,52 @@ var migrations = []Migration{
...
 		toVersion:   semver.MustParse("0.4.0"),
 		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+
 			if e.DriverName() == model.DATABASE_DRIVER_MYSQL {
-				if _, err := e.Exec("ALTER TABLE IR_Incident ADD PropsJSON TEXT"); err != nil {
-					return errors.Wrapf(err, "failed adding column PropsJSON to table IR_Incident")
+				if _, err := e.Exec(`
+					CREATE TABLE IF NOT EXISTS IR_StatusPosts (
+						IncidentID VARCHAR(26) NOT NULL REFERENCES IR_Incident(ID),
+						PostID VARCHAR(26) NOT NULL,
```
agarciamontoro, line 258: Do we need a reference to the `Posts` table here? I recall that there may be performance concerns with that, right?

server/sqlstore/migrations.go
```diff
@@ -249,22 +250,52 @@ var migrations = []Migration{
 		fromVersion: semver.MustParse("0.3.0"),
 		toVersion:   semver.MustParse("0.4.0"),
 		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+
 			if e.DriverName() == model.DATABASE_DRIVER_MYSQL {
-				if _, err := e.Exec("ALTER TABLE IR_Incident ADD PropsJSON TEXT"); err != nil {
-					return errors.Wrapf(err, "failed adding column PropsJSON to table IR_Incident")
+				if _, err := e.Exec(`
+					CREATE TABLE IF NOT EXISTS IR_StatusPosts (
```
agarciamontoro, line 256: Should we add a unique constraint on both columns here, as in the table in postgres?

server/incident/incident.go
```diff
@@ -22,7 +22,8 @@ type Incident struct {
 	PostID         string               `json:"post_id"`
 	PlaybookID     string               `json:"playbook_id"`
 	Checklists     []playbook.Checklist `json:"checklists"`
-	StatusPostsIDs []string             `json:"status_posts_ids"`
+	StatusPostIDs  []string             `json:"status_post_ids"`
+	StatusPosts    []StatusPost         `json:"status_posts"`
```
agarciamontoro, line 26: Why do we need both?

server/sqlstore/incident.go
```diff
@@ -311,18 +318,33 @@ func (s *incidentStore) GetIncident(incidentID string) (out *incident.Incident,
...
+		From("IR_StatusPosts as ir").
+		Join("Posts as p ON ir.PostID = p.Id").
+		Where(sq.Eq{"IncidentID": incidentID}).
+		OrderBy("p.CreateAt")
+
+	// SELECT ir.IncidentID, p.Id, p.CreateAt, p.DeleteAt
+	//    FROM IR_StatusPosts as ir
+	//             JOIN Posts as p
+	//                  ON ir.PostID = p.id
+	//    ORDER BY p.CreateAt;
```
agarciamontoro, line 338: Old commented code?

server/sqlstore/incident.go
```diff
@@ -146,21 +161,26 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
...
 			return nil, err2
 		}
-		incidents = append(incidents, *incident)
+		incidents = append(incidents, *asIncident)
+		incidentIDs = append(incidentIDs, asIncident.ID)
 	}
 
 	var statusPosts incidentStatusPosts
 
+	//subclause := queryForIncidentIDs.Prefix("ir.IncidentID IN (").Suffix(")")
```
agarciamontoro: Old commented code? (and some lines below, too)

server/api/incidents.go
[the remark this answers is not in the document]
agarciamontoro: Looks good to me! (I had missed this comment before)

agarciamontoro (approved):
Looks great! I left some questions below, but nothing blocking.

Thank you :tada:

server/sqlstore/incident.go
agarciamontoro: This is outdated, sorry (I reviewed commit by commit)

server/sqlstore/incident.go
agarciamontoro: (This says it's outdated because I reviewed it in an older commit, but the question is still the same for the newer code)

cpoile: It's true, we'll probably never delete status posts, because we'll probably never save an incident that has removed an id from `incident.StatusPostIDs`. But I didn't want to make it impossible to do so, you know?

server/sqlstore/migrations.go
cpoile: Good question! Personally I don't think we'll gain anything from a reference here, and it feels wrong to me to introduce a constraint on the `Posts` table because we don't 'own' it. If we have that constraint, and then the server wants to delete or change an ID, they might get an error and not understand that it's our plugin that's causing it. What do you think @lieut-data ?

server/sqlstore/migrations.go
cpoile: Ah, good catch, missed that.

server/sqlstore/migrations.go
agarciamontoro: Us not owning it is a good enough point for me.

server/incident/incident.go
agarciamontoro: We already have that data in the frontend with each element in `StatusPosts` having the ID :thinking:

server/sqlstore/incident.go
agarciamontoro: I cannot think of a use case where we want to do that, and maybe this hides it in case we do it by mistake? Hm, I'm not sure.

server/incident/incident.go
cpoile: Right, but we only save the `StatusPostIDs` to the db, and the rest of the server code will only append to `StatusPostIDs` (e.g., `service/UpdateStatus(...)`) --- so we still need it. The frontend will probably only use `StatusPosts`. I guess it's unfortunate that we have a part of the incident struct (`StatusPosts`) that is only used on the frontend.

agarciamontoro: Yeah, you're right. We may need to decouple the server struct from the frontend struct sometime in the future (not for this PR, I'm not asking for more changes here)

server/sqlstore/incident.go
cpoile: Okay, I can remove the code -- we can add it later if we find we need it. 😄 
Done in 6dbf430

agarciamontoro: Oh great, thanks!

server/sqlstore/incident.go
```diff
@@ -117,27 +128,55 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
...
 	pageCount := int(math.Ceil(float64(total) / float64(options.PerPage)))
 	hasMore := options.Page+1 < pageCount
 
 	incidents := make([]incident.Incident, 0, len(rawIncidents))
+	incidentIDs := make([]string, 0, len(rawIncidents))
 	for _, rawIncident := range rawIncidents {
-		incident, err := toIncident(rawIncident)
-		if err != nil {
-			return nil, err
+		asIncident, err2 := s.toIncident(rawIncident)
```
lieut-data, line 152: nit: pre-declare `asIncident` instead of `err2`?

server/sqlstore/incident.go
```diff
@@ -117,27 +128,55 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
 
 	queryForResults = queryForResults.OrderBy(fmt.Sprintf("%s %s", options.Sort, options.Direction))
 
+	tx, err := s.store.db.Beginx()
```
lieut-data, line 131: Is the transaction necessary here? (It's read-only, and if we're looking for an isolation level, I think we have to be explicit.)

server/sqlstore/incident.go
```diff
@@ -377,6 +479,49 @@ func (s *incidentStore) buildPermissionsExpr(info incident.RequesterInfo) sq.Sql
...
+INSERT INTO IR_StatusPosts(IncidentID, PostID)
+    SELECT ?, ?
+    WHERE NOT EXISTS (
+        SELECT 1 FROM IR_StatusPosts
+            WHERE IncidentID = ? AND PostID = ?
+    );`
+	if s.store.db.DriverName() == model.DATABASE_DRIVER_MYSQL {
+		insertExpr = `
+INSERT INTO IR_StatusPosts(IncidentID, PostID)
+    SELECT ?, ? FROM DUAL
```
lieut-data, line 506: Can we avoid `FROM DUAL` and the database-specific query? Is it necessary?

server/sqlstore/migrations.go
```diff
@@ -18,6 +18,23 @@ type Migration struct {
 	migrationFunc func(sqlx.Ext, *SQLStore) error
 }
 
+const MySQLCharset = "DEFAULT CHARACTER SET utf8mb4"
+
+// 'IF NOT EXISTS' syntax is not supported in Postgres 9.4, so we need
+// this workaround to make the migration idempotent
+var createPGIndex = func(indexName, tableName, columns string) string {
+	return fmt.Sprintf(`
+						DO
```
lieut-data, line 27: nit: can we unindent this somewhat?

server/sqlstore/migrations.go
lieut-data: Definitely don't want a foreign key on `Posts` -- that can creates locks on the `Posts` table when inserting into this table and dramatically kill performance for the rest of the app. (Foreign keys ~= bad for high-volume tables)

server/sqlstore/migrations.go
```diff
@@ -249,14 +250,53 @@ var migrations = []Migration{
...
 				}
+
+				if _, err := e.Exec("ALTER TABLE IR_Incident ADD ReminderPostID TEXT"); err != nil {
+					return errors.Wrapf(err, "failed adding column ReminderPostID to table IR_Incident")
+				}
+			}
+
+			incidentUpdate := sqlStore.builder.
+				Update("IR_Incident").
+				Set("ReminderPostID", "").
```
lieut-data, line 296: Instead of forcing a write to the table, would it make sense to guard our read code to `COALESCE(ReminderPostId, "")` instead, guaranteeing the code will never break for a future `NULL` value (i.e. if someone clears it out incorrectly or some such).

A practical situation where this might occur is if the plugin gets downgraded for some reason, and then later upgraded again -- records created in the interim won't have a non-NULL value.

lieut-data (approved):
Really great work moving this into the database, @cpoile -- a handful of comments below, but I think they aren't strictly blocking.

server/incident/incident.go
```diff
@@ -35,8 +36,12 @@ func (i *Incident) Clone() *Incident {
 	newIncident.Checklists = newChecklists
 
 	var newStatusPostsIDs []string
-	newStatusPostsIDs = append(newStatusPostsIDs, i.StatusPostsIDs...)
-	newIncident.StatusPostsIDs = newStatusPostsIDs
+	newStatusPostsIDs = append(newStatusPostsIDs, i.StatusPostIDs...)
+	newIncident.StatusPostIDs = newStatusPostsIDs
```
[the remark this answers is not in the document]
cpoile: Yah, I had tried that before, but remember when we were having trouble with nil slices vs empty slices in the tests? `require.Equal` says they're different. But, regardless of what the test package says, we wanted to make sure every call to sqlstore returns the consistent type -- remember we decided on nil (however, we check for nil and return an empty array in the api layer, because the frontend chokes on a nil array). 

So, long story short, this is here so that `sqlstore/CreateIncident` (which calls clone) returns a nil slice. If we use the `append([]string{}` version, `CreateIncident` will return an empty slice, and a lot of the tests fail.

Same thing for `[]StatusPost` just below it. :)

server/sqlstore/incident.go
```diff
@@ -147,25 +147,35 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
 	hasMore := options.Page+1 < pageCount
 
 	incidents := make([]incident.Incident, 0, len(rawIncidents))
+	incidentIDs := make([]string, 0, len(rawIncidents))
 	for _, rawIncident := range rawIncidents {
-		incident, err2 := s.toIncident(rawIncident)
+		asIncident, err2 := s.toIncident(rawIncident)
```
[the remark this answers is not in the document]
cpoile: Yep! But I'm 0/5 on the extra line vs the `err2`; both are a little ugly to me. 🤷

server/sqlstore/incident.go
```diff
@@ -499,7 +524,7 @@ INSERT INTO IR_StatusPosts(IncidentID, PostID)
     );`
```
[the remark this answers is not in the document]
cpoile: Yep, old MySQL versions need this, I discovered.

server/sqlstore/incident.go
cpoile: Right, I guess it's okay to not use a transaction. The reason for it was to make sure the statusPosts query was seeing the same version as the previous queries. But since we're using statusPosts as append-only, I guess it's okay to not use it. 

Just for my own education, @lieut-data is there a better way to ensure a consistent version across 2 or more queries?

server/sqlstore/incident.go
cpoile: Yep, old MySQL needs this.

server/sqlstore/migrations.go
cpoile: Whoa, don't know why it looks like that on github, it was much more reasonable in the ide. Will fix.

server/sqlstore/migrations.go
cpoile: That's a good solution, thanks! I learned something new today. :)
Just a note -- sqlx won't be able to map a function to a field, so you have to put the field name next to the function.

server/incident/incident.go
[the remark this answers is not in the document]
lieut-data: I admit I don't remember the specific conversation, but ideally we could shift complexity to the tests here. For example, if `require.Equal` doesn't work for us, let's write a wrapper that handles `len(slice) == 0` specially. 

The only other touchpoint was how the REST API responds, but that's a "rendering" concern that definitely shouldn't be handled in the store.

server/sqlstore/incident.go
lieut-data: A transaction is the right way to do this, but it quickly gets into the weeds as to which isolation level to use and what's actually supported by the database in question, etc. I sometimes find that seeing an "inconsistent" view may not actually matter, so long as there isn't a surprise coupling between the two queries. In this case, seeing a slightly older set of status posts (or seeing a newer one) /shouldn't/ break anything, I think?

server/incident/incident.go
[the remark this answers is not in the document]
cpoile: Right, I agree that we should make the tests not distinguish between them. But IMO, I think it's probably more important that the various sqlstore methods are returning the same empty type, and not returning a `[]string(nil)` and a `[]string{}` for the same field depending on what method you're calling.

server/sqlstore/incident.go
cpoile: I don't /think/ it will break anything, but I left it in just in case -- I'd rather be safe and correct, and taking out the transaction felt a bit like premature optimization. I figure we can take it out if we see it causing any problems.

server/incident/incident.go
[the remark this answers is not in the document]
lieut-data: Right, but in `Go`, `[]string(nil)` and `[]string{}` are semantically the same type, and it should not be necessary to distinguish them. The "oddities" are at the edge cases here.

PR #381, opened 2020-11-17
MM-29713: Broadcast status update

agarciamontoro:
#### Summary
This PR adds the ability to broadcast every status update to a channel specified per playbook.

I have added a `BroadcastChannelID` field to both the `Playbook` and `Incident` (we need to copy the channel ID to the incident because the playbook may change after the incident is created) models, which means I've edited the last SQL migration to add a `BroadcastChannelID` column to both the `IR_Playbook` and `IR_Incident` tables.

The channel selector in the backstage's playbook edit screen is good enough to be usable (mostly copied from the `ProfileAutocomplete` component). But it's not perfect: I did not want to spend too much time on this given that we're moving it to a new location with a new design in a following ticket.

The broadcasted post adds a header with the following info:
- Title `Incident Update: <Channel name>`, where `<Channel name>` is a link to the original post.
- Stats line:
  - author handle
  - duration
  - status

After a horizontal line, it copies the original post verbatim:

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29713


agarciamontoro:
Feel free to review even if it's in a draft state. I'll ping you afterwards if I change critical things.

itao (approved):
🙌  Nice! 

As an aside, could you also make a Help Wanted ticket for propagating message edits to the broadcasted message?

crspeller (approved):
Looks good!
Little nervous about only having one broadcast channel. Not sure if multiple channels are a likely future addition. But we can always migrate if we need to.

cpoile (approved):
Awesome!!

PR #383, opened 2020-11-20
MM-30191: handle leaving channel without teamId

lieut-data:
#### Summary
When we leave a channel, the websocket event doesn't include the channel metadata: we only know the channel id. Avoid trying to reference a null `channel` object and instead rely solely on the `channelId` in the event.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-30191

### DIFF
--- a/webapp/src/reducer.test.ts
+++ b/webapp/src/reducer.test.ts
@@ -1,0 +1,64 @@
+import {
+    REMOVED_FROM_INCIDENT_CHANNEL,
+} from 'src/types/actions';
+import reducer from 'src/reducer';
+
+describe('myIncidentsByTeam', () => {
+    // @ts-ignore
+    const initialState = reducer(undefined, {}); // eslint-disable-line no-undefined
+
+    describe('REMOVED_FROM_INCIDENT_CHANNEL', () => {
+        const makeState = (myIncidentsByTeam: any) => ({
+            ...initialState,
+            myIncidentsByTeam,
+        });
+
+        it('should ignore a channel not in the data structure', () => {
+            const state = makeState({
+                teamId1: {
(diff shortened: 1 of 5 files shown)


webapp/src/reducer.test.ts
```diff
@@ -0,0 +1,64 @@
+import {
+    REMOVED_FROM_INCIDENT_CHANNEL,
+} from 'src/types/actions';
+import reducer from 'src/reducer';
+
+describe('myIncidentsByTeam', () => {
+    // @ts-ignore
+    const initialState = reducer(undefined, {}); // eslint-disable-line no-undefined
```
lieut-data, line 8: This combination of directives to permit me to simulate the [default invocation of Redux](https://github.com/reduxjs/redux/blob/3cf3b0f48c4093aaa094eedb11efa8656e9b0309/examples/shopping-cart/src/reducers/cart.spec.js#L10-L12) is amusing :)

webapp/src/reducer.ts
```diff
@@ -114,12 +114,17 @@ const myIncidentsByTeam = (state: Record<string, Record<string, Incident>> = {},
     }
     case REMOVED_FROM_INCIDENT_CHANNEL: {
         const removedFromChannelAction = action as RemovedFromIncidentChannel;
-        const teamId = removedFromChannelAction.teamId;
+        const channelId = removedFromChannelAction.channelId;
+        const teamId = Object.keys(state).find((t) => Boolean(state[t][channelId]));
```
cpoile, line 118: That's clever, nice.

cpoile (approved):

agarciamontoro (approved):
Nice! :tada: The reducer tests are great :)

PR #384, opened 2020-11-20
#15761

SBagaria2710:
#### Summary
Fix step description overlap in the RHS by closing popover on clicking outside the description


lieut-data:
Thanks, @SBagaria2710! I've asked our team to help review your changes :)

webapp/src/components/checklist_item.tsx
```diff
@@ -98,7 +98,10 @@ interface StepDescriptionProps {
 const StepDescription = (props: StepDescriptionProps) : React.ReactElement<StepDescriptionProps> => {
     const [showTooltip, setShowTooltip] = useState(false);
     const target = useRef(null);
-
+    useClickOutsideRef(target, () => {
```
agarciamontoro, line 101: This `target` references the info icon, used by the `Overlay` component to place the popover: see how it's declared as `ref` in the `InfoIcon` component and used as the `target` field in the `Overlay component`. So the effect this has right now is that clicking anywhere outside the info icon (included the popover itself) will close the popover.

What we should do instead is create a second reference with `useRef` and use it in the `StyledPopover` component itself. You may need to create a `div` container wrapping everything inside the `StyledPopover` component, so you can add this new reference as its `ref`.

webapp/src/components/checklist_item.tsx
```diff
@@ -98,7 +98,10 @@ interface StepDescriptionProps {
...
+        setShowTooltip(false);
+    });
+
```
agarciamontoro, line 104: Minor thing: the linter is complaining that there is some blank space here. Could you remove it? Thanks!

agarciamontoro (changes requested):
Thank you for your contribution, @SBagaria2710 :tada: Your idea of using that hook is great, we only need to use the correct reference, as I describe below.

SBagaria2710 (commented):
I made the required changes. It seems to be working fine now.

webapp/src/components/checklist_item.tsx
SBagaria2710: Made the required changes. Also, it seems to be working fine in local.
Let me know if any other changes are required from my end.

agarciamontoro: > Now what happens is when I click on the info-icon, popover won't close once opened. What is happening here is that after triggering the handler function of the useClickOutsideRef, it is triggering the onClick set at InfoIcon. How do we overcome this behaviour?

I'd say it's ok to keep the popup open when the user clicks on the info icon. We have an X button and clicking outside of it would close it, so I'd say it's a good solution.

webapp/src/components/checklist_item.tsx
```diff
@@ -98,7 +98,10 @@ interface StepDescriptionProps {
 const StepDescription = (props: StepDescriptionProps) : React.ReactElement<StepDescriptionProps> => {
     const [showTooltip, setShowTooltip] = useState(false);
     const target = useRef(null);
-
+    const popoverRef = useRef(null);
+    useClickOutsideRef(popoverRef, () => {
+        setTimeout(() => setShowTooltip(false), 10);
```
agarciamontoro, line 103: Why do we need the timeout here? Can we remove it?

agarciamontoro (approved):
Looks good, thank you! :tada: I left a non-blocking comment below.

webapp/src/components/checklist_item.tsx
SBagaria2710: Like I mentioned above, on `useClickOutsideRef` it is working as expected, but then it is triggering the `onClick` on the `InfoIcon`.

Which is why I put the `useClickOutsideRef`  as a callback for `setTimeout` so that it trigger the `onClick` prior to `useClickOutsideRef`.

I don't think it'll be a problem, but I am open to learn about the alternative if you'd want to get into this.

SBagaria2710: Also, I made this change as I felt it wasn't a good user experience. I'd expect to click on the same button/icon to toggle.
I can remove it tho if you want 👍

agarciamontoro: I'm thinking now that another solution for this would be to change the current `useClickOutsideRef` hook to `useClickOutsideRefs`, and pass a list of refs instead of a single one. In the code of the hook, check that none of the refs is being clicked on. In the caller code, use it simply with `useClickOutsideRef([popoverRef, target], setShowTooltip(false))`.
I have not tested it, but it may work. Are you open to try it?

SBagaria2710: I don't see any harm in trying. Let me raise another PR in sometime :)

cpoile: I agree -- I don't think we need it. You can probably do:
```suggestion
        setShowTooltip(false);
```

cpoile (changes requested):
Looks great @SBagaria2710 ! One more thing then we're good to go.

webapp/src/components/checklist_item.tsx
SBagaria2710: So, do we want to try `useClickOutsideRefs` which @agarciamontoro  suggested? Or should I just remote `setTimeout` and update the PR?

cpoile (approved):
Thanks for the contribution @SBagaria2710, looks great!

cpoile:
@prapti I've tested it on my own machine, but would feel better if you could build from this branch and test on your own machine. It's your call if you think this can make the code freeze deadline (I would vote yes, if it passes your testing). Thanks!

prapti (approved):
Tested locally and on cloud and can verify that the descriptions no longer overlap each other. This is a great improvement. Thank you @SBagaria2710 for working on this fix!

Thanks for the heads up @cpoile! I too think this one should make the code freeze deadline :)

SBagaria2710:
Thank you all @lieut-data  @agarciamontoro  @cpoile @prapti ! 🙇 

This was my **first contribution** to mattermost. Looking forward to making more in near future 😄  🚀 

Until next time...

PR #386, opened 2020-11-24
MM-30262 - Checking and unchecking the checklist items throws JS error

cpoile:
#### Summary
- `doPut` was expecting something back from these endpoints, so it was throwing a JSON parsing error.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-30262

### DIFF
--- a/webapp/src/client.ts
+++ b/webapp/src/client.ts
@@ -17,16 +17,16 @@ import {CommanderInfo} from 'src/types/backstage';
 import {
     FetchIncidentsParams,
     FetchIncidentsReturn,
+    Incident,
+    isIncident,
     isMetadata,
     Metadata,
-    isIncident,
-    Incident,
 } from 'src/types/incident';
 import {
-    Playbook,
     ChecklistItem,
     ChecklistItemState,
     FetchPlaybooksNoChecklistReturn,
+    Playbook,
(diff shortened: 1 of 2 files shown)


webapp/src/client.ts
```diff
@@ -187,21 +187,18 @@ export async function setCommander(incidentId: string, commanderId: string) {
...
-        return data;
+        return await doPut(`${apiUrl}/incidents/${incidentId}/checklists/${checklistNum}/item/${itemNum}/assignee`, body);
     } catch (error) {
         return {error};
     }
 }
 
 export async function setChecklistItemState(incidentID: string, checklistNum: number, itemNum: number, newState: ChecklistItemState) {
-    const data = await doPut(`${apiUrl}/incidents/${incidentID}/checklists/${checklistNum}/item/${itemNum}/state`,
+    return doPut(`${apiUrl}/incidents/${incidentID}/checklists/${checklistNum}/item/${itemNum}/state`,
```
lieut-data, line 197: Is there a missing `await` here?

cpoile: We don't need to await if you return the promise (it's redundant).  (I learned that today. :)

crspeller (approved):

lieut-data (approved):

PR #387, opened 2020-11-24
update nuke-db

cpoile:
#### Summary
- fix nuke-db by adding the new tables we're using

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -124,17 +124,20 @@ func (p *Plugin) OnActivate() error {
 		return errors.Wrapf(err, "failed creating the SQL store")
 	}
 
+	incidentStore := sqlstore.NewIncidentStore(apiClient, p.bot, sqlStore)
+
 	mutex, err := cluster.NewMutex(p.API, "IR_dbMutex")
 	if err != nil {
 		return errors.Wrapf(err, "failed creating cluster mutex")
 	}
 
-	// Cluster lock: only one plugin will perform the migration when needed
-	if err = p.UpgradeDatabase(sqlStore, mutex); err != nil {
+	mutex.Lock()
+	if err = incidentStore.RunMigrations(); err != nil {
+		mutex.Unlock()
 		return errors.Wrapf(err, "failed to run migrations")
 	}
+	mutex.Unlock()
 
-	incidentStore := sqlstore.NewIncidentStore(apiClient, p.bot, sqlStore)
 	playbookStore := sqlstore.NewPlaybookStore(apiClient, p.bot, sqlStore)
 
 	p.handler = api.NewHandler()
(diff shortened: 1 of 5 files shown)


crspeller (approved):

server/sqlstore/incident.go
```diff
@@ -430,6 +430,14 @@ func (s *incidentStore) NukeDB() (err error) {
 		return errors.Wrap(err, "could not delete IR_Playbook")
 	}
 
+	if _, err := tx.Exec("DELETE FROM IR_StatusPosts"); err != nil {
+		return errors.Wrap(err, "could not delete IR_StatusPosts")
+	}
+
+	if _, err := tx.Exec("DELETE FROM IR_System"); err != nil {
```
lieut-data, line 437: To double check, this will also remember the record of what migrations have been applied, even though the tables are all still there -- is that ok? Should we `DROP TABLE` instead everywhere?

lieut-data (approved):

server/sqlstore/incident.go
cpoile: Yep. I moved the migration code around so it can be called right after (so that the self-test won't break in the middle).

cpoile: I hate the way we have to Unlock the mutex in two places -- but making an anonymous function and handling the error is even worse. :/

server/sqlstore/incident.go
```diff
@@ -418,20 +418,29 @@ func (s *incidentStore) NukeDB() (err error) {
 	}
 	defer s.store.finalizeTransaction(tx)
 
-	if _, err := tx.Exec("DELETE FROM IR_PlaybookMember"); err != nil {
-		return errors.Wrap(err, "could not delete IR_Playbook")
+	if _, err := tx.Exec("DROP TABLE IF EXISTS IR_PlaybookMember, IR_Incident, IR_Playbook, IR_StatusPosts, IR_System CASCADE"); err != nil {
```
lieut-data, line 421: As a matter of SQL hygiene, I propose we never `DROP TABLE ... CASCADE` and prefer being explicit about what tables we have to drop, even if it means defining them in the right `DROP` order.

server/plugin.go
```diff
@@ -124,17 +124,20 @@ func (p *Plugin) OnActivate() error {
 		return errors.Wrapf(err, "failed creating the SQL store")
 	}
 
+	incidentStore := sqlstore.NewIncidentStore(apiClient, p.bot, sqlStore)
```
lieut-data, line 127: It feels weird to host `RunMigrations` on `incidentStore`, which in turn sets up things used by the `playbookStore` as well. That leads to weirdness like this, where we initialize one store, run migrations, then initialize another store, and hope there aren't any implicit assumptions about the database already being setup.

Is there a reason we don't just host this directly on `sqlStore`? Maybe `nukeDB` needs to be there too?

server/sqlstore/incident.go
```diff
@@ -418,20 +418,29 @@ func (s *incidentStore) NukeDB() (err error) {
...
+		return errors.Wrap(err, "could not delete all IR tables")
 	}
 
-	if _, err := tx.Exec("DELETE FROM IR_Incident"); err != nil {
-		return errors.Wrap(err, "could not delete IR_Incident")
+	if err := tx.Commit(); err != nil {
+		return errors.Wrap(err, "could not delete all rows")
```
lieut-data, line 426: nit: error here is about committing, not deleting

lieut-data (changes requested):
Thanks, @cpoile -- a few comments below. Sorry for the thrashing!

server/sqlstore/incident.go
cpoile: Yep, that makes sense, thanks!

server/plugin.go
cpoile: Yep, that's where it should be, don't know why I didn't think of that.

lieut-data (approved):
Nice!

PR #389, opened 2020-11-24
MM-30517 - Update Tasks header to new design

cpoile:
#### Summary
- Adds the task header and help text
- NOTE for @itao and Alex: the designs have the drag and drop handle a bit more to the right, but we've always had it more to the left because it only appears on hover and without it there is a lot of empty space on the left. Take a look at with and without.

Without:
image

With:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-30520


crspeller (approved):

webapp/src/components/backstage/stages_and_steps_edit.tsx
```diff
@@ -145,65 +163,71 @@ export const StagesAndStepsEdit = (props: Props): React.ReactElement => {
...
-                    onRemoveChecklist(confirmRemoveChecklistNum);
-                    setConfirmRemoveChecklistNum(-1);
-                }}
-                onCancel={() => setConfirmRemoveChecklistNum(-1)}
-            />
-        </DragDropContext>
+        <>
+            <Header>
+                <HeaderTitle>{'Tasks'}</HeaderTitle>
+                <HeaderHelpText>{'Stages allow you to group your tasks. Tasks are meant to be completed by members of the workflow channel.'}</HeaderHelpText>
```
agarciamontoro, line 169: Should be `workflow` or `incident`?

agarciamontoro (approved):
Neat! :tada: I left a question about the never-ending wording issue.

webapp/src/components/backstage/stages_and_steps_edit.tsx
cpoile: @itao I copied the copy from the design, but I think Alejandro is right. Can you confirm?

itao (approved):
Nice. The drag handle spacing looks fine to me, good job.

webapp/src/components/backstage/stages_and_steps_edit.tsx
cpoile: Changing to "incident"

PR #390, opened 2020-11-25
[MM-29685]: Change text of the ephemeral message in command.go

haardikdharma10:
#### Summary 

Updated the ephemeral message in command.go

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/16400
JIRA: https://mattermost.atlassian.net/browse/MM-29685

### DIFF
--- a/server/command/command.go
+++ b/server/command/command.go
@@ -468,7 +468,7 @@ func (r *Runner) actionList() {
 func (r *Runner) actionInfo() {
 	incidentID, err := r.incidentService.GetIncidentIDForChannel(r.args.ChannelId)
 	if errors.Is(err, incident.ErrNotFound) {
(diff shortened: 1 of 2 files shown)


lieut-data (approved):
Thanks, @haardikdharma10! Can you update the e2e tests that verify this behaviour as part of your change? Otherwise -- looks good :)

agarciamontoro (commented):
Change looks good, thank you! :tada: As @lieut-data pointed out, you'll need to change the same word in the [info_spec.js](https://github.com/mattermost/mattermost-plugin-incident-management/blob/master/tests-e2e/cypress/integration/frontstage/slash_command/info_spec.js) file.

lieut-data:
Skipping `3: QA Review`, since already covered by the e2e tests. Thanks, @haardikdharma10!

PR #394, opened 2020-12-01
MM-30259: Fix backstage bug

agarciamontoro:
#### Summary
Calling `clientExecuteCommand` relies on the Redux store containing a current channel. If the first URL to be loaded is the one in the backstage, such channel does not exist. The proposed fix is simple: just defaulting to town-square when no current channel exists. I think this is a good solution, but I'm open to any other ideas you have (the best solution would be not to have a mandatory channel in the API endpoint, of course, but that would be a weeks long solution, probably).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30259

### DIFF
--- a/tests-e2e/cypress/integration/backstage/incident_list_spec.js
+++ b/tests-e2e/cypress/integration/backstage/incident_list_spec.js
@@ -56,6 +56,23 @@ describe('backstage incident list', () => {
         cy.get('#root').findByText('What are Incidents?').should('be.visible');
     });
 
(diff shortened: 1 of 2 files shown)


lieut-data (approved):
This seems reasonable -- and clever! I was going to ask about the message posted in the channel from which `/incident start` gets run, but I realize it's only ephemeral anyway so it's not as though we leak it out to a wider audience.

cpoile (approved):
Cool, nice solution!

PR #401, opened 2020-12-03
MM-31039 - Previous reminder time becomes default for next reminder time

cpoile:
#### Summary
- Added a SQL column, migrations, pretty simple to put the reminder as the default. 
- Added e2e tests for it so Prapti doesn't have to do it.
- Based off #400 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-31039

### DIFF
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -303,7 +303,7 @@ func (s *ServiceImpl) OpenUpdateStatusDialog(incidentID string, triggerID string
 	}
 	// TODO: if there is no newestPost, use the message template: https://mattermost.atlassian.net/browse/MM-30519
 
-	dialog, err := s.newUpdateIncidentDialog(message, currentIncident.BroadcastChannelID)
+	dialog, err := s.newUpdateIncidentDialog(message, currentIncident.BroadcastChannelID, currentIncident.PreviousReminder)
 	if err != nil {
 		return errors.Wrap(err, "failed to create update status dialog")
 	}
@@ -378,6 +378,7 @@ func (s *ServiceImpl) UpdateStatus(incidentID, userID string, options StatusUpda
 	}
 
 	incidentToModify.StatusPostIDs = append(incidentToModify.StatusPostIDs, post.Id)
+	incidentToModify.PreviousReminder = options.Reminder
 	if err = s.store.UpdateIncident(incidentToModify); err != nil {
 		return errors.Wrap(err, "failed to update incident")
 	}
@@ -1057,7 +1058,7 @@ func (s *ServiceImpl) newIncidentDialog(teamID, commanderID, postID, clientID st
 	}, nil
 }
 
-func (s *ServiceImpl) newUpdateIncidentDialog(message, broadcastChannelID string) (*model.Dialog, error) {
+func (s *ServiceImpl) newUpdateIncidentDialog(message, broadcastChannelID string, previousReminder time.Duration) (*model.Dialog, error) {
 	introductionText := "Update your incident status."
 
 	broadcastChannel, err := s.pluginAPI.Channel.Get(broadcastChannelID)
@@ -1122,7 +1123,7 @@ func (s *ServiceImpl) newUpdateIncidentDialog(message, broadcastChannelID string
 				Type:        "select",
 				Options:     options,
 				Optional:    true,
-				Default:     "0",
+				Default:     fmt.Sprintf("%d", previousReminder/time.Second),
 			},
 		},
 		SubmitLabel:    "Update Status",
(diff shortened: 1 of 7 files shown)


server/sqlstore/migrations.go
```diff
@@ -308,6 +308,17 @@ var migrations = []Migration{
...
+			return nil
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.4.0"),
+		toVersion:   semver.MustParse("0.5.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if _, err := e.Exec("ALTER TABLE IR_Incident ADD PreviousReminder BIGINT NOT NULL DEFAULT 0"); err != nil {
+				return errors.Wrapf(err, "failed adding column PreviousReminder to table IR_Incident")
+			}
```
agarciamontoro, line 320: I noticed yesterday that these `ALTER TABLE` migrations are not idempotent. I'm working on this ticket to fix it: https://mattermost.atlassian.net/browse/MM-31086

tests-e2e/cypress/integration/frontstage/rhs/latest_update_spec.js
```diff
@@ -130,6 +130,135 @@ describe('incident rhs > latest update', () => {
         });
     });
 
+    describe('the default reminder', () => {
+        it('shows \'none\' when we have not made a previous update', () => {
+            // # Run the /incident status slash command.
```
agarciamontoro, line 135: Nit: 
```suggestion
            // # Run the /incident update slash command.
```

The typo is also in the following tests.

agarciamontoro (approved):
Cool, thank you! :tada:

server/incident/service.go
```diff
@@ -1122,7 +1123,7 @@ func (s *ServiceImpl) newUpdateIncidentDialog(message, broadcastChannelID string
 				Type:        "select",
 				Options:     options,
 				Optional:    true,
-				Default:     "0",
+				Default:     fmt.Sprintf("%d", previousReminder/time.Second),
```
lieut-data, line 1126: If `previousReminder` -- for some reason -- isn't in the options, does this "work"? Should we triple-check that it's in the list, and if not, add it again.

server/sqlstore/incident.go
```diff
@@ -45,7 +45,7 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 	incidentSelect := sqlStore.builder.
 		Select("ID", "Name", "Description", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
 			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "ActiveStageTitle", "PostID", "PlaybookID",
-			"ChecklistsJSON", "COALESCE(ReminderPostID, '') ReminderPostID", "BroadcastChannelID").
+			"ChecklistsJSON", "COALESCE(ReminderPostID, '') ReminderPostID", "PreviousReminder", "BroadcastChannelID").
```
lieut-data, line 48: nit: `PreviousReminder` doesn't tell me anything about the units in question. Is it worth calling this `ReminderSeconds` or some such?

lieut-data (approved):
Looks great, just two comments!

server/sqlstore/migrations.go
cpoile: @agarciamontoro Do you want to do this one too, since you'll know exactly what to do?

cpoile: (after this PR is merged, I mean)

agarciamontoro: Yup!

server/incident/service.go
cpoile: I am pretty confident this will do the right thing -- it defaults to zero, and any other value will had to have been selected by the user, so it will be a valid value. Plus -- the e2e tests test each possibility.  :). If it wasn't a valid value (maybe editing the database), then the default option will be blank--no big deal.

server/sqlstore/incident.go
cpoile: That's a good question. It's actually stored as a `time.Duration`. I thought it was self explanatory that way (the type tells us it's a standard Go `time.Duration`, so nanoseconds). But, that's not standard for our server (which is... an int64 but in milliseconds? I can never remember.). What do you think? I can see the argument to keep it in the same units as our `CreateAt` etc. values. (even though this isn't a `CreateAt` kind of type...) It's a judgement call I guess. I'm 1/5 to use `time.Duration` because it's unambiguous and the Go standard.

@lieut-data Tagging before I merge this, for your feedback.

lieut-data: I'm not sure there is a standard to follow for persisting duration here. `time.Duration` seems fine -- thanks for explaining :)

PR #402, opened 2020-12-04
MM-31086: Make SQL migrations idempotent

agarciamontoro:
#### Summary

The SQL migrations were not idempotent, only because of the `ALTER TABLE` statements. This PR adds a couple of functions (for PostgreSQL and MySQL) to add columns in an idempotent way. And a couple of tests to prove it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31086


server/sqlstore/migrations.go
```diff
@@ -35,6 +36,44 @@ var createPGIndex = func(indexName, tableName, columns string) string {
 	`, indexName, indexName, tableName, columns)
 }
 
+var addColumnToPGTable = func(e sqlx.Ext, tableName, columnName, columnType string) error {
```
lieut-data, line 39: nit: worth moving some of these helper functions to a separate file in the `sqlstore`?

server/sqlstore/system.go
```diff
@@ -26,7 +26,21 @@ func (sqlStore *SQLStore) getSystemValue(q queryer, key string) (string, error)
...
+	// The value is already as specified, so do nothing
+	if err == nil && currentValue == value {
+		return nil
+	}
+
+	// There was an unexpected error querying for the IR_System table
+	if err != nil && err != sql.ErrNoRows {
+		return err
+	}
+
```
agarciamontoro, line 43: I needed to add this because the test that runs each migration twice were trying to update the `DatabaseVersion` key to the value it already had, so the check below for `rowsAffected > 0` was false, and the following `Insert` was failing (because the key did exist, it was just that it already had the value we were trying to update)

server/sqlstore/migrations.go
```diff
@@ -35,6 +36,44 @@ var createPGIndex = func(indexName, tableName, columns string) string {
...
+var addColumnToMySQLTable = func(e sqlx.Ext, tableName, columnName, columnType string) error {
+	var result int
+	err := e.QueryRowx(
+		"SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = ? AND COLUMN_NAME = ?",
+		tableName,
+		columnName,
+	).Scan(&result)
+
+	// Column exists, so we don't need to alter the table
+	if err == nil {
```
lieut-data, line 64: nit: maybe rearrange the rest of this as:
```
// Only alter the table if we don't find the column
if err == sql.ErrNoRows {
    _, err = e.Exec(...)
}

return err
```

server/sqlstore/system.go
```diff
@@ -26,7 +26,21 @@ func (sqlStore *SQLStore) getSystemValue(q queryer, key string) (string, error)
 }
 
 // setSystemValue updates the IR_System table for the given key.
-func (sqlStore *SQLStore) setSystemValue(e execer, key, value string) error {
+func (sqlStore *SQLStore) setSystemValue(e queryExecer, key, value string) error {
+	var currentValue string
```
lieut-data, line 30: Could we solve (what I think is the intent of this change) using https://github.com/mattermost/mattermost-server/blob/d137cf0aa5b2a179dd293c5db46df54975b59eb5/store/sqlstore/channel_store.go#L479 instead?

lieut-data (changes requested):
Nice! Just a few comments.

server/sqlstore/system.go
agarciamontoro: @lieut-data Oh that's cool! But I don't fully understand what happens with postgres there. There's no native support for upsert in Postgres 9.4, so we still need some workaround, isn't it?

agarciamontoro: Hm, it looks like it works for Postgres :thinking:

server/sqlstore/migrations.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-management/pull/402/commits/80f03b4e66b57659cac6c488d0522f6f5bd70853.

server/sqlstore/system.go
agarciamontoro: I checked and postgres does report 1 row affected even when the row exists, so we can keep the same behaviour for it. I added the native upsert for MySQL in https://github.com/mattermost/mattermost-plugin-incident-management/pull/402/commits/b3885f94ca72703ec34c9f9441bce9e6ea55dfa1

server/sqlstore/migrations.go
lieut-data: I go to `migrations.go` to find the meat of the changes being made, but I'm presently forced to scroll through a ton of helper functions to get there. I wouldn't object to a `*utils.go` file (just a `utils` package ;P).

agarciamontoro: Three functions is not a ton of functions, Jesse! :joy: Just kidding, I did the change in https://github.com/mattermost/mattermost-plugin-incident-management/pull/402/commits/726a58ca034297ef1018132474938ec34d61c8dc

lieut-data (approved):
So beautiful!

agarciamontoro:
@cpoile, we either merge this PR after you review it so you can use the new function in #401, or if yours is merged before, I'll add the changes to your migration. Whatever works best for you!

cpoile (approved):
Looks great, thanks @agarciamontoro !

PR #403, opened 2020-12-07
Update incident-response-slash-commands.md

justinegeffen:
Updated slash commands content to add test data commands.


agarciamontoro (approved):
Nice! :tada:

PR #405, opened 2020-12-07
Fix msql migrations

cpoile:
#### Summary
- Essentially: mysql needs table_schema set when querying if a column exists.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-31140

### DIFF
--- a/server/sqlstore/migrations_utils.go
+++ b/server/sqlstore/migrations_utils.go
@@ -41,6 +41,9 @@ var addColumnToPGTable = func(e sqlx.Ext, tableName, columnName, columnType stri
 var addColumnToMySQLTable = func(e sqlx.Ext, tableName, columnName, columnType string) error {
 	var dbName string
 	err := e.QueryRowx("SELECT DATABASE();").Scan(&dbName)
+	if err != nil {
+		return err
+	}
 
 	var result int
 	err = e.QueryRowx(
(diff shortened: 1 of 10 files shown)


lieut-data (commented):
Quick comment: is it worth just querying `SELECT DATABASE();` when needed vs. the signature change?

cpoile:
@lieut-data 🤦 I feel like an idiot. That's much easier.

server/sqlstore/migrations_utils.go
```diff
@@ -39,9 +39,16 @@ var addColumnToPGTable = func(e sqlx.Ext, tableName, columnName, columnType stri
...
+	err := e.QueryRowx("SELECT DATABASE();").Scan(&dbName)
+	if err != nil {
+		return err
+	}
+
 	var result int
-	err := e.QueryRowx(
-		"SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = ? AND COLUMN_NAME = ?",
+	err = e.QueryRowx(
+		"SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = ? AND TABLE_NAME = ? AND COLUMN_NAME = ?",
```
lieut-data, line 50: Could also inline this as `TABLE_SCHEMA = DATABASE()` to avoid the extra query.

lieut-data (approved):
Thanks for getting to the root problem either way, @cpoile :)

crspeller (approved):

PR #406, opened 2020-12-07
MM-30519 Moving broadcast channel, adding reminder timer and message template to preferences tab.

crspeller:
#### Summary
Implements the preferences tab of the playbook backstage.
- Moves existing broadcast channel functionality. 
- Adds a default reminder timer. Will be used until the user selects another time, then that time will be remembered. 
- Adds default message template. Will be used until the user changes the message, then the last message will be used. 

Ending up going with the existing pattern of copying preferences to the struct.

#### Screenshot
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30519


server/sqlstore/incident.go
```diff
@@ -45,7 +45,8 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 	incidentSelect := sqlStore.builder.
 		Select("ID", "Name", "Description", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
 			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "ActiveStageTitle", "PostID", "PlaybookID",
-			"ChecklistsJSON", "COALESCE(ReminderPostID, '') ReminderPostID", "PreviousReminder", "BroadcastChannelID").
+			"ChecklistsJSON", "COALESCE(ReminderPostID, '') ReminderPostID", "PreviousReminder", "BroadcastChannelID",
```
lieut-data, line 48: Can we harden the queries here using `COALESCE` for all the nullable columns? I realize we set them to `''` on migration, but if anyone goes in and sets these to `NULL` manually (or there's a bug), I think we completely fail on queries here instead of handling that gracefully.

lieut-data (approved):
Nice -- just one comment, but not strictly blocking.

server/incident/service.go
```diff
@@ -302,8 +302,9 @@ func (s *ServiceImpl) OpenUpdateStatusDialog(incidentID string, triggerID string
 			return errors.Wrap(err, "failed to find newest post")
 		}
 		message = post.Message
+	} else {
+		message = currentIncident.ReminderMessageTemplate
```
cpoile, line 306: This probably means we'll use the template if they submitted an update, then deleted it. I guess that's okay?

cpoile (approved):
Nice! One non-blocking question/observation.

server/sqlstore/incident.go
```diff
@@ -46,7 +46,7 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 		Select("ID", "Name", "Description", "IsActive", "CommanderUserID", "TeamID", "ChannelID",
 			"CreateAt", "EndAt", "DeleteAt", "ActiveStage", "ActiveStageTitle", "PostID", "PlaybookID",
 			"ChecklistsJSON", "COALESCE(ReminderPostID, '') ReminderPostID", "PreviousReminder", "BroadcastChannelID",
-			"ReminderMessageTemplate").
+			"COALESCE(ReminderMessageTemplate, '')").
```
cpoile, line 49: Have to add the destination col name

server/incident/service.go
crspeller: Yup, that's what will happen.  I feel like that's good behavior.

server/sqlstore/incident.go
crspeller: That's what I get for not testing my changes. 😆

prapti (approved):
Tested and verified on cloud server.
LGTM.

PR #408, opened 2020-12-08
MM-31187 Enable broadcast to private channels.

crspeller:
#### Summary
Enables selection of private channels in the broadcast option. 

Also fixes bug where users could possibly select channels they are not a part of by guessing an ID. 

If another member of the playbook is not a member of the channel selected, they will see "Unknown Channel" in the dropdown. They will still be able to change the values of other fields, or select a channel they are a member of. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31187

### DIFF
--- a/webapp/src/websocket_events.ts
+++ b/webapp/src/websocket_events.ts
@@ -5,7 +5,7 @@ import {Dispatch} from 'redux';
 
 import {GetStateFunc} from 'mattermost-redux/types/actions';
 import {Post} from 'mattermost-redux/types/posts';
-import {WebSocketMessage} from 'mattermost-redux/actions/websocket';
+import {WebSocketMessage} from 'mattermost-redux/types/websocket';
 import {getCurrentTeam, getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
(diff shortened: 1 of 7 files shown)


lieut-data (approved):
Nice!

webapp/src/websocket_events.ts
```diff
@@ -34,7 +34,7 @@ export function handleReconnect(getState: GetStateFunc, dispatch: Dispatch) {
 }
 
 export function handleWebsocketIncidentUpdated(getState: GetStateFunc, dispatch: Dispatch) {
-    return (msg: WebSocketMessage): void => {
+    return (msg: WebSocketMessage<{payload: string}>): void => {
```
cpoile, line 37: Just curious, why are we adding this (and the ones below)?

cpoile (approved):
Thanks! Just a non-blocking question.

webapp/src/websocket_events.ts
crspeller: There was a breaking change in the type exported by mattermost-redux. I updated mattermost-redux, so I needed to fix the types here.

prapti (approved):

PR #410, opened 2020-12-10
MM-31241, MM-31237: Fix update status dialog

agarciamontoro:
#### Summary
Check that the broadcast channel is public, and add information about it to the interactive dialog only in that case. If it is a private channel, a DM or a GM, then a generic message "This message will be broadcasted to a private channel." is added instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31237
https://mattermost.atlassian.net/browse/MM-31241


crspeller (approved):

lieut-data (approved):
Bravo!

prapti (approved):
Looks great! Thank you for this quick fix @agarciamontoro!

PR #412, opened 2020-12-10
MM-28969 Enabled open RHS with incident list when issuing /incident list from webapp

shuang2411:
#### Summary
Added a new if statement to catch '/incident list' at the client side of the webapp

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16403


agarciamontoro (changes requested):
Thank you for your contribution, @shuang2411, the code looks great! :tada:

We can still improve the behavior in a couple of cases:
- when the RHS is already open, it looks like nothing has happened. To inform the user, we can simply send an ephemeral post stating something like "The list of your incidents is already open in the RHS".
- there is also an edge case I had not anticipated: we can be in an incident channel and have the RHS open, but its state may be `ViewingIncident`. To overcome this, I think that once the RHS is open, `setRHSViewingList` can do the trick, but I have not tested it.

Could you try these two things and see if they work? Thank you!

agarciamontoro (changes requested):
We're almost there! Thank you, @shuang2411 :tada:

I left a couple of comments below, mostly related to the changes in other files we probably don't need. But I also discovered an edge case while testing your changes:

If the user is in an incident channel, and the RHS is closed, executing `/incident list` opens the RHS, but shows the incident details instead of the list. I think that this happens because the default when the RHS opens in an incident channel is to view the incident details, so that's the one shown. Even if we later dispatch the `setRHSViewingList`, I think that we are probably competing with the `setRHSViewingIncident` action dispatched [here](https://github.com/mattermost/mattermost-plugin-incident-management/blob/master/webapp/src/components/rhs/rhs_main.tsx#L36) :thinking: I cannot think of a sensible way to solve this without modifying the main behaviour of the RHSMain component. Any idea on this, @lieut-data?

tests-e2e/package.json
```diff
@@ -6,7 +6,7 @@
   },
   "devDependencies": {
     "babel-eslint": "10.1.0",
-    "cypress": "4.1.0",
+    "cypress": "^4.1.0",
```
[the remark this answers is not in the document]
shuang2411: No, I do not think this is needed. It must be added unintentionally when I try to install cypress.

shuang2411:
> We're almost there! Thank you, @shuang2411 🎉
> 
> I left a couple of comments below, mostly related to the changes in other files we probably don't need. But I also discovered an edge case while testing your changes:
> 
> If the user is in an incident channel, and the RHS is closed, executing `/incident list` opens the RHS, but shows the incident details instead of the list. I think that this happens because the default when the RHS opens in an incident channel is to view the incident details, so that's the one shown. Even if we later dispatch the `setRHSViewingList`, I think that we are probably competing with the `setRHSViewingIncident` action dispatched [here](https://github.com/mattermost/mattermost-plugin-incident-management/blob/master/webapp/src/components/rhs/rhs_main.tsx#L36) 🤔 I cannot think of a sensible way to solve this without modifying the main behaviour of the RHSMain component. Any idea on this, @lieut-data?

@agarciamontoro 
Thank you for your feedback!
I think you are right. All the changes you commented are not necessary. Sorry, I should have kept my working tree clean.

shuang2411 (commented):
@agarciamontoro 
Thank you for your feedback!
I think you are right. All the changes besides the first one are not necessary. Sorry, I should have kept my working tree clean. 

Btw, is their any thing I need to do from my end? I am new to open source. Please definitely let me know if there is anything I can do!

agarciamontoro:
Hi @shuang2411! You can simply add a new commit reverting those changes. For example, removing the files not needed, commiting that and pushing again. Ping me in Mattermost if you need any more help with it!

agarciamontoro (changes requested):
This looks great, @shuang2411! :tada:

I just left a comment on a change in package-lock.json that is still there, but otherwise, it is ready! :tada:

Jesse and I talked about the concern I had about this edge case where `/incident list` opens the incident details in an incident channel with the RHS closed. Solving that is not trivial, and it will be a bit more involved, so we can merge this PR as it is (after removing the change in package-lock.json) and I will create another ticket to address that specific problem, so we can tackle it afterwards. If you're open to continue contributing, please feel free to take it (I'll link to it here in this PR when I create it)!

webapp/package-lock.json
```diff
@@ -13082,7 +13082,7 @@
         "compass-mixins": "0.12.10",
         "core-js": "3.6.5",
         "css-vars-ponyfill": "2.3.1",
-        "custom-protocol-detection": "github:ismailhabib/custom-protocol-detection",
+        "custom-protocol-detection": "github:ismailhabib/custom-protocol-detection#e51da5c97d37d4ea04d872fb48467d793e7741ea",
```
agarciamontoro, line 13085: This is the change we need to remove.

lieut-data (approved):
Looks great -- thanks @shuang2411!

agarciamontoro (approved):
Woohoo! Thank you, @shuang2411, and sorry for all the back and forth in this!

Let's move into the QA process so the PR can get merged. There is no need for you to do anything else for the moment, unless QA requests some changes! :tada:

shuang2411:
@agarciamontoro Thank you so much!! Thank you for all your helps during the process!

agarciamontoro:
@prapti, heads-up that this PR covers most of the functionality expected in the ticket, except for one edge case that has been postponed for a future ticket: if the user is in an incident channel, and the RHS is closed, `/incident list` shows the incident details instead of the incident list. We'll fix this in https://mattermost.atlassian.net/browse/MM-31484.

prapti (changes requested):
Thank you for your work on this PR @shuang2411!
The improvement in the functionality of the command works great on the webapp, and still works as expected in mobile apps.
The only suggestion I have is to show the message `The list of your incidents is already open in the right hand side of the channel` when the RHS is already showing the incident list. Or perhaps getting rid of the work "already" so that it does not read strange when issuing the command when RHS is not open or open with a different view.
Other than that, I think it's great.

shuang2411:
> Thank you for your work on this PR @shuang2411!
> The improvement in the functionality of the command works great on the webapp, and still works as expected in mobile apps.
> The only suggestion I have is to show the message `The list of your incidents is already open in the right hand side of the channel` when the RHS is already showing the incident list. Or perhaps getting rid of the work "already" so that it does not read strange when issuing the command when RHS is not open or open with a different view.
> Other than that, I think it's great.

Hi @prapti,
I think I have already implemented the function you suggested.
An ephemeral post will appear after you input `/incident list`.
Is this what you want?
image

If you some how does not see this, can you let me know how to reproduce it? Thank you!

server/command/command.go
```diff
@@ -406,6 +406,18 @@ func (r *Runner) actionList() {
...
 
+	session, err := r.pluginAPI.Session.Get(r.context.SessionId)
+	if err != nil {
+		r.warnUserAndLogErrorf("Error retrieving session: %v", err)
+		return
+	}
+
+	if !session.IsMobileApp() {
+		// The RHS was opened by the webapp, so inform the user
+		r.postCommandResponse("The list of your incidents is already open in the right hand side of the channel.")
```
prapti, line 417:
```suggestion
		r.postCommandResponse("The list of your incidents is open in the right hand side of the channel.")
```

prapti (commented):
Hi @shuang2411, yeah removing "already" should be fine in my opinion. However, wondering if you've pushed the change after removing it? I'm still seeing the older ephemeral message.

server/command/command.go
shuang2411: I have not pushed yet. I just want to make such the ephemeral post is working fine for you. I will push now!

shuang2411: @prapti Please review the latest commit！

prapti (approved):
This is awesome! Thank you for your contribution @shuang2411 and really appreciate your patience through the PR review. Your changes are going in to the Incident Management product 🎉

prapti:
@agarciamontoro This one is ready for merge (I think you're the only dev available who can merge this, so TIA!) 😄

agarciamontoro:
Thank you, @prapti!

And thank you for the contribution @shuang2411 :tada: It's awesome :)

PR #414, opened 2020-12-14
MM-31240: Set min height on backstage header

lieut-data:
#### Summary
Set a minimum height on the headers of the playbook backstage edit screen to prevent their collapsing under flex rules.

Before:
image
image

After:
image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-31240

### DIFF
--- a/webapp/src/components/backstage/backstage.tsx
+++ b/webapp/src/components/backstage/backstage.tsx
@@ -24,11 +24,9 @@ import PlaybookIcon from '../assets/icons/playbook_icon';
 import IncidentIcon from '../assets/icons/incident_icon';
 
 const BackstageContainer = styled.div`
-    overflow: hidden;
     background: var(--center-channel-bg);
-    display: flex;
-    flex-direction: column;
     height: 100%;
+    overflow-y: auto;
 `;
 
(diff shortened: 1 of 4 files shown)


crspeller (approved):

cpoile (approved):

prapti:
@lieut-data The original issue is indeed fixed, and the tabs no longer render on top of the line; however, I'm seeing a slight movement when switching between the tabs (first screengrab below), which wasn't there originally (2nd screengrab below). Should this be addressed?

Movement when changing tabs:
screen flicker

Without movement:
no flicker

lieut-data:
Thanks, @prapti -- good eyes :)

Fixing this was a bit more involved, but I've tested my changes across Firefox/Chrome/Safari, and things are looking better. In short, I've reworked the container / header paradigm to leverage `position: sticky` consistently, making the header sticky in all contexts instead of just some of the time.

I also took the liberty of aligning the the "Tasks" header in the playbook edit screen since it was bugging me:
image

prapti (approved):
Agreed, this is much better for sure!
Thanks Jesse!

Approving.

PR #415, opened 2020-12-14
MM-31242: fix unusual backstage shadows

lieut-data:
#### Summary
Safari's treatment of `linear-gradient` has `transparent` meaning black transparent (namely `rgba(0,0,0,0)` -- [reference](https://stackoverflow.com/questions/30673204/css-linear-gradient-transparency-misbehaving-only-in-safari)).

This is apparently according to spec, but results in the undesirable behaviour below as we try to fade out the image assets in the backstage:
image

Fix this by making a transparent version of the background colour over which to apply the gradient ([reference](https://stackoverflow.com/questions/30673204/css-linear-gradient-transparency-misbehaving-only-in-safari)):
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31242

### DIFF
--- a/webapp/src/components/assets/left_fade.tsx
+++ b/webapp/src/components/assets/left_fade.tsx
@@ -6,7 +6,7 @@ const LeftFade = styled.div`
     top: 85px;
     left: 0;
     height: 100%;
-    background: linear-gradient(90deg, var(--center-channel-bg) 0%, rgba(255, 255, 255, 0) 94.89%);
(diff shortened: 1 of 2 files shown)


crspeller (approved):

cpoile (approved):

prapti (approved):
Verified that this PR fixes the shadow in Safari; regression tested in Chrome and Firefox. LGTM!

PR #416, opened 2020-12-14
MM-31243: broadcast channel tweaks

lieut-data:
#### Summary
Keep showing `Unknown Channel` when a private channel is configured as the broadcast channel in the backstage, but restore showing `Select a channel` when none is selected at all.

As part of writing the unit tests here, I discovered it was impossible to type a channel name: search was broken. I restored this behaviour with a simple `includes` scan on id, name and display name.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-31243

### DIFF
--- a/tests-e2e/cypress/integration/backstage/playbook_details_spec.js
+++ b/tests-e2e/cypress/integration/backstage/playbook_details_spec.js
@@ -139,7 +139,7 @@ describe('backstage playbook details', () => {
             cy.get('#root').findByText('Preferences').click();
 
             // # Open the broadcast channel widget and select a public channel
-            cy.findByTestId('playbook-preferences-broadcast-channel').click().type('off-topic{enter}');
+            cy.findByTestId('playbook-preferences-broadcast-channel').click().type('off-topic{enter}', {delay: 200});
 
             // # Save the playbook
             cy.findByTestId('save_playbook').click();
@@ -162,7 +162,7 @@ describe('backstage playbook details', () => {
             cy.get('#root').findByText('Preferences').click();
 
             // # Open the broadcast channel widget and select a public channel
-            cy.findByTestId('playbook-preferences-broadcast-channel').click().type('autem-2{enter}');
+            cy.findByTestId('playbook-preferences-broadcast-channel').click().type('autem-2{enter}', {delay: 200});
 
             // # Save the playbook
             cy.findByTestId('save_playbook').click();
@@ -185,7 +185,7 @@ describe('backstage playbook details', () => {
             cy.get('#root').findByText('Preferences').click();
 
             // # Open the broadcast channel widget and select the private channel
-            cy.findByTestId('playbook-preferences-broadcast-channel').click().type(privateChannelId + '{enter}');
+            cy.findByTestId('playbook-preferences-broadcast-channel').click().type(privateChannelId + '{enter}', {delay: 200});
 
             // # Save the playbook
             cy.findByTestId('save_playbook').click();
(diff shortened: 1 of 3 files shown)


webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -393,11 +393,13 @@ const PlaybookEdit: FC<Props> = (props: Props) => {
                                             {'Broadcast the incident status to an additional channel. All status posts will be shared automatically with both the incident and broadcast channel.'}
                                         </BackstageSubheaderDescription>
                                     </BackstageSubheaderText>
-                                    <ChannelSelector
-                                        onChannelSelected={handleBroadcastInput}
-                                        playbook={playbook}
-                                        isClearable={true}
-                                    />
+                                    <div data-testid='playbook-preferences-broadcast-channel'>
```
lieut-data, line 396: I didn't like have to wrapping this component, but I updating the chain of code to pass through the `data-testid` directly seemed ominous. Suggestions welcome!

crspeller (approved):

webapp/src/components/backstage/playbook_edit.tsx
agarciamontoro: This is only to retrieve the component in the E2E tests? Can't we use a selector to get it?

webapp/src/components/backstage/channel_selector.tsx
```diff
@@ -39,9 +39,20 @@ const ChannelSelector: FC<Props> = (props: Props) => {
     };
 
     const channelsLoader = (term: string, callback: (options: OptionsType<Channel>) => void) => {
-        callback(selectableChannels);
+        callback(selectableChannels.filter((channel) => {
+            if (term.trim().length === 0) {
+                return true;
+            }
```
agarciamontoro, line 45: Nit: can we move this outside of the filter? Otherwise, it will simply return true for every channel in selectableChannels, right? I mean something like
```js
if (term.trim().length === 0) {
    return selectableChannels;
} else {
    return selectableChannels.filter(.....);
}
```

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
```diff
@@ -84,4 +84,127 @@ describe('backstage playbook details', () => {
...
+            cy.visit('/ad-1/com.mattermost.plugin-incident-management/playbooks/' + playbookId);
+
+            // # Switch to Preferences tab
+            cy.get('#root').findByText('Preferences').click();
+
+            // * Verify placeholder text is present
+            cy.findByTestId('playbook-preferences-broadcast-channel').should('have.text', 'commodi');
+        });
+
+        it('shows "Unknown channel" when private broadcast channel configured and user it not a member', () => {
```
agarciamontoro, line 180: Nit
```suggestion
        it('shows "Unknown channel" when private broadcast channel configured and user is not a member', () => {
```

agarciamontoro (approved):
Nice! :tada: I left a couple of non-blocking comments, but it looks good!

webapp/src/components/backstage/playbook_edit.tsx
lieut-data: I just caught myself up on https://developers.mattermost.com/contribute/webapp/end-to-end-tests/#writing-specs -- in turns out that while `data-*` is the Cypress-recommended way of doing things, our internal preference seems to be `id` attributes. I'll fix that, but I think the answer to using only selectors remains the same advice from the Cypress docs:

> Every test you write will include selectors for elements. To save yourself a lot of headaches, you should write selectors that are resilient to changes.

webapp/src/components/backstage/channel_selector.tsx
lieut-data: Good idea! Fixed.

prapti (approved):
Verified that the changes work as expected: the channel selection box now shows "Select a channel" when no channel is selected, and the channel search now works when typed. The private vs public channel message in the update modal appears as expected.
LGTM!

PR #419, opened 2020-12-18
MM-31205 - Refactoring & use the redux as the source of truth for the frontend

cpoile:
#### Summary
- Getting rid of Header since we don't use it anymore
- Using the redux store for all rhs work now, including the title. We no longer have to wait for a call to the backend when switching to an incident channel, which means no loading spinner, and a much snappier feeling to the UI -- it doesn't lag while switching, and the title changes instantly instead of showing the previous incident name for a split second like it used to.
- This also fixed MM-31205 -- the 404 console error when switching to non-incident channels: we no longer hit the backend to check if the channel is an incident, therefore no 404.
- I split the commits into two self-contained chunks for easier reviewing

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-31205


agarciamontoro (approved):
Oh, getting rid of the `Header` and having the store as the source of truth is the greatest Christmas present!

One question about the latter: are we sure the store is synced with the state in the server? Or have you seen other places where it may not be?

webapp/src/components/rhs/rhs_details_view.tsx
```diff
@@ -3,29 +3,25 @@
...
-    if (incidentFetchState === IncidentFetchState.Loading) {
-        return spinner;
-    } else if (incidentFetchState === IncidentFetchState.NotFound || incident === null) {
-        // This should not happen--if incident is not found or null, we should be viewing the list.
-        // Returning the spinner so that if it ever happens, we at least show something.
-        return spinner;
+    // This shouldn't happen, because we're only rendering this view after calling inIncidentChannel,
+    // but we need to guard against the undefined that could be returned by currentIncident:
+    if (!incident) {
+        return <></>;
```
lieut-data, line 24: Can we return `null` to make this more explicit (at least as I understand React semantics).

webapp/src/components/rhs/rhs_details_view.tsx
```diff
@@ -3,29 +3,25 @@
...
+    const incident = useSelector<GlobalState, Incident | undefined>(currentIncident);
     const currentTabState = useSelector<GlobalState, RHSTabState>(currentRHSTabState);
 
-    if (incidentFetchState === IncidentFetchState.Loading) {
-        return spinner;
-    } else if (incidentFetchState === IncidentFetchState.NotFound || incident === null) {
-        // This should not happen--if incident is not found or null, we should be viewing the list.
-        // Returning the spinner so that if it ever happens, we at least show something.
-        return spinner;
+    // This shouldn't happen, because we're only rendering this view after calling inIncidentChannel,
```
lieut-data, line 21: nit: This comment /feels/ superfluous. It's true that it can't happen in the larger code base as currently used, but it seems perfectly possible that someone in the future might try to use a `RHSDetailsView` in the wrong way, and this is exactly how the component would be expected to repond: i.e. it's not unusual.

lieut-data (approved):
I love the simplifications! Do we still have a mechanism to keep the incidents in sync in the event we miss websocket messages?

webapp/src/components/rhs/rhs_details_view.tsx
cpoile: We can, but then TS turns it into a `JSX.Element | null` -- I suppose we could widen it to the type that includes null--I'll check that!

cpoile:
@agarciamontoro @lieut-data -- about keeping the incidents in sync, I looked through the code, and it looks like we're updating the frontend on any changes to the incident with websockets, but I'm not sure if there's a way to be certain that it's 100%reliable. I think the best we can do is apply the change and then use it and see if it misses anything. It's not a great solution though... 

An alternative is to keep the code that hits the backend whenever changing channel, but update the store with what we receive (instead of waiting on it, like we used to). That would lead to a responsive UI, and fill in anything that was maybe missed with websockets. Unfortunately, that would bring back the 404 error for any channel that isn't an incident.

I think we rely on websockets in the server code for keeping posts up to date, don't we? In that case, can we assume they're reliable? We already refetch incidents on a websocket disconnect and on changing teams, so we don't have to worry about those edge cases.

webapp/src/components/rhs/rhs_details_view.tsx
lieut-data: There are days I feel like TS adds an unnecessary burden to frontend programming... 😕

lieut-data: (But I'm flexible either way -- non-blocking comment.)

lieut-data:
I support "letting it soak" to vet the theory, and agree that the reconnection logic /should/ cover us here. Nice analysis -- thanks for confirming :)

cpoile:
Cool! I added @prapti because it is a big change. Prapti, for this it's basically all the RHS interactions/functionality... So maybe it would be more convenient to fold this into the 1.3 release testing in general? Your call. :)

webapp/src/components/rhs/rhs_details_view.tsx
cpoile: Looks like we can return a null and TS just propogates the union null type, and nothing seems to break, so that was easy. :)

prapti:
I'm still testing these changes and would like to test it on a fresh (cleaned and nuked) local server one more time tomorrow, since I'm seeing some JS errors that I can't repro in cloud servers. Two errors that are quite sporadic only on my local server are the `Uncaught SyntaxError` when checking/unchecking tasks in RHS (something that was fixed previously) and a possible memory leak when switching licenses, that I'd like to try and understand the scenario for better, and also try to repro a plugin crash. I really appreciate all your patience! I'll definitely try and provide more clarity on the errors tomorrow if they keep happening. Thank you!

prapti (approved):
Can verify that the errors I was seeing are not related to this PR. Approving this PR as all IM plugin features tested (mainly pertaining to the Incident RHS) are working normally. I will continue to investigate the issues outside this set of changes.

PR #420, opened 2020-12-29
MM-25755: promote incident commander to channel admin

Hassall:
#### Summary
promotes incident commander to channel admin

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/16518

### DIFF
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -984,6 +984,10 @@ func (s *ServiceImpl) createIncidentChannel(incdnt *Incident, public bool) (*mod
 		return nil, errors.Wrapf(err, "failed to add user to channel")
 	}
 
+	if _, err := s.pluginAPI.Channel.UpdateChannelMemberRoles(channel.Id, incdnt.CommanderUserID, model.CHANNEL_ADMIN_ROLE_ID); err != nil {
+		return nil, errors.Wrapf(err, "failed to promote incident commander to channel_admin role")
+	}
+
 	return channel, nil
 }
 


lieut-data (approved):
Code looks good here -- thanks, @Hassall! The e2e tests are failing due to upstream changes. I'll open a PR to tackle this and we should be able to get this build passing.

crspeller (approved):

prapti (changes requested):
Hi @Hassall, I'm testing this PR and am running into a slight problem -- your changes work great for a user that has the system administrator privileges, however, throws a `403 forbidden` error for a non-system admin user. What I'm seeing is that when creating an incident as a non-admin user, even though the incident gets started and the incident channel gets created successfully, the app loads a "Channel Not Found" page with the 403 error. Would you be able to look a bit more into this?

@lieut-data Wanted to double check that the updating a user to channel admin when changing commanders will be handled in a separate ticket/PR based on your last comment in the Jira ticket.

lieut-data:
@prapti, confirmed re: a separate ticket to consider promoting commanders to channel admin. Right now, that's "not safe", because it can violate the default system permission scheme, whereas doing so on incident start is equivalent to just creating a channel by that user. Thanks for checking!

Hassall:
@prapti 
Thanks

@lieut-data @crspeller 
I've spent a bit of time and can't seem to find out why this issue occurs for non-system admin users.

The issue occurs on line 405 in incidents.go : The user does not have permission to enter the channel. (on line 47 permissions.go, it doesn't manage to hit any conditions and returns `ErrNoPermissions` error) 

I've confirmed that for both the system admin user and non-system admin user, they have the same privileges (`channel_admin` as role and `schemeAdmin` is true).

lieut-data:
@Hassall, sorry it took so long to get back to you! I had a chance to figure out what's going on, and it turns out the plugin API under the covers is extremely unintuitive here: setting only `model.CHANNEL_ADMIN_ROLE_ID` /unsets/ `model.CHANNEL_USER_ROLE_ID`, making them an admin but apparently not giving them access to the channel they administrate. It /seems/ like this is probably a bug, but I've fixed it in your proposed changes by simply setting both.

As an aside, our CircleCI servers seem to be unable to build this commit. It's an obscure issue documented over at https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization- that involves your user account following your fork on CircleCI instead of the upstream. Can you try unfollowing your fork on CircleCI to see if this fixes the build issue?

Hassall:
@lieut-data 
No problem. Thanks for the clarity. 

I've unfollowed my fork on CircleCI.

prapti:
On test queue.

prapti (approved):
Thank you for looking this @Hassall and @lieut-data!
The incident reporter is appropriately being promoted to a channel admin now, for both system admins and non-admin team members. Couldn't break it this time :)
Approving!

p.s. @lieut-data, For the incidents that have already started prior to this PR, where the reporter is still not a channel admin, do we want to promote them to channel admins as well in the future? I'm guessing probably not, but wanted to make sure to keep it in mind if we take that approach.

lieut-data:
@prapti, I think it's fine to leave existing incidents untouched, and only promote in new ones. Thanks for confirming the behaviour, and thanks for the assist @Hassall!

lieut-data:
Thanks again, @Hassall!

PR #421, opened 2021-01-05
MM-31634 Remove stages and restore single checklist.

crspeller:
#### Summary
Removes stages and restores a single checklist. Prelude to incident status. 

#### Screenshot
image
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31634

#### Testing Notes
Try and see if you can find places we still use the word `Stage` in the UI.


crspeller:
Note that the e2e tests are broken, so this will require some updates to the tests once I am able to do so.

server/api/incidents.go
```diff
@@ -137,10 +136,11 @@ func (h *IncidentHandler) createIncidentFromPost(w http.ResponseWriter, r *http.
 	ReturnJSON(w, &newIncident, http.StatusCreated)
 }
 
+// Note that this current does nothing. This is temporary given the removal of stages. Will be used by status.
```
lieut-data, line 139:
```suggestion
 // Note that this currently does nothing. This is temporary given the removal of stages. Will be used by status. 
```

server/incident/incident_test.go
```diff
@@ -1,238 +1 @@
 package incident
-
```
lieut-data, line 2: It's not obvious why this file is being removed?

server/telemetry/rudder.go
```diff
@@ -27,7 +27,6 @@ const (
 	actionCreate          = "create"
 	actionEnd             = "end"
 	actionRestart         = "restart"
-	actionChangeStage     = "change_stage"
```
lieut-data, line 30: Let's make sure we queue up docs changes to reflect this change.

lieut-data (approved):
Looks good!

server/telemetry/rudder.go
crspeller: https://mattermost.atlassian.net/browse/MM-31732

server/incident/incident_test.go
crspeller: Because these tests break for any change and are annoying to update. I thought it better for everyone's sanity to remove. Also not sure if they are guaranteed stable between go versions.

lieut-data: If we didn't have a custom `MarshalJSON` in play, I'd tend to agree, but it seems like these are here to assert desired behaviour around the handling of `nil` vs. empty slices for frontend rendering.

crspeller: Thought of a way to do it without the maintenance burden. Pushed a simple test.

server/incident/incident_test.go
```diff
@@ -1 +1,17 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestIncident_MarshalJSON(t *testing.T) {
+	testIncident := &Incident{}
+	result, err := json.Marshal(testIncident)
+	require.NoError(t, err)
+	// Should not contain null. Triggering this?
+	// Add your new nullable thing to one of the MarshalJSONs in incident/incident.go
+	require.NotContains(t, string(result), "null")
```
lieut-data, line 16: :tada:

webapp/src/components/rhs/rhs_incident_summary.tsx
```diff
@@ -39,50 +37,46 @@ const RHSIncidentSummary: FC<Props> = (props: Props) => {
...
+                    <div className='first-title'>
+                        {'Duration'}
+                        <Duration
+                            from={props.incident.create_at}
+                            to={props.incident.end_at}
+                        />
                     </div>
                 </div>
-            </Scrollbars>
-            <RHSFooterSummary/>
```
cpoile, line 84: At the moment there's no way to end the incident via the RHS, but that will be added back in with the new features, right?

cpoile (approved):
Looks great. Non-blocking questions.

cpoile:
I found `Stage` in the UI:
image

lieut-data:
(Removing request for @prapti until merge conflicts resolved.)

crspeller:
@prapti Merge conflicts resolved.

webapp/src/components/rhs/rhs_incident_summary.tsx
prapti: @crspeller Wondering the same ☝️

prapti (approved):
Couldn't break it, looks good!
Thanks for taking care of the e2e tests too @crspeller!

PR #422, opened 2021-01-05
[MM-31683] Fixing e2e and postgres

cpoile:
#### Summary
- upgraded postgres in the e2e tests
- added a bunch of little things to stop the e2e tests from failing. E.g., the "Search" bar and its drop down were activating after clicking on the incident-management channel header button, and hiding components below it. Why now? I don't know.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-31683

### DIFF
--- a/tests-e2e/cypress/integration/frontstage/incident_rhs_list_spec.js
+++ b/tests-e2e/cypress/integration/frontstage/incident_rhs_list_spec.js
@@ -94,6 +94,9 @@ describe('rhs incident list', () => {
             // # Click the incident icon
             cy.get('#channel-header').within(() => {
                 cy.get('#incidentIcon').should('exist').click();
+
+                // # Remove the Search dropdown that covers the rhs
+                cy.get('#searchBox').click().type('{esc}');
             });
 
             // * Verify we see the welcome screen when there are no incidents
@@ -129,6 +132,9 @@ describe('rhs incident list', () => {
                 // # Click the incident icon
                 cy.get('#channel-header').within(() => {
                     cy.get('#incidentIcon').should('exist').click();
+
+                    // # Remove the Search dropdown that covers the rhs
+                    cy.get('#searchBox').click().type('{esc}');
                 });
 
                 // * Verify the rhs list is open the incident is visible.
@@ -161,6 +167,9 @@ describe('rhs incident list', () => {
             // # Click the incident icon
             cy.get('#channel-header').within(() => {
                 cy.get('#incidentIcon').should('exist').click();
+
+                // # Remove the Search dropdown that covers the rhs
+                cy.get('#searchBox').click().type('{esc}');
             });
 
             // # start 2 incidents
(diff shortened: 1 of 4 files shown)


tests-e2e/cypress/integration/frontstage/incident_rhs_list_spec.js
```diff
@@ -94,6 +94,9 @@ describe('rhs incident list', () => {
             // # Click the incident icon
             cy.get('#channel-header').within(() => {
                 cy.get('#incidentIcon').should('exist').click();
+
+                // # Remove the Search dropdown that covers the rhs
```
lieut-data, line 98: Is this reproducible when using the core product? It reads like an actual regression we should fix, vs. patch the test.

tests-e2e/cypress/support/ui_commands.js
```diff
@@ -95,7 +95,14 @@ Cypress.Commands.add('verifyPostedMessage', (message) => {
 
 // verifyEphemeralMessage verifies the receipt of an ephemeral message containing the given
 // message substring. An exact match is avoided to simplify tests.
-Cypress.Commands.add('verifyEphemeralMessage', (message, isCompactMode) => {
+Cypress.Commands.add('verifyEphemeralMessage', (message, isCompactMode, needsToScroll) => {
+    if (needsToScroll) {
+        // # Scroll the ephemeral message into view
```
lieut-data, line 100: Is this another regression in the core product? Does it reproduce manually, or just in CI?

lieut-data (changes requested):
Postgres changes look great -- thanks! But I'm a bit wary about the proposed working around of issues, with comments below.

tests-e2e/cypress/integration/frontstage/incident_rhs_list_spec.js
cpoile: Yes, it happens on master, but not on community. Try by clicking on the incident management button, the search should be selected the dropdown will appear. I suppose I should file a ticket for this?

tests-e2e/cypress/support/ui_commands.js
cpoile: This is appears to be something that messes up cypress, it doesn't seem to be happening to me manually.

crspeller (approved):
LGTM except the search dropdown thing, but I am OK with just reverting that when it gets fixed.

tests-e2e/cypress/integration/frontstage/incident_rhs_list_spec.js
cpoile: Thanks @crspeller , that's exactly the issue. I guess SET will get a crack at it tomorrow. 

I think we should merge this PR so the tests aren't blocking our other PRs. Then revert the changes specific to the search box after https://mattermost.atlassian.net/browse/MM-31708 is fixed. (I can do that)

Sound good?

cpoile:
Ah, didn't see your comment here. Yep, that's a good plan.

lieut-data (approved):

PR #424, opened 2021-01-06
MM-29698: use channel name only

lieut-data:
#### Summary
As previously discussed, this PR deprecates the `Name` property of an incident and relies exclusively on the corresponding channel name. At the query level, it introduces a `JOIN` to the `Channels` table when querying incidents. When a channel name is updated, the frontned receives this websocket event and triggers an incident fetch if the channel is a known incident.  

I encountered a number of issues to be resolved before I could effect the desired changes, with the necessary refactoring broken up into separate commits:
* consistently use table aliases instead of unscoped columns (to facilitate the future `JOIN`)
* refactor tests to avoid global variables and isolate changes to the test in question
* clear user preferences for the sysadmin, `user-1` and `user-2` changes before running the test suite, so as to prevent the tutorial or "what's new" modals from interfering with the tests

I'd recommend reviewing the individual commits in question first. The final commit contains the changes specific to the linked ticket.

Note that these changes also currently depend on @cpoile's https://github.com/mattermost/mattermost-plugin-incident-management/pull/419, though I intend to wait for that to merge and rebase this PR accordingly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29698


lieut-data:
The failing tests here /appear/ to be spurious: something to do with the bulk export slash commands.

tests-e2e/cypress/support/api/user.js
```diff
@@ -0,0 +1,456 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
crspeller, line 1: I assume this is copy paste from server cypress tests? Wish there was way to share this stuff.

crspeller (approved):

tests-e2e/cypress/support/api/user.js
lieut-data: Yes, I should have called that out -- a ton of the "new" code is the existing work in the server. @prapti, perhaps this is something we can raise with the e2e team to avoid this duplication going forward? (i.e. release an NPM package dedicated to such an effort?)

tests-e2e/cypress/support/env.js
```diff
@@ -0,0 +1,17 @@
...
+
+export function getAdminAccount() {
+    return {
+        username: Cypress.env('adminUsername'),
+        password: Cypress.env('adminPassword'),
+        email: Cypress.env('adminEmail'),
+    };
+}
+
+export function getDBConfig() {
```
cpoile, line 12: Was this included by mistake?

tests-e2e/cypress/support/index.js
```diff
@@ -10,9 +10,23 @@ import './server_api_commands';
...
 
+before(() => {
+    // Disable the tutorial, cloud onboarding, and whats new modal for specific users.
+    const userNames = ['sysadmin', 'user-1', 'user-2'];
+
+    cy.apiAdminLogin().then(() => {
+        cy.apiGetUsersByUsernames(userNames).then(({users}) => users.forEach((user) => {
+            cy.apiSaveTutorialStep(user.id, '999');
+            cy.apiSaveCloudOnboardingPreference(user.id, 'hide', 'true');
+            cy.apiHideSidebarWhatsNewModalPreference(user.id, 'true');
```
cpoile, line 25: I couldn't find these (because I wanted to see how they did it) -- are they in this PR?

tests-e2e/cypress/support/index.js
```diff
@@ -10,9 +10,23 @@ import './server_api_commands';
 import './ui_commands';
 import './plugin_ui_commands';
 import './api/preference';
+import './api/user';
 
 require('cypress-terminal-report/src/installLogsCollector')();
 
+before(() => {
+    // Disable the tutorial, cloud onboarding, and whats new modal for specific users.
```
cpoile, line 18: this is great, thank you!

cpoile (approved):
Nice, thank you!  Just some non-blocking questions

cpoile:
I thought those had been fixed by the scrolling... I'm running them again and will keep an eye on them.

tests-e2e/cypress/support/env.js
lieut-data: I believe it's included because it gets referenced by some of the other "new" code copied from the webapp -- I kept it, even if we strictly don't need all the functions, because of my assumption that we'll eventually move towards replicating the webapp model of generating dynamic test data (instead of requiring `make test-data`).

tests-e2e/cypress/support/index.js
lieut-data: They actually already existed in a previous copy from the webapp -- for example: https://github.com/mattermost/mattermost-plugin-incident-management/blob/3faabb36bdf78774cf3f56c49d0925d9cdd190e8/tests-e2e/cypress/support/api/preference.js#L226

PR #425, opened 2021-01-06
MM-31007: handle err fetching incident by channel

lieut-data:
#### Summary
When a user is added to a channel, we query the server to determine if this is an incident. If not, the client dutifully throws an 404 error, but we failed to handle that from the caller and treat it as a non-error. Fix this.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-31007


crspeller (approved):

cpoile (approved):
This is interesting. I wonder why it's able to catch the error here, but it didn't when the try/catch was in the `client.ts` file. Very weird. This prevents the 404 from being logged in the console?

lieut-data:
@cpoile, it doesn't prevent logging a 404 -- I think that's default (and OK) behaviour by the browser --  but it does prevent the uncaught promise error.

cpoile:
@lieut-data That's what catching in `client.ts` does too, but I really wanted to get rid of the console 404 notice. Too bad. :)

PR #426, opened 2021-01-06
Remove search bar bug workarounds

cpoile:
#### Summary
- Remove the search bar bug workarounds

#### Ticket Link
- none


crspeller (approved):

lieut-data (commented):
Commented in channel, but I'm actually inclined to keep things the way they are. Ideally, this would have been caught by the webapp's own e2e test pipeline before it hit us, but in the grand scheme of things, I'd rather catch this failure more quickly than end up with a regression that's entirely outside our product for which we have no visibility.

cpoile:
@lieut-data I agree that we want to surface when changes break our code, but if the changes are bugs then we're left with the choice between patching our tests temporarily, or halting merging all our PRs until the bug in master is fixed. Personally, I think both of those options are worse than the danger of missing a real breaking change. (But that's debatable.)

What if we pinned to the most recent RC, and to the most recent release when we don't have an RC? That should give us some time for us to surface problems with new features breaking our code, but will stop bugs on master from preventing us merging our PRs.

lieut-data:
I think my specific concern is about cloud -- do we really want to be updating our pins every two weeks to discover unexpected breakage? Honestly, this particular issue is "so bad", it should have been reverted from `-daily` immediately upon being discovered. At the risk of sounding like I'm shifting the blame, I think this is an upstream problem that we shouldn't be trying to work around. We just happen to have an e2e pipeline that runs more frequently (?) than the webapp's own tests.

crspeller:
@lieut-data 
> Ideally, this would have been caught by the webapp's own e2e test pipeline before it hit us

That's true. Why didn't this break the main e2e tests?

lieut-data:
Assuming there was coverage, I /suspect/ the delay is because the webapp e2e pipeline runs nightly, and not on each PR. With rounding, it could be days before it gets reverted or fixed. @prapti, I'm curious on your opinions here in case my mental model for e2e testing is invalid.

cpoile:
Switched back to master. Now this PR is only about removing the search bar bug workarounds.

lieut-data (approved):

PR #427, opened 2021-01-08
Fix e2e tests again

cpoile:
#### Summary
- This time the webapp's cypress code was a bit different than our previous code, and cypress needed to be updated to get the `ensureScrollable` option. 
- Updated cypress and related packages to match what we use in webapp.

#### Ticket Link
- none


lieut-data (approved):
Nice -- thank you!

crspeller (approved):

PR #428, opened 2021-01-14
MM-31919: set id when rendering Textbox

lieut-data:
#### Summary
The `Textbox` component from webapp requires an `id`, throwing an exception in development environments when we validate props at runtime.

This doesn't appear in a cloud/production environment, where runtime prop checking is disabled, and thus must be verified locally.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-31919

### DIFF
--- a/webapp/src/components/backstage/step_edit.tsx
+++ b/webapp/src/components/backstage/step_edit.tsx
@@ -4,7 +4,7 @@ import styled, {createGlobalStyle} from 'styled-components';
 
 import {ChecklistItem} from 'src/types/playbook';
 import {TertiaryButton} from 'src/components/assets/buttons';
-import {uniqueId} from 'src/utils';
+import {useUniqueId} from 'src/utils';
 
 import {StyledTextarea} from './styles';
 
@@ -236,7 +236,7 @@ const StepCommand: FC<StepCommandProps> = (props: StepCommandProps) => {
     const [focus, setFocus] = useState(false);
     const [hover, setHover] = useState(false);
     const ref = useRef(null);
-    const [id] = useState(uniqueId('step-command-'));
+    const id = useUniqueId('step-command-');
 
     useEffect(() => {
         if (focus && ref && ref.current) {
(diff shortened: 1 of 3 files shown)


agarciamontoro:
It looks like the linter does not find the import :thinking:

lieut-data:
Oops! Looks like I might be relying on the implicit presence of `lodash` from internal dependencies. I'll find a replacement instead.

lieut-data:
I've replaced the lodash import with a trivial implementation of my own that should suffice. Sorry for the thrashing!

webapp/src/components/backstage/step_edit.tsx
```diff
@@ -235,6 +236,7 @@ const StepCommand: FC<StepCommandProps> = (props: StepCommandProps) => {
     const [focus, setFocus] = useState(false);
     const [hover, setHover] = useState(false);
     const ref = useRef(null);
+    const [id] = useState(uniqueId('step-command-'));
```
crspeller, line 239: This will actually call `uniqueId` every render, incrementing the idCounter as a side effect. The result will be ignored though. See: https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

Also it would be nice and clean if you make a custom hook like `useUniqueID` and used that. That way all the implementation details like the call to `uniqueId` having to be in a callback and the destructuring of id can be hidden.

crspeller (changes requested):

webapp/src/components/backstage/step_edit.tsx
lieut-data: Ooh, good catch! Will create a custom hook instead -- thanks :)

crspeller (approved):

agarciamontoro (approved):
Nice! :tada:

prapti (approved):
The "Failed prop type" error no longer appears when creating a playbook with the "Incident Management" template.
LGTM!

PR #432, opened 2021-01-21
MM-31940 - RHS Timeline with status

cpoile:
#### Summary
- V0 of the timeline, only showing status for now
- Based on #429 

Looks like: 
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-31940


webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -0,0 +1,206 @@
...
+    const getPostFn = (postId: string) => getPostAction(postId)(dispatch as DispatchFunc, getStateFn);
+    const getUserFn = (userId: string) => getUserAction(userId)(dispatch as DispatchFunc, getStateFn);
+
+    const selectPost = useSelector<GlobalState, IdToPostFn>((state) => (postId: string) => getPost(state, postId));
+    const selectUser = useSelector<GlobalState, IdToUserFn>((state) => (userId: string) => getUser(state, userId));
+
+    const [events, setEvents] = useState<Event[]>([]);
+
+    useEffect(() => {
+        Promise.all(props.incident.status_posts.map(async (p) => {
```
crspeller, line 137: This might generate multiple requests per entry in the timeline correct? 
Maybe we should make an endpoint that returns the appropriate data for the timeline. Might be nice as we expand the timeline to include more data. The downside is that it would have to assume the client doesn't have anything it requires.

agarciamontoro (changes requested):
This is neat, I really like it! :tada:

My comments below are not blocking, but I tested this locally and I saw a couple of weird behaviours:
- When a status message is deleted, the whole timeline disappears after a refresh (maybe we need to filter out the deleted status posts?)
- In the previous case, or if there are no status updates at all (as in a brand new incident), the timeline renders a tiny bar:
image
A simple solution for this would be to always have a first event in the timeline: the incident creation.

Another comment, but maybe for v0.1: could we remove the bar below the last event? I mean converting this:
image
into this:
image
(maybe this a question for @alexsiclari)

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+        if (isMobile()) {
+            dispatch(toggleRHS());
+        }
+    };
+
+    return (
+        <Timeline>
+            {
+                events.map((event) => {
```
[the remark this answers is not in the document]
cpoile: Yep, that's better.

webapp/src/components/rhs/rhs_timeline.tsx
cpoile: I would worry about it, but we're not going to be using this code for v2 -- we're going to send over a pre-populated list of Events from the server, so no client-side fetching will be needed. All this code in the `useEffect` will go away.

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -0,0 +1,206 @@
...
+    const [events, setEvents] = useState<Event[]>([]);
+
+    useEffect(() => {
+        Promise.all(props.incident.status_posts.map(async (p) => {
+            let post = selectPost(p.id) as Post | undefined;
+
+            if (!post) {
+                const ret = await getPostFn(p.id);
+                if (!ret.data) {
+                    return Promise.reject(new Error('no post'));
```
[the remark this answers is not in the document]
cpoile: Yep, that's a good idea, done.

cpoile:
@agarciamontoro Nice catch on the line when there was no status posts -- I never even noticed it on the light theme. I changed around how the rendering works, and I'm putting in a `Created` event now, so that fixes that problem.

I'm also adding the event types, which are the same as the ones I have in the v2 branch. v2 will send the event list as an array in the incident, it will always be up to date in the store.

webapp/src/components/rhs/rhs_timeline.tsx
crspeller: Ok. If it's going away that's fine :D

crspeller:
> (funny--we currently don't record who created the incident anywhere :))

Not yet, but soon: https://github.com/mattermost/mattermost-plugin-incident-management/pull/430

crspeller (approved):

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -0,0 +1,116 @@
...
+            const displayName = displayUsername(user, displayPreference);
+            return {
+                type: TimelineEventType.StatusUpdated,
+                create_at: p.create_at,
+                post_id: p.id,
+                display_name: displayName,
+                status: p.status || 'unset',
+            } as TimelineEvent;
+        })).then((eventArray) => {
+            setEvents(eventArray.filter((e) => e) as TimelineEvent[]);
```
agarciamontoro, line 90: Hm, does this filter out the nulls? I don't understand how it works.

agarciamontoro (approved):
Great job here :tada:

webapp/src/components/rhs/rhs_timeline.tsx
cpoile: Yep, it filters everything that isn't null/undefined `(e) => e`

agarciamontoro: Oh god, it's crystal clear now... Friday evening is not a good time to review a PR.

prapti (approved):
Working great!
I've added and tested some 20 test cases relevant to the RHS Timeline in the Incident Collaboration test plan including tests for timeline event logs for incident creation, status update, change of commander, events as links to update-messages, event logging for reopened incidents and timestamp & duration verification, and all of them passed! 🎉  Couldn't break anything this time around.
The only thing I noticed that's worth mentioning is that when clicking on the timeline events to "jump" to the status update message, it takes noticeably longer for the page to load as opposed to clicking on permalinks. Is this a known issue?
Otherwise the rest looks great.

cpoile:
Thank you for the thorough review!

> The only thing I noticed that's worth mentioning is that when clicking on the timeline events to "jump" to the status update message, it takes noticeably longer for the page to load as opposed to clicking on permalinks. Is this a known issue?

@prapti Can you tell a difference in the time it takes to click on a timeline event and the time it takes to click on a status update timestamp (in the Summary tab)? It should be the same, since it's the same code. And the comparison is with a permalink to the same message in the same channel? (Just so I can try to repro)

prapti:
@cpoile I think the lag might be a server thing. It's not doing it on my local nor on cloud. I'll just keep an eye out to see if I see it again.

cpoile:
@prapti Awesome, please let me know if you see it again!

PR #433, opened 2021-01-21
MM-32148 - Generate version metadata from tags

cpoile:
#### Summary
- If the current commit has a tag, use that as the version (without the prefix 'v'). Otherwise, use the latest tag and add the short hash, eg. "1.2.0+9fcd0b92".
- Q: should we also update the `release_notes_url` in `plugin.json`?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32148


cpoile:
Doing some experimentation with e2e tests, will request reviews when that's all over with

build/setup.mk
```diff
@@ -4,8 +4,17 @@ ifeq ($(GO),)
     $(error "go is not available: see https://golang.org/doc/install")
 endif
 
+# Gather build variables to inject into the manifest tool
+BUILD_HASH_SHORT = $(shell git rev-parse --short HEAD)
+BUILD_TAG_LATEST = $(shell git describe --tags --abbrev=0)
```
lieut-data, line 9: Ooh, I haven't encountered `git describe` before. Should we leverage `--match v*` to match only version tags?

plugin.json
```diff
@@ -6,14 +6,15 @@
     "support_url": "https://github.com/mattermost/mattermost-plugin-incident-management/issues",
     "release_notes_url": "https://github.com/mattermost/mattermost-plugin-incident-management/releases/tag/v1.2.0",
     "icon_path": "assets/incident_plugin_icon.svg",
-    "version": "1.2.0",
+    "version": "1.2.0+9fcd0b92",
```
lieut-data, line 9: Propose we remove the `version` attribute from the checked-in file otherwise we'll always have a permanently old version checked in.

server/manifest.go
```diff
@@ -19,7 +19,7 @@ const manifestStr = `
   "support_url": "https://github.com/mattermost/mattermost-plugin-incident-management/issues",
   "release_notes_url": "https://github.com/mattermost/mattermost-plugin-incident-management/releases/tag/v1.2.0",
   "icon_path": "assets/incident_plugin_icon.svg",
-  "version": "1.2.0",
+  "version": "1.2.0+9fcd0b92",
```
lieut-data, line 22: Should we remove these files from the repo, and let them get generated only at build time? (Do we also `.gitignore` them?)

build/setup.mk
```diff
@@ -4,8 +4,17 @@ ifeq ($(GO),)
...
+BUILD_TAG_CURRENT = $(shell git tag --points-at HEAD)
+LDFLAGS += -X "main.BuildHashShort=$(BUILD_HASH_SHORT)"
+LDFLAGS += -X "main.BuildTagLatest=$(BUILD_TAG_LATEST)"
+LDFLAGS += -X "main.BuildTagCurrent=$(BUILD_TAG_CURRENT)"
```
lieut-data, line 13: nit: LDFLAGS seems like it might "have side effects" for downstream invocations. Should we inline these instead?

lieut-data (changes requested):
Nice! A few comments below.

build/setup.mk
cpoile: Nice, added!

plugin.json
cpoile: I tried to do that, but couldn't think of a way to do that--manifest will output the whole plugin.json, including the `version` field

server/manifest.go
cpoile: I was thinking of that, but then we would have to have the "base" `plugin.json` somewhere, so that we can read it in. I was thinking of putting it in the build directory. What do you think?

build/setup.mk
cpoile: Sure, will do.

server/manifest.go
cpoile: I made that change so we can see what it looks like. It's not perfect though, because we still update the manifest files with the version info, and I can't think of a way to not have to check that in. Maybe ignore the manifest files as well since they're generated?   Hmm, let's see if that will work.

cpoile: There, I think that works. Now the three generated files are not checked in. Only thing that feels a little "wrong" is putting the base `plugin.json` file in the build directory. Open to alternatives there. But if we keep it in the root directory it will be updated and we will start to check in old version numbers.

plugin.json
lieut-data: Can we just manually edit the file in this PR to remove the `version` attribute? I guess then the `manifest` will always leave a modified `plugin.json` in the repo... which would be annoying. Can we write the output to a temporary `plugin.json` (i.e. the one that's being put into the bundle path)?

server/manifest.go
lieut-data: Oops, commented above before I read this thread. Yeah, wondering if we can keep the `root` one the `base` one, and change `manifest` to write to the temporary one in the bundle directory at build time?

plugin.json
cpoile: Yep, we can write the plugin.json anywhere, and change the bundle instructions.  Will do that.

cpoile: There, the `plugin.json` is back in the root, but without the `version` field. The version field is generated and put into the build directory, and the make bundle copies that generated one to the dist folder. Looks good now, thanks!

server/manifest.go
cpoile: As a compromise we're writing into the build directory, because we rm the dist directory before copying the manifest, and I don't want to mess around with that. :)

cpoile:
@lieut-data @crspeller ready for re-review, then we can start testing the cloud build process. :)(

Makefile
```diff
@@ -2,7 +2,7 @@ GO ?= $(shell command -v go 2> /dev/null)
 NPM ?= $(shell command -v npm 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MM_DEBUG ?=
-MANIFEST_FILE ?= plugin.json
+MANIFEST_FILE ?= build/plugin.json
```
lieut-data, line 5: Hmm, this doesn't "smell right" to me -- at least from the context of reading the `Makefile`, `MANIFEST_FILE` is about defining the source of truth. This file doesn't exist yet: I'd love to see whatever temporary path we use be an implementation detail vs. something that's configured like this.

Maybe we just make pass the desired output path to `bin/manifest` and invoke it manually within the `bundle` target? Or, actually, "applying" `plugin.json` only needs to happen within the `bundle` target, so maybe it's just a different `bin/manifest apply_version` command or some such.

lieut-data (changes requested):
Thanks, @cpoile! One comment below about `build/plugin.json`.

Makefile
cpoile: Yep, that does work better, nice idea. I added a `manifest dist` command to write the `plugin.json` to the dist directory. Thanks!

lieut-data (approved):
:tada:

plugin.json
```diff
@@ -6,14 +6,14 @@
...
-    "version": "1.2.0",
     "min_server_version": "5.28.0",
     "server": {
         "executables": {
             "linux-amd64": "server/dist/plugin-linux-amd64",
             "darwin-amd64": "server/dist/plugin-darwin-amd64",
             "windows-amd64": "server/dist/plugin-windows-amd64.exe"
-        }
+        },
+        "executable": ""
```
crspeller, line 16: Why did this come back?

crspeller (approved):
Awesome!

plugin.json
cpoile: Looks like it was generated by marshalling the manifest struct. I can not check it in if it's causing confusion?

crspeller: @cpoile Nope its fine.

PR #440, opened 2021-01-26
MM-31560: Invite users on incident creation

agarciamontoro:
#### Summary

This PR begins the work on the automation features, adding the ability to invite users when an incident is created.

automation_sneakpeek

##### Changes in the backstage
- There's a new `Automation` tab that contains the settings for all automations (right now, only the first one, although I tried to organize the components so it's easy to add new ones).
- The list of invited users in the new setting is a bit different than what we had before: I implemented it as a row of overlapping avatars, something like the user selector in Jira. When we were working on the design, I did not notice we did not design how to render that (my bad!), so I improvised that UI (and that's why I'm adding @alexsiclari as reviewer too, mainly).
- There is a known bug with the user selector (https://mattermost.atlassian.net/browse/MM-32313), which I bypassed in the invite users selector by checking that the added user is not already in the list of users.

##### Changes in the DB
  - `ConcatenatedInvitedUserIDs`, which, unsurprisingly, contains a concatenated (comma-separated) string with the IDs of the users to invite.
  - `InviteUsersEnabled`, that stores the on/off state of the setting.
  - `ConcatenatedInvitedUserIDs`, which stores a copy of the data in the playbook's column when the incident is created. If the playbook's `InviteUsersEnabled` contains `false`, the data is not copied, so there's no need to add a `InviteUsersEnabled` column to the `IR_Incident` table.

##### Changes in the service
When the incident is created, every member in the `InvitedUserIDs` field is added to the channel; if there is some error, a message is posted in the incident channel.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31560
https://mattermost.atlassian.net/browse/MM-31565


agarciamontoro:
I'm still working on the E2E tests, but the code is ready for review.

agarciamontoro:
Interesting, all tests were passing locally :thinking:

alexsiclari (approved):
The pattern works for now. We will need to work on this to see if we have a different UI pattern for something like this. One side note, we will need to think through how this would look with adding multiple automation.

agarciamontoro:
> One side note, we will need to think through how this would look with adding multiple automation.

Agree, I tried to have a toggle to collapse it, but I was not 100% sure it would work, so I left that for the next iteration

server/incident/incident.go
```diff
@@ -38,6 +38,7 @@ type Incident struct {
 	PreviousReminder        time.Duration        `json:"previous_reminder"`
 	BroadcastChannelID      string               `json:"broadcast_channel_id"`
 	ReminderMessageTemplate string               `json:"reminder_message_template"`
+	InvitedUserIDs          []string             `json:"invited_user_ids"`
```
crspeller, line 41: Why do we need to copy the user ids over to the incident? We invite the users on incident creation, can't we just look it up from the playbook?

server/incident/service.go
```diff
@@ -104,6 +104,38 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
 	s.poster.PublishWebsocketEventToChannel(incidentUpdatedWSEvent, incdnt, incdnt.ChannelID)
 	s.telemetry.CreateIncident(incdnt, userID, public)
 
+	usersFailedToInvite := []string{}
+	for _, userID := range incdnt.InvitedUserIDs {
+		_, err = s.pluginAPI.Team.GetMember(incdnt.TeamID, userID)
```
crspeller, line 109: This is checking if they are a member of the team? Can we make this clear that's what this is for?

server/incident/service.go
```diff
@@ -104,6 +104,38 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
...
+			continue
+		}
+	}
+
+	if len(usersFailedToInvite) != 0 {
+		usernames := make([]string, 0, len(usersFailedToInvite))
+		for _, userID := range usersFailedToInvite {
+			user, userErr := s.pluginAPI.User.Get(userID)
+			if userErr != nil {
+				// User may not exist anymore, ignore the error
```
crspeller, line 127: Should we maybe append "a deleted user" to the list? 0/5

server/playbook/playbook.go
```diff
@@ -25,6 +25,8 @@ type Playbook struct {
 	BroadcastChannelID          string      `json:"broadcast_channel_id"`
 	ReminderMessageTemplate     string      `json:"reminder_message_template"`
 	ReminderTimerDefaultSeconds int64       `json:"reminder_timer_default_seconds"`
+	InvitedUserIDs              []string    `json:"invited_user_ids"`
+	InviteUsersEnabled          bool        `json:"invite_users_enabled"`
```
crspeller, line 29: Do we need both a list of user ids and an enabled flag? Isn't the presence of a list of user IDs enough to imply they want to be invited?

server/sqlstore/incident.go
```diff
@@ -18,7 +18,8 @@ import (
 
 type sqlIncident struct {
 	incident.Incident
-	ChecklistsJSON json.RawMessage
+	ChecklistsJSON             json.RawMessage
+	ConcatenatedInvitedUserIDs string
```
crspeller, line 22: Why not JSON like the checklists?

server/sqlstore/playbook.go
```diff
@@ -15,7 +16,8 @@ import (
 
 type sqlPlaybook struct {
 	playbook.Playbook
-	ChecklistsJSON json.RawMessage
+	ChecklistsJSON             json.RawMessage
+	ConcatenatedInvitedUserIDs string
```
crspeller, line 20: Same comment as incident.

webapp/src/components/backstage/automation/invite_users.tsx
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
crspeller, line 1: This client code is very clean and nice! Good job! 😀

crspeller (changes requested):

server/incident/incident.go
cpoile: Depends, would it be more or less surprising to use the list of users as of incident creation?

server/playbook/playbook.go
cpoile: I think it's for the enabled button -- you wouldn't want to lose your list of users when you click that on and off, right?

server/sqlstore/incident.go
cpoile: I don't know, I think since it's just userIds it's simpler to make it a concatenated string. What would making it json give us?

webapp/src/components/backstage/automation/invite_users.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    position: relative;
+    transition: transform .4s;
+    :hover {
+        z-index: 1;
+        transform: translateY(-8px);
+
+        ${Cross} {
+            visibility: visible;
+        }
```
cpoile, line 108: very cool

webapp/src/components/backstage/automation/toggle.tsx
```diff
@@ -0,0 +1,75 @@
...
+        content: "";
+        transition: .4s;
+    }
+
+    input:checked + & {
+        background-color: var(--button-bg);
+    }
+
+    input:checked + &:before {
+        transform: translateX(16px);
```
cpoile, line 64: also cool

cpoile (approved):
Very nice work Alejandro! Non-blocking comments.

Very slick ui too, nicely done. 🎉

server/incident/incident.go
agarciamontoro: The reasoning here was: all data in the automation features need to live in the incident table, except for this first feature, which happens to use the playbook table because it is looked up on incident creation. So instead of having this as an exception, we treat this one in the same way we treat the others. The result is our future brains being happier because the only source of truth for automation features is the incident table., whatever the specific feature is.

server/incident/service.go
agarciamontoro: I'll add a comment here, thanks!

server/incident/service.go
agarciamontoro: Yeah, why not, thanks!

server/playbook/playbook.go
agarciamontoro: Yup! We want to keep the list of users in place even if the feature is disabled (maybe the user is disabling it temporarily, or toggled the button by mistake). So we track the two things separately.

server/sqlstore/incident.go
agarciamontoro: Yeah, I think a concatenated string is enough. Maybe @crspeller is referring to the long name?

server/incident/incident.go
crspeller: But the automation features aren't really related to each other? There is no automation system, just a bunch of one offs. I don't understand how this helps us build other features? I feel like it's more confusing because it raises the obvious question of why. 

If we want to make a pattern of this, why not just make a copy of the playbook that existed at the time of incident creation we can reference?

server/playbook/playbook.go
crspeller: I can't see that being particularly useful. But it is in the design.

server/sqlstore/incident.go
crspeller: Meh, just to make it fit the pattern of using json.RawMessage 1/5

agarciamontoro: Leaving it as it is now, we can change it in the future if needed.

server/incident/incident.go
agarciamontoro: I'm 2.5/5 on leaving it as it is now for a couple of reasons:
- The one stated above (we treat every automation the same way, even if they're a bunch of one offs they are semantically grouped)
- I could see it being useful in the future from the incident itself, as a historic data on how the playbook was configured.

@crspeller, if you're x/5 with x > 2.5, I'll change it! :)

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
```diff
@@ -20,68 +20,70 @@ describe('backstage playbook details', () => {
         cy.url().should('include', '/ad-1/com.mattermost.plugin-incident-management/error?type=playbooks');
     });
 
-    describe('slash command', () => {
```
agarciamontoro, line 23: All the changes from here until line 86 are whitespace changes: for the sake of consistency, I grouped the `slash command` `describe` inside a `tasks` `describe`, so every tab is in its own `describe`.

agarciamontoro:
@prapti, I'd say that most of the behaviour, if not all, is covered by the E2E tests I just pushed, so (when (if?) they pass) it's up to you if you wanna rely on that or give this a more manual test, I'm ok with both!

server/incident/incident.go
crspeller: Not sure about it. Copying doesn't harm anything so I am fine with leaving it as is.

crspeller (approved):

prapti:
On test queue.

prapti (commented):
Great to see the automation feature coming along 🌮  and thank you so much for including the E2E tests with this PR @agarciamontoro!
I tested a few more cases in addition to what you have and there is one scenario that I ran into that I should probably have you look at before approving. The test steps below might make it more clear but the gist is: when a user who is already in the automated invite list, removed from the team, the playbook can no longer be edited and saved, and this error gets thrown: 
`Uncaught (in promise) Error: {"error":"Not authorized"}`

**Test steps:**

1. Create a playbook “PB-1”.
2. In the Automation tab, enable the “Invite members” setting.
3. Add user-A and user-B to the invite list.
4. Save playbook.
5. Go back to the main channel view.
6. Remove user-2 from the team.
    a. Click the main hamburger menu.
    b. Click “Manage Members”.
    c. Locate user-2.
    d. Click the “Member” chevron for user-2.
    e. Click “Remove from Team”.
7. Go to Playbook backstage.
8. Find and open playbook “PB-1”.
9. Make any edit in the playbook (add a new task or add a task description).
10. Click “Save”.
11. Observe that the playbook cannot be saved and the above mentioned error gets thrown.

agarciamontoro:
Thank you, @prapti, I had not thought about that! I added a new test covering your test steps, and fixed the bug by filtering out the invited users that are no longer part of the team, when updating the playbook. So, instead of returning an error if the user is not part of the team, it is removed from the list and a warning is logged.

agarciamontoro:
Waiting for @alexsiclari's input on the user selection UX. Prapti, I'll ask for your review when that's done.

agarciamontoro:
https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/440/commits/a91ceaf98b164f27fa072b949e3a3c0d7515bf70 update the selector for inviting users following @alexsiclari's design:

| Disabled 	| Enabled with no invited members 	| Enabled with invited members 	|
|-	|-	|-	|
| image 	| image 	| image 	|

The new code is simply an adaptation of the `ProfileAutocomplete` component, so I'm not asking for a new dev review. I created [a ticket](https://mattermost.atlassian.net/browse/MM-32861) to merge `InviteUsersSelector` and `ProfileAutocomplete`: they share a lot of functionality and styles, but have some differences that makes the merge not completely trivial here. I'll need some design input on that one, too.

agarciamontoro:
For some reason Github does not let me request a new review from you, @alexsiclari. Please, take a quick look at the screenshots above, thanks!

prapti:
@agarciamontoro Sorry, I have more issues now 😬 
I can't save any new playbook, and the incident backstage view loads blank.
Tried on both local and cloud server.. and both are doing the same.. I can look more into it on Monday, but I get the `Uncaught (in promise) Error: {"error":"An internal error has occurred. Check app server logs for details."}` error.

agarciamontoro:
@prapti, that sounds like migrations conflicts again... Sorry for that, let me merge master and then try to test it from a clean database. The E2E tests check that the playbook can be saved, so that should be covered (assuming the test is correct, of course).

agarciamontoro:
@prapti I locally tested, in a clean database, that playbooks could be saved, apart from the E2E tests passing. Please tell me if you still see those errors, thank you!

prapti (approved):
Thank you @agarciamontoro! Tested and verified that the latest version is working great now! Also added the following tests in the test plan for reference:
- “Automation” tab is highlighted when selected
- “Invite Members” in Automation is disabled by default
- Members cannot be added to invite list when the “Invite members” setting is disabled
- Members can be added to invite list when the “Invite members” setting is enabled
- Non-team members cannot be added to the invite list
- Team member removed from the team cannot be added to the invite list
- With “Invite members” disabled, no user is invited to a new incident channel when there is no member in the playbook automation invite list
- With “Invite members” enabled, no user is invited to a new incident channel when there is no member in the playbook automation invite list
- With “Invite members” enabled, users in the invite list are invited to the incident channel
- With “Invite members” disabled, no user is invited to a new incident channel when there are members in the playbook automation invite list
- When a user who is already in the invite list removed from the team, the user is also removed from the invite list
- Invited members can be successfully removed from the invite list
- Invited members cannot be removed front he invite list when “Invite members” is disabled

agarciamontoro:
Thank you so much, @prapti!

PR #441, opened 2021-01-26
Add telemetry on reminder and broadcast features

agarciamontoro:
#### Summary
Add missing information in the telemetry data.

@Adovenmuehle, one of the fields I just added to the incident structure (`ReminderTimerSeconds`) is only present in a specific event, as it does not make much sense to send it with the others that relate to incidents. Is that ok for your data processing, or do you guys need us to always add the same fields for the events that refer to the same data type?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32051

### DIFF
--- a/server/telemetry/rudder.go
+++ b/server/telemetry/rudder.go
@@ -147,6 +147,7 @@ func (t *RudderTelemetry) ChangeCommander(incdnt *incident.Incident, userID stri
 func (t *RudderTelemetry) UpdateStatus(incdnt *incident.Incident, userID string) {
 	properties := incidentProperties(incdnt, userID)
 	properties["Action"] = actionUpdateStatus
+	properties["ReminderTimerSeconds"] = int(incdnt.PreviousReminder)
 	t.track(eventIncident, properties)
 }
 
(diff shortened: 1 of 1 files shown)


server/telemetry/rudder.go
```diff
@@ -228,14 +229,17 @@ func playbookProperties(pbook playbook.Playbook, userID string) map[string]inter
...
+		"UserActualID":                userID,
+		"PlaybookID":                  pbook.ID,
+		"TeamID":                      pbook.TeamID,
+		"NumChecklists":               len(pbook.Checklists),
+		"TotalChecklistItems":         totalChecklistItems,
+		"IsPublic":                    pbook.CreatePublicIncident,
+		"NumMembers":                  len(pbook.MemberIDs),
+		"NumSlashCommands":            totalChecklistItemsWithCommands,
+		"ReminderTimerDefaultSeconds": pbook.ReminderTimerDefaultSeconds,
+		"BroadcastChannelID":          pbook.BroadcastChannelID,
```
cpoile, line 241: Don't know if it's a question we need to wait for @itao to answer, but do we really need to record things like the `BroadcastChannelID`? Wouldn't something like "UsesBroadcastChannel" be both more useful and less intrusive?

cpoile (approved):

server/telemetry/rudder.go
lieut-data: I remember raising something /like/ this with @Adovenmuehle (or @enelson720), and the consensus was to use explicit IDs whenever possible. It might be overkill in this case, but it's also pretty interesting to see if all playbooks get announced in a single channel or multiple channels to drive changes in this part of the product.

lieut-data (approved):

lieut-data:
@agarciamontoro, can you plan to update the documentation to reflect the new telemetry additions?

Adovenmuehle:
@agarciamontoro it's fine to have fields that are specific to only one event. 👍

server/telemetry/rudder.go
Adovenmuehle: Yeah it helps to have the actual IDs so that deeper analysis can be done with the rest of telemetry that we receive.

agarciamontoro: I agree with @cpoile here. Would that be useful in this specific case? What's the use case for having that information tracked?

agarciamontoro:
> @agarciamontoro, can you plan to update the documentation to reflect the new telemetry additions?

@lieut-data, heads-up that I created https://mattermost.atlassian.net/browse/MM-32353 for this sprint.

PR #442, opened 2021-01-28
MM-32357 Telemetry for timeline tab

cpoile:
#### Summary
- For v1 of Timelines, Jason wanted to have the ability to check whether or not people were using it.
- Added a `frontend` event (open to suggestions on naming) for all telemetry coming from the frontend. Added `timeline_tab_clicked` to record this one (also open to suggestions).
- Why not handle this on the frontend? I did at first, but I found I was duplicating the boilerplate info we want to have everywhere, and I wanted to be absolutely sure we were using the identical identifier ids, field names, etc. as the backend was sending, so I think it's easier if we do it this way. Less error prone.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32357

### DIFF
--- a/server/api/incidents.go
+++ b/server/api/incidents.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -28,17 +29,19 @@ type IncidentHandler struct {
 	pluginAPI       *pluginapi.Client
 	poster          bot.Poster
 	log             bot.Logger
+	telemetry       incident.Telemetry
 }
 
 // NewIncidentHandler Creates a new Plugin API handler.
 func NewIncidentHandler(router *mux.Router, incidentService incident.Service, playbookService playbook.Service,
-	api *pluginapi.Client, poster bot.Poster, log bot.Logger) *IncidentHandler {
+	api *pluginapi.Client, poster bot.Poster, log bot.Logger, telemetry incident.Telemetry) *IncidentHandler {
 	handler := &IncidentHandler{
 		incidentService: incidentService,
 		playbookService: playbookService,
 		pluginAPI:       api,
 		poster:          poster,
 		log:             log,
+		telemetry:       telemetry,
 	}
 
 	incidentsRouter := router.PathPrefix("/incidents").Subrouter()
@@ -78,6 +81,10 @@ func NewIncidentHandler(router *mux.Router, incidentService incident.Service, pl
 	checklistItem.HandleFunc("/assignee", handler.itemSetAssignee).Methods(http.MethodPut)
 	checklistItem.HandleFunc("/run", handler.itemRun).Methods(http.MethodPost)
 
+	telemetryRouterAuthorized := router.PathPrefix("/telemetry").Subrouter()
+	telemetryRouterAuthorized.Use(handler.checkViewPermissions)
+	telemetryRouterAuthorized.HandleFunc("/incident/{id:[A-Za-z0-9]+}", handler.telemetryForIncident).Methods(http.MethodPost)
+
 	return handler
 }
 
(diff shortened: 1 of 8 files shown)


webapp/src/components/rhs/rhs_tab_view.tsx
```diff
@@ -65,7 +68,10 @@ const RHSTabView = () => {
             </TabItem>
             <TabItem
                 active={currentTabState === RHSTabState.ViewingTimeline}
-                onClick={() => setTabState(RHSTabState.ViewingTimeline)}
+                onClick={() => {
+                    setTabState(RHSTabState.ViewingTimeline);
+                    telemetryEventForIncident(incident.id, 'timeline_tab_clicked');
```
lieut-data, line 73: Should we add one for tasks too?

webapp/src/client.ts
```diff
@@ -244,6 +244,12 @@ export async function clientReorderChecklist(incidentID: string, checklistNum: n
     return data;
 }
 
+export async function telemetryEventForIncident(incidentID: string, action: string, props?: Record<string, string>) {
+    const params = qs.stringify({action}, {addQueryPrefix: true});
```
lieut-data, line 248: I'm not a /huge/ fan of the mix of semantic URL, query string parameters, and an (unused?) body for more parameters. If we're going all-in on a body, can we use that exclusively?

server/telemetry/rudder.go
```diff
@@ -150,6 +152,15 @@ func (t *RudderTelemetry) UpdateStatus(incdnt *incident.Incident, userID string)
 	t.track(eventIncident, properties)
 }
 
+func (t *RudderTelemetry) FrontendTelemetryForIncident(incdnt *incident.Incident, userID, action string, props map[string]string) {
```
lieut-data, line 155: This feels a bit coarse. At the risk of eventually exploding the number of methods here, would it make sense to have a `IncidentTimelineOpened` event instead?

server/api/incidents.go
```diff
@@ -876,6 +901,44 @@ func (h *IncidentHandler) reorderChecklist(w http.ResponseWriter, r *http.Reques
...
+		return
+	}
+
+	incdnt, err := h.incidentService.GetIncident(id)
+	if err != nil {
+		HandleError(w, err)
+		return
+	}
+
+	var props map[string]string
```
lieut-data, line 923: As near as I can tell, nothing in this PR actually uses the `props`. Should we consider removing support for this in favour of more explicit types / parameters when we need them in the future?

lieut-data (changes requested):
Nice addition! A few thoughts below.

webapp/src/components/rhs/rhs_tab_view.tsx
cpoile: Maybe? But I don't think we need it -- all the task-related actions are already tracked. Do we want to know if the tasks tab was clicked, in the same way we want to know if the timeline tab was clicked?

lieut-data: Ah, good point. Yeah, I guess until there are discrete actions within timeline, just opening it suffices.

webapp/src/client.ts
cpoile: Sorry, I agree it's not pretty, so I should have added a bit of my reasoning: The action is absolutely required, and it's the 'verb' that the event is tracking. And it's in the context of an incident, so it reads like: "incident does this action".

Whereas, the props are truly optional, and we will probably need them in the future (like we do in the existing telemetry). We /could/ put it in the props, but the action really isn't a prop. That's why I chose this way.

I'm only 2/5 on this design (just like every other rest api design, right?). So if you're stronger on it we can change it.

server/telemetry/rudder.go
cpoile: That wouldn't fit with how we're doing the telemetry right now. We have big "topic" level events like `incident` `tasks` and `playbook`. It's the `action` that narrows it down to the actual thing we're tracking.

server/api/incidents.go
cpoile: The way we send over the telemetry it basically is a `props` bag that we add to when we need a bit more context. So adding types or parameters to this endpoint might just add more work than benefit? But maybe I'm not understanding what you mean?

webapp/src/client.ts
lieut-data: I'm 0/5 as well, and I consider this an internal API so it wouldn't be subject to any breaking changes rule, but I wonder if we should just encode the body payload as:
```
{
    "action": action,
    "props": props,
}
```

server/api/incidents.go
lieut-data: I mean that `telemetryEventForIncident` is only ever called with an empty `props` bag, making it feel like we're designing for a future use that might change.

server/api/incidents.go
```diff
@@ -99,6 +106,24 @@ func (h *IncidentHandler) checkEditPermissions(next http.Handler) http.Handler {
 	})
 }
 
+func (h *IncidentHandler) checkViewPermissions(next http.Handler) http.Handler {
```
agarciamontoro, line 109: Nice!

server/api/incidents.go
agarciamontoro: For what is worth, I had the exact same thought, I was trying to see where it was being used.

webapp/src/client.ts
agarciamontoro: I think I agree with Jesse here, because we're not using query parameters elsewhere, other than for filtering in the collections endpoints (correct me if I'm wrong!). It feels a bit weird to have the action as a query parameter. We could even do something like `/telemetry/incident/{id}/{action}`, which would also fulfill your argument on "the action is absolutely required".

agarciamontoro (approved):
Nice job! No blocking comments from my side, but I added my two cents to the discussions opened by Jesse.

webapp/src/client.ts
cpoile: Okay, makes sense -- changed to sending it all in the body, and removed the props until we need them. Thanks!

server/api/incidents.go
cpoile: Sounds good to me. :)

lieut-data (approved):

PR #446, opened 2021-02-01
rip out unused source plugin id checking

lieut-data:
#### Summary
In doing some rework in support of the Incidents Service client API, I found what appears to be legacy code. Removing it simplifies things for the cleanup, but I'd like to do it out of band for clarity.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32495


crspeller (approved):
That code was for handling plugin to plugin communication. Copy paste from other plugins that where using it. If we do use that in the future we can just add the ID to the context in `server/plugin.go` so removing it is fine.

server/plugin.go
```diff
@@ -43,9 +43,9 @@ type Plugin struct {
 	bot                 *bot.Bot
 }
 
-// ServeHTTP demonstrates a plugin that handles HTTP requests by greeting the world.
+// ServeHTTP routes incoming HTTP requests to the plugin's REST API.
```
agarciamontoro, line 46: Good catch :joy:

agarciamontoro (approved):
Yay!

lieut-data:
@prapti, I can't (yet) assign you to this pull request (due to the accidental org reset), but heads up that this one should just requires a quick smoketest. (Something I think the e2e tests actually already do well in this case.)

lieut-data:
This PR will require merging with upstream before merging.

lieut-data:
@prapti, for testing purposes, there really isn't anything "special" here that shouldn't already be covered by the unit tests. Your call on the coverage :)

prapti (approved):
Ran a quick smoke test and everything is passing as expected.

PR #451, opened 2021-02-03
MM-32516: selectively align profile selector to right

lieut-data:
#### Summary
To facilitate using the profile selector with the new "assignee" button, a `right: 0` property was unconditionally added. Unfortunately, this shifted all the profile selectors as such, causing problems for the commander dropdown when the RHS shrinks in size for smaller windows:

image

image

image

As a workaround, pass in a new `showOnRight` prop to conditionally set
this field:

image

image

image

image

Ideally, we'll revisit this component in the future.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-32516

### DIFF
--- a/webapp/src/components/checklist_item.tsx
+++ b/webapp/src/components/checklist_item.tsx
@@ -311,6 +311,7 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
                                 onCustomReset: resetAssignee,
                             }}
                             controlledOpenToggle={profileSelectorToggle}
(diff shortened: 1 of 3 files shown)


cpoile (approved):
This feels... ugly. But I guess we need it.

crspeller (approved):
Needs more CSS magic.

prapti (approved):
The commander dropdown no longer renders behind the main channel. Tested and verified in all three major browsers and desktop app.
Approving.

PR #452, opened 2021-02-03
Update plugin-api and mattermost-sever hashes

cpoile:
#### Summary
- Because the ScheduleOnce job scheduler was merged to master, we can't depend on a PR branch that doesn't exist anymore

#### Ticket Link
- none


lieut-data (approved):

agarciamontoro (approved):

lieut-data:
@cpoile, given the various dependency bumps, should we still QA Review this one for confidence?

cpoile:
@lieut-data I think the e2e tests should be good for that. But if you feel we need it, I don't mind it of course (just want to be efficient with Prapti's time)

cpoile:
@prapti I think this just needs a very brief "are interactions that require the plugin-api working" kind of test. Start a few incidents, make a few reminders, broadcast channel working. If those work, then the dependency updates didn't break functionality.  The e2e tests should also be a good proxy for most of this. (except the reminder functionality)

prapti (approved):
Thank you for the test specs @cpoile, that helped! Verified that incident creation, playbook creation, broadcast and reminder work. Most E2E tests passed but verified that the ones failing passed when tested manually (looking into the failures).
Approving.

PR #457, opened 2021-02-05
added newid if pure multibyte character set

arjunagl:
#### Summary
If the incident name has only multibyte characters its does not create the channel, this change will create a new id and it will allow channel creation.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/16834#event-4282247227


lieut-data (changes requested):
This looks great, @arjunagl! Can you help us test this my updating the unit tests with a case that covers this path? 

If you're feeling really adventurous, you can even try your hand at an e2e test that programmatically tests this from the frontend as well.

arjunagl:
thanks @lieut-data, I have included an unit test as well into the .go code.

server/incident/service_test.go
```diff
@@ -187,6 +187,49 @@ func TestCreateIncident(t *testing.T) {
...
+		}
+
+		store.EXPECT().CreateIncident(gomock.Any()).Return(incdnt, nil)
+		store.EXPECT().CreateTimelineEvent(gomock.AssignableToTypeOf(&incident.TimelineEvent{}))
+		mattermostConfig := &model.Config{}
+		mattermostConfig.SetDefaults()
+		pluginAPI.On("GetConfig").Return(mattermostConfig)
+		pluginAPI.On("CreateChannel", mock.MatchedBy(func(channel *model.Channel) bool {
+			return channel.Name != ""
+		})).Return(&model.Channel{Id: "channel_id"}, nil)
```
agarciamontoro, line 216: I'm a bit concerned about this part because this is exactly what we want to test, and mocking it kind of bypasses that. I see no other way of testing it, though, and at least we check that the function is called with the correct parameters (a channel with a non-empty name).

agarciamontoro (approved):
Thank you for your contribution,@arjunagl :tada:

The code looks good to me! I'd love to see an E2E test (I explained my specific concerns on the mocking below), but I'm approving it because it does solve the bug.

Thank you!

server/incident/service_test.go
lieut-data: Maybe if we assert `model.IsValidId` here to verify?

lieut-data (approved):
I think this makes sense -- thanks again, @arjunagl!

@prapti, this one could use an e2e test for full coverage. Is this something you might be able to take a stab at helping write for added coverage? (No immediate hurry :))

server/incident/service_test.go
agarciamontoro: That's a good idea!

prapti (approved):
Thank you for your contribution on this @arjunagl! I have tested and verified that your solution works great!
Approving to merge 🎉

lieut-data:
Just a heads up that this change may find an edge case in the bug filed over at https://mattermost.atlassian.net/browse/MM-32912, but it's something for the Web Platform team to investigate.

PR #458, opened 2021-02-05
log a warning only when failing to promote commander to channel admin

lieut-data:
#### Summary
While the automatic promotion to channel admin doesn't work with managed team schemes, it does still work in the vanilla case. Keep the "base" functionality for customers not using advanced permissions by just logging a warning when this promotion fails.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-25755


crspeller (approved):

agarciamontoro (approved):
:tada:

prapti (approved):
Tested and verified in local and cloud servers that the reporter of a new incident is also the channel admin by default (however promoting a different channel member to commander does not promote them to a channel admin).

PR #459, opened 2021-02-05
MM-32545: Timeline display all events; clean up design/wording

cpoile:
#### Summary
- All built-in events are being tracked, filtering out the three events we don't want to see right now (but keeping the code for when we add them back, but open to deleting them completely)
- Worked with @itao for which events to show, how to word them, and the way to display the time for each event
- Added scrollbars

Looks like: 
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32545


agarciamontoro (approved):
Nice, I'm loving the shape the timeline is taking, it's so cool! I left a non-blocking comment below. About the events being filtered out, I'm ok with having that code in place, I can see it will be useful when continuing the work on the post-mortem features.

webapp/src/types/incident.ts
```diff
@@ -93,6 +98,21 @@ export function isMetadata(arg: any): arg is Metadata {
         typeof arg.total_posts === 'number');
 }
 
+export function isTimelineEvent(arg: any): arg is TimelineEvent {
+    return Boolean(arg &&
+        typeof arg.id === 'string' &&
+        typeof arg.id === 'string' &&
```
lieut-data, line 104: duplicate test?

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -46,70 +53,80 @@ interface Props {
...
+            autoHideDuration={500}
+            renderThumbHorizontal={renderThumbHorizontal}
+            renderThumbVertical={renderThumbVertical}
+            renderView={renderView}
+            style={{position: 'absolute'}}
+        >
+            <Timeline data-testid='timeline-view'>
+                {
+                    events.map((event) => {
+                        return (
```
lieut-data, line 117: nit: can omit `return` with just `events.map((event) => (<RHSTimelineEventItem ... />))`.

server/incident/service.go
```diff
@@ -102,7 +102,6 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
 		return nil, errors.Wrapf(err, "failed to create incident")
 	}
 
-	s.poster.PublishWebsocketEventToChannel(incidentUpdatedWSEvent, incdnt, incdnt.ChannelID)
```
lieut-data, line 105: Why do we no longer broadcast this event?

lieut-data (approved):
Looks good! A few comments, but probably nothing blocking.

webapp/src/components/rhs/rhs_timeline_event_item.tsx
```diff
@@ -95,31 +106,66 @@ const RHSTimelineEventItem = (props: Props) => {
...
         break;
     case TimelineEventType.StatusUpdated:
         iconClass = 'icon icon-flag-outline';
-        summaryTitle = 'Incident Status Update';
-        summaryDetail = <SummaryDetail>{props.event.display_name + ' updated incident to ' + props.event.status}</SummaryDetail>;
+        if (props.event.summary === '') {
+            summaryTitle = props.event.subject_display_name + ' posted a status update';
+        } else {
+            summaryTitle = props.event.subject_display_name + ' changed status from ' + props.event.summary;
+        }
```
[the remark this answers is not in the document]
cpoile: It's a good suggestion, I was thinking of something like that myself. Do you mind if we add it later, since I need to think it over a bit?

server/incident/service.go
cpoile: It was never being used on the frontend. It's bugged me that we send an updated event, but the frontend only cares about the created event that is sent in the api code.

webapp/src/components/rhs/rhs_timeline_event_item.tsx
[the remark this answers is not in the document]
agarciamontoro: Sure, it can p[erfectly wait

cpoile:
To be QA reviewed as part of #432

PR #460, opened 2021-02-09
MM-32408 Moving some stuff to use styled components.

crspeller:
#### Summary
This might fix the issue Prapti was seeing in the ticket. But I am unable to reproduce myself. Good to have ether way. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32408


cpoile (approved):

cpoile:
This is weird. This PR is failing the e2e tests because the `incident_rhs_list_spec.js` file is different from master, but that difference isn't showing up in the diff. See the diff from this PR (this is the line breaking the e2e tests):
https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/57dd9c702ec1781775298e135b4e349fb597a585/tests-e2e/cypress/integration/frontstage/incident_rhs_list_spec.js#L204

and master: https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/bc3259b14962e51a953ca46c490f6b7f59a048fa/tests-e2e/cypress/integration/frontstage/incident_rhs_list_spec.js#L208

crspeller:
@cpoile Maybe a merge from master will fix it?

cpoile:
Yah, probably -- I was assuming you had branched off master.

webapp/src/components/checklist_item.tsx
```diff
@@ -280,8 +455,8 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
             onMouseLeave={() => setShowMenu(false)}
             data-testid='checkbox-item-container'
         >
-            <div
-                className={'checkbox-container live'}
+            <CheckboxContainer
+                className={'live'}
```
agarciamontoro, line 459: Nit: could we get rid of this class by doing something like
```js
const CheckboxContainerLive = styled(CheckboxContainer)`
    height: 35px;
`;
```

Just for being consistent using styled-components.

agarciamontoro (approved):
It looks great! I left a non-blocking suggestion, feel free to dismiss it.

webapp/src/components/checklist_item.tsx
crspeller: Done!

crspeller:
@prapti Looking to see if this fixes the ticket mentioned. Even if it doesn't, we can merge this as code cleanup and try again. Unless of course it makes things worse somehow.

prapti (approved):
@crspeller It's definitely not making anything worse, but I can still see the weirdness (and this time in Safari as well). Clearing cache however made things look normal for a little bit -- and then back to the big labels again. But approving since you've noted that it's a good thing to have anyway, even though it doesn't necessarily fix the original issue :D
Screen Shot 2021-02-10 at 5 47 15 PM
Screen Shot 2021-02-10 at 5 47 54 PM

PR #461, opened 2021-02-10
MM-32615 - Add to Timeline from post dropdown menu

cpoile:
#### Summary
- Can add a post to an incident using the post's dropdown menu

#### Q's for @justinegeffen / @itao :
- What do you want the post dropdown to read? As of now it looks like:
image

- What do you want the interactive dialog to read? Currently looks like:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32615


server/incident/service.go
```diff
@@ -1079,6 +1155,46 @@ func (s *ServiceImpl) newUpdateIncidentDialog(message, broadcastChannelID, statu
...
+		Title: "Add to Incident Timeline",
+		Elements: []model.DialogElement{
+			{
+				DisplayName: "Incident",
+				Name:        DialogFieldIncidentKey,
+				Type:        "select",
+				Options:     options,
+			},
+			{
+				DisplayName: "Summary",
```
lieut-data, line 1184: Should we pre-populate with something from the post itself?

server/sqlstore/incident.go
```diff
@@ -99,6 +99,11 @@ func (s *incidentStore) GetIncidents(requesterInfo incident.RequesterInfo, optio
 		queryForTotal = queryForTotal.Where(sq.Eq{"i.CurrentStatus": options.Status})
 	}
 
+	if len(options.Statuses) != 0 {
```
lieut-data, line 102: Should we fail if `Status` and `Statuses` are both non-empty?

justinegeffen:
Thanks, @cpoile!

I propose referring to timeline in lowercase for consistency with other aspects of incident collaboration (e.g., playbooks, incidents). I'm 5/5 on this as well as naming any other new features going forward.

In this vein, I suggest "Add to incident timeline" for the drop-down/context menu. 

The interactive dialog can remain as-is, because the H1 capitalization is correct, as is the CTA. And "timeline" is already lowercase so nothing needs to change there.

server/telemetry/rudder.go
```diff
@@ -159,6 +160,13 @@ func (t *RudderTelemetry) FrontendTelemetryForIncident(incdnt *incident.Incident
 	t.track(eventFrontend, properties)
 }
 
+// AddEventFromPost tracks userID creating a timeline event from a post.
```
lieut-data, line 163:
```suggestion
// AddPostToTimeline tracks userID creating a timeline event from a post. 
```

lieut-data (approved):

server/incident/service.go
cpoile: I can't think of anything that would be foolproof -- we don't want to do any kind of search and replace, regex, or parsing, so I just left it blank. Open to suggestions if you can think of anything that won't have edge cases?

server/sqlstore/incident.go
cpoile: Sure, that makes sense--done. Note: I'm not adding this validation onto the API side, since we're not adding `statuses` as an API parameter (until we need to do so). I want to keep this confined to the backend. Sound good?

justinegeffen (approved):

justinegeffen:
Thanks for making that change, @cpoile. Looks good to me. :)

crspeller (approved):

server/incident/service.go
lieut-data: My only thought was to take the first 30 characters and truncate -- it feels like it's useful most of the time, even if occasionally wrong (and broken, if halfway through some markdown).

lieut-data:
@cpoile, did you want @itao to block review here, or should we let @prapti take a look now?

itao (approved):

server/incident/service.go
cpoile: Yep, that sounds good to me. I added it for up to 40 chars from the original post. It adds an ellipsis if the original text is longer than the summary. The placeholder shows if the post had no text.

cpoile:
Added default summary (as demoed in the R&D meeting), and adding the default incident if the post is from a non-archived incident channel.

@lieut-data Asking for a re-review because there's a few pieces of new code.

lieut-data (approved):

prapti (approved):
Tested and verified that events can be added to an incident timeline from the post menu.
For reference, added the following tests in the test plan:
1. Events can be added to timeline from a post menu in main channel
2. Events can be added to timeline from post menu by a system admin user
3. Events can be added to timeline from post menu by a normal user
4. Events can be added to timeline from post menu by guest user (this will need to get verified later since Christopher has added some guest user permission improvements)
5. Events can be added to timeline from a post menu reply in RHS
6. Events added to timeline from post menu appear in chronological order of the time added
7. Events already added can be added again from post menu
8. Events added to timeline link to the associated message in main channel
9. Events added from post menu show ‘the pencil’ icon in timeline
10. Events can be added from a private incident to another private incident timeline
11. Events can be added from a public incident to another public incident timeline
12. Events can be added from a private incident to a public incident timeline
13. Events can be added from a public incident to a private incident timeline
14. Long post messages added to timeline get ellipsised to fit the RHS width
15. “Summary” field in Add to Incident Timeline modal can be edited to up to 64 chars

@cpoile Just one clarifying question -- can the same post be added again to timeline? It looks like the following right now, but wanted to check that that's fine for now:
Screen Shot 2021-02-22 at 6 36 09 PM

cpoile:
@prapti Yep, the same post can be added to the timeline -- I didn't think putting logic or restrictions on it would add any value (if they want to add a post twice, go for it). Thanks for the very thorough test plan, it looks amazing!

PR #466, opened 2021-02-12
Use filterOption in ProfileAutocomplete

agarciamontoro:
#### Summary
React-select has a prop `filterOption` for doing what we were incorrectly doing when grabbing the data from the server. This PR simply moves the filtering logic where it should be.

I found this while working on the new user selector in https://github.com/mattermost/mattermost-plugin-incident-management/pull/440, but I did not want to add more noise there. Hence, this independent PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32313

### DIFF
--- a/webapp/src/components/backstage/profile_autocomplete.tsx
+++ b/webapp/src/components/backstage/profile_autocomplete.tsx
@@ -105,8 +105,7 @@ const ProfileAutocomplete: FC<Props> = (props: Props) => {
 
         //@ts-ignore
         profiles.then(({data}) => {
-            const profilesWithoutAlreadyAdded = data.filter((profile: UserProfile) => !props.userIds.includes(profile.id));
-            callback(profilesWithoutAlreadyAdded);
+            callback(data);
         }).catch(() => {
             // eslint-disable-next-line no-console
             console.error('Error searching user profiles in custom attribute settings dropdown.');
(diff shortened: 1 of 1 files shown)


cpoile (approved):
cool!

crspeller (approved):

agarciamontoro:
@prapti, testing this should come down to making sure that the user selector in the right hand side of the playbook backstage still works, and that the users selected are removed from the list of users to select.

prapti (commented):
@agarciamontoro Verified that the user selector in the RHS of the playbook backstage still works and the users selected are removed from the list but only after the playbook is saved. I saw this in the earlier plugin versions too so opened https://mattermost.atlassian.net/browse/MM-33117 thinking that could be addressed separately. But now thinking if that's what you wanted me to test. If so, I can close that ticket and wait for the fix in this PR, otherwise, I can approve this and let that bug be handled separately. Please let me know.

agarciamontoro:
Argh, I guess I only tested this when filtering, so i

agarciamontoro:
With respect to the ticket you opened, I think it's a duplicate of the ticket this PR was trying to solve: https://mattermost.atlassian.net/browse/MM-32313, although much better explained. I'll close https://mattermost.atlassian.net/browse/MM-32313 with this PR and we can track yours (doing https://mattermost.atlassian.net/browse/MM-32861 will fix it)

agarciamontoro:
I just noticed I had not requested re-review. This PR can wait, but just so you have it on your radar.

prapti (approved):
Thank you @agarciamontoro!
I quickly re-reviewed this and is looking good, thus approving keeping in the mind that there will be more improvements coming soon :)
Thanks again!

PR #467, opened 2021-02-12
MM-32452 Allow guest users access to their own incidents.

crspeller:
#### Summary
Guests will now have full access to the incidents they are a member of. They will now not have access to playbooks or the ability to create incidents. 

This is functionality side only. The UI still makes it seem like they can do some of this, but will error (with a message)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32452

### DIFF
--- a/server/api/incidents.go
+++ b/server/api/incidents.go
@@ -334,6 +334,10 @@ func (h *IncidentHandler) createIncident(newIncident incident.Incident, userID s
 	return h.incidentService.CreateIncident(&newIncident, userID, public)
 }
 
+func (h *IncidentHandler) getRequesterInfo(userID string) (incident.RequesterInfo, error) {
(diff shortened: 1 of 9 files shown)


lieut-data (approved):
Looks good -- heads up, @prapti, that this might be a PR to test against the advanced permissions stuff and experiment with a few permutations.

cpoile (approved):
Nice and clean. Thanks!

prapti (approved):
Thank you for the detailed note on guests not having access to playbook or to create incidents. With that in mind, the rest works great. Tested scenarios in local and cloud servers:
Guest user has access to incidents when
    - Guest user is added to a public incident
    - Guest user is added to a private incident
    - A normal user is added to a public incident and then converted to a guest user
    - A normal user is added to a private incident and then converted to a guest user

PR #468, opened 2021-02-12
MM-32547 - Remove any event from Timeline

cpoile:
#### Summary
- Can remove any event from the timeline
- Looks like:
image
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32547

### DIFF
--- a/server/api/incidents.go
+++ b/server/api/incidents.go
@@ -64,6 +64,7 @@ func NewIncidentHandler(router *mux.Router, incidentService incident.Service, pl
 	incidentRouterAuthorized.HandleFunc("/update-status-dialog", handler.updateStatusDialog).Methods(http.MethodPost)
 	incidentRouterAuthorized.HandleFunc("/reminder/button-update", handler.reminderButtonUpdate).Methods(http.MethodPost)
 	incidentRouterAuthorized.HandleFunc("/reminder/button-dismiss", handler.reminderButtonDismiss).Methods(http.MethodPost)
+	incidentRouterAuthorized.HandleFunc("/timeline/{eventID:[A-Za-z0-9]+}", handler.removeTimelineEvent).Methods(http.MethodDelete)
 
 	channelRouter := incidentsRouter.PathPrefix("/channel").Subrouter()
 	channelRouter.HandleFunc("/{channel_id:[A-Za-z0-9]+}", handler.getIncidentByChannel).Methods(http.MethodGet)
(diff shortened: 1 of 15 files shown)


webapp/src/components/rhs/rhs_timeline_event_item.tsx
```diff
@@ -75,6 +77,24 @@ const SummaryDetail = styled.div`
     color: var(--center-channel-color-64)
 `;
 
+const HoverMenu = styled.div`
```
crspeller, line 80: Should we share this? So when we get a new design we don't have to change it in multiple places?

crspeller (approved):
Nice! One non-blocking

webapp/src/components/rhs/rhs_timeline_event_item.tsx
cpoile: I thought about it, but I wasn't sure if we wanted to make changes specific to just this (if I'm asked to from UI)... But, yah, let's worry about that if/when it happens. Will moved it to shared.

cpoile: Done, thanks!

server/api/incidents.go
```diff
@@ -714,6 +715,21 @@ func (h *IncidentHandler) reminderButtonDismiss(w http.ResponseWriter, r *http.R
 	ReturnJSON(w, nil, http.StatusOK)
 }
 
+// removeTimelineEvent handles the DELETE /incidents/{id}/timeline/{eventID} endpoint.
+// User has been authenticated to edit the incident.
+func (h *IncidentHandler) removeTimelineEvent(w http.ResponseWriter, r *http.Request) {
```
agarciamontoro, line 720: We should probably update the api.yaml with this one

agarciamontoro (changes requested):
Nice! :tada: The only thing I'm missing is the update of the API docs.

server/api/incidents.go
cpoile: Good call, added. Thanks!

agarciamontoro (approved):
Great, thanks!

prapti (approved):
Tested and verified that removing the incident from timeline works as expected. LGTM!
Added the following tests:
1. Hovering over timeline event shows a delete hover button
2. Clicking on delete hover button shows a “Confirm Entry Delete” confirmation dialog
3. Clicking “Cancel” in “Confirm Entry Delete” confirmation dialog does not delete the timeline event
4. Clicking “Delete Entry” in “Confirm Entry Delete” confirmation dialog deletes the timeline event
5. Timeline event can be deleted by any incident channel member
6. A guest user can delete a timeline event (this will need to be tested later)

PR #469, opened 2021-02-12
MM-32387: activate but functionally disable without license

lieut-data:
#### Summary
Allow incident collaboration to start without a valid E20 license, but disable all user facing functionality (including the REST API and slash commands) until a license is installed. Dynamically transition across any license changes.

To make this work, I had to pull in the store functionality from the pluginapi and remove the built-in license check there (since I wanted to start up the store without a valid license in place). We should consider revisiting this artificial limitation, but for now I've at least put that in a separate commit.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32387


go.mod
```diff
@@ -5,12 +5,15 @@ go 1.14
 require (
 	github.com/Masterminds/squirrel v1.4.0
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/go-sql-driver/mysql v1.5.0
```
lieut-data, line 8: These are being pulled in as a result of incorporating the pluginapi's store support.

plugin.json
```diff
@@ -18,7 +18,7 @@
         "bundle_path": "webapp/dist/main.js"
     },
     "settings_schema": {
-        "header": "",
+        "header": "Incident Collaboration requires a Mattermost Cloud or Mattermost E20 License.",
```
lieut-data, line 21: Just in case someone is confused in the system console.

server/sqlstore/database.go
```diff
@@ -0,0 +1,135 @@
+package sqlstore
```
lieut-data, line 1: This is a lightly modified version pulled from the pluginapi.

server/sqlstore/unexport_test.go
```diff
@@ -0,0 +1,3 @@
+package sqlstore
+
+var NewDatabase = newDatabase
```
lieut-data, line 3: Let me use `NewDatabase` from `sqlstore_test` without actually exporting it (`_test.go` files are unconditionally ignored.)

webapp/src/index.tsx
```diff
@@ -41,17 +41,21 @@ import {
...
 } from './types/websocket_events';
+import RegistryWrapper from './registry_wrapper';
+import {isE20LicensedOrDevelopment} from './license';
 
 export default class Plugin {
     public initialize(registry: PluginRegistry, store: Store<GlobalState>): void {
         registry.registerReducer(reducer);
 
         let mainMenuActionId: string | null;
         const updateMainMenuAction = () => {
```
lieut-data, line 52: Kept this section discrete since it has special handling.

lieut-data:
Lastly, I discussed this briefly with @prapti, but I don't know how to e2e test this given the current licensing architecture -- we don't seem to have any tests that remove/add as such.

server/api/incidents_test.go
```diff
@@ -47,8 +50,62 @@ func TestIncidents(t *testing.T) {
...
+			CommanderUserID: "testUserID",
+			TeamID:          "testTeamID",
+			Name:            "incidentName",
+			PlaybookID:      withid.ID,
+			Checklists:      withid.Checklists,
+		}
+
+		retI := testIncident
+		retI.ID = "incidentID"
+		retI.ChannelID = "channelID"
```
cpoile, line 93: Not sure if we need these, right?

server/sqlstore/database.go
cpoile: Not reviewing it too closely then. :)

webapp/src/registry_wrapper.tsx
```diff
@@ -0,0 +1,87 @@
+import {Store} from 'redux';
+
+//@ts-ignore Webapp imports don't work properly
+import {PluginRegistry} from 'mattermost-webapp/plugins/registry';
+import {GlobalState} from 'mattermost-redux/types/store';
+
+// RegistryWrapper intercepts component and event handling registrations with the WebApp
```
cpoile, line 7: This is awesome.

cpoile (approved):

server/api/incidents_test.go
lieut-data: Good catch! Fixed :)

crspeller (changes requested):

server/command/command.go
```diff
@@ -1172,6 +1175,11 @@ func (r *Runner) Execute() error {
 		return err
 	}
 
+	if !r.configService.IsLicensed() {
```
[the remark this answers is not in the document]
lieut-data: It's not my ideal experience, but it turns out there is no event telling plugins when the license changes, so there's no simply way to make this come back automatically. I /think/ it's close enough for now, but defer to @itao.

server/sqlstore/database.go
```diff
@@ -0,0 +1,135 @@
...
+	// import sql drivers
+	_ "github.com/go-sql-driver/mysql"
+	_ "github.com/lib/pq"
+	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+// database exposes the underlying database.
+type database struct {
```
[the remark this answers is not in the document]
lieut-data: That's reasonable -- I'll try to drive that, but it may require "upstream approval".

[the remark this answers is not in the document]
lieut-data: Raised over at https://community.mattermost.com/core/pl/3akxk61ebjfqmbbh588pn5mrjr

lieut-data:
@itao, can I get your eyes on the proposed changes here? I'd like to get buy-in before asking for a second developer review just to avoid thrashing.

itao (approved):
Thanks for the ping @lieut-data, I read through the Jira ticket and this approach seems fine to me.

crspeller (approved):

prapti:
@lieut-data Not The plugin built from this branch works normally with all the UI functionalities available both on TE and E0 locally; I was thinking that even though it can get installed and enabled without error, the functionalities should not be available to users. Wondering if I understood it correctly?

lieut-data:
No, you definitely understood the intent correctly. That's very odd: let me take a double look to see what's gone (horribly) wrong! :)

lieut-data:
@prapti, a few updates. First, I merged master and pushed a change to depend on the formal `mattermost-plugin-api` that's since been updated to not enforce the E20 check on connecting to the store. This, however, is unrelated to your observation.

I was able to partially reproduce what you describe. For TE, I ran `BUILD_ENTERPRISE=false BUILD_ENTERPRISE_READY=false make run-server` and discovered that the existing license is still loaded, even though the `enterprise` code is not present to do anything advanced. So it seems possible to put TE into a semi-licensed state, which incident response honours. I'm unsure if we should do anything here, since the license is present and valid and we don't actually depend on anything enterprise-specific at that point.

But for both TE with the license removed, and E0 with the license removed, the plugin behaves as advertised: disabling itself when the license is not present on startup (or removed thereafter), and enabling otherwise. Can you help me reproduce the symptoms you were originally describing?

prapti (commented):
@lieut-data It is looking much better now. I can't reproduce the issue I was seeing last week. However, TE (running locally as you mentioned) did show this error when trying to activate: `"Unable to activate plugin","plugin_id":"com.mattermost.plugin-incident-management","error":"a valid Mattermost Enterprise E20 license is required to use this plugin"`. I can see this happening in this server too although this server is more of an E0 than TE: https://ic-te.test.mattermost.cloud/
One other thing I noticed too was that, although the rest of the features are not available as expected, the `/incident` slash commands are still available
Screen Shot 2021-03-01 at 4 12 06 PM

lieut-data:
Thanks, @prapti -- let's plan to chat through this one later today as I bet there's still a mismatch between my environment and yours, and I want to get to the bottom of that to make sure we've covered the cases here.

prapti (approved):
Thank you for taking a deeper look into this today @lieut-data. I tested again on a fresh server and can verify that there is no such issues that I had mentioned earlier. The plugin features stay hidden (except for the slash commands) in servers that don't have license but activate without any issue. The features turn on seamlessly when an E20 license is provided, or migrated from E0 or E10 to E20.
Approving.

PR #473, opened 2021-02-16
Use only one version when naming the build

cpoile:
#### Summary
- There's no way to narrow down the `git tag` command to filter (with a glob or regex) so using the simple rule:
  - If there are multiple tags, we care most about tags that start with `v`, and for simplicity just use the first one we find.

#### Ticket Link
- none


crspeller (approved):

lieut-data (approved):
Thanks, @cpoile!

lieut-data:
Skipping `QA Review` since this only involves build tooling.

PR #474, opened 2021-02-17
MM-30878 Change format of dates that are not this year

berkeka:
#### Summary

Changes incident create date format if year is different than the current year.

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/16399


crspeller (commented):
Great! But I think we need to do the same for `endedAt` below.

berkeka:
> Great! But I think we need to do the same for `endedAt` below.

@crspeller Do we still want show "--" for incidents that were created before 2020-01-01?

crspeller:
@berkeka I think that's a backwards compatibility thing. So we should keep that.

crspeller (approved):
Looks good! Thanks for the contribution!

cpoile (approved):
Looks great, thanks @berkeka !

prapti:
The only easy way to test this is using the the `/incident test` tool that @agarciamontoro developed I believe. The command however is not working as expected, and @agarciamontoro is looking into fixing a bug. Will check back again once there is a way to create incidents in the past dates.
Thank you for your patience!

agarciamontoro:
Ready to test!

prapti (approved):
Thank you for your patience through the review process @berkeka! And thank you to @agarciamontoro for helping with the foundational work into place to make testing this set of changes possible!
Everything looks great. I ran some tests to ensure that the newly created incidents and the old data migration is normal.
Approving to merge for the v1.7 release 🎉

PR #475, opened 2021-02-18
MM-32358 - /incident timeline slash command for mobile

cpoile:
#### Summary
- Show a table of the incident timeline for mobile (and desktop)
- I'm making a couple design decisions here. 
  1. Display as a table, because it's more space efficient than using slack attachments (which would be pages long for any serious incident)
  2. Display for desktop as well as mobile, because this is what Corey originally wanted, and if you're taking the time to type `/incident timeline` I imagine you want to see a cut-and-pastable summary of the incident events, rather than a message saying "look in the RHS"
- Will ask for Justine's review after dev review

Looks like:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32358


agarciamontoro (approved):
I like this idea better than what we do now for `/incident list` on mobile. I may steal it (or open a Help Wanted ticket?).
Just an idea: would it make sense to link the different events in the table to the corresponding posts, as the timeline does?

crspeller (approved):

cpoile:
Took @agarciamontoro 's idea, the timestamp is the link now. Notice that the slash commands don't have posts attached:
image

cpoile:
@justinegeffen Looking for review on the help text and feedback text (most of what is in quotations in the code).

justinegeffen (approved):

justinegeffen:
LGTM! Thank you! :D

prapti (approved):
@cpoile Just wanted to double check however that `/incident timeline` table shows all events including status updates, checking, unchecking, task assignments etc., whereas the RHS timeline view, for now, only shows the status updates between different statuses?
If that's true, the rest looks good to me. Added the following test cases to the test plan:
- `/incident timeline` shows  a fresh new incident a table with a “Reported” event
- `/incident timeline` after a few checks and unchecked shows a table with the check and uncheck events
- `/incident timeline` after status update shows a table with the status update event
- `/incident timeline` in a non-incident channel shows a “You can only run the timeline command from within an incident channel” message
- `/incident timeline` works in reported incident
- `/incident timeline` work in active incident
- `/incident timeline` works in resolved incident
- `/incident timeline` works in archived incidents
- Incident timeline mobile view works in native mobile app
- Incident timeline mobile view works in mobile web

prapti:
@cpoile Actually, the timestamp link does not work in mobile apps (native mobile app in both Android and iOS). Should I open a different ticket for this?
29E3347D-802E-47FC-BE71-8F72426A3056

cpoile:
@prapti Thank you for the awesome review and new test cases. That looks perfect.

I'm adjusted the list to display the same events as the RHS (no tasks checked/unchecked, assignee changed, or slash commands run). 

re: the mobile redirect links will be working in v1.41.0, added by @lieut-data here: https://github.com/mattermost/mattermost-mobile/pull/5136

We /could/ remove the links if the user is on mobile, but I don't like that solution because then v1.41+ users couldn't use them. I think we should keep the links and users of the older versions will get the error (until they upgrade). Is that an okay user experience?

prapti:
@cpoile Thank you for the update. I quickly tested again to verify the latest changes and they look good.

Regarding the mobile redirect links, if they'll work in 1.41 that'll be fine imo, the only concern is that 1.40 is still pending release, so it might be a little while before 1.41 is out the door. Is there any way the message on mobile could be changed so that it doesn't say "the link belongs to a deleted team or to a team to which you don not have access"? That is quite misleading. If that's not an easy task to do, I'd maybe defer it to @itao's decision depending on how many customers are likely to use the feature.

lieut-data:
@prapti, unfortunately there's no way to change the experience in v1.40 of `_redirect`. But, @cpoile, we /could/ avoid the `_redirect` semantics altogether and just resolve the team name into the URL (unless it's a DM). Might that reduce the scope?

cpoile:
@prapti Thanks for finding this -- the new commit does what Jesse suggested, it uses the team name for the post link. @lieut-data it does work with DMs. :)

prapti (approved):
@cpoile Yeah, this is a much better experience !  🙌  Thank you!
Approving again :)

lieut-data:
@cpoile, I may have merged this pre-maturely -- not sure if the target branch was correct. Apologies!

cpoile:
@lieut-data hmm, it will collapse the separate (rather large) PRs into a single commit on master -- is that okay? I can reopen and redo it to keep it separate (my preference)

PR #477, opened 2021-02-19
MM-32853 & MM-32951 - Hovermenu tweaks and timeline notice

cpoile:
#### Summary
- Moving the hovermenu up a bit so it doesn't hide the `details` icon. This makes it a tad more difficult to click, but it's consistent with how hovermenus work in UIs in general. Please try it out (it's tough to describe the interaction in photos).
- The Incident Timeline notice shows when there are no events (either because the incident had activity before v1.4.0, or because the events were removed). 
- The timeline was using the `reported` event to figure out when the event time starts, but if that's manually removed it wouldn't exist. So using the incident `create_at` field instead (should have used that from the start anyway). I can't think of why that wouldn't be the `reported at` time, can anyone else think of why that would cause a problem?
- e2e tests are broken, probably because cloud branch is broken (community reverted yesterday). Hopefully will be fixed soon.

Looks like:
image
image
image
image

#### Ticket Link
- Fixes: https://mattermost.atlassian.net/browse/MM-32853
- https://mattermost.atlassian.net/browse/MM-32951

### DIFF
--- a/tests-e2e/cypress/integration/frontstage/rhs/timeline-spec.js
+++ b/tests-e2e/cypress/integration/frontstage/rhs/timeline-spec.js
@@ -243,6 +243,22 @@ describe('timeline', () => {
             removeTimelineEvent('commander_changed', 1, 0, 'Commander changed from @aaron.peterson to @user-1');
         });
     });
+
+    describe('timeline notice', () => {
+        it('shows when there are no events', () => {
+            // * Verify incident created message is visible in the timeline
+            verifyTimelineEvent('incident_created', 1, 0, 'Incident Reported by user-1');
+
+            // * Delete the incident created event
+            removeTimelineEvent('incident_created', 1, 0, 'Incident Reported by user-1');
+
+            // * Verify notice is shown
+            cy.get('#rhsContainer').within(() => {
+                cy.findByText('Timeline events are displayed here as they occur. Hover over an event to remove it.')
+                    .should('exist');
+            });
+        });
+    });
 });
 
 const verifyTimelineEvent = (expectedEventType, expectedNumberOfEvents, expectedEventIndex, expectedEventSummary) => {
--- a/webapp/src/components/checklist_item.tsx
+++ b/webapp/src/components/checklist_item.tsx
@@ -48,17 +48,6 @@ const HoverableIcon = styled.i`
     }
 `;
 
-const InfoIcon = styled(HoverableIcon)`
-    position: relative;
-    top: 2px;
-`;
-
-const CloseIcon = styled(HoverableIcon)`
-    position: absolute;
-    right: 13px;
-    top: 13px;
-`;
-
 const StyledPopover = styled(Popover)<PopoverProps>`
     min-width: 180px;
     border-radius: 8px;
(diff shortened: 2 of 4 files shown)


cpoile:
Here's a test server to try it out: https://timelinetest.test.mattermost.cloud/test/channels/test-1. 
L: sysadmin P: Sys@dmin123

justinegeffen (approved):

tests-e2e/cypress/integration/frontstage/rhs/timeline-spec.js
```diff
@@ -241,6 +241,22 @@ describe('timeline', () => {
...
+        it('shows when there are no events', () => {
+            // * Verify incident created message is visible in the timeline
+            verifyTimelineEvent('incident_created', 1, 0, 'Incident Reported by user-1');
+
+            // * Delete the incident created event
+            removeTimelineEvent('incident_created', 1, 0, 'Incident Reported by user-1');
+
+            // * Verify notice is shown
+            cy.get('#rhsContainer').within(() => {
+                cy.findByText('Timeline events are displayed here as they occur. Hover over an event to remove it.')
```
justinegeffen, line 255:
```suggestion
                cy.findByText('Timeline events are displayed here as they occur. Hover over an event to permanently remove it.')
```

justinegeffen: @cpoile, wondering if it's obvious that it's permanent or whether we should add this?

agarciamontoro (approved):
Nice, thanks! :tada:

This solution works perfectly for now, but I would not be opposed to revisit this at some point, being "consistent with how hovermenus work in UIs in general" is not a good thing per se :joy:

tests-e2e/cypress/integration/frontstage/rhs/timeline-spec.js
cpoile: @justinegeffen I don't thinkwe need the `permanently`, personally. 🤷 But I'll defer to your decision on the wording.

server/sqlstore/incident.go
```diff
@@ -472,6 +472,23 @@ func (s *incidentStore) getTimelineEventsForIncident(q sqlx.Queryer, incidentIDs
 	return timelineEvents, nil
 }
 
+// GetTimelineEvent returns the timeline event for incidentID by the timeline event ID.
+func (s *incidentStore) GetTimelineEvent(incidentID, eventID string) (*incident.TimelineEvent, error) {
```
crspeller, line 476: Why include both the incident and event IDs? Why not just select by event ID?

cpoile: Gah, this PR is showing a lot of code (like this) from other PRs. 

Anyway: True, we could just select by ID (which I originally did). But then we have to check that this event (which we're going to delete or modify) is attached to the incident that the user has been verified to be able to access (otherwise they could call an API point with an `incidentID` they have access to, but using an `eventID` they don't have access to). So I checked that the event's `incidentID` matched the incidentID we were using in the route. Then I thought: just add that into the query and then you don't have to do that manual check.

cpoile: There, fixed the rebase

crspeller: Makes sense!

crspeller:
Waiting on the hover changes for approval.

cpoile:
@crspeller Ready. Also fixed the positioning of the icons and the profile selector button (notice that it isn't aligned and the background isn't the same as the trashcan on community-daily).
image
image
image
image

crspeller (approved):

PR #481, opened 2021-02-26
Fix height of checklist item containers

agarciamontoro:
#### Summary
This PR fixes the height of the checklist item containers. I simply removed the `ChecklistItemContainerLive` component, replacing its only usage with `ChecklistItemContainer` (which was not used anywhere else that was not dead code, actually).

| Before 	| After 	|
|-	|-	|
| image 	| image 	|

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33301

PR #482, opened 2021-02-26
Remove dead code

agarciamontoro:
#### Summary
Just that, while doing #481 I found some dead code in `checklist_item.tsx`, so this PR removes it. I'm not sure why the eslint rule for `no-unused-vars` does not apply in the code I removed, but instead of going down that rabbit hole, I simply removed what I found.

I also ran `npm run fix` and some indentations were changed as a result :man_shrugging:

#### Ticket Link
--


cpoile (approved):
I'll let Christopher decide is the `ChecklistItemDetailsEdit` is something he was planning on using later, and that's the reason it's still in (he's pretty ruthless about removing dead code). :)

crspeller (approved):
Fine to remove this. Always in the history. Not sure what we are going to do about editable checklist items yet.

agarciamontoro:
@prapti skipping QA here because this just removes dead code

PR #483, opened 2021-02-26
MM-31561: Commander automation

agarciamontoro:
#### Summary

This PR continues the work on the automation features, adding the ability to assign a user as a commander when an incident is created.

assign-commander

##### Changes in the DB
  - `DefaultCommanderID`, which, unsurprisingly, contains the ID of the commander.
  - `DefaultCommanderEnabled`, that stores the on/off state of the setting.
  - `DefaultCommanderID`, which stores a copy of the data in the playbook's column when the incident is created. If the playbook's `DefaultCommanderEnabled` contains `false`, the data is not copied, so there's no need to add a `DefaultCommanderEnabled` column to the `IR_Incident` table.

##### Changes in the service
When the incident is created, the user selected is assigned as commander (if the user is not invited, it is automatically invited to the channel). I changed the messages from the bot to adapt to the cases where the commander and reporter are the same or different,

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31561


agarciamontoro:
(Oops, added Prapti too soon, sorry!)

server/incident/service.go
```diff
@@ -151,12 +160,22 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
...
+	if incdnt.CommanderUserID != incdnt.ReporterUserID {
+		commander, err2 := s.pluginAPI.User.Get(incdnt.CommanderUserID)
+		if err2 != nil {
+			return nil, errors.Wrapf(err2, "failed to resolve user %s", incdnt.CommanderUserID)
+		}
+
+		startMessage = fmt.Sprintf("This incident has been started by @%s and is commanded by @%s.", reporter.Username, commander.Username)
+	}
+
+	newPost, err := s.poster.PostMessage(channel.Id, startMessage)
```
cpoile, line 178: The `SubjectUserID` on line 189 should be set to the `ReportedUserID` -- thanks!

webapp/src/components/backstage/automation/assign_commander_selector.tsx
```diff
@@ -0,0 +1,218 @@
...
+    getProfiles: () => ActionFunc;
+    isDisabled: boolean;
+}
+
+const AssignCommanderSelector: FC<Props> = (props: Props) => {
+    const [options, setOptions] = useState<UserProfile[]>([]);
+    const [searchTerm, setSearchTerm] = useState('');
+    const commanderUser = useSelector<GlobalState, UserProfile>((state: GlobalState) => getUser(state, props.commanderID));
+
+    // Update the options whenever the passed user IDs or the search term are updated
```
cpoile, line 29:
```suggestion
    // Update the options whenever the commanderID or the search term are updated
```

cpoile (approved):
Looks really nice, great tests as always, thanks Alejandro!

server/incident/service.go
agarciamontoro: Oh, good catch, thanks!

webapp/src/components/backstage/automation/assign_commander_selector.tsx
agarciamontoro: Fixed, thank you!

server/incident/service.go
```diff
@@ -84,6 +84,16 @@ func (s *ServiceImpl) GetIncidents(requesterInfo RequesterInfo, options FilterOp
 
 // CreateIncident creates a new incident. userID is the user who initiated the CreateIncident.
 func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public bool) (*Incident, error) {
+	if incdnt.DefaultCommanderID != "" {
+		if s.pluginAPI.User.HasPermissionToTeam(incdnt.DefaultCommanderID, incdnt.TeamID, model.PERMISSION_LIST_TEAM_CHANNELS) {
```
crspeller, line 88: We check "view team" when saving, but we check "list team" here. It should probably be the same permission. 
The thing is it seems like its the permission to be invited to a channel. Which really anyone on the team should. Maybe we check for membership? View team is probably good, since Guests won't have the list team channels permission.

crspeller (changes requested):

server/incident/service.go
agarciamontoro: Argh, I never get this right, thank you for catching it! I think we catch sys admins with `model.PERMISSION_VIEW_TEAM`, which I guess we want to?

agarciamontoro: But we list all the members of a team in the selector, so we should probably be consistent with that.

crspeller (approved):
That works I think 👍

agarciamontoro:
It actually does not :( If the user was a member of the team and they were later removed from it, the `GetTeamMember` function returns a valid `model.TeamMember` struct, only with the `DeleteAt` set to something different than 0.
I'm trying to figure out the best way to do this, because I could simply filter the ones that have the `DeleteAt` field set, but was trying to see if there were more robust solutions.

crspeller:
@agarciamontoro Hmm. Just checking for delete at would probably work fine. Maybe wrap that in a "Is current member of team" function. Forgot about that particular gotcha.

agarciamontoro:
By the way, the refactor is in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/483/commits/0b74ac009902497fd2da022d4b6169294e705b29.
The fix for the permissions, using the new isolated `permissions`, is in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/483/commits/f743d624d057568e803c0d74d7cbac6604399ae9

Thank you!

server/permissions/permissions.go
```diff
@@ -109,17 +105,26 @@ func CanPostToChannel(userID, channelID string, pluginAPI *pluginapi.Client) boo
...
 
-	return incident.RequesterInfo{
+	return RequesterInfo{
 		UserID:  userID,
 		IsAdmin: isAdmin,
 		IsGuest: isGuest,
 	}, nil
 }
+
+func IsMemberOfTeamID(userID, teamID string, pluginAPI *pluginapi.Client) bool {
```
crspeller, line 123: I remembered the reason this doesn't work as a "Permissions" check. If a system admin is not a member of the team they should probably still have permissions to do things. 
I think it's OK for where we are using it, since it's not permissions it's just filtering the list of users to only the ones on the team. (being the commander grants no permissions) 
However in the future that might not be true. 
Can we maybe move this out of the permissions package to avoid confusion? Add a comment explaining this behavior?

crspeller (changes requested):

server/permissions/permissions.go
agarciamontoro: I am not sure I understand the confusion.
I would agree with you in case the function was named something similar to `permissions.ViewTeam`, but I think that the name is self-explanatory as it is: it just checks if a user is member of a team, given its ID. The reason I put it in the `permissions` package is because I consider it a permissions check, as being a member of a team grants some permissions over that team (it's true that an admin also has some of those permissions, even if they are not a member, but I would say that this is not relevant because the function is scoped to checking the team membership).
Where would you put the function?

cpoile (approved):
Looks good, thanks @agarciamontoro !

server/permissions/permissions.go
crspeller: As long as the person knows about this gotcha, then it would be ok where it is. I guess I am just worried because this has been a problem in the past.

agarciamontoro:
@prapti, the E2E tests here cover most of the expected behaviour, and this PR is very similar to the one for the invited users, so your review here should be similar to the one in #440.
There's also a refactor on the permissions package. I tried to make sure I did not break anything, and the tests passing give me some peace of mind, but if you want to do a smoke test around permissions all around, that would be nice.
Also, #490 is probably way more straightforward, as it's more isolated, so feel free to start there and get back to this one *if* you have time.

prapti (approved):
Commander automation is looking great! 🎉  
FYI, I've added the following test cases to the test plan, most of which have already been automated, so thank you for adding the e2e tests.. they are great! But just for reference:
1. “Automation” tab in playbook creation page has “Assign commander” setting
2. “Assign commander” is disabled by default
3. Member cannot be added to “Assign commander” when setting is disabled
4. Member can be added to “Assign commander” when setting is enabled
5. Only one member can be added to “Assign commander” when setting is enabled
6. Member added to “Assign commander” can be edited to change to a different member
7. Member info of assigned commander persists even after the setting is disabled
8. Member added to “Assign commander” is auto-invited to incident channel
9. Member added to “Assign commander” is auto assigned as the incident commander when starting a new incident with the playbook
10. A bot message confirms the member added to “Assign commander” as the commander in the incident channel
11. Reporter of the incident is assigned commander when member added to “Assign commander” is removed from team after being added
12. Reporter of the incident is assigned as commander if no member is added to “Assign commander” when enabled
13. Reporter of the incident is assigned as commander if “Assign commander” is disabled
14. Commander is removed and “Assign commander” disabled when the commander initially assigned is removed from team
15. Reporter of the incident is assigned commander when member added to “Assign commander” is deactivated
16. Member who is not a team member cannot be searched to assigned commander

Test # 15 fails right now, but I don't think it's a blocker for this feature (and I need to test a few more features with deactivated users). I can open up a ticket to address the "bug".
Approving for 1.6!

agarciamontoro:
Thank you for the thorough review, Prapti! About test 15, I also think it can wait, ping me when you've tried those features with deactivated users!
I made a quick review myself and happened to have the dark theme activated, and saw that the selected commander had the wrong colors. I have added a new commit with single addition to the CSS. I'm gonna merge it without a re-review because it is a change as isolated as it can get!
Thanks again :)

agarciamontoro:
(and a go mod tidy to make the linter happy)

PR #484, opened 2021-02-26
MM-33335  Use CurrentStatus as source of truth. 

crspeller:
#### Summary
Since we are unable to distinguish between the cases of a brand new incident and a migrated incident with no status updates, the method of using status updates as the source of truth for current status is not viable. So to support the migration case I have simplified and now use the CurrentStatus field in the incident as the source of truth. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33335


prapti (approved):
Plugin upgrade now changes all previously ongoing incidents' status to "Active" regardless of whether the status was updated or not. Looks good to me!
Approving this prior to dev reviews since it needs to go in 1.5. However, if there is any change request in this PR, please reassign me for a second review and I'll be happy to do it. Otherwise let's merge it 🙂

cpoile (approved):
This feels better anyway. I didn't really like that way of computing status, a little too error-prone.

lieut-data (approved):
Nice -- glad we were writing the field already, simplifying this migration!

lieut-data:
@crspeller, leaving you to merge and handle any release branches necessary.

PR #485, opened 2021-03-02
MM-32113: Fix banner width

agarciamontoro:
#### Summary
The styling of the banner dates back to when we had a left hand side bar in the backstage (https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/70/)!

This PR moves the old CSS to a shared style component and updates its width to simply `100%`. I also had to add a `left: 0` so the banner for the exported channel in the incident details view is correctly placed. The `banner__text` class was not defined anywhere, so I ignored it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32113

### DIFF
--- a/webapp/src/components/backstage/shared.tsx
+++ b/webapp/src/components/backstage/shared.tsx
@@ -4,9 +4,8 @@
 import styled from 'styled-components';
 
 export const Banner = styled.div`
-    color: #155724;
-    background-color: #d4edda;
-    border-color: #c3e6cb;
+    color: var(--button-color);
+    background-color: var(--button-bg);
     position: fixed;
     top: 0;
     left: 0;
--- a/webapp/src/components/backstage/shared.tsx
+++ b/webapp/src/components/backstage/shared.tsx
@@ -1,0 +1,18 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import styled from 'styled-components';
+
+export const Banner = styled.div`
+        color: #155724;
+        background-color: #d4edda;
+        border-color: #c3e6cb;
+        position: fixed;
+        top: 0;
+        left: 0;
+        width: 100%;
+        z-index: 8;
+        overflow: hidden;
+        padding: 1rem 2.4rem;
+        text-align: center;
+`;
--- a/webapp/src/components/backstage/shared.tsx
+++ b/webapp/src/components/backstage/shared.tsx
@@ -4,15 +4,15 @@
 import styled from 'styled-components';
 
 export const Banner = styled.div`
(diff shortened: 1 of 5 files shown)


webapp/src/components/backstage/shared.tsx
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import styled from 'styled-components';
+
+export const Banner = styled.div`
+    color: #155724;
+    background-color: #d4edda;
+    border-color: #c3e6cb;
```
cpoile, line 9: Should these be themed?

cpoile (approved):
One suggestion, otherwise looks good, thanks Alejandro!

webapp/src/components/backstage/shared.tsx
agarciamontoro: I also thought about it, but I checked and the colors look good on dark themes. @alexsiclari, should we theme these colors for the banners that appear when deleting a playbook or exporting an incident's channel from the backstage?

image

crspeller (approved):

webapp/src/components/backstage/shared.tsx
agarciamontoro: Maybe we can use the same colors of the unread messages banner?
image
image

alexsiclari:
@agarciamontoro, I would stick to the full-width banner here to inform the user of changes based on their actions. The other banner is used more in the context of allowing the user to scroll to something by clicking on it or the user has to scroll to the location for it to go away, where as the full-width banner should dismiss on its own after a set time.

Regarding theming, for now, I would theme this because eventually, we will be updating platform-wide to have a true semantic color palette built into the themes.

alexsiclari (approved):
@agarciamontoro please refer to my comment that I made in regards to display type and theming in the comments here.

agarciamontoro:
Thank you, @alexsiclari! And sorry for the confusion, I posted the screenshots only to show the colors in that banner, not the banner itself, whose shape and behaviour is not changed in the backstage. The doubt I have is that I don't know which variables to use when theming.

agarciamontoro:
In the end, I styled these following the unread message toast colors:
image
image

agarciamontoro:
@prapti, heads up that this not only changes the banner shown when deleting a playbook, but also when exporting a channel from the incident details in the backstage. I updated the QA steps in the ticket itself.

prapti (approved):
Tested and verified locally in all three browsers at varying widths that the banner now stretches across the entire browser.
Approving.

agarciamontoro:
Thank you, Prapti!

PR #486, opened 2021-03-03
MM-33439: Fetch invited users independently

agarciamontoro:
#### Summary
The component was filtering the invited members out of the list that is fetched when the component is mounted. This list is capped at 100 users (the default for getUsers), so if the invited member was not originally there, it was not shown in the component.

This PR fixes the bug by pre-fetching all the invited members. In particular:
1. Fetch all the invited members on component mount.
2. If the user has not typed anything in the search box, just pass the fetched members to react-select as the group of invited members. All the invited members will be there.
3. If the user has typed something in the search box, rely on the server-side filtering. This is the tricky part: if there are more than 100 hits for your search term, it is possible that the invited member is still lost, but this also happens with non-invited members. The alternative is to do the filtering client-side, but I took a look at the filtering on the server side to try and replicate it and it's quite complex: we're filtering not only by username, but also by the first and last name, and I am sure t]at if we try to replicate that here, it will come back to bite us. That's why I decided it was an acceptable trade-off to rely on the server filtering in this case.

This is a working fix for this dot release, but we still need something more robust for the future (for example, we don't even have pagination in the fetching of the users). I added this to my now long list of things I want to tackle as part of https://mattermost.atlassian.net/browse/MM-32861

About the release itself: I created a branch from the v1.5.1 tag, release-v1.5.2, which is the base for this PR. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33439

### DIFF
--- a/webapp/src/components/backstage/automation/invite_users_selector.tsx
+++ b/webapp/src/components/backstage/automation/invite_users_selector.tsx
@@ -1,4 +1,5 @@
 import React, {FC, useState, useEffect} from 'react';
+import {useDispatch, useSelector} from 'react-redux';
 
 import ReactSelect, {GroupType, ControlProps, OptionsType, MenuListComponentProps} from 'react-select';
 
@@ -7,6 +8,9 @@ import {Scrollbars} from 'react-custom-scrollbars';
 import styled from 'styled-components';
 import {ActionFunc} from 'mattermost-redux/types/actions';
 import {UserProfile} from 'mattermost-redux/types/users';
+import {GlobalState} from 'mattermost-redux/types/store';
+import {getUser} from 'mattermost-redux/selectors/entities/users';
+import {getUser as fetchUser} from 'mattermost-redux/actions/users';
 
 import Profile from 'src/components/profile/profile';
 
@@ -20,11 +24,22 @@ interface Props {
 }
 
 const InviteUsersSelector: FC<Props> = (props: Props) => {
+    const dispatch = useDispatch();
+
     // When there are no users invited, options is UserProfile[], a plain list. When there is at least one user invited,
     // options contains two groups: the first with invited members, the second with non invited members. This is needed
     // because groups are rendered in the selector list only when there is at least one user invited.
     const [options, setOptions] = useState<UserProfile[] | GroupType<UserProfile>[]>([]);
     const [searchTerm, setSearchTerm] = useState('');
+    const invitedUsers = useSelector<GlobalState, UserProfile[]>((state: GlobalState) => props.userIds.map((id) => getUser(state, id)));
+
+    useEffect(() => {
+        invitedUsers.forEach((user, idx) => {
+            if (!user) {
+                dispatch(fetchUser(props.userIds[idx]));
+            }
+        });
+    }, [props.userIds]);
 
     // Update the options whenever the passed user IDs or the search term are updated
     useEffect(() => {
@@ -38,16 +53,23 @@ const InviteUsersSelector: FC<Props> = (props: Props) => {
 
             //@ts-ignore
             profiles.then(({data}: { data: UserProfile[] }) => {
-                const invitedProfiles: UserProfile[] = [];
-                const nonInvitedProfiles: UserProfile[] = [];
+                let invitedProfiles: UserProfile[] = [];
+                let nonInvitedProfiles: UserProfile[] = [];
 
-                data.forEach((profile: UserProfile) => {
-                    if (props.userIds.includes(profile.id)) {
-                        invitedProfiles.push(profile);
-                    } else {
-                        nonInvitedProfiles.push(profile);
-                    }
-                });
+                if (term.trim().length === 0) {
+                    invitedProfiles = invitedUsers.filter((user) => user);
+                    nonInvitedProfiles = data.filter(
+                        (profile: UserProfile) => !props.userIds.includes(profile.id),
+                    );
+                } else {
+                    data.forEach((profile: UserProfile) => {
+                        if (props.userIds.includes(profile.id)) {
+                            invitedProfiles.push(profile);
+                        } else {
+                            nonInvitedProfiles.push(profile);
+                        }
+                    });
+                }
 
                 if (invitedProfiles.length === 0) {
                     setOptions(nonInvitedProfiles);
@@ -64,7 +86,7 @@ const InviteUsersSelector: FC<Props> = (props: Props) => {
         };
 
         updateOptions(searchTerm);
-    }, [props.userIds, searchTerm]);
+    }, [props.userIds, searchTerm, JSON.stringify(invitedUsers)]);
 
     let badgeContent = '';
     const numInvitedMembers = props.userIds.length;
--- a/webapp/src/components/backstage/automation/invite_users_selector.tsx
+++ b/webapp/src/components/backstage/automation/invite_users_selector.tsx
@@ -25,13 +25,9 @@ interface Props {
 
 const InviteUsersSelector: FC<Props> = (props: Props) => {
     const dispatch = useDispatch();
-
-    // When there are no users invited, options is UserProfile[], a plain list. When there is at least one user invited,
-    // options contains two groups: the first with invited members, the second with non invited members. This is needed
-    // because groups are rendered in the selector list only when there is at least one user invited.
-    const [options, setOptions] = useState<UserProfile[] | GroupType<UserProfile>[]>([]);
     const [searchTerm, setSearchTerm] = useState('');
     const invitedUsers = useSelector<GlobalState, UserProfile[]>((state: GlobalState) => props.userIds.map((id) => getUser(state, id)));
+    const [searchedUsers, setSearchedUsers] = useState<UserProfile[]>([]);
 
     useEffect(() => {
         dispatch(getProfilesByIds(props.userIds));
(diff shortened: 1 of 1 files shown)


webapp/src/components/backstage/automation/invite_users_selector.tsx
```diff
@@ -64,7 +86,7 @@ const InviteUsersSelector: FC<Props> = (props: Props) => {
         };
 
         updateOptions(searchTerm);
-    }, [props.userIds, searchTerm]);
+    }, [props.userIds, searchTerm, JSON.stringify(invitedUsers)]);
```
agarciamontoro, line 89: If using the raw array, the component enters a loop where this effect is called infinitely. This is because even though the contents are the same, the array is changing. I am relying on this stringify call, as stated in https://github.com/facebook/react/issues/14476#issuecomment-471199055

agarciamontoro: (if you have other ideas on how to approach this, please tell me so!)

webapp/src/components/backstage/automation/invite_users_selector.tsx
```diff
@@ -38,16 +53,23 @@ const InviteUsersSelector: FC<Props> = (props: Props) => {
...
-                        invitedProfiles.push(profile);
-                    } else {
-                        nonInvitedProfiles.push(profile);
-                    }
-                });
+                let invitedProfiles: UserProfile[] = [];
+                let nonInvitedProfiles: UserProfile[] = [];
+
+                if (term.trim().length === 0) {
+                    invitedProfiles = invitedUsers.filter((user) => user);
```
lieut-data, line 60: This appears to just copy `invitedUsers`? Should we make that more explicit with:
```suggestion
                    invitedProfiles = [...invitedUsers];
```

webapp/src/components/backstage/automation/invite_users_selector.tsx
```diff
@@ -38,16 +53,23 @@ const InviteUsersSelector: FC<Props> = (props: Props) => {
...
+                    nonInvitedProfiles = data.filter(
+                        (profile: UserProfile) => !props.userIds.includes(profile.id),
```
lieut-data, line 62: Just a caution that this is `O(n^2)` in terms of the number of invited users. /Probably/ not an issue, but just want to make sure it's expected that the list stays relatively small.

webapp/src/components/backstage/automation/invite_users_selector.tsx
lieut-data: Could we avoid the dependency by having the effect maintain a set of `searchedUsers` and then just compute `invitedProfiles` and `nonInvitedProfiles` for each render from the available data?

webapp/src/components/backstage/automation/invite_users_selector.tsx
```diff
@@ -20,11 +24,22 @@ interface Props {
...
     // options contains two groups: the first with invited members, the second with non invited members. This is needed
     // because groups are rendered in the selector list only when there is at least one user invited.
     const [options, setOptions] = useState<UserProfile[] | GroupType<UserProfile>[]>([]);
     const [searchTerm, setSearchTerm] = useState('');
+    const invitedUsers = useSelector<GlobalState, UserProfile[]>((state: GlobalState) => props.userIds.map((id) => getUser(state, id)));
+
+    useEffect(() => {
+        invitedUsers.forEach((user, idx) => {
+            if (!user) {
+                dispatch(fetchUser(props.userIds[idx]));
```
lieut-data, line 39: To avoid a future problem, is there a convenient `fetchUsers` we could use instead?

lieut-data (changes requested):
Good catch, @agarciamontoro! A few thoughts below.

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: I'm filtering out the undefined ones! We can make it more explicit with `user !== undefined`, if you want to.

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: Oh there is a getProfilesByIds, I'll check if it does what it says.

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: But we still need to depend on the list of invitedUsers being updated when the invited users are finally fetched, right? Maybe I misunderstood your proposal here.

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: It is, but I cannot think of a simple way to reduce the complexity here. For every member in the list, we need to decide if it is invited or not.

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: It does!

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: Typescript complains about using undefined. I see even more confusing to add a typescript ignore comment, so I let it as it was. The other cool option is doing `invitedUsers.filter(Boolean)` :sunglasses: But I see that it's equally confusing.

agarciamontoro:
@lieut-data, addressed one (1) of your comments, I replied to the other ones in place to follow the conversation.

webapp/src/components/backstage/automation/invite_users_selector.tsx
```diff
@@ -20,11 +24,18 @@ interface Props {
...
     // When there are no users invited, options is UserProfile[], a plain list. When there is at least one user invited,
     // options contains two groups: the first with invited members, the second with non invited members. This is needed
     // because groups are rendered in the selector list only when there is at least one user invited.
     const [options, setOptions] = useState<UserProfile[] | GroupType<UserProfile>[]>([]);
     const [searchTerm, setSearchTerm] = useState('');
+    const invitedUsers = useSelector<GlobalState, UserProfile[]>((state: GlobalState) => props.userIds.map((id) => getUser(state, id)));
+
+    useEffect(() => {
+        dispatch(getProfilesByIds(props.userIds));
+    }, [props.userIds]);
```
cpoile, line 38: Inside here, for example. Turn `invitedUsers` into a state hook above, and fill them inside this effect hook.

webapp/src/components/backstage/automation/invite_users_selector.tsx
cpoile: The `(u) => u` is fine by me (we do a similar filtering elsewhere in the code). Maybe leave a comment for future us who won't remember this idiom? :)

cpoile (changes requested):
Nice catch! This is a tricky one for sure. 

Not sure if the comment I gave will help or not, but let me know.

webapp/src/components/backstage/automation/invite_users_selector.tsx
```diff
@@ -20,11 +24,18 @@ interface Props {
...
 
 const InviteUsersSelector: FC<Props> = (props: Props) => {
+    const dispatch = useDispatch();
+
     // When there are no users invited, options is UserProfile[], a plain list. When there is at least one user invited,
     // options contains two groups: the first with invited members, the second with non invited members. This is needed
     // because groups are rendered in the selector list only when there is at least one user invited.
     const [options, setOptions] = useState<UserProfile[] | GroupType<UserProfile>[]>([]);
     const [searchTerm, setSearchTerm] = useState('');
+    const invitedUsers = useSelector<GlobalState, UserProfile[]>((state: GlobalState) => props.userIds.map((id) => getUser(state, id)));
```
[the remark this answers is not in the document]
agarciamontoro: I vaguely remember this discussion in other parts of the codebase, but not the specifics on what caused the re-renders.

I took a look at the docs and, according to [this](https://react-redux.js.org/api/hooks#equality-comparisons-and-updates), the component will re-render only if the returned object is different (as in `===` returning false). It's true that the selector is executed with every change in the state, but the reason for the re-render in this specific case would be that `invitedUsers` changes. As `invitedUseres` is an array that gets created every time the selector is executed, it will definitely re-render, but it looks like we can prevent that by using a custom comparison operator, such as `shallowEqual` from `react-redux`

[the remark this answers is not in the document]
agarciamontoro: Or maybe not: https://github.com/reduxjs/react-redux/issues/1654 I don't get it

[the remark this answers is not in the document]
agarciamontoro: Anyway, I tested this with the paint flashing tool and I am not able to reproduce what you say. The component does not re-render when I make changes to the store (I tried removing a user, creating a new channel, inviting the user that has the backstage open to a new channel and then kicking them, and posting messages in channels where the user is). 
In none of the cases did I get a repaint, so I guess it is not being re-rendered. Maybe I'm missing something else here, of course.

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: Added a comment in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/486/commits/5d0531150a2484814bde10fc027b0f61a0db7e72, resolving this one!

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: I think I know now what you mean, let me push a new commit with this.

agarciamontoro:
I've simplified the useEffect hook by moving all the logic to build the `options` outside of it, so the dependency on the invitedUsers is implicit. I empirically checked that there is no re-painting happening, but another look into that would be appreciated, as I still don't understand it completely, I think.

webapp/src/components/backstage/automation/invite_users_selector.tsx
lieut-data: One way out would be to convert the individual lookups to `O(1)` using a dictionary, but I suspect that's not necessary at this time.

webapp/src/components/backstage/automation/invite_users_selector.tsx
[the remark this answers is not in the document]
lieut-data: I don't think the issue is the inability to track the selector's dependencies -- Redux always calls all `mapStateToProps` every time the store changes -- but the fact that this selector always returns a new array even if the underlying conditions haven't changed:
```js
const userIds = ['userid1', 'userid2', 'userid3'];
getUser = (userId) => userId;

const call1 = userIds.map((id) => getUser(id));
const call2 = userIds.map((id) => getUser(id));

// returns false
console.log(call1 === call2);
```

Redux selectors are /supposed/ to be pure, so I think the pattern above will unconditionally trigger React's "render" phase whenever anything in the state changes. However, React won't actually render anything new to the DOM, since the output of this function will be the same as before. This does come at a cost, however, over skipping the render phase altogether.

I add this comment not that I think we necessarily need to touch anything here -- I see this as a low-bandwidth page whose components won't be mounted very often -- but just to share what I imagine is going on under the covers.

webapp/src/components/backstage/automation/invite_users_selector.tsx
lieut-data: I /think/ this is perhaps an anti-pattern, having a component juggle second-order state like this. If this is a concern, perhaps we could introduced a memoized selector instead?

lieut-data (approved):
Nice simplifications, @agarciamontoro!

webapp/src/components/backstage/automation/invite_users_selector.tsx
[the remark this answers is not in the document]
agarciamontoro: Thank you for the insight! That's what I thought, but it's true that checking the repainting is not enough to test that the rendering phase is triggered again. Is there any easy way to test that? (just for the sake of knowing it for the future)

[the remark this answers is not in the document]
agarciamontoro: But I still don't get why the issue I linked before is like that: even when using a custom comparison function, the reload is triggered (the useEffect hook in the example in the issue).

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/486/commits/5b7f98c04c4b89ee3f0923925e364014d3071ee6.

webapp/src/components/backstage/automation/invite_users_selector.tsx
agarciamontoro: That's a good point, and an obvious one, actually (note to self: not think about complexity after 7 pm). I'll check if it's a good addition when doing the merge of the components, thanks!

webapp/src/components/backstage/automation/invite_users_selector.tsx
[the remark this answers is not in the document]
cpoile: Hmm, this is an interesting issue... Definitely we're good to go for this fix, but I really want to know what's going on under the hood here. Wonder if we should have webapp (or someone from our team) do a deep dive and then a bit of a presentation (a brown bag?) to the rest of the dev team

cpoile (approved):
Nice, thanks @agarciamontoro !

webapp/src/components/backstage/automation/invite_users_selector.tsx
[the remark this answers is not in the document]
agarciamontoro: I've been wanting to do that for a while (also for the sake of my understanding, because I feel like I lack some knowledge on the low level aspects of react and redux)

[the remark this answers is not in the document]
lieut-data: I might be mistaken, but I think you can just `console.log` within the functional component. If it's called, that's React's "render" phase. But React reserves the right to render at any time, so it's not a bug to have it render more than strictly necessary.

A deep dive here would be awesome!

prapti (approved):
Tested and verified that invited members and non-invited members still work as expected.
Approving.

PR #487, opened 2021-03-03
MM-32546 - Filtering events

cpoile:
#### Summary
- Can filter events using a dropdown. Your filter settings are remembered (until you reload the redux store).
- The dropdown is a modified dot-menu. By injecting a little bit of custom CSS we can reuse the dot-menu code pretty nicely. I like how this turned out. I think the custom css is cleaner than doing a complicated if-else inside the styled component (pro), but then you get some css living outside the base component (con). I think the pro outweighs the con here, because the existing uses of dot-menu don't change (they aren't injecting anything).
- One change: had to make the dropdown menu only be hidden when clicking outside the dot-menu/dot-menu-button, or else it would disappear whenever you clicked on an option. That doesn't affect the other uses of it, but it's something to keep in mind for the future.
- Based on designs by @alexsiclari here: https://www.figma.com/file/IHIkIwjUaZJGRdZZl9U5uZ/MM---32671?node-id=5%3A64
- I made some adjustments to the design, but I think it feels a bit better with less vertical space. Let me know @alexsiclari ?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-32546


webapp/src/actions.ts
```diff
@@ -23,11 +23,11 @@ import {
     ReceivedToggleRHSAction,
     REMOVED_FROM_INCIDENT_CHANNEL,
     RemovedFromIncidentChannel,
-    SET_CLIENT_ID,
+    SET_CLIENT_ID, SET_RHS_EVENTS_FILTER,
```
lieut-data, line 26: nit: why on the same line?

webapp/src/components/dot_menu.tsx
```diff
@@ -54,27 +61,29 @@ const DropdownMenu = styled.div<DropdownMenuProps>`
...
     left?: boolean;
+    menuCSS?: string;
+    buttonCSS?: string;
 }
 
 const DotMenu: FC<DotMenuProps> = (props: DotMenuProps) => {
     const [isOpen, setOpen] = useState(false);
     const toggleOpen = () => {
-        setOpen(!isOpen);
+        setOpen(true);
```
lieut-data, line 86: Intentional?

webapp/src/components/multi_checkbox.tsx
```diff
@@ -0,0 +1,74 @@
...
+    value: string;
+    selected: boolean;
+}
+
+interface Props {
+    options: CheckboxOption[];
+    onselect: (value: string, checked: boolean) => void;
+}
+
+const MultiCheckbox = (props: Props) => {
```
lieut-data, line 44: nit: can we eliminate the return and just have:
```js
 const MultiCheckbox = (props: Props) => (
    <DotMenu
```

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -59,30 +78,66 @@ interface Props {
...
-        case TimelineEventType.AssigneeChanged:
-        case TimelineEventType.TaskStateModified:
-        case TimelineEventType.RanSlashCommand:
-            return true;
+    const [allEvents, setAllEvents] = useState<TimelineEvent[]>([]);
+    const [filteredEvents, setFilteredEvents] = useState<TimelineEvent[]>([]);
+    const eventsFilter = useSelector<GlobalState, TimelineEventsFilter>(currentRHSEventsFilter);
+    const [filterOptions, setFilterOptions] = useState<CheckboxOption[]>([]);
+
+    useEffect(() => {
```
lieut-data, line 93: Can you elaborate on the need to `useEffect` here and why `filterOptions` is state and not just computed at run-time?

webapp/src/reducer.ts
```diff
@@ -144,11 +144,24 @@ const tabStateByChannel = (state: Record<string, RHSTabState> = {}, action: SetR
     }
 };
 
+const eventsFilterByChannel = (state: Record<string, TimelineEventsFilter> = {}, action: SetRHSEventsFilter) => {
+    switch (action.type) {
+    case SET_RHS_EVENTS_FILTER:
+        return {
+            ...state,
+            [action.channelId]: action.nextState,
```
lieut-data, line 152: I'm not sure I'd want this setting to be per-channel, and then only transient until the next reload anyway. I feel like it would be confusing on how to explain what I might see at any given moment. Should we just have this be a global RHS setting for any incident channels (no need to persist past reload).

lieut-data (changes requested):
Neat stuff! A few comments below, but I also feel like the proposed CSS approach is "wrong", and wonder if we might leverage @deanwhillier and the UI Platform team's input on how to structure things using styled components.

webapp/src/actions.ts
cpoile: That's just the way the IDE automatically placed it. I'll put it on a new line.

webapp/src/components/dot_menu.tsx
cpoile: Yes, that's how the menu stays open while clicking on the options (the third comment in the PR description).

webapp/src/components/multi_checkbox.tsx
cpoile: Yep!

webapp/src/components/rhs/rhs_timeline.tsx
cpoile: That was a refactor I should have kept refactoring -- fixed. Thanks for pointing that out.

webapp/src/reducer.ts
cpoile: That's a good point! But here's how I was thinking about it: There are two UX paths,
1. switching to a new incident and only seeing some events based on a filter set in another incident (possibly a long time ago)
2. or switching to a new incident and showing all events (the default).
 
I was thinking 2 would be less confusing, since it's the default.

Maybe @itao should weigh in, since it's kind of a UX experience?

cpoile:
@deanwhillier Thanks for taking a look at this! I know you're busy, so I'll try to summarize what the issue is: The `MultiCheckbox` component (in `multi_checkbox.tsx`) is customizing the `DotMenu` component's `DotMenuButton` and `DropdownMenu`. Normally I would style those components directly, which would let me adjust their CSS. But since `DotMenuButton` and `DropdownMenu` are owned by `DotMenu` I can't do that. A couple options would be to:
1. Require users of `DotMenu` to provide the `DotMenuButton` and `DropdownMenu` components they want the `DotMenu` to use.
2. Make more flags for `DotMenuButton` and `DropdownMenu` to adjust their css internally. (We're already doing this with the `top` and `left` flag in `DropdownMenu`, for example.)

I didn't want to do 1, for obvious reasons. I didn't want to do 2 because it seems like this would just pile on and we would be adding flags for every bit of css customization. It's okay once or maybe twice, but then it gets too much. If we add this `extraCSS` prop we can add customization without adding a new flag. (I was thinking of moving the `top` and `left` flags into the `extraCSS` too, to clean up the flags.)

Anyway, that was the thinking behind it. Let me know if there's a better way that I don't know about, thanks!

webapp/src/components/dot_menu.tsx
lieut-data: Should we eliminate `toggle*` altogether and just `() => setOpen(true)` or some such?

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -165,11 +134,41 @@ const RHSTimeline = (props: Props) => {
         );
     }
 
+    const makeFilterOptions = () => {
```
lieut-data, line 137: nit: can we just assign `filterOptions` directly and avoid the wrapper?

lieut-data (approved):

webapp/src/components/rhs/rhs_timeline.tsx
cpoile: 🤦 Of course!

webapp/src/components/dot_menu.tsx
```diff
@@ -54,27 +61,29 @@ const DropdownMenu = styled.div<DropdownMenuProps>`
...
     z-index: 1;
+    ${(props) => (props.extraCSS && css`${props.extraCSS}`)};
 `;
 
 interface DotMenuProps {
     children: JSX.Element[] | JSX.Element;
     icon: JSX.Element;
     top?: boolean;
     left?: boolean;
+    menuCSS?: string;
```
crspeller, line 79: I think the offical way to do this is with a className prop then extending the style https://styled-components.com/docs/advanced#styling-normal-react-components
Also sometimes its easier to just pass a prop that changes the CSS appropriately.

webapp/src/components/multi_checkbox.tsx
```diff
@@ -0,0 +1,72 @@
...
+        buttonCSS={'margin: 0 16px 0 auto'}
+    >
+        {props.options.map((option) => {
+            return (
+                <FilterCheckboxContainer
+                    key={option.value}
+                    onClick={() => props.onselect(option.value, !option.selected)}
+                >
+                    <input
+                        className='checkbox'
```
crspeller, line 61: Why `className`?

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -59,30 +78,35 @@ interface Props {
...
+    const [filteredEvents, setFilteredEvents] = useState<TimelineEvent[]>([]);
+    const eventsFilter = useSelector<GlobalState, TimelineEventsFilter>(currentRHSEventsFilter);
+
+    useEffect(() => {
+        setFilteredEvents(allEvents.filter((e) => showEvent(e.event_type, eventsFilter)));
+    }, [eventsFilter, allEvents]);
+
+    const selectOption = (value: string, checked: boolean) => {
+        if (value === 'all' && checked) {
+            dispatch(setRHSEventsFilter(currentChannelId, TimelineEventsFilterDefault));
```
crspeller, line 98: Are we sure this is the behavior we want? Like this it only mostly remembers your settings except annoyingly forgets them sometimes. Probably confusing to the user. 
Probably would rather it never remembers, considering the default is what you want most of the time?
If we want it to remember always, we could stuff it in local storage rather than using redux.

crspeller (changes requested):

webapp/src/components/dot_menu.tsx
cpoile: @crspeller are you sure the link is what you meant? It doesn't address this issue...
About passing the prop: that's what I mention to Dean above, point 2 -- we're piling on the special-cases. A `menuCSS` string seems cleaner to me, and then you don't have to make another special-case prop for the next customization of the DotMenu.

webapp/src/components/multi_checkbox.tsx
cpoile: Ah, good catch, thanks!

webapp/src/components/rhs/rhs_timeline.tsx
cpoile: What do you mean `Like this it only mostly remembers your settings except annoyingly forgets them sometimes. Probably confusing to the user.`?

webapp/src/components/dot_menu.tsx
crspeller: Not directly. But it's the official way to extend the css of a component. So framing the solution around that seems like a good idea. Possibly in combination with solution 1 above?

webapp/src/components/rhs/rhs_timeline.tsx
crspeller: I mean because we use the redux store and have it per channel. I could setup a bunch of filter settings on a bunch of channels, have it seem like they save, yhen I close my browser and reopen and they are all gone.

cpoile: Ah I see. But that's not unexpected, that's how all of our RHS state behaves now.

webapp/src/components/dot_menu.tsx
cpoile: I don't think that's what that section of the docs is talking about. We're not having trouble attaching a className or applying styles on a component.

It feels like Solution 1 is piling on special-cases, and I thought the css string would be a way to reduce the special case prop flags. I'm totally open to a better way, but more special case props seems like a code smell to me.

crspeller: I think it would just make it more flexible. Split up the main component and allow customizable sub components? Not going to block on this. It works fine as is.

webapp/src/components/rhs/rhs_timeline.tsx
crspeller: If that's how we want it to work.

crspeller (approved):

alexsiclari (approved):
The spacing looks okay. I think it feels a little better. I just want to ensure that we are using a consistent 8px grid spacing and not some random numbers. I think the spacing looks like it is 20px? If so, I would increase it to 24px so we can stay consistent. If it is 10px, then I would bump it up to 16px.

cpoile:
@alexsiclari It's 38px, I can bump it to 40px. Thanks!

deanwhillier:
Apologies for the delayed response @cpoile! I wanted to make sure I had a good idea of how things were working for your specific question and also in general with the incident plugin but I kept getting pulled away from my exploring.

So, there isn't really a "no-brainer" answer to this question and it depends on several factors. I'm honestly still learning Styled Components myself so am not yet 100% sure on what the 'best practice' would be from their perspective, though I do think that there is perhaps a better way to handle this kind of situation.

Unless I'm missing something, in both uses of the dot menu you're trying to:

1. Ensure there is spacing between the right side of the calling button and the edge of the screen.
2. Affect the position of the dropdown menu relative to the calling button.

Generally speaking, I like to think of the styling responsibilities of components like this:
- A component is ultimately responsible for its own visual styling and positioning of child elements within itself as the component should be fully aware of itself and how it should look and behave. Of course, the component can provide mechanisms to allow for flexibility depending on usage context/needs.
- A component's parent is ultimately responsible for the positioning and spacing of said component as the component itself doesn't and shouldn't need to have any idea of the context in which it will be used.

So with that in mind for #1, positioning the calling button away from the edge of the screen would probably best be handled by adding default padding to the containing `Header` component so the button doesn't need to worry about where it is positioned relative to the screen edge. If in the future you needed to have more than one button in the header with an appropriate amount of spacing between each, you could still handle that from the `Header` component (flexbox, grid, clever use of selectors, etc.).

Positioning the menu based on the button in this use case could probably be handled more generically without the need for passing in positioning data by removing the specific `left: ###px` and instead adding this to the components default styles for the drop down menu:

```css
right: 0;
white-space: nowrap;
```

Of course if you needed more flexibility around positioning, you could pass in the direction the menu should open as a prop (or even determine that at runtime) and then adjust the default CSS accordingly by using `left: 0;` vs `right: 0;` styles depending on the direction you needed the menu to open.

Alternatively, what I might do in this case is convert the 'Dot Menu' into a more generic set of menu components that support your various needs for menu's and then [extend those with styled-components](https://styled-components.com/docs/basics#styling-any-component) (`const DotMenu = styled(Menu)...`) in context as needed to create more specific components, like different menus with different button icons and menu elements.

Another approach I might generically take in situations where I legitimately need a parent component to affect the internal styling of a child component is to pass a new classname to the child component prefixed with the parent components name (using BEM naming) and then add styles to the parent component using precise selectors that target child elements with the provided classname to override default styling. In the case of styled-components, it's possibly not necessary to pass a new classname as you can do [SCSS-like css selectors](https://styled-components.com/docs/basics#pseudoelements-pseudoselectors-and-nesting) that are scoped to the parent element and target the child element.

All that said, with the effort underway in my new team to finally develop a common set of reusable components, menus will be one such component and so at some point in the near future you will be able to import our very own MM Compass Components to handle situations like this with hopefully less hassle on your end. :)

deanwhillier:
Gah, I directly commented instead of providing as a review comment. Removing myself for now, but please feel free to re-add me again if you would like me to review again!

cpoile:
@itao Here is the opinionated filter default:
image

@justinegeffen Can you suggest any other names for the filters? @itao expressed concern that the mix of past tense and plurality might be inconsistent, but I couldn't think of a better option--this was the least awkward of the possibilities, imo. What do you think?

itao:
I recommend using noun rather than past tense verb to describe the type of event as this may extend better to future event types such as `Pull request`, `Opsgenie alert`. So for this list here specifically, I recommend:
- Commander change
- Status update
- Channel message
- Task completion
- Task assignment
- Slash command

Also a suggestion for UI, it may be helpful to separate the `All events` with something like a divider and when it's selected, disable all the other options but preserve their value. This way it makes it clear that when it is selected, everything else is ignored but if the user unselects it, they don't lose the previous configuration.

justinegeffen:
Thaks, @cpoile. I see what you mean but the example provided reads very clearly and I don't think it's problematic to have that combination in this context. One suggestion would be to arrange them by tense/plurality. So you'd have all the "Changes" ones together, then "Updated", and then the rest so there's an obvious flow. Does that make sense?

justinegeffen:
I didn't see @itao's comment and the additional points about scalability aren't ones I considered. I'm not entirely sold on "Task completion" but the others do work, and I like the idea of a separator.

cpoile:
@itao @justinegeffen Here's what it looks like now. I realized we have to include the `Incident reported` event in the list, so these are the new defaults. 

I'm not sure if I like the tense now. I think I prefer past tense, @itao?
Also, I think `Task state change` is better than `Task assignment`, because the event can be a task-unassignment as well.

image

itao:
Mm maybe we should squash all task-related events into just `Task changes`? I'm 1/5, good with what you got there. Only other suggestion now that I look at it is to make them all plural, just like `All events`.

itao (approved):

justinegeffen:
So if it's past tense then it'd be:

- Incident reported
- Commander changed
- Status updated
- Channel messaged (?)
- Task state changed
- Task assigned
- Slash command

Not sure if "Channel message" works well if I have the tense correct. Maybe "Channel notified" would work? Otherwise I like the past tense because you're looking for historical information for the most part. 

Other comments: Slash command would work better under "All Events" and "All Events" should be "All events" to match the rest of the casing.

cpoile:
@itao @justinegeffen Thanks for the feedback! I decided to treat the `Incident Reported` event as part of the `Status updates` events (since it's a change in the status of the event), so I could remove the `Incident report` filter (it would only show or hide a single event, which is almost useless).  And I changed the filter labels to plural, which matched the `All events` and makes more sense to me. I think this looks good, but let me know if you have any concerns!

image

itao:
Looks sweet, ship ship away! 🚢   🎉

cpoile:
@deanwhillier Thank you for taking a look, Dean! I decided to continue to use boolean props, since that seems to be clearest, and it keeps the css grouped together in one place. I'm looking forward to seeing what your team comes up with wrt the correct way to deal with this type of problem! Thanks again.

cpoile:
@crspeller Could I get a quick look at the changes I've made since the last review? Shouldn't be anything blocking, but I'd feel better with another pair of eyes on it. Thanks!

webapp/src/components/multi_checkbox.tsx
```diff
@@ -0,0 +1,95 @@
...
+                    <input
+                        type='checkbox'
+                        checked={option.selected}
+                        disabled={option.disabled}
+                    />
+                    <OptionDisplay>{option.display}</OptionDisplay>
+                </FilterCheckboxContainer>
+            );
+
+            if (idx === 1) {
```
crspeller, line 81: This kinda ruins this components generic feel.

crspeller (approved):
Looks good. One non blocking suggestion.

webapp/src/components/multi_checkbox.tsx
cpoile: Good point, I fixed it. Thanks!

cpoile:
@prapti This is ready now, thanks!
I have e2e tests for all the combinations, so maybe it deserves just a smoketest. The settings will be remembered for each channel (until a webapp refresh).

prapti (approved):
Thanks for the heads up again @cpoile! Since you already added most of the tests that I would otherwise run manually, things went super fast than usual 🚀  Thank you so much for the great Cypress test coverage.
I've only added a few more to the test plan but most of them are covered by the tests you've written anyway; but just for reference:

1. Timeline tab has a filter button in the subheader
2. Hovering over the filter button highlights the button
3. Clicking on the filter button brings up a filter menu
4. Clicking on the filter button again closes the filter menu
5. Filter menu has the following option checkboxes:
    - All events
    - Commander changes
    - Status updates
    - Events from posts
    - Task state changes
    - Task assignments
    - Slash commands
6. By default, only the following are selected for an incident:Commander changesStatus updatesEvents from posts
7. Checking “All events” shows timeline events for all events regardless of previous filter
8. Checking “All events” deactivates the previously selected filter options
9. Unchecking “All events” shows timeline with events for previously selected filters
10. Unchecking all options hides all events from the timeline
11. Checking a filter option shows all events of that type in the timeline
12. Unchecking a filter option hides all events of that type from the timeline
13. Checking multiple filter options shows all events of the selected event types in the timeline
14. Filter settings are remembered when navigating away to a different channel
15. Filter settings are reset when the app is reloaded

Also did some sanity tests in dark and light themes in Chrome, Safari and Firefox.
The one tiny thing I noticed was that with dark theme, when all filters are unchecked (and nothing is there to show in the timeline), there is a tiny "|" sort of line that shows (as seen in the screenshot below). Not a blocker, but I think I'll open a different ticket later sometime. Also, the filter dropdown does not close when the button is clicked again (which I think is what you've mentioned in the PR description above). Would you also want me to open a ticket for that? The rest looks great.. the feature is awesome!

Screen Shot 2021-03-22 at 4 12 28 PM

cpoile:
@prapti Thanks! Arg, that one little piece of css won't go away. I'll look into that. Since it's only cosmetic and an edge case I'll merge after seeing why the e2e tests are failing.
Thanks!

PR #489, opened 2021-03-09
MM-33408 Basic Stats View

crspeller:
#### Summary
Basic statistics view. 
Single number stats included:
- The number of incidents in the reported state (clicking on this one opens incident list filtered by reported)
- The number of incidents in the active state (clicking on this on opens incident list filtered by active)
- The number of unique people in incident channels right now
- The average duration of incidents that are currently active (not resolved or archived)

Graphs included:
- The number of incidents that where active at any time on each day.
- The number of people currently in incidents that where created on each day.
- The average reported to active time for incidents created on each day
- The average reported to resolved time for incidents created on each day

"each day" begins at the current instant and are the 24 hour periods preceding regardless of day boundaries in the current time zone. 

This data is created using SQL queries on our existing data. Efficiency of these queries has not been looked at closely. `UniquePeopleInIncidents` is especially bad as it has to make separate queries for each day. 

If this expands, we may want to look into keeping this data as a time series rather than using these queries.

#### UI Screenshot
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33408


crspeller:
@itao Updated with your text changes except the `Stats` vs `Dashboard` one. Waiting on @alexsiclari 

We could totally include one. Make a ticket including the text you want.

crspeller:
> Total Incidents per day and total participants per day is showing a high total when it may not be that actual high of an amount.

@alexsiclari That data is correct for my test server.

itao:
> Total Incidents per day and total participants per day is showing a high total when it may not be that actual high of an amount.

I think @alexsiclari meant the chart give the impression that the an alarmingly high number of people (ie. 100% of employees) are stuck in incidents. Please correct me if that's not the case.

@alexsiclari Could you provide guidance on the visualization as well as fit and polish with a high fidelity screenshot?

cpoile:
@crspeller Should we wait until iterating on design is finished before code review?

crspeller:
@cpoile No, this is ready for dev review. Polish can come in another PR.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+	days := make([]int, 14)
+	days[0] = activeNow
+	for day := 1; day < 14; day++ {
+		days[day] = days[day-1] + ended[day] - started[day-1]
+	}
+
+	return days
+}
+
+// Inefficent. Calculates the number of people in the incident using
```
cpoile, line 205: For these queries, do we have any idea /how/ inefficient they are? They could be "bad" but not bad enough to worry about, considering we're not going to be running them that often. Or the could be really bad, and then we should just cache them. Just from eyeballing I can't tell, but it doesn't seem that horrible...

webapp/src/components/backstage/incidents/incident_list/incident_list.tsx
```diff
@@ -163,13 +166,24 @@ const BackstageIncidentList: FC = () => {
         },
     );
 
+    useEffect(() => {
+        // This mess makes typescript happy because string | null can't be assigned to string | undefined
+        const queryForStatus = new URLSearchParams(query).get('status');
+        let status: string | undefined;
+        if (queryForStatus) {
+            status = queryForStatus;
```
cpoile, line 174: `const status = queryForStatus | undefined`?

webapp/src/components/backstage/incidents/incident_list/incident_list.tsx
```diff
@@ -179,11 +193,17 @@ const BackstageIncidentList: FC = () => {
...
+                setIncidents(incidentsReturn.items);
+                setTotalCount(incidentsReturn.total_count);
+            }
         }
 
         fetchIncidentsAsync();
+
+        return () => {
+            isCanceled = true;
+        };
```
cpoile, line 206: ah, forgot about that. Probably got a "not mounted" react error, right?

webapp/src/components/backstage/stats.tsx
```diff
@@ -0,0 +1,301 @@
...
+    && {
+        color: inherit;
+    }
+`;
+
+const StatisticCount: FC<Props> = (props: Props) => {
+    const match = useRouteMatch<URLParams>('/:team/:plugin');
+    const inner = (
+        <div className='col-lg-3 col-md-4 col-sm-6'>
+            <div className='total-count'>
```
cpoile, line 47: Why not styled components (for this, and a lot of below)?

webapp/src/components/backstage/stats.tsx
```diff
@@ -0,0 +1,301 @@
...
+                            count={stats?.total_active_participants}
+                        />
+                        <StatisticCount
+                            id={'AverageDuration'}
+                            title={'Average Duration'}
+                            icon={'fa-clock-o'}
+                            count={renderDuration(moment.duration(stats?.average_duration_active_incidents_minutes, 'minutes'))}
+                        />
+                    </div>
+                    <GraphBox>
```
cpoile, line 140: Would it clean things up if we made a `GraphBox` FC? (similar to the `StatisticCount`)

cpoile (changes requested):
Nothing really major, just suggestions and a few questions. Nice work!

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+		From("ChannelMembers as cm").
+		Join("IR_Incident AS i ON i.ChannelId = cm.ChannelId").
+		Join("Channels AS c ON (c.Id = i.ChannelId)")
+
+	query = applyFilters(query, filters)
+
+	peopleInIncidents, err := s.MovingWindowQueryActive(query, 14)
+	if err != nil {
+		s.log.Warnf("Unable to get people in incidents %w", err)
+		return []int{}
```
lieut-data, line 219: nit: `return nil` instead?

server/sqlstore/stats.go
lieut-data: If we can avoid using any fields from the `channel` table /except/ for the `Id`, I think I'm ok with this approach -- running the queries 14 times is suboptimal, but bounded.

The chief problem I'm worried about is a table scan, but if we can rely on just using the indexes instead, we should be fine.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+
+	return results, nil
+}
+
+func (s *StatsStore) ActiveIncidents(filters *StatsFilters) []int {
+	now := model.GetMillis()
+
+	// Get the number of incidents started on each day
+	startQuery := s.store.builder.
+		Select(fmt.Sprintf("COUNT(i.Id), (%v - c.CreateAt) / 86400000 as DayStarted", now)).
```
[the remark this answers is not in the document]
lieut-data: Also, I commented about this below, but let's avoid using `c.*` except for the `Id` to avoid the inevitable table scan.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+	peopleInIncidents, err := s.MovingWindowQueryActive(query, 14)
+	if err != nil {
+		s.log.Warnf("Unable to get people in incidents %w", err)
+		return []int{}
+	}
+
+	return peopleInIncidents
+}
+
+// Average times from CreateAt to the first non reported update for the last number of days.
```
lieut-data, line 225: It wasn't obvious what `non-reported update` meant here until I read the query and parsed this as `non-"Reported"` update.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+// Averages are for incidents created on that day. Days with no created incidents use the last day.
+func (s *StatsStore) AverageStartToActive(filters *StatsFilters) []int {
```
lieut-data, line 227: This seems like the kind of query where we might consider adding explicit timestamps on the incident objects to record these instead.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+	daysToQuery := 42
+	firstNonReportedStatusPost := `(
+		SELECT p.CreateAt 
+		FROM IR_StatusPosts sp 
+		JOIN Posts AS p ON sp.PostId = p.Id
```
lieut-data, line 232: Is it feasible to avoid the `JOIN Posts`? That would remove most of the performance concerns.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+		WHERE sp.Status != 'Reported' AND sp.IncidentId = i.Id 
+		ORDER BY p.CreateAt ASC 
+		LIMIT 1
+	)`
+	now := model.GetMillis()
+
+	query := s.store.builder.
+		Select(fmt.Sprintf("COALESCE(FLOOR(AVG(%s - c.CreateAt)), 0) as Average, (%v - c.CreateAt) / 86400000 as DayStarted", firstNonReportedStatusPost, now)).
+		From("IR_Incident as i").
+		Join("Channels AS c ON (c.Id = i.ChannelId)").
```
lieut-data, line 242: Noted earlier, but if we can use `i.CreateAt` we won't need the `JOIN` at all.

lieut-data (changes requested):
A handful of thoughts and comments -- I think the biggest win here in terms of performance is to minimize joins to any of `Channels`, `Posts` or `ChannelMembers`, and then only on indexed fields (like `Id`).

I'd love to do a second pass after we decide on next steps here.

server/api/stats.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		HandleErrorWithCode(w, http.StatusBadRequest, "Bad filters", err)
+		return
+	}
+
+	if !permissions.CanViewTeam(userID, filters.TeamID, h.pluginAPI) {
+		HandleErrorWithCode(w, http.StatusForbidden, "permissions error", errors.Errorf(
+			"userID %s does not have view permission for teamID %s", userID, filters.TeamID))
+		return
+	}
```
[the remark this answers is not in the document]
crspeller: Good call on the guests. Guests actually do have the view team permissions. They don't have the list team channels permission. So I will modify to check that instead. 

> And who knows, in the future we might accidentally include confidential info in the stats.

Agreed, this will need tightening in the future, but for right now all the stats are fine to share with everyone.

> What happens if I omit the TeamID?

That gets caught in two places. The parseStatsFilters will return an error if the team_id is blank. Also the HasPermissionToTeam check does a blank check internally. You never have permissions to invalid IDs.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+	TeamID string
+}
+
+func applyFilters(query sq.SelectBuilder, filters *StatsFilters) sq.SelectBuilder {
+	return query.Where(sq.Eq{"i.TeamId": filters.TeamID})
+}
+
+func (s *StatsStore) TotalReportedIncidents(filters *StatsFilters) int {
+	query := s.store.builder.
+		Select("COUNT(ID)").
```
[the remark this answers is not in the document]
crspeller: I believe there is a slight one. Probably optimized away in most cases, but `COUNT` excludes NULL rows, so counting a single column can eliminate some checks.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+	var total int
+	if err := s.store.getBuilder(s.store.db, &total, query); err != nil {
+		s.log.Warnf("Error retriving stat %w", err)
+		return -1
+	}
+
+	return total
+}
+
+func (s *StatsStore) TotalActiveParticipants(filters *StatsFilters) int {
```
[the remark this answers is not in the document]
crspeller: That's a good point. I will ask Tao.

webapp/src/components/backstage/incidents/incident_list/incident_list.tsx
crspeller: Love to, but the linter won't let me use the keyword `undefined`.

webapp/src/components/backstage/incidents/incident_list/incident_list.tsx
crspeller: That, and if the `fetchParams` changed quickly, the correct request would sometimes not win the race.

webapp/src/components/backstage/stats.tsx
crspeller: Because I am stealing these from the system console. https://github.com/mattermost/mattermost-webapp/blob/master/components/analytics/statistic_count.tsx#L24

When we get our own styles we should replace these with styled components.

webapp/src/components/backstage/stats.tsx
crspeller: Yes! I should have done that before!

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+	query := s.store.builder.
+		Select("COUNT(DISTINCT cm.UserId)").
+		From("ChannelMembers as cm").
+		Join("IR_Incident AS i ON i.ChannelId = cm.ChannelId").
+		Where("i.EndAt = 0")
```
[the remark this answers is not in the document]
crspeller: To easily include both `Ended` and `Archived` status.

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+		s.log.Warnf("Error retriving stat %w", err)
+		return -1
+	}
+
+	return total
+}
+
+func (s *StatsStore) AverageDurationActiveIncidentsMinutes(filters *StatsFilters) int {
+	query := s.store.builder.
+		Select("AVG(c.CreateAt)").
```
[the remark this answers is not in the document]
crspeller: Yes. Will assume https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/491 is merged and modify queries as such.

server/sqlstore/stats.go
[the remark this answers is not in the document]
crspeller: No way to avoid the `fmt.Sprintf` here as the `Select` function takes a raw string.

server/sqlstore/stats.go
crspeller: That would work until we had custom statuses or something like that. Something to consider though.

server/sqlstore/stats.go
crspeller: I don't think so. Love to be proved wrong though. The status post is what we record for when an incident leaves "Reported". 
Also we already do this join in `GetIncidents`: https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/678a3c1fe38dc809186410f9cb5d815883b0f5ab/server/sqlstore/incident.go#L183 probably our hottest route? 
I think it even has worse performance characteristics there.

crspeller:
Awaiting #491

server/sqlstore/stats.go
```diff
@@ -0,0 +1,318 @@
...
+		Select("COUNT(ID)").
+		From("IR_Incident as i").
+		Where("CurrentStatus = 'Reported'")
+
+	query = applyFilters(query, filters)
+
+	var total int
+	if err := s.store.getBuilder(s.store.db, &total, query); err != nil {
+		s.log.Warnf("Error retriving stat %w", err)
+		return -1
```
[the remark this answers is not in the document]
crspeller: It doesn't. But it simplifies the code. Otherwise there would be a huge block of err checks doing this elsewhere. `-1` provides some warning that the stat is invalid.
1/5 I could move all these over.

webapp/src/components/backstage/incidents/incident_list/incident_list.tsx
```diff
@@ -2,11 +2,13 @@
 // See LICENSE.txt for license information.
 
 import React, {FC, useEffect, useState} from 'react';
+
```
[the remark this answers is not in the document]
crspeller: No idea. Artifact of the auto importer. Removed.

webapp/src/types/stats.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export interface PlaybookUse {
```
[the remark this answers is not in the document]
crspeller: Considering it's not used, that's not surprising 😆 Removed.

webapp/src/index.tsx
```diff
@@ -53,10 +53,10 @@ export default class Plugin {
                 mainMenuActionId = null;
             } else if (!mainMenuActionId && !isMobile()) {
                 mainMenuActionId = registry.registerMainMenuAction(
-                    'Playbooks & Incidents',
+                    'Incident Collaboration',
                     () => {
                         const team = getCurrentTeam(store.getState());
-                        navigateToTeamPluginUrl(team.name, '/playbooks');
+                        navigateToTeamPluginUrl(team.name, '/stats');
```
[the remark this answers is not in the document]
crspeller: Good question. Will ask Tao.

server/sqlstore/stats.go
crspeller: Same as above.

server/sqlstore/stats.go
lieut-data: Yeah, that's fair. I was imagining a solution that lets us track `CreateAt` directly on our tables which might alleviate some of the need to interop with the core Mattermost schema.

webapp/src/components/backstage/incidents/incident_list/incident_list.tsx
cpoile: I would add an `ignore` -- that's an "incorrect" linter error anyway, since a lot of js and js packages use `undefined` to specifically mean "don't include this field."

cpoile (approved):
🚢 🇮🇹

PR #490, opened 2021-03-09
MM-29101 - Fix for Calling Users endpoint too often

cpoile:
#### Summary
- I ended up settling on what @lieut-data and I discussed, essentially ensuring that the profiles are loaded.
- I gave up trying to find a clean way to batch or debounce calls to the users endpoint (since each hook is independent). 
  - Practically this doesn't matter, since the default RHS tab is `Summary`, the commander picker loads the profiles for the channel. Switching to the Tasks tab loads all 20 checklist items (for example), each with its own profile selector. But since the commander selector loaded the channel's profiles, no other selector needs to make the call. Cool.
  - In the weird chance that someone clears their store and somehow jumps to the Tasks tab before loading the Summary tab (not likely, or even possible, I think) each checklist item loads its own profile selector, each of those makes a call when it sees that there are no profiles in the store for the current channel. Once the profiles are loaded, they won't make the call again, but the server is still hit with 20 calls to the server.
  - Very rare (impossible?) that this will happen. But I still want to figure out how to make a debounced call to the server that would work across hooks. Maybe for a future project.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-29101


webapp/src/hooks.ts
```diff
@@ -95,3 +105,36 @@ export function useTimeout(callback: () => void, delay: number | null) {
...
+                },
+            ]));
+        };
+
+        if (profilesInChannel.length > 0) {
+            return;
+        }
+
+        getProfiles();
+    }, [currentChannelId, currentUserId, profilesInChannel]);
```
agarciamontoro, line 137: I am not sure I understand why we depend on the `profilesInChannel` here. If it changes, it's because the store has changed, so we would not need to update the store again by fetching the profiles one more time, right? Maybe I'm missing something here.

agarciamontoro (approved):
Nice use of custom hooks! I like this pattern here.

I left a couple of questions below, but nothing blocking.

webapp/src/hooks.ts
lieut-data: I too wondered about this -- intuitively, it seems like this hook will effectively double the number of API calls until the results "settle" (and our reducers don't just create new values each time: otherwise it's an infinite loop).

lieut-data (changes requested):
Same questions as @agarciamontoro.

webapp/src/hooks.ts
cpoile: Good questions. It doesn't double the amount of calls, because when `profilesInChannel` is updated, it will have a length > 0. But there's a possibility it could be updated to an empty list, so we would need to refetch. Not sure if that would ever happen, but it's a possibility.
Regardless, we /should/ put all the external dependencies in the effect's dependency list.

webapp/src/hooks.ts
```diff
@@ -95,3 +105,36 @@ export function useTimeout(callback: () => void, delay: number | null) {
...
+
+            dispatch(batchActions([
+                {
+                    type: UserTypes.RECEIVED_PROFILES_LIST_IN_CHANNEL,
+                    data: profiles,
+                    id: currentChannelId,
+                },
+                {
+                    type: UserTypes.RECEIVED_PROFILES_LIST,
+                    data: removeUserFromList(currentUserId, [...profiles]),
```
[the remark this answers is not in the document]
lieut-data: Instead of invoking the Client4 API ourselves, is it feasible to dispatch the same action used by the webapp to populate this list and avoid having to "juggle Redux state" manually?

webapp/src/hooks.ts
lieut-data: Ah, I see -- I read the async block as executed code instead of just being a definition.

lieut-data: Ah, I see -- I read the async block as executed code instead of just being a definition.

webapp/src/hooks.ts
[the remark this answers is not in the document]
cpoile: Thanks for pushing me on this; I had tried this earlier and it hadn't worked, but I think it's because I was importing one that webapp defined (and ran into the dependency import problem we get when trying to import components).

I can't help but think there's a more elegant way of using ActionFuncs with hooks. But this works. :)

webapp/src/hooks.ts
```diff
@@ -122,26 +121,15 @@ export function useClientRect() {
...
-                    type: UserTypes.RECEIVED_PROFILES_LIST_IN_CHANNEL,
-                    data: profiles,
-                    id: currentChannelId,
-                },
-                {
-                    type: UserTypes.RECEIVED_PROFILES_LIST,
-                    data: removeUserFromList(currentUserId, [...profiles]),
-                },
-            ]));
+            getProfilesInChannel(currentChannelId, 0, PROFILE_CHUNK_SIZE)(dispatch, store.getState);
```
lieut-data, line 132: I think it's possible to write this as the following, without the need to explicitly tie ourselves to the store:
```
            dispatch(getProfilesInChannel(currentChannelId, 0, PROFILE_CHUNK_SIZE));
```

cpoile: TIL, thanks! That simplifies everything.

lieut-data (approved):

prapti (approved):
Thank you for explaining the areas this PR touches @cpoile, that definitely helped understand and test!
Everything looks good, and didn't notice any regression bugs yet. Will keep an eye out after this becomes available in community and cloud.
Approving.

PR #491, opened 2021-03-10
Bring back IR_Incident's CreateAt and DeleteAt fields

agarciamontoro:
#### Summary
Undo part of what #429 did; namely, get back to use the `CreateAt` and `DeleteAt` fields from `IR_Incident`, instead of the same ones from `Channels`:
- https://github.com/mattermost/mattermost-plugin-incident-collaboration/commit/285ce1aa17b7a228b02897a7f1872cf135e23461 changes the incidents' select query to use the fields from the `IR_Incident` table, bringing back the old `CreateAt` and `DeleteAt` to the insert query, which were set to 0.
- As those fields have been set to 0 for the incidents created after #429 was merged, https://github.com/mattermost/mattermost-plugin-incident-collaboration/commit/9848849c499485d3aaa443b53c668d7a6d13af87 adds a new data migration that updates the incidents that have those fields set to 0, and copy the data back from the channel info.
- As this migration adds a dependency on the Channels table existing, I needed to adjust the tests so they call `setupChannelsTable` before testing the migrations.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33646


cpoile (approved):
Cool, looks great, thanks @agarciamontoro !

crspeller (approved):

agarciamontoro:
We need to test a couple of things here, @prapti:
1. That the new incidents use the creation date. I think the easiest way to test that is using #474 with these changes to make sure it works.
2. Check that the migration of old data works, and that it does not conflict with anything else. The data migration introduced by this PR changes all incidents that were created after the changes in #429 were released. I think that the easiest way to test this is to create some incidents with the current master, inspect the db to make sure their CreateAt and DeleteAt are set to 0, then deploy the plugin with this PR and make sure that those same incidents had their CreateAt and DeleteAt updated to the values of the corresponding fields in the incident's channel (DeleteAt will be different than 0 only if the channel was archived, which I did not test myself). It would be also nice to have a quick smoke test checking that the data of the incidents that were already in the database are not corrupted in any way because of the data migration.

Please, ping me if you need anything from me!

prapti (approved):
Looks good with where this PR right now. The DeleteAt field gets its data from the when the channel is archived, instead of the incident getting archived. So there isn't much we can do about.
Thank you for help with testing this @agarciamontoro! Approving!

agarciamontoro:
Thank you, Prapti!

PR #493, opened 2021-03-12
MM-33512 Create the updateStatus API endpoint

cpoile:
#### Summary
- Created the updateStatus API endpoint
- had to make some changes to the client library, will tag a new version after merged
- tests

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-33512


client/client.go
```diff
@@ -169,11 +169,16 @@ func checkResponse(r *http.Response) error {
...
+		Error string
+	}
 	if data != nil {
 		// Try to extract a structured error from the body, otherwise fall back to using
 		// the whole body as the error message.
-		if err := json.Unmarshal(data, errorResponse); err != nil {
+		if err2 := json.Unmarshal(data, &errorText); err2 != nil {
 			errorResponse.Err = errors.New(string(data))
+		} else {
+			errorResponse.Err = errors.New(errorText.Error)
```
cpoile, line 181: @lieut-data This was a weird little bug that tripped me up for awhile -- is it correct as originally written? Because it doesn't make much sense that way. This version makes more sense, since we return `{"error": "message"}` from our API endpoints.

cpoile:
I'm not sure why `go mod tidy` is different on CircleCI -- there's no difference on my local machine...

client/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.15
 require (
 	github.com/google/go-querystring v1.0.0
 	github.com/mattermost/mattermost-plugin-incident-collaboration v1.4.0
-	github.com/mattermost/mattermost-server/v5 v5.28.0
+	github.com/mattermost/mattermost-server/v5 v5.32.0
```
agarciamontoro, line 8: Why the bump here?

server/api/incidents.go
```diff
@@ -613,6 +614,61 @@ func (h *IncidentHandler) changeCommander(w http.ResponseWriter, r *http.Request
...
+		return
+	}
+
+	err = h.incidentService.UpdateStatus(incidentID, userID, options)
+	if err != nil {
+		HandleError(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
agarciamontoro, line 669: If we're not returning any data, we should consider responding with a `http.StatusNoContent`.

client/incident.go
```diff
@@ -163,3 +176,11 @@ type GetIncidentsResults struct {
 	HasMore    bool       `json:"has_more"`
 	Items      []Incident `json:"items"`
 }
+
+// StatusUpdateOptions are the fields required to update an incident's status
+type StatusUpdateOptions struct {
+	IncidentID        string `json:"incident_id"`
```
agarciamontoro, line 182: This is not needed in the JSON body, right? As far as I understand, it is only used when building the URL. If so, we can ignore it when marshaling:
```suggestion
	IncidentID        string `json:"-"`
```

server/api/api.yaml
```diff
@@ -674,6 +674,60 @@ paths:
...
+        description: Payload to change the incident's status.
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                status:
+                  type: string
+                  description: The new status. May be 'Reported', 'Active', 'Resolved', or 'Archived'
+                  example: Active
```
agarciamontoro, line 703: We can use the `enum` keyword here:
```suggestion
                status:
                  type: string
                  enum:
                    - Reported
                    - Active
                    - Resolved
                    - Archived
                  description: The new status.
                  example: Active
```

server/api/api.yaml
```diff
@@ -674,6 +674,60 @@ paths:
...
+                - status
+                - message
+      x-codeSamples:
+        - lang: curl
+          source: |
+            curl -X PUT 'http://localhost:8065/plugins/com.mattermost.plugin-incident-management/api/v0/incidents/1igmynxs77ywmcbwbsujzktter/update-status' \
+            -H 'Authorization: Bearer 9g64ig7q9pds8yjz8rsgd6e36e'
+            -d '{"status": "Resolved", "message": "Resolving incident because the issue was solved."}'
+      responses:
+        200:
```
agarciamontoro, line 722: (we should change this too if we finally decide to send the NoContent code, as proposed in the comment in `server/api/incidents.go`)

server/api/incidents_test.go
```diff
@@ -1168,4 +1169,141 @@ func TestIncidents(t *testing.T) {
...
+
+		incidentService.EXPECT().GetIncidentIDForChannel(testIncident.ChannelID).Return(testIncident.ID, nil)
+		pluginAPI.On("HasPermissionTo", mock.Anything, model.PERMISSION_MANAGE_SYSTEM).Return(false)
+		incidentService.EXPECT().GetIncident(testIncident.ID).Return(&testIncident, nil).Times(2)
+		pluginAPI.On("HasPermissionToChannel", mock.Anything, mock.Anything, model.PERMISSION_READ_CHANNEL).Return(true)
+		pluginAPI.On("HasPermissionToChannel", mock.Anything, mock.Anything, model.PERMISSION_CREATE_POST).Return(true)
+
+		err := c.Incidents.UpdateStatus(context.TODO(), icClient.StatusUpdateOptions{
+			IncidentID:        "incidentID",
+			Status:            "Arrrrrrrctive",
```
agarciamontoro, line 1225: This made me laugh :joy:

agarciamontoro (changes requested):
This is great, thank you! The tests are super thorough :pray: 

I left some minor comments below, all of them more than open to discussion!

client/client.go
lieut-data: Hmm, `ErrorResponse` defines the appropriate field into which to receive that `{"error": "message"}`:

https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/fa9fedde728f2f71319038a89343076925c3191d/client/error_response.go#L14-L15

Can you elaborate on what exactly was going wrong?

client/incident.go
```diff
@@ -82,6 +82,46 @@ type IncidentCreateOptions struct {
 	PlaybookID      string `json:"playbook_id"`
 }
 
+// IncidentSort enumerates the available fields we can sort on.
+type IncidentSort string
+
+const (
+	// SortByCreateAt sorts by the "create_at" field. It is the default.
+	SortByCreateAt IncidentSort = "create_at"
```
lieut-data, line 90: I realize this is the same as before, but a tangential observation nonetheless: `SortByCreateAt`, despite being an `IncidentSort`, is effectively a global name in this package. If we support sorting playbooks by similarly named fields, should we disambiguate either with a more precise name `SortIncidentByCreateAt`, or by making the sortable fields of a more generic, `Sort` type?

client/incident.go
```diff
@@ -82,6 +82,46 @@ type IncidentCreateOptions struct {
...
+	SortByCommanderUserID IncidentSort = "commander_user_id"
+
+	// SortByTeamID sorts by the "team_id" field.
+	SortByTeamID IncidentSort = "team_id"
+
+	// SortByEndAt sorts by the "end_at" field.
+	SortByEndAt IncidentSort = "end_at"
+
+	// DirectionAsc sets the direction of the sort to ascending (lower to higher).
+	DirectionAsc = "asc"
```
lieut-data, line 108: Do we need these /and/ `SortDirection` below?

client/incident.go
lieut-data: I think this is for marshalling to /send/ to the server.

client/incidents.go
```diff
@@ -113,3 +113,18 @@ func (s *IncidentsService) Create(ctx context.Context, opts IncidentCreateOption
 
 	return incident, nil
 }
+
+func (s *IncidentsService) UpdateStatus(ctx context.Context, opts StatusUpdateOptions) error {
```
lieut-data, line 117: Again, somewhat tangential given this interface is still changing, but I recall an observation from the plugin-api-wrapper days where we agreed to make required parameters explicit, and leave the options for optional ones. So for example, in this case I wonder if we should:

```
func (s *IncidentsService) UpdateStatus(ctx context.Context, incidentId string, status Status, Message string, opts StatusUpdateOptions) error {
```

The rationale was that this made it much more obvious to the author of the code what to supply, avoiding common mistakes like forgetting required fields. We don't /have/ to do this in this PR -- just surfacing for discussion. (We should probably do something similar for `List`, with `page` and `perPage` being required instead of optional to force the pagination problem into the forefront.)

server/api/incidents.go
```diff
@@ -613,6 +614,61 @@ func (h *IncidentHandler) changeCommander(w http.ResponseWriter, r *http.Request
 	w.WriteHeader(http.StatusOK)
 }
 
+// updateStatusD handles the POST /incidents/{id}/update-status endpoint, user has edit permissions
```
lieut-data, line 617:
```suggestion
// updateStatus handles the POST /incidents/{id}/update-status endpoint, requiring edit permissions from the user
```

lieut-data: Would it make more sense to anchor this at `POST /incidents/{id}/status`, since we're effectively creating a status for the given incident, vs. `update-status` which reads more like a `PUT`?

server/api/incidents.go
lieut-data: 3/5, but while semantically correct, I bet some clients trip over "weird" status codes. I can't recall ever seeing `StatusNoContent` in the wild.

lieut-data (approved):
Nothing blocking, but a handful of comments.

server/api/incidents.go
agarciamontoro: I'd say that in that case, it's a client bug. The standard [states](https://tools.ietf.org/html/rfc7231#section-6.3.1) that "Aside from responses to CONNECT, a 200 response always has a payload, though an origin server MAY generate a payload body of zero length.If no payload is desired, an origin server ought to send 204 (No Content) instead"

client/incident.go
agarciamontoro: Yes, but the server only reads the Message, Reminder and Status, right?

client/client.go
cpoile: Because the `"error"` field as parsed from the response is a `string`, so the marshaller throws an incompatible type error when it tries to stuff it into an `error` type.

lieut-data: Oh, that makes sense! Can we customize the unmarshalling of `ErrorResponse` to handle this correctly and avoid doing it here?

cpoile: Yep, I did it. Take a look to see if you agree with how I brought the "otherwise fall back to using the whole body as the error message" into `ErrorResponse`'s `UnmarshalJSON`. I like how that reduces the caller's code. But open to suggestions.

client/go.mod
cpoile: It's a dependency bump from plugin-api v0.0.14

server/api/incidents.go
cpoile: As a compromise I'm returning `{"status":"OK"}`. We're not doing that for every 200 endpoint, but we do for some (e.g.: the frontend telemetry, because javascript's fetch expects a body and will error out when it doesn't get it).

client/incident.go
cpoile: Right -- I'm removing `incident_id` and also making the argument list change below. Thanks!

server/api/api.yaml
cpoile: Awesome, thanks!

client/incident.go
cpoile: I think `IncidentSort` -> `Sort` makes sense, since we do keep the field name consistent. Done.

client/incident.go
cpoile: oops, I forgot to remove those (and these weren't being used)

client/incidents.go
cpoile: That's a good point, I like the arguments being a list of what is necessary, and prefer that over the struct. Done.

client/go.mod
agarciamontoro: Ooooh ok, thanks!

server/api/incidents.go
agarciamontoro: Don't want to block this PR, so I'm ok with the compromise suggested. However, we're using 204 in other parts of the codebase, and I still don't understand why we should not use them :thinking: We can continue the discussion elsewhere!

agarciamontoro: We can always return the status itself, which also complies with the standard.

client/incident.go
agarciamontoro: Cool, thanks!

agarciamontoro (approved):
Thank you for addressing all the changes :pray:

client/error_response.go
```diff
@@ -15,6 +19,25 @@ type ErrorResponse struct {
...
+	temp := &struct {
+		Err string `json:"error"`
+		*Alias
+	}{
+		Alias: (*Alias)(e),
+	}
+
+	// Try to extract a structured error from the body, otherwise fall back to using
+	// the whole body as the error message.
+	if err := json.Unmarshal(data, &temp); err != nil || temp.Err == "" {
```
lieut-data, line 33: Why the `temp.Err == ""`? Won't that be the "successful" case, meaning we'll always copy the body of every payload into `e.Err`?

lieut-data (changes requested):
Nice cleanup! Just one comment on something that seems broken.

client/error_response.go
cpoile: if `temp.Err == ""` it means we couldn't unmarshal the error message into the `error` field (maybe it wasn't formatted correctly, e.g.: `{"errors":"the error message"}`), so that's when the fall back is triggered (use the whole body as the error message).

lieut-data: Ah, my apologies, I had forgotten we were already in the context of "an error occurred, we're trying to figure out what it is."

lieut-data (approved):

PR #494, opened 2021-03-16
MM-33309 - Update confirm modal text

cpoile:
#### Summary
- Small update to the confirm modal text. See the linked Ticket for the original, and the changes for the new text. Let me know if you can think of a better way to explain the effects, thanks!

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-33309

### DIFF
--- a/webapp/src/components/checklist_item.tsx
+++ b/webapp/src/components/checklist_item.tsx
@@ -446,7 +446,7 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
             <ConfirmModal
                 show={showDeleteConfirm}
                 title={'Confirm Task Delete'}
-                message={`Are you sure you want to delete this task? "${title}"?`}
+                message={'Are you sure you want to delete this task? Deleted tasks will be removed from this incident\'s task list.'}
                 confirmButtonText={'Delete Task'}
                 onConfirm={() =>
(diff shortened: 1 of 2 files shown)


webapp/src/components/checklist_item.tsx
```diff
@@ -446,7 +446,7 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
             <ConfirmModal
                 show={showDeleteConfirm}
                 title={'Confirm Task Delete'}
-                message={`Are you sure you want to delete this task? "${title}"?`}
+                message={'Are you sure you want to delete this task? Deleted tasks will be removed from this incident\'s task list.'}
```
justinegeffen, line 449:
```suggestion
                message={'Are you sure you want to delete this task? Deleted tasks will be permanently removed from this incident\'s task list.'}
```

justinegeffen (commented):
Thanks, @cpoile! Added my suggestions - LMK if they work for you. :)

webapp/src/components/checklist_item.tsx
cpoile: I kind of don't want to put `permanently` here. For one, it's not permanent (we set the `delete_at` value, but we can always undo that elsewhere and re-add the event). But also because it seems overly harsh. But I'm 1/5. What do you think @justinegeffen ?

agarciamontoro (approved):
It sounds good!

PR #495, opened 2021-03-16
MM-31562: Incident creation announcement

agarciamontoro:
#### Summary
As with #440 and #483, this keeps on adding automations! This one is somewhat similar to the feature of broadcasting the status updates to a channel, but now with the incident creation. The announcement channel is independent of the one to broadcast the status updates.

##### Changes in the DB
  - `AnnouncementChannelID`, that contains the ID of the channel where the announcement is made
  - `AnnouncementChannelEnabled`, that stores the on/off state of the setting.
  - `AnnouncementChannelID`, which stores a copy of the data in the playbook's column when the incident is created. If the playbook's `AnnouncementChannelEnabled` contains `false`, the data is not copied, so there's no need to add a `AnnouncementChannelEnabled` column to the `IR_Incident` table.

##### Changes in the service
When the incident is created, it is announced in the configured channel. If the channel does not exist or is archived/deleted, the bot posts a message in the channel warning the user about it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31562


server/incident/service.go
```diff
@@ -83,6 +83,34 @@ func (s *ServiceImpl) GetIncidents(requesterInfo permissions.RequesterInfo, opti
...
+	if err != nil {
+		return err
+	}
+
+	announcementChannel, err := s.pluginAPI.Channel.Get(theIncident.AnnouncementChannelID)
+	if err != nil {
+		return err
+	}
+
+	if announcementChannel.DeleteAt != 0 {
```
crspeller, line 97: If we do care about the channel is in the right team thing, we need to check it here too. Channels can move between teams. (rare, but happens)

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
```diff
@@ -785,6 +785,241 @@ describe('backstage playbook details', () => {
...
+                    cy.visit(`/ad-1/com.mattermost.plugin-incident-management/playbooks/${playbookId}`);
+
+                    // # Switch to Automation tab
+                    cy.get('#root').findByText('Automation').click();
+
+                    // * Verify that the toggle is unchecked
+                    cy.get('#announcement-channel label input').should('not.be.checked');
+                });
+
+                it('can be enabled', () => {
```
crspeller, line 801: Is it necessary to check if a checkbox works?

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
```diff
@@ -785,6 +785,241 @@ describe('backstage playbook details', () => {
...
+
+                        // # Select a channel
+                        cy.selectChannel('Town Square');
+
+                        // * Verify that the control shows the selected commander
+                        cy.get('.channel-selector__control').contains('Town Square');
+                    });
+                });
+
+                it('allows changing the channel', () => {
```
crspeller, line 864: Is it necessary to verify the channel can be selected?

webapp/src/components/backstage/automation/announcement.tsx
```diff
@@ -0,0 +1,119 @@
...
+interface Props {
+    enabled: boolean;
+    onToggle: () => void;
+    channelId?: string;
+    onChannelSelected: (channelID: string | undefined) => void;
+}
+
+export const Announcement: FC<Props> = (props: Props) => {
+    return (
+        <>
```
crspeller, line 27: Remove extra wrapper?

crspeller (changes requested):
Good stuff!

server/api/playbooks.go
```diff
@@ -90,6 +90,29 @@ func (h *PlaybookHandler) createPlaybook(w http.ResponseWriter, r *http.Request)
...
+			HandleErrorWithCode(w, http.StatusBadRequest, fmt.Sprintf("announcement channel with ID %s does not exist", pbook.AnnouncementChannelID), err)
+			return
+		}
+
+		if channel.DeleteAt != 0 {
+			HandleErrorWithCode(w, http.StatusBadRequest, "channel is archived", errors.Errorf("channel with ID %s is archived", channel.Id))
+			return
+		}
+
+		if channel.TeamId != pbook.TeamID {
```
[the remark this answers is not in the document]
agarciamontoro: This makes it consistent with what we show in the UI, which is channels in the same team. I'd rather err in the side of being consistent instead of giving too much flexibility in the API side. What do you think?

server/api/playbooks.go
```diff
@@ -194,6 +217,27 @@ func (h *PlaybookHandler) updatePlaybook(w http.ResponseWriter, r *http.Request)
 		pbook.DefaultCommanderEnabled = false
 	}
 
+	if pbook.AnnouncementChannelID != "" {
```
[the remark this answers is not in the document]
agarciamontoro: Yeah, this was a design decision I made with the invite users: if one of the invited users is deleted/removed from the team, they are removed from the automation. Same with the commander, but in this case it is different: if the commander is gone, should we show the setting with no commander selected? That is not supported otherwise, and I think it would be more misleading having the setting enabled but completely useful (because we will not be able to assign the commander). Same for this setting with the announcement channel.

I agree with the helper function, thanks for catching that!

server/incident/service.go
agarciamontoro: Oh, I was not aware of that. Good catch!

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
agarciamontoro: Common things can be broken for the most strange reasons. I would not say it is super necessary, but it gives me some peace of mind. I am not concerned about my time developing this, so the only missing concern is the time spent in the tests. I don't think this specific test adds that much time, although the solution for that would be to parallelize the tests.

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
agarciamontoro: Same answer as above!

webapp/src/components/backstage/automation/announcement.tsx
agarciamontoro: Oops, you're right, thanks!

server/api/playbooks.go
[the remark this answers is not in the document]
crspeller: My intuition says the API should be as flexible as we can make it. I understand the need to simplify the UI. But what if a customer has the use case of a cross-team announcement and we can enable that for them using the API? I think someone using the API would probably appreciate the flexibility.

server/api/playbooks.go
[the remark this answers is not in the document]
crspeller: The part that gets me is removing it as a side effect of saving. I wonder if we could just forget about checking on update, and just put an error message in the UI warning that it's out of date? Like <Red highlight>"This user is disabled" or whatever.
This shouldn't block this PR though.

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
crspeller: My concern is mostly the overall maintenance cost of all these e2e tests. It might be simple to write. But when Tao goes and has us rename something, or we refactor, or the webapp changes, these tests break because they are super brittle most of the time. The less and more focused the tests, the less breakage we have to deal with. 
0/5 on this specific case. This is more something I have been thinking about overall.

server/api/playbooks.go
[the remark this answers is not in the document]
agarciamontoro: That's a good point. I'll remove this restriction, and check instead that the user is in the channel or can join it (the channel is public in a team the user is a member of, or the user is an admin). Do you happen to know if the PERMISSION_READ_CHANNEL would work for that use case?

tests-e2e/cypress/integration/backstage/playbook_details_spec.js
agarciamontoro: I agree that E2E tests are now very brittle. We need to solve that, but I'd vote for a different approach than writing less tests. I'd vote for leaving these ones here, but I'll take this into account for future tests. Does that sound good?

server/api/playbooks.go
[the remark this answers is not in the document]
agarciamontoro: I agree we need some error checking in the client side. That is something we've been postponing for a while, now, but that would let us simply return an error here. We need to discuss this to tackle it holistically, I think.

server/api/playbooks.go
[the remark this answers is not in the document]
agarciamontoro: @crspeller, I was having a hard time changing this because it was not clearly defined, so I got back to the ticket and saw that this restriction was there by design. What do you think about leaving this as is and reviewing it in a future PR, with a better defined permission system?

webapp/src/components/backstage/automation/announcement.tsx
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/495/commits/c5bf76ff373ef19c485ed3544d8255a8a314d8ed.

server/incident/service.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/495/commits/52ce165e96027496130681cf9d63e66a3e4d9d47.

server/api/playbooks.go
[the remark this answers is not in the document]
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/495/commits/52ce165e96027496130681cf9d63e66a3e4d9d47.

server/api/playbooks.go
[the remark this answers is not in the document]
crspeller: @agarciamontoro Maybe just do the same check as the broadcast channel ID above? PERMISSION_CREATE_POST sounds about right. It's still restricted in the UI so we can do what the ticket says.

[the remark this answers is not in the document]
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/495/commits/647abb0a0f3e2eddb8cda4dd1f5838581a3b443e

crspeller (approved):
Looking good!

server/incident/service.go
```diff
@@ -83,6 +83,29 @@ func (s *ServiceImpl) GetIncidents(requesterInfo permissions.RequesterInfo, opti
...
+	if err != nil {
+		return err
+	}
+
+	if err := permissions.IsChannelActiveInTeam(theIncident.AnnouncementChannelID, theIncident.TeamID, s.pluginAPI); err != nil {
+		return err
+	}
+
+	announcementMsg := fmt.Sprintf("#### New Incident: ~%s\n", incidentChannel.Name)
+	announcementMsg += fmt.Sprintf("**Commander**: @%s\n", commander.Username)
```
justinegeffen, line 97:
```suggestion
	announcementMsg += fmt.Sprintf("**Commander:** @%s\n", commander.Username)
```

server/incident/service.go
```diff
@@ -83,6 +83,29 @@ func (s *ServiceImpl) GetIncidents(requesterInfo permissions.RequesterInfo, opti
...
+	if theIncident.Description != "" {
+		announcementMsg += fmt.Sprintf("**Description**: %s\n", theIncident.Description)
```
justinegeffen, line 99:
```suggestion
		announcementMsg += fmt.Sprintf("**Description"** %s\n", theIncident.Description)
```

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
```diff
@@ -423,5 +423,149 @@ describe('incident automation', () => {
...
+                            playbookId = playbook.id;
+                        });
+
+                        // # Delete channel
+                        cy.apiDeleteChannel(channel.id);
+                    });
+                }).then(() => {
+                    cy.apiLogin('user-1');
+
+                    // # Create a new incident with the playbook.
```
justinegeffen, line 547:
```suggestion
                    // # Create a new incident with the playbook
```

server/incident/service.go
```diff
@@ -83,6 +83,29 @@ func (s *ServiceImpl) GetIncidents(requesterInfo permissions.RequesterInfo, opti
...
+	}
+
+	if _, err := s.poster.PostMessage(theIncident.AnnouncementChannelID, announcementMsg); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // CreateIncident creates a new incident. userID is the user who initiated the CreateIncident.
```
justinegeffen, line 109:
```suggestion
// CreateIncident creates a new incident. userID is the user who initiated the CreateIncident
```

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
```diff
@@ -423,5 +423,149 @@ describe('incident automation', () => {
...
+
+                    cy.apiStartIncident({
+                        teamId,
+                        playbookId: playbook.id,
+                        incidentName,
+                        commanderUserId: userId,
+                        description: 'Incident description.',
+                    });
+
+                    // # Navigate to the incident channel.
```
justinegeffen, line 455:
```suggestion
                    // # Navigate to the incident channel
```

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
```diff
@@ -423,5 +423,149 @@ describe('incident automation', () => {
...
+                    cy.visit(`/ad-1/channels/${incidentChannelName}`);
+
+                    // * Verify that the channel is created and that the first post exists.
```
justinegeffen, line 458:
```suggestion
                    // * Verify that the channel is created and that the first post exists
```

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
```diff
@@ -423,5 +423,149 @@ describe('incident automation', () => {
...
+                        teamId,
+                        playbookId: playbook.id,
+                        incidentName,
+                        commanderUserId: userId,
+                    });
+
+                    // # Navigate to the incident channel
+                    cy.visit(`/ad-1/channels/${incidentChannelName}`);
+
+                    // * Verify that the channel is created and that the first post exists.
```
justinegeffen, line 505:
```suggestion
                    // * Verify that the channel is created and that the first post exists
```

justinegeffen (commented):
I gave this a quick review and added some minor suggestions. :)

cpoile (approved):
Can't think of anything, looks great, thanks @agarciamontoro!

PR #500, opened 2021-03-23
MM-34207: css issue with RHS timeline

cpoile:
#### Summary
- @prapti noticed an edge case when the filter filtered out all events, a piece of the timeline's line still showed. So don't return the Timeline element if there are no filtered events. https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/487#pullrequestreview-618079651

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-34207
- https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/487#pullrequestreview-618079651


agarciamontoro (approved):
Looks good! :tada:

crspeller (approved):

PR #503, opened 2021-03-26
MM-31558: Outgoing webhook on incident's creation

agarciamontoro:
#### Summary
One more automation! This is a bit more fun, as we are talking to the outside world now via a webhook!

##### Changes in the backstage
There's a new "Send a webhook" entry in the automations group, with a simple text box. The text box validates itself via some cool `input` attributes (`required` and `pattern`) and CSS pseudo-classes (`:invalid`) I did not know about.
| Correct webhook 	| Misconfigured webhook 	|
|-	|-	|
| good 	| wrong 	|

##### Changes in the DB
  - `CreationWebhookURL`, which contains the, ehem, webhook URL.
  - `CreationWebhookEnabled`, that stores the on/off state of the setting.
  - `CreationWebhookURL`, which stores a copy of the data in the playbook's column when the incident is created. If the playbook's `CreationWebhookEnabled` contains `false`, the data is not copied, so there's no need to add a `CreationWebhookEnabled` column to the `IR_Incident` table.

##### Changes in the service
When the incident is created, a POST request is sent to the URL with a payload consisting of the incident data plus a couple of absolute URLs: one to the incident's channel, and another one to the incident's details in the backstage. I decided to isolate this work in a goroutine because I did not want to block the incident creation with the response from an external server we know nothing about.

If the response to the request is not in the 2XX range, the bot posts a message in the incident's channel warning about it. There are no retries whatsoever.

image

##### Changes in api.yaml
I noticed I had not updated the api.yaml file with any of the automations. I fixed it here! I also learned that we can add "callbacks" to document outgoing requests, which is nice. The only downside is that you need to tie it to one of the `path`s, so I added it to the playbook creation, which is the one that defines the URL.

image

##### Tests
I tested that the webhook is sent and that the payload is correct in `service_test.go`. I added E2E tests for the overall feature (basically, I'm checking that the error message is not posted if the webhook is correctly configured, and viceversa). I did not add any more E2E tests per our latest discussions on the mindset change!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31563


agarciamontoro:
@justinegeffen, I wanted to know your opinion on the bot's message. And, if you have time, on the changes in the `api.yaml` file

justinegeffen (commented):
I've taken a look through the PR and made some suggestions. I think I missed the bot messages though - could you let me know whether I did and I'll give it a second pass.

server/api/api.yaml
```diff
@@ -1424,12 +1424,67 @@ paths:
...
+              operationId: creationWebhook
+              requestBody:
+                required: true
+                content:
+                  application/json:
+                    schema:
+                      $ref: "#/components/schemas/CreationWebhookPayload"
+              responses:
+                "2XX":
+                  description: Your server returns a 2XX code if it successfuly received the request.
```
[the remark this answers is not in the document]
agarciamontoro: - I used the uppercase X because it's like that in the official docs: https://swagger.io/specification/#patterned-fields-1
- I never  trust myself when writing so many double consonants in that word. I'll add it in a separate comment not to add the lowercase, too, if that's ok!

server/incident/service.go
```diff
@@ -216,6 +273,15 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
 		}
 	}
 
+	if incdnt.CreationWebhookURL != "" {
+		go func() {
+			if err = s.sendCreationWebhook(incdnt); err != nil {
+				s.pluginAPI.Log.Warn("failed to send a POST request to the creation webhook URL", "webhook URL", incdnt.CreationWebhookURL, "error", err)
+				_, _ = s.poster.PostMessage(channel.Id, "Failed to communicate the incident's creation through the outgoing webhook.")
```
agarciamontoro, line 280: This is the bot message, @justinegeffen

agarciamontoro:
Sorry for the forced push, this was based on #495 and a simple merge was a bit cumbersome. All good to go now.

server/incident/service.go
```diff
@@ -106,6 +108,61 @@ func (s *ServiceImpl) broadcastIncidentCreation(theIncident *Incident, commander
...
+		theIncident.ID,
+	)
+
+	payload := struct {
+		Incident
+		ChannelURL string `json:"channel_url"`
+		DetailsURL string `json:"details_url"`
+	}{
+		Incident:   *theIncident,
+		ChannelURL: channelURL,
```
crspeller, line 145: These are just conveniences for the webhook receiver? We are imagining the receiver want's to display the incident to the user and these are the back-links?

server/api/api.yaml
```diff
@@ -1424,12 +1424,67 @@ paths:
...
+                  example: true
       x-codeSamples:
         - lang: curl
           source: |
             curl -X POST 'http://localhost:8065/plugins/com.mattermost.plugin-incident-management/api/v0/playbooks' \
             -H 'Authorization: Bearer 9g64ig7q9pds8yjz8rsgd6e36e'\
             -d '{"title": "Cloud Incidents", "description": "A playbook to follow when there is an incident regarding the availability of the cloud service.", "team_id": "p03rbi6viyztysbqnkvcqyel2i","create_public_incident": true,"checklists": [{"title": "Triage issue","items": [{"title": "Gather information from customer."}]}]}'
+      callbacks:
+        incidentCreation:
+          "{$request.body#/creation_webhook_url}":
```
cpoile, line 1474: Is this specific `api.yaml` code? I'm not sure how to parse it, to tell if it's correct or not. :)

server/incident/service.go
```diff
@@ -106,6 +108,61 @@ func (s *ServiceImpl) broadcastIncidentCreation(theIncident *Incident, commander
...
+		DetailsURL: detailsURL,
+	}
+
+	body, err := json.Marshal(payload)
+	if err != nil {
+		return err
+	}
+
+	resp, err := http.Post(theIncident.CreationWebhookURL, "application/json", bytes.NewReader(body))
```
crspeller, line 154: It's really tempting to use this, but we should make an `http.Client` and use that. At least the last time I checked `http.Post` has many many issues. 
I think the biggest things are having a constant connection pool and specifying a timeout. Have a look at how the server uses `services/httpservice` for inspiration.

server/incident/service.go
```diff
@@ -106,6 +108,61 @@ func (s *ServiceImpl) broadcastIncidentCreation(theIncident *Incident, commander
...
+	if err != nil {
+		return err
+	}
```
crspeller, line 157: Need to close the response body. See: https://golang.org/pkg/net/http/#Client.Post

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
```diff
@@ -567,5 +567,75 @@ describe('incident automation', () => {
...
+            it('with webhook correctly configured and setting enabled', () => {
+                const playbookName = 'Playbook (' + Date.now() + ')';
+
+                // # Create a playbook with a correct webhook and the setting enabled
+                cy.apiCreatePlaybook({
+                    teamId,
+                    title: playbookName,
+                    createPublicIncident: true,
+                    memberIDs: [userId],
+                    creationWebhookURL: 'https://httpbin.org/post',
```
crspeller, line 581: This will fail if this URL is unavailable. Or when there is an inevitable glitch.

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
```diff
@@ -567,5 +567,75 @@ describe('incident automation', () => {
...
+            it('with webhook misconfigured and setting enabled', () => {
+                const playbookName = 'Playbook (' + Date.now() + ')';
+
+                // # Create a playbook with a wrong webhook and the setting enabled
+                cy.apiCreatePlaybook({
+                    teamId,
+                    title: playbookName,
+                    createPublicIncident: true,
+                    memberIDs: [userId],
+                    creationWebhookURL: 'http://example.com/not-an-actual-endpoint',
```
crspeller, line 616: What happens if the owner of example.com decides this URL should return success?

webapp/src/components/backstage/automation/creation_webhook.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <div>{'Send a webhook'}</div>
+        </AutomationTitle>
+        <SelectorWrapper>
+            <TextBox
+                enabled={props.enabled}
+                type={'url'}
+                required={true}
+                value={props.enabled ? props.url : ''}
+                onChange={(e) => props.onChange(e.target.value)}
+                pattern={'https?://.*'}
```
crspeller, line 34: Nice!

crspeller (changes requested):

server/incident/service.go
```diff
@@ -106,6 +108,61 @@ func (s *ServiceImpl) broadcastIncidentCreation(theIncident *Incident, commander
 	return nil
 }
 
+// sendCreationWebhook sends a POST request to the creation webhook URL.
+// It blocks until a response is received.
```
cpoile, line 112: iirc, there's a client method you can use with a `Context` that will fire a timeout and close the request. Maybe we should use that? I agree that we don't need to worry about retries, but we 1/5 should try to protect against opening connections without a timeout.

server/incident/service.go
agarciamontoro: Yes, they are! It's an ask from Tao so that the users don't need to do it themselves. We do know it's redundant, as you can derive this string from the other fields (and the base URL) but it's a couple of strings and it can make their lives easier.

server/sqlstore/migrations.go
```diff
@@ -570,6 +570,45 @@ var migrations = []Migration{
...
 
+			return nil
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.12.0"),
+		toVersion:   semver.MustParse("0.13.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DATABASE_DRIVER_MYSQL {
+				if err := addColumnToMySQLTable(e, "IR_Incident", "CreationWebhookURL", "TEXT"); err != nil {
```
cpoile, line 581: 0/5 naming: should we prefix all webhooks with a standard, like `On` or `Webhook`, e.g.: `OnCreationWebhookURL` or `WebhookCreationURL`

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
agarciamontoro: Yup, that's true, but it was way easier to do it this way than to spin up a new server listening to this for the sake of testing. I'd vote for fixing it if/when it fails.

tests-e2e/cypress/integration/frontstage/incident_automation_spec.js
agarciamontoro: The test will fail. As before, I think we should fix this if the need arise.

cpoile (changes requested):
Minor suggestions, but requesting changes because I'd like to know what you decide. :)

Nice work @agarciamontoro!

server/incident/service.go
agarciamontoro: I did not know about those issues. Thanks for that!

server/api/api.yaml
agarciamontoro: I got this from https://swagger.io/docs/specification/callbacks/?sbsearch=callback%20SmartBear It displays literally like that in the rendered docs, so I'm not that sure it's correct. I'll take another look, I had forgotten I had added that.

server/sqlstore/migrations.go
agarciamontoro: I like the second option, it's true that it will be easier to parse it in the future. Thanks!

server/incident/service.go
agarciamontoro: What I meant with the code is that it will block until we get a response, even with a timeout. I was assuming there was a default timeout in http.Post, but I actually did not check. I'll take a look, along with the suggestion from Christopher to use an actual client.

server/api/api.yaml
agarciamontoro: It is actually part of the spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#runtimeExpression Sadly, redocly shows the literal string. I'd vote for leaving it as it is. I think it is understandable, and it is correct. We can also use a simple string such as "URL specified at creation_webhook_url".
What do you think?

server/sqlstore/migrations.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/503/commits/f8ec2f6612d7d58927d019e5d698f42e61e81457. I ended up prefixing everything by `Webhook`, then `OnCreation`, then whatever.

server/incident/service.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/503/commits/3307d570f4e333e0f818d8f74eb4b23c70dcd3df
I ended up doing something similar to https://github.com/mattermost/mattermost-server/blob/master/app/webhook.go#L150-L167, but with [the default transport](https://github.com/golang/go/blob/master/src/net/http/transport.go#L37-L53) instead of a custom one. I also thought about maintaining a client in the service, but I think that, for now, we are ok creating one.
Thoughts?

server/incident/service.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/503/commits/3307d570f4e333e0f818d8f74eb4b23c70dcd3df. Thanks!

server/incident/service.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/503/commits/3307d570f4e333e0f818d8f74eb4b23c70dcd3df. I specified a timeout of 30 seconds, which is [the default](https://github.com/mattermost/mattermost-server/blob/master/services/httpservice/client.go#L19) used by the `service/httpservice` in the server. Thoughts?

server/api/api.yaml
cpoile: Sounds good to me, just checking. :)

server/incident/service.go
cpoile: Awesome!

server/incident/service.go
justinegeffen: Thanks, @agarciamontoro! Are there remediation steps provided or is it a once-off message as-is? I wonder if something like "Check the logs for more information" would be helpful, and then elaborating on the log as to what exactly failed in terms of the POST request? Helpful error messages are few and far between and this is a good opportunity to provide helpful guidance. :)

justinegeffen: Something like: "Incident creation using an outgoing webhook was unsuccessful. Review the logs for more details."

^^ if that works, I can do another review for any other warnings/errors and we can set a standard for them going forward.

agarciamontoro: There would be more information in the logs, yes. Only available to admins, of course, while this message is readable by everyone on the incident's channel.

justinegeffen: Is there an associated error message? For example, we can say "Incident creation using the outgoing webhook was unsuccessful with error xxx. Contact your System Admin for more information." That way it's easier to diagnose and fix the issue?

justinegeffen: Sorry, error code, not error message. Like 200, 504, etc.

agarciamontoro: Mmm it can fail for other reasons we may not want to surface in the channel to everyone, so I'd err on the safe side and simply point to the logs. What do you think?

justinegeffen: Yep, I agree. How about: "Incident creation using the outgoing webhook was unsuccessful. Contact your System Admin for more information."

server/incident/service.go
```diff
@@ -225,6 +295,15 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
 		}
 	}
 
+	if incdnt.WebhookOnCreationURL != "" {
+		go func() {
+			if err = s.sendWebhookOnCreation(incdnt); err != nil {
+				s.pluginAPI.Log.Warn("failed to send a POST request to the creation webhook URL", "webhook URL", incdnt.WebhookOnCreationURL, "error", err)
+				_, _ = s.poster.PostMessage(channel.Id, "Incident creation using the outgoing webhook was unsuccessful. Contact your System Admin for more information.")
```
agarciamontoro, line 302: @justinegeffen To me, this sounds like the creation of the incident was unsuccessful. But the creation was successful. The only error occurred when sending the request to the configured webhook (i.e., kind of announcing the creation of the incident to the configured webhook), which does not prevent the incident from being created. Maybe I'm misunderstanding something?

justinegeffen: Ah, I see, I misunderstood! So something like:

"Incident creation announcement through the outgoing webhook failed. Contact your System Admin for more information."

For reference, the original was: "Failed to communicate the incident's creation through the outgoing webhook." which imo also works if my suggestion over-complicates things! :)

agarciamontoro: I think that works perfectly, thank you! I'll add it myself, I also need to change a test so it does not fail.

agarciamontoro:
@crspeller, ready for re-review!

server/incident/service.go
crspeller: @agarciamontoro Looks good, but optimally we should keep the client on the service. It keeps a connection pool for subsequent requests.

crspeller (changes requested):
Just the one issue with the client.

server/incident/service.go
agarciamontoro: Ok, let me move it to the incident service, thanks!

crspeller (approved):
Awesome!

agarciamontoro:
Skipping QA review because this feature is fairly isolated.

PR #504, opened 2021-03-26
MM-31564: backend support for inviting groups

lieut-data:
#### Summary
Extend the data model and REST API to support automatically inviting LDAP groups. No frontend logic is exposed at this time.

To test this in development, you'll need to be running a server with https://github.com/mattermost/mattermost-server/pull/17232 merged. First enable sign-in and sync with ldap (`MM_LDAPSETTINGS_ENABLE=true`, `MM_LDAPSETTINGS_ENABLESYNC=true`). The default `Makefile` scripts create a set of groups and users for authenticating with LDAP; leverage this next by logging in (and then out) of at least two LDAP users (`board.one` and `board.two`, using password `Password1`).

In the System Console under `/admin_console/user_management/groups`, link the `board` group:

image

then `Configure` that group to enable Group Mentions:

image

While you're in here, make sure the group shows the two users previously logged in:

image

Now create a playbook and look up its `id` for the team in question (here `agg39hruzjyrjpibo56ndsq93r`). I prefer using https://httpie.io/, but use any HTTP client as necessary:
```
> http :8065/plugins/com.mattermost.plugin-incident-management/api/v0/playbooks team_id==agg39hruzjyrjpibo56ndsq93r
HTTP/1.1 200 OK
Content-Length: 540
Content-Type: text/plain; charset=utf-8
Date: Fri, 26 Mar 2021 15:21:47 GMT

{
    "has_more": false,
    "items": [
        {
            "broadcast_channel_id": "",
            "checklists": [],
            "create_at": 1616605115659,
            "create_public_incident": false,
            "default_commander_enabled": false,
            "default_commander_id": "",
            "delete_at": 0,
            "description": "",
            "id": "4e67x71xr3rpix1wwos6sccszw",
            "invite_users_enabled": false,
            "invited_group_ids": [],
            "invited_user_ids": [],
            "member_ids": [],
            "num_stages": 1,
            "num_steps": 1,
            "reminder_message_template": "",
            "reminder_timer_default_seconds": 0,
            "team_id": "agg39hruzjyrjpibo56ndsq93r",
            "title": "Group Invite Updated"
        }
    ],
    "page_count": 1,
    "total_count": 1
}
```

Note that automation configuration does /not/ show up when querying playbooks in bulk, so fetch the specific playbook in detail before modifying any fields:
```
> http :8065/plugins/com.mattermost.plugin-incident-management/api/v0/playbooks/4e67x71xr3rpix1wwos6sccszw
HTTP/1.1 200 OK
Content-Length: 808
Content-Type: text/plain; charset=utf-8
Date: Fri, 26 Mar 2021 15:23:12 GMT

{
  "id": "4e67x71xr3rpix1wwos6sccszw",
  "title": "Group Invite Updated",
  "description": "",
  "team_id": "agg39hruzjyrjpibo56ndsq93r",
  "create_public_incident": false,
  "create_at": 1616605115659,
  "delete_at": 0,
  "num_stages": 1,
  "num_steps": 1,
  "checklists": [
    {
      "id": "",
      "title": "New stage",
      "items": [
        {
          "id": "",
          "title": "Test 2",
          "state": "",
          "state_modified": 0,
          "state_modified_post_id": "",
          "assignee_id": "",
          "assignee_modified": 0,
          "assignee_modified_post_id": "",
          "command": "",
          "command_last_run": 0,
          "description": ""
        }
      ]
    }
  ],
  "member_ids": [
    "1y119zymp3fh8gkejbtazm5wmc"
  ],
  "broadcast_channel_id": "",
  "reminder_message_template": "",
  "reminder_timer_default_seconds": 0,
  "invited_user_ids": [
    "1y119zymp3fh8gkejbtazm5wmc"
  ],
  "invited_group_ids": [],
  "invite_users_enabled": true,
  "default_commander_id": "",
  "default_commander_enabled": false
}
```

Note that the playbook above has `invite_users_enabled` enabled and even a user id to invite, but an empty set of `invited_group_ids`. Looking up the right group is perhaps easiest via SQL:
```
mattermost_test=# select * from usergroups;
             id             | name  | displayname | description | source |               remoteid               |   createat    |   updateat    | deleteat | allowreference
----------------------------+-------+-------------+-------------+--------+--------------------------------------+---------------+---------------+----------+----------------
 rgjnp9pmuj8ajrxfp6y1fc6sfa | board | board       |             | ldap   | df14c416-2107-103b-88b5-35b46ae25e10 | 1616604455558 | 1616771977335 |        0 | t
(1 row)
```

Now issue a PUT command to update the playbook. This can be done as a one-liner:
```
http :8065/plugins/com.mattermost.plugin-incident-management/api/v0/playbooks/4e67x71xr3rpix1wwos6sccszw team_id==agg39hruzjyrjpibo56ndsq93r | jq ' .invited_group_ids = ["rgjnp9pmuj8ajrxfp6y1fc6sfa"] | http PUT :8065/plugins/com.mattermost.plugin-incident-management/api/v0/playbooks/4e67x71xr3rpix1wwos6sccszw
```

Now, go and start an incident using this playbook and confirm that `board.one` and `board.two` are added automatically to the channel:
image

Edge cases to verify, most of which are covered by unit tests:
* It should not be possible to configure a group for which mentions are disabled, otherwise one could leak the contents of that group. (Enabling mentions for a group warns the sysadmin that the contents are effectively now public.)
* While the frontend doesn't support changing the groups right now (it will later!), it shouldn't clobber the group setting when a playbook is saved.
* It should be possible to create a playbook with a group automation via the API.
* Multiple groups, only some of which support mentions.
* Configuring a group should work even without https://github.com/mattermost/mattermost-server/pull/17232 merged, but an error will be logged if the plugin can't invoke the new Group APIs to enumerate the contents of the group. This should have no visible side effects, meaning we don't need to bump the minimum server version at this time.

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-31564


client/go.mod
```diff
@@ -4,9 +4,8 @@ go 1.15
 
 require (
 	github.com/google/go-querystring v1.0.0
-	github.com/mattermost/mattermost-server/v5 v5.32.0
+	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20210325162458-7e0f20de4cff
```
lieut-data, line 7: Pseudo-version likely until v5.35 ships, though I'll bump the commit hash here once merged.

server/incident/incident.go
```diff
@@ -61,6 +62,8 @@ func (i *Incident) Clone() *Incident {
 
 	newIncident.StatusPosts = append([]StatusPost(nil), i.StatusPosts...)
 	newIncident.TimelineEvents = append([]TimelineEvent(nil), i.TimelineEvents...)
+	newIncident.InvitedUserIDs = append([]string(nil), i.InvitedUserIDs...)
```
lieut-data, line 65: Found this mismatch during development.

server/sqlstore/support_for_test.go
```diff
@@ -20,7 +20,7 @@ var driverNames = []string{model.DATABASE_DRIVER_POSTGRES, model.DATABASE_DRIVER
 func setupTestDB(t testing.TB, driverName string) *sqlx.DB {
 	t.Helper()
 
-	sqlSettings := storetest.MakeSqlSettings(driverName)
+	sqlSettings := storetest.MakeSqlSettings(driverName, false)
```
lieut-data, line 23: This comes from the updated server module.

server/incident/incident.go
```diff
@@ -47,6 +47,7 @@ type Incident struct {
 	BroadcastChannelID      string               `json:"broadcast_channel_id"`
 	ReminderMessageTemplate string               `json:"reminder_message_template"`
 	InvitedUserIDs          []string             `json:"invited_user_ids"`
+	InvitedGroupIDs         []string             `json:"invited_group_ids"`
```
crspeller, line 50: It seems like a huge pain to have to embed group support everywhere like this. Is there a world where we make this one list and have functions that proxy the "Add user to channel" or other user actions and transparently translate group ids? That way we seamlessly add group support everywhere at once.

crspeller (changes requested):

server/incident/incident.go
lieut-data: In principle, yes, but there's nothing that guarantees a user and group won't share the same identifier. While that's unlikely to happen in real-life, I'd be hesitant to bake in that assumption just to simplify the code.

It's also useful to have discrete lists on the frontend: we already distinguish the "invited" from "non-invited" members, and it would be a pain to design a widget that can show all four combinations in one place. I also want to be able to add group-invitation specific instructions (i.e. "Only available with LDAP, v5.35, ...").

lieut-data:
Oops, should have kept reviewers consistent -- removing @cpoile and adding @agarciamontoro. (@cpoile, if you had feedback, I'm still all ears!)

lieut-data:
Also, @agarciamontoro, on starting with the frontend changes, I realized that the `enabled` flag was meant to apply to each "line item" in the automation screen, so it might make sense to have a discrete `invite_group_enabled` or some such. I may push that here after the initial review.

agarciamontoro (approved):
Looks good! Nothing blocking :tada:

I checked that everything was working and got to invite the two members to the incident!
image

server/incident/service.go
```diff
@@ -123,8 +123,41 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
...
+				s.pluginAPI.Log.Warn("failed to query group", "group_id", groupID, "err", err)
+				break
+			}
+			if len(users) == 0 {
+				break
+			}
+
+			for _, user := range users {
+				invitedUserIDs = append(invitedUserIDs, user.Id)
+			}
```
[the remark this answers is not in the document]
lieut-data: Ooh, really good catch actually!

server/incident/incident.go
crspeller: Alright. It's probably a useful distinction client side. Wish we had that build into groups somehow so we didn't have to do it here. If we had designed this feature for general availability we might have caught a design consideration like this. This is going to be a pain for every plugin author and API user, not just us. It means groups won't be supported very widely, which is a shame. However I think the ship has sailed on this.

crspeller (approved):

agarciamontoro:
> Also, @agarciamontoro, on starting with the frontend changes, I realized that the enabled flag was meant to apply to each "line item" in the automation screen, so it might make sense to have a discrete invite_group_enabled or some such. I may push that here after the initial review.

@lieut-data, I think there is no need to add a new invite_group_enabled. The current toggle, invite_users_enabled, has the state of the invite users feature in general. Adding groups through the UI should occur on the same member selector we already have in the backstage, so we should rely on the state of invite_users_enabled.

lieut-data:
@agarciamontoro, I'd love to be able to use the same member selector in the backstage, but I'm worried this going to be "hard". Besides the fact that the existing component is designed to consume profiles, not groups -- so we'd need to refactor to be more general -- it also has a usability problem of needing to search for and enumerate groups in addition to users all in a single scrolling interface.

I'm inclined to just fork the current selector and offer the ability to pick a set of groups discretely from a set of users. This goes directly to @crspeller's wish that groups were treated more like users directly by the system, so we wouldn't have to make the distinction at all.

Thoughts?

agarciamontoro:
> @agarciamontoro, I'd love to be able to use the same member selector in the backstage, but I'm worried this is going to be "hard". Besides the fact that the existing component is designed to consume profiles, not groups -- so we'd need to refactor to be more general -- it also has a usability problem of needing to search for and enumerate groups in addition to users all in a single scrolling interface.

That's true, although I think people usually search in the selectors by typing, not scrolling, so that does not concern me a lot from a UX perspective, as I believe a group mention should behave similar to a user mention in every possible aspect.  Maybe something for @alexsiclari to assess, but I think the first ideas we came up with at the beginning involved adding different "sections" in the selector, one for users and other for groups.

> I'm inclined to just fork the current selector and offer the ability to pick a set of groups discretely from a set of users.

Whilw the effort for doing this would be a bit lower than the one needed to add this to the existing component, I don't think it will save you a lot of time.

But, honouring our new mindset, probably better to go with the easiest path here.

prapti (approved):
Thank you so much for the detailed setup steps and edge test cases Jesse, that was very helpful!
Verified the main functionality and edge cases as described in the PR, as well as some additional ones.
Approving to merge!

PR #507, opened 2021-03-29
Retrieve only active users

agarciamontoro:
#### Summary
The default when searching users through `GET /users/search` is to return only the active ones, but we need to be specific when retrieving all users in a team via `GET /users`, as the default is to return active and inactive users.

It is actually a bit weird, as there are [two boolean query parameters, `active` and `inactive`](https://api.mattermost.com/#tag/users/paths/~1users/get), mutually exclusive, that both default to false. They actually mean "only active" and "only inactive", so both being false means "both active and inactive".

Anyway, the change just adds the `active: true` option to get only active users, and I had to fill the other precedent parameters with their default values.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33809


crspeller (approved):

cpoile (approved):

agarciamontoro:
@prapti, this can be easily verified with the same QA test steps you wrote on the ticket. Thank you!

prapti (approved):
LGTM; the original bug is fixed. Thanks Alejandro!

PR #509, opened 2021-04-01
stop testing build/sync

lieut-data:
#### Summary
We don't use this utility, but it takes upwards of 60s+ when running `make test`. Eliminate it.

#### Ticket Link
None.


cpoile (approved):
Yesss!

PR #512, opened 2021-04-06
MM-34604 Better stats colours

crspeller:
#### Summary
Fixes the theme colors in the stats view. Not totally the best, but better designs for the stats view are coming. At least this doesn't burn your retinas. 

#### Screenshots
image
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34604

### DIFF
--- a/webapp/src/components/backstage/stats.tsx
+++ b/webapp/src/components/backstage/stats.tsx
@@ -120,6 +120,7 @@ const GraphBoxContainer = styled.div`
 
 const GraphBox: FC<GraphBoxProps> = (props: GraphBoxProps) => {
     const style = getComputedStyle(document.body);
+    const centerChannelFontColor = style.getPropertyValue('--center-channel-color');
     return (
         <GraphBoxContainer>
             <Line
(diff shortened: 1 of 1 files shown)


webapp/src/components/backstage/stats.tsx
```diff
@@ -97,18 +128,24 @@ const GraphBox: FC<GraphBoxProps> = (props: GraphBoxProps) => {
...
                             },
                         }],
                         xAxes: [{
                             scaleLabel: {
                                 display: true,
                                 labelString: props.xlabel,
+                                fontColor: style.getPropertyValue('--center-channel-color'),
+                            },
+                            ticks: {
+                                fontColor: style.getPropertyValue('--center-channel-color'),
```
agarciamontoro, line 148: Non-blocking nit: maybe move the four callls to `getPropertyValue` into a single variable?

agarciamontoro (approved):
This is much much better, thank you!

webapp/src/components/backstage/stats.tsx
crspeller: Makes sense!

cpoile (approved):

PR #515, opened 2021-04-12
MM-34772: Fix broadcast status update

agarciamontoro:
#### Summary
We were using the previous status when broadcasting a status update. Which was obviously wrong.

I also adapted the specific E2E test, which was not changing the status itself, so the important part was left unchecked.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34772

### DIFF
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -526,7 +526,7 @@ func (s *ServiceImpl) RemoveTimelineEvent(incidentID, eventID string) error {
 	return nil
 }
 
-func (s *ServiceImpl) broadcastStatusUpdate(statusUpdate string, theIncident *Incident, authorID, originalPostID string) error {
+func (s *ServiceImpl) broadcastStatusUpdate(statusUpdate string, theIncident *Incident, newStatus, authorID, originalPostID string) error {
 	incidentChannel, err := s.pluginAPI.Channel.Get(theIncident.ChannelID)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


server/incident/service.go
```diff
@@ -526,7 +526,7 @@ func (s *ServiceImpl) RemoveTimelineEvent(incidentID, eventID string) error {
 	return nil
 }
 
-func (s *ServiceImpl) broadcastStatusUpdate(statusUpdate string, theIncident *Incident, authorID, originalPostID string) error {
+func (s *ServiceImpl) broadcastStatusUpdate(statusUpdate string, theIncident *Incident, newStatus, authorID, originalPostID string) error {
```
cpoile, line 529: If we're using the new `theIncident` here, shouldn't it have the new status as its current status already? (no need to pass in `newStatus` explicitly)

cpoile (approved):

server/incident/service.go
agarciamontoro: Mmmmmm maybe we should just update it in place, you're right

agarciamontoro: Oh my god and I had not updated the status itself. I cannot do three things at the same time, it shows :joy:

crspeller (approved):

cpoile (approved):
👍

agarciamontoro:
Skipping QA, the fix is straightforward and the test (if it passes) is already covering it.

PR #516, opened 2021-04-13
MM-34098, MM-34047, MM-34482, MM-34567 - New Incident backstage: overview

cpoile:
#### Summary
- These are mostly webapp changes, so not too complicated to review I think.  Like I mentioned before, I tried to keep the components as small and compartmentalized as possible.
- The design hasn't been finalized, so I'm using a kind of Whimsical style. That will change.
- The Metrics component is placeholder for now.
- The backend changes are mostly related to adding the "changes since last update" textarea to the `Update Status` dialog, adding the defaults to the playbook templates, and connecting the playbook templates to the incident instance.
- I'm using the description field as the "changes since last update", this way we don't have to add anything to the SQL, and we weren't using it anyway. Tao doesn't want us to do anything with the channel headers anymore.

#### Notes / TODOs
- ~~This won't build / pass CI until we figure out what to do about npm 7 in CI~~ Builds and passes.
- ~~Do not merge -- there will be at least one more PR adding to this. This will be a feature branch until then.~~ Now that it fixes e2e, we should merge this asap.

#### This changes quite a few parts of the flow:
image

image

image

image

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-34047 - In `Update Status` dialog, separate the Current Status and Changes Since Last Update
- https://mattermost.atlassian.net/browse/MM-34482 - Description (changes since last update) in the RHS and backstage overview
- https://mattermost.atlassian.net/browse/MM-34567 - Participants list in backstage overview
- https://mattermost.atlassian.net/browse/MM-34098 - Prototyping the backstage overview


webapp/src/components/backstage/backstage.tsx
```diff
@@ -191,10 +193,12 @@ const Backstage: FC = () => {
                         <PlaybookList/>
                     </Route>
                     <Route path={`${match.url}/incidents/:incidentId`}>
-                        <BackstageIncidentDetails/>
+                        <IncidentBackstage/>
+                        {/*<BackstageIncidentDetails/>*/}
```
cpoile, line 197: `BackstageIncidentDetails` will be removed in final cleanup PR. I need them for now for reference.

webapp/src/components/backstage/backstage.tsx
```diff
@@ -191,10 +193,12 @@ const Backstage: FC = () => {
...
                     </Route>
                     <Route path={`${match.url}/incidents`}>
                         <BackstageIncidentList/>
+                        {/*<Dashboard/>*/}
```
cpoile, line 201: this is coming in a future PR

webapp/src/components/backstage/incidents/incident_backstage/overview/participants.tsx
```diff
@@ -0,0 +1,76 @@
...
+function Participant(props: { userId: string }) {
+    const [showMessage, setShowMessage] = useState(false);
+
+    return (
+        <ParticipantRow
+            onMouseEnter={() => setShowMessage(true)}
+            onMouseLeave={() => setShowMessage(false)}
+        >
+            <ProfileWithPosition userId={props.userId}/>
+            {showMessage && <SecondaryButtonRight>{'Message'}</SecondaryButtonRight>}
```
cpoile, line 73: Todo

webapp/src/hooks.ts
```diff
@@ -132,3 +131,61 @@ export function useProfilesInChannel() {
...
+export function useProfilesInChannel(channelId: string) {
+    const dispatch = useDispatch() as DispatchFunc;
+    const profilesInChannel = useSelector((state) => getProfileSetForChannel(state as GlobalState, channelId));
+
+    useEffect(() => {
+        if (profilesInChannel.length > 0) {
+            return;
+        }
+
+        dispatch(getProfilesInChannel(channelId, 0, PROFILE_CHUNK_SIZE));
```
agarciamontoro, line 144: This gets only a few of the profiles, right?

webapp/src/selectors.ts
```diff
@@ -110,3 +118,26 @@ const findLastUpdated = (incident: Incident) => {
...
+    } else if (profileSet === PROFILE_SET_ALL) {
+        currentProfiles = Object.keys(profiles).map((key) => profiles[key]);
+    } else {
+        currentProfiles = Array.from(profileSet).map((p) => profiles[p]);
+    }
+
+    currentProfiles = currentProfiles.filter((profile) => Boolean(profile));
+
+    return currentProfiles.sort(sortByUsername);
+}
```
agarciamontoro, line 137: Not sure about what this function does. Maybe add a comment?

agarciamontoro (approved):
The code looks good! I left a handful of comments, all of them nits, so nothing blocking. I also found some commented code in the CSS, but assumed it was because the design is not finished.

I like the small components, and *love* the new hooks, thank you!

agarciamontoro (changes requested):
I had skipped over the non-themed CSS before, because I assumed you were consciously not doing that until the design is finished. But it looks like some change here also broke the dark mode in the whole backstage:
image

server/incident/service.go
```diff
@@ -635,6 +635,14 @@ func (s *ServiceImpl) UpdateStatus(incidentID, userID string, options StatusUpda
 		s.pluginAPI.Log.Warn("failed to broadcast the status update to channel", "ChannelID", incidentToModify.BroadcastChannelID)
 	}
 
+	// Update the incident's description (tl;dr summary/customer impact)
+	if options.Description != incidentToModify.Description {
+		incidentToModify.Description = options.Description
+		if err = s.store.UpdateIncident(incidentToModify); err != nil {
+			return errors.Wrap(err, "failed to update incident")
+		}
+	}
```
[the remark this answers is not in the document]
cpoile: Good idea, don't know why I didn't do that originally. Done!

webapp/src/components/backstage/incidents/incident_backstage/overview/updates.tsx
```diff
@@ -0,0 +1,52 @@
...
+    incident: Incident;
+}
+
+const Updates = (props: Props) => {
+    const statusPosts = props.incident.status_posts.sort((a, b) => b.create_at - a.create_at);
+
+    return (
+        <TabPageContainer>
+            <Title>{'Updates'}</Title>
+            {statusPosts?.map((sp) => (
```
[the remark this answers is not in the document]
cpoile: Thanks -- I've changed this around a bit too to handle the case where there's no status posts.

webapp/src/components/backstage/incidents/shared.tsx
```diff
@@ -0,0 +1,92 @@
...
+    flex: 1;
+    margin-left: 20px;
+    //border: 1px solid green;
+`;
+
+export const TabPageContainer = styled.div<{first?: boolean}>`
+    font-size: 12px;
+    font-weight: normal;
+    margin-top: ${(props) => (props.first ? 0 : '20px')};
+`;
```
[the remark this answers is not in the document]
cpoile: Ah, yes! But I should have removed all of that logic anyway. Fixed!

webapp/src/components/rhs/rhs_footer.tsx
```diff
@@ -2,18 +2,25 @@
...
+import {currentIncident} from 'src/selectors';
 
 interface Props {
     incident: Incident;
 }
 
 const RHSFooter: FC<Props> = (props: Props) => {
     const dispatch = useDispatch();
+    const currentTeam = useSelector<GlobalState, Team>(getCurrentTeam);
+    const incident = useSelector(currentIncident);
```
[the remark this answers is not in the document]
cpoile: Thanks good catch! (was for a convenience 'go to backstage' button that I removed before submitting the PR)

webapp/src/hooks.ts
cpoile: 200, which I thought would be enough. Should we get thousands just in case?

webapp/src/selectors.ts
cpoile: Ok, done -- it's a function used in mm-redux (but which isn't exported) whenever a list of user profiles is returned by a selector. Here we're making our own selector (getProfileSetForChannel) so we have to populate the profiles before returning them.

cpoile:
@agarciamontoro Thanks! Fixed the css. The rest of the css is still WIP, will be done in a design-pass PR. Here's what it looks like for now:
image
image
image

webapp/src/hooks.ts
agarciamontoro: 0/5 on the limit, just wondering whether we need to get *all* of them, regardless of the limit. Probably something to solve if there's a need to.

webapp/src/selectors.ts
agarciamontoro: Oh, I see, thank you!

agarciamontoro (approved):
:tada:

webapp/src/hooks.ts
cpoile: @crspeller let me know what you think -- I don't think there's a way to "get all" other than using a crazy-high chunk size.

webapp/src/components/backstage/incidents/incident_backstage/overview/updates.tsx
```diff
@@ -0,0 +1,62 @@
...
+`;
+
+interface Props {
+    incident: Incident;
+}
+
+const Updates = (props: Props) => {
+    const statusPosts = props.incident.status_posts.sort((a, b) => b.create_at - a.create_at);
+
+    let updates: JSX.Element | JSX.Element[] = <EmptyBody>{'There are no updates available.'}</EmptyBody>;
```
crspeller, line 28: Shouldn't typescript infer this type?

webapp/src/hooks.ts
crspeller: Probably make this clear that it doesn't get all the profiles in the channel. 
Just getting the first few really depends on context. If it's just to populate one of those profile selector boxes with search, then it's fine. But in other situations it might be unexpected. I think this just gets used in context where you don't really need every profile, right?

crspeller (approved):
Non blocking comments. Good work!

server/incident/incident.go
```diff
@@ -140,10 +140,12 @@ type UpdateOptions struct {
 }
 
 // StatusUpdateOptions encapsulates the fields that can be set when updating an incident's status
+// NOTE: changes made to this should be reflected in the client/incident StatusUpdateOptions struct
 type StatusUpdateOptions struct {
-	Status   string        `json:"status"`
-	Message  string        `json:"message"`
-	Reminder time.Duration `json:"reminder"`
+	Status      string        `json:"status"`
+	Description string        `json:"description"`
```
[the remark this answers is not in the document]
cpoile: Agreed, it's a bad name. It's only named `Description` to match the field we're using to store it in the incident struct/sql table. Otherwise I would have called it `CurrentSummary` or something like that. I didn't want to alter the table, so we're repurposing the `description` field.

webapp/src/components/backstage/incidents/incident_backstage/overview/updates.tsx
cpoile: No, it would infer a single element, but below we might need to store an array of elements.

server/incident/incident.go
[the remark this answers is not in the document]
crspeller: @cpoile What's wrong with just altering the table?

client/incidents.go
```diff
@@ -118,10 +118,11 @@ func (s *IncidentsService) Create(ctx context.Context, opts IncidentCreateOption
 	return incident, nil
 }
 
-func (s *IncidentsService) UpdateStatus(ctx context.Context, incidentID string, status Status, message string, reminderInSeconds int64) error {
+func (s *IncidentsService) UpdateStatus(ctx context.Context, incidentID string, status Status, description, message string, reminderInSeconds int64) error {
```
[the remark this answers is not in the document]
cpoile: From a previous review Jesse suggested we make the required fields be parameters, not an options struct -- that way we don't have to manually verify the required fields.

server/incident/incident.go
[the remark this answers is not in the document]
cpoile: @crspeller For speed. We originally wanted to get this out really quickly (before npm and circleCI delayed it), and we're not using the `description` field at all, so I just used it. I can make a new field if you really want, I'm only 1/5.

client/incidents.go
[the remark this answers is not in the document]
crspeller: You can still pass "" to a function parameter though? Maybe I don't understand.
I guess if forces a break in the client every time we update? Maybe that's desirable to remember to update?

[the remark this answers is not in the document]
cpoile: I think it's that, and it's like documentation. True, you could pass "", but then you're choosing to, and you will get the error from the server that you need to provide those. I guess it makes it more explicit, rather than just forgetting to fill in a field in the options (and not knowing which are required).

PR #519, opened 2021-04-17
circleci: store_test_results for Cypress

lieut-data:
#### Summary
Configure multiple reporters for Cypress, so that we can collect results programmatically in CircleCI. Conveniently, this branch is spurious failing right now:

image

image

#### Ticket Link
None.


cpoile (approved):
Tests will work once #516 is merged. 
What can view the xml report files? How will we use them?

lieut-data:
I was torn on the benefit of keeping the artifacts instead of just using the CircleCI interface to view them. At times, I've found it helpful to extract the results programmatically, but honestly I suspect we'll ignore them most of the time.

PR #520, opened 2021-04-17
MM-34718: fix incident welcome when filtering

lieut-data:
#### Summary
When the incident list loads, check for incidents without filtering to reliably decide when to show the welcome screen.

(`master` is failing spuriously right now, but the specific test in question passes locally for me.)

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-34718


agarciamontoro (approved):
Looks good!

Only thing I noticed while re-reading this code is that this screen will not update real-time, if someone else creates an incident. But that already happened before this PR, so nothing else to do here.

Thanks!

crspeller (approved):

prapti (approved):
The initial bug is fixed: the links in stats page open the incident list view with correct filters even when there aren't any incidents matching that filter.
Thanks Jesse!

PR #523, opened 2021-04-20
MM-34775 Playbook open by default

crspeller:
#### Summary
Modifies the playbook permissions to be in a tab. Creates a toggle to enable and disable playbook restrictions. Default to open. Admin privileges no longer apply to the restrictions. For now, works the same as it did before with only one level of permissions, everything or nothing. More granular permissions in the future. 

Draft only because this is based on https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/522

UI:
image

Default all access:
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34483


itao:
Please request review from @justinegeffen as well for copy to clarify the feature. For example, it's currently not clear that by not hitting the toggle, everyone in the team can access it.

Btw @abhijit-singh I wanted to tag you for heads up. This pattern will likely be used in a bunch of additional places like incident permission and becomes a wee bit more complex when we expand to workspace-scope as opposed to the current team-scope. So it's worth more design input in the near future when we cross that bridge.

itao:
Why is the `Permissions` tab placed between `Preferences` and `Automations`?

If I think out loud from the user perspective of "how do I get the most value with least amount of work", I'd order them as
- Automation
- Preferences (which will soon be merged as Automations that are on by default)
- Tasks
- Permissions

crspeller:
@justinegeffen Tagging you on this. See @itao's comment. Maybe just expand the help text to include an explanation? I was hoping it would be clear by the disabled state that there was no restrictions unless activated. Hence "Enable access restriction". Maybe there is a more clear name for the toggle too?

crspeller:
@itao I was actually debating if it should go:

- Tasks
- Permissions
- Preferences
- Automation
or 
- Tasks
- Preferences
- Permissions
- Automation

I was thinking the natural order would be the most important to least important. So the tasks obviously wins there. Then it's probably preferences or permissions, then automatons are completely optional 0/5. I am 2/5 on tasks first though. Landing on Automation would be strange. Especially since once you have those setup you are probably not changing them often.

server/sqlstore/playbook_test.go
```diff
@@ -822,7 +848,7 @@ func TestGetPlaybooksForTeam(t *testing.T) {
 				Items:      []playbook.Playbook{pb08},
 			},
 			expectedErr: nil,
-		},
+		},*/
```
agarciamontoro, line 851: Are all the commented tests not valid anymore?

agarciamontoro (changes requested):
Cool! Only concerned about the tests being removed.

I also think that disabling the setting to allow everyone to use the playbook is a bit confusing.

server/sqlstore/playbook_test.go
crspeller: Ya, those are testing if the Admin has their special powers, which this PR removes. One of the tests added makes sure they don't. 
I didn't remove them because I am a bit worried we will immediately backtrack on that.

abhijit-singh:
Suggesting some changes based on the UI recommended for Settings here (https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/522#pullrequestreview-642498335) and to make it clear what does it mean to 'not' restrict access.

Design link: https://www.figma.com/file/2EKxGlxlYFMkgXLQqv9cUz/Playbook-Permissions?node-id=6%3A998

01 04 - IM - New Playbook - Permissions

crspeller:
@abhijit-singh Awesome! Applied the tweaks here too.
image

webapp/src/components/backstage/share_playbook.tsx
```diff
@@ -1,28 +1,88 @@
...
 }
 
+const UserSelectorWrapper = styled.div`
+    margin-left: 24px;
+    width: 400px;
+    height: 40px;
+`;
+
 const SharePlaybook: FC<SharePlaybookProps> = (props: SharePlaybookProps) => {
+    const currentTeamName = useSelector<GlobalState, string>((state: GlobalState) => getCurrentTeam(state).name);
```
cpoile, line 31: I think when you use a function inside `useSelector` redux can't tell when it needs to refresh the component, so it does on every store state change.

cpoile (changes requested):

webapp/src/components/backstage/share_playbook.tsx
crspeller: Yup. Because the reference to the arrow function changes every time, we would have to re-run the selector when the redux store has not changed and see we got the same result. Modified to use a single function.

cpoile (approved):
Cool, thanks. The whole backstage is rerendering on any store change, so we need to fix something higher up in the hierarchy.

agarciamontoro (approved):
Looks good, thank you! :tada:

PR #524, opened 2021-04-20
MM-34568 - Design Polish on Incident Overview tab

cpoile:
#### Summary
- CSS polish
- removed unused components
- added an `Overview` button for convenience
- `Message` buttons work; the one for the commander is 'always on'

#### Looks like:

With narrow margins on Update boxes:
image

With the same 24px margins as Description:
image

image

image

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-34568


crspeller (approved):
Delicious and nutritious!

cpoile:
@abhijit-singh I made the assumption that the navigation bar keeps its 32px left/right margin no matter how wide the client, but the body has a max-width of 1140px (incl. a 20px padding, so a max effective 1120px of body). Let me know if you want me to change that. Thanks!

agarciamontoro (approved):
Much nicer! :tada:

One small thing I just noticed: the commander is both in the About and the Participants boxes. So, heretic idea: what if we get rid of the About box and only show the duration?

abhijit-singh (approved):
Looks good! Can you please verify a few visual nits though - 

- In the Overview page, the `font-weight` for the section headings (Description, Updates, About) should be `Semibold` or `600` and the `font-size` ideally be `18px`.
- The spacing inside each Update card seems tight. Can we make it the same as the Description card?
img-2

- In the RHS Summary view, there seems to be some extra space between the title `Description` and the content in it `Incident Update...`. Just wanted to make sure it is the same as for the `Recent Update` block.

cpoile:
@abhijit-singh Thanks! I've fixed the font weights (font size was already 18px) and the spacing.

itao (approved):
LGTM!

PR #525, opened 2021-04-21
MM-34665: Pricing differentiation plan: backend support

agarciamontoro:
#### Summary
This is the first of a series of PRs for supporting the pricing differentiation plan features. This one actually does not change the behaviour of the product at all.

It just adds a couple of new things, that are not yet used: a `GetNumPlaybooksForTeam` method to the playbook service and a new  `/playbooks/count` endpoint, as well as support for new websocket events and a store entry that maintain the number of playbooks created per team. These will support the limits for E0/E10 users in the coming PRs.

The only commit that changes something important is the first one, which moves the license-check middleware to every API subrouter. This commit also adds a "feature flag", which is a constant set to false for now, that controls the restriction of (for now) the playbook subrouter. As the feature flag is false, this subrouter is still restricted to E20 users.

There's also a minor change in the code of the backstage navbar to remove the Switch component, which was not needed.

I encourage you to review this PR commit by commit, I tried to make them as isolated and self-explanatory as possible.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34665


server/api/playbooks.go
```diff
@@ -163,6 +170,10 @@ func (h *PlaybookHandler) createPlaybook(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	h.poster.PublishWebsocketEventToTeam(playbookCreatedWSEvent, map[string]interface{}{
```
crspeller, line 173: These probably make more sense in the service like we have them for incidents.  We might also consider including the playbook itself like we do for incidents.

crspeller (approved):
Good stuff!

cpoile (approved):
Looks great!  Agree with Christophers review comment.

server/api/playbooks.go
agarciamontoro: @crspeller: moved to the service for consistency, thanks for catching that! Although we also have a couple of PublishWebsocket in the incidents API. We can change that later.

>  We might also consider including the playbook itself like we do for incidents.

I cannot do this here, because I'm broadcasting to the whole team (the restriction on the number of playbooks is scoped by team). If I include the playbook itself, everyone in a team would have info on potentially sensitive playbooks.

cc/ @cpoile

agarciamontoro:
Skipping QA, this PR does not change anything in the product, it's just isolated backend work.

PR #527, opened 2021-04-21
MM-33850 - Event for when participants joined/left the incident

cpoile:
#### Summary
- Added a timeline event `user_joined_left`, added the filter item, all the supporting backend work to track the user joined/left hooks.

#### Editor review
@justinegeffen @abhijit-singh 
- The timeline checkbox name is `Users joined/left`
- If they left/joined themselves, the timeline message is `@username joined the channel` `@username left the channel`
- If they were added/removed by someone else, the message is `@username  was added to channel by @username` `@username was removed from the channel by @username`
- I used the `at` icon, please suggest a more appropriate one. @abhijit-singh is there a contact sheet with all our icons? That would be so useful.

image

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-33850


agarciamontoro (approved):
Looks cool!

(I like the @ icon, by the way)

server/plugin.go
```diff
@@ -209,3 +209,19 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 
 	return &model.CommandResponse{}, nil
 }
+
+func (p *Plugin) UserHasJoinedChannel(c *plugin.Context, channelMember *model.ChannelMember, actor *model.User) {
```
crspeller, line 213: I think this works, even in HA. However it's a bit inefficient and might be error prone. I think it's fine for now. 
In the future, it might be worth considering using the channel member history table to derive this info.

crspeller (approved):

abhijit-singh (commented):
I think the user/account icon might work better if we're not using it for something else in the timeline already 
icon

And yes there's a sheet with all the icons we're currently using - [Icon Sheet](https://drive.google.com/open?id=1fEKMDa3hdaAunc7g8-inVKxH50PGYymO&authuser=michael.gamble%40mattermost.com&usp=drive_fs)

itao:
In the filter list, everything is a noun (eg. posts, updates, changes) except this one which is verb in past tense (joined/left). We should make it consistently noun, @justinegeffen welp?

justinegeffen:
Thanks, @itao. 

It's definitely possible to change it for consistency and have it read "User activity". This will possibly be best in terms of scalability if we want to add more user activity going forward?

itao:
Oo yea that could work well. I'd just recommend plural for consistency

itao:
Might not be important right now, just pointing out that in the future an incident could be have multiple channels and also this timeline will be accessed from retro where the viewer may not have context in which channel it's referring to. 

Might be worth putting in channel name and link as opposed to "the channel". 1/5 for this PR

justinegeffen:
@itao, by "plural" do you mean "User activities" or "Users activity"?

justinegeffen:
With regard to "@username was added to channel by @username @username was removed from the channel by @username", I agree with @itao. The channel name/link would provide much better context.

itao:
"User activities"

justinegeffen:
Could "User actions" work instead of "activities"? "Activities" isn't sitting well with me for some reason.

server/plugin.go
cpoile: yep, makes sense. thanks!

itao (approved):

cpoile:
@itao If we want to add information such as the channel name and a link to the channel, it would be easier to do that in the event summary (the line below the main title of the event). 
So it would be something like:
```
@user joined
@user was added to [Release: May 5th Cloud Release](link...) by @otheruser
```

Is that what we want to explore in this PR? Or should we wait until we have multiple channels and a need for that amount of detail?  Right now it might be unnecessary detail, considering there's only one incident channel.

itao:
Mmm that's a good idea, that info can go in the line below if additional context like someone else added them is needed. My suggestion would be this below so that the format is kept consistent as <actor> <action> <object>:
```
@user joined (or left) the channel
@otheruse added @user to [channel name](link)
```

itao:
I think it's worth doing now because we know we want to bring timeline to retro and that'll happen before we have support for multiple channels.

cpoile:
@itao Ok. I can make the title specific, we don't need the `(or left)`. The channel links will go dead (turn into plain text) if the channel is renamed, but that is consistent with how channel links work in MM chat. Here's what it will look like:

image

itao:
Haha I just realized how confusing that I put `(or left)` in verbatim text, I assumed it'll be how you showed in the screenshot. It looks a little redundant when someone joins on their own but I think it's OK because it'll be the minority case. LGTM

PR #528, opened 2021-04-22
switch back to Node 14 and NPM 6

lieut-data:
#### Summary
This PR switches the CI pipeline back to Node 14 and NPM 6 (basically, the defaults in the CircleCI golang image). I'm proposing that we stay in lockstep with the webapp to simplify developer setup.

In regards to my developer setup, I simply have the following (as installed by `brew` in my case, though `nvm` would work too):
```sh
> node --version
v14.16.1
> npm --version
6.14.12
```

And a `make clean && make` successfully installs everything with the updated `package-lock.json` file included in this PR.


lieut-data:
@cpoile, can you give this branch a whirl locally to see if this fixes the issues you were having previously?

agarciamontoro (approved):
Just tested, with your exact same versions for node and npm, and a `make clean && make` worked smoothly.

As a side note, in case someone had the same problem: I manage the versions in my machine with `nvm`, and doing `nvm use 14` set node to version 14 but npm to version 7, instead of 6. A simple reinstall with `nvm uninstall 14` and `nvm install 14` did set npm to version 6 when using node 14. I guess I had my environment messed up, but a reinstall always fixes things.

cpoile (approved):
Yep, works for me!

crspeller (approved):

PR #529, opened 2021-04-22
MM-35074: don't initialize args parameter

lieut-data:
#### Summary
This change complements (well, duplicates) the solution in https://github.com/mattermost/mattermost-webapp/pull/7969, but ships in the plugin to get it into the hands of customers relatively quickly. We're only planning on patching the webapp ESR right now.

I've based this pull request off of the yet-to-be-merged `node14` branch, since I needed to add a developer dependency.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35096


agarciamontoro (approved):
Does this fix things for our plugin, or for other plugins in case ours registers first? I am not sure I understand the issue here, actually.

lieut-data:
This only prevents our plugin from causing a problem for other plugins. It's conceivable that another plugin exists that causes this issue as well.

cpoile (approved):

PR #531, opened 2021-04-22
Temporarily hide the /stats page

lieut-data:
#### Summary
We're planning on putting the stats page behind an experimental feature flag, but no time for that in v1.9, so just effectively hide navigation to it for now.

#### Ticket Link
None.

### DIFF
--- a/webapp/src/components/backstage/backstage.tsx
+++ b/webapp/src/components/backstage/backstage.tsx
@@ -126,6 +126,8 @@ const Backstage: FC = () => {
         navigateToTeamPluginUrl(currentTeam.name, '/playbooks');
     };
 
+    const experimentalFeaturesEnabled = false;
+
     return (
         <BackstageContainer>
             <Switch>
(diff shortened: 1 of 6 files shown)


webapp/src/components/backstage/backstage.tsx
```diff
@@ -126,6 +126,8 @@ const Backstage: FC = () => {
         navigateToTeamPluginUrl(currentTeam.name, '/playbooks');
     };
 
+    const experimentalFeaturesEnabled = false;
```
lieut-data, line 129: "This was their finest hour..."

webapp/src/components/backstage/backstage.tsx
```diff
@@ -202,6 +206,12 @@ const Backstage: FC = () => {
                     <Route path={`${match.url}/stats`}>
                         <StatsView/>
                     </Route>
+                    <Route
```
lieut-data, line 209: This is mostly to support the simpler e2e tests.

crspeller (approved):

lieut-data:
Tests passing on master, but not on cloud, and don't appear to be related to these changes, so I'm calling this one "merge ready" ;P

cpoile (approved):

PR #532, opened 2021-04-26
Validate webhook URL only when enabled

agarciamontoro:
#### Summary
Make sure that the playbook creation is not rejected because of the webhook validation when that automation is disabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35161


crspeller (approved):

cpoile (approved):

prapti:
Tested and verified that the initial bug is fixed -- playbooks can now be saved once an invalid webhook is entered but the webhook setting is turned off 👍

PR #533, opened 2021-04-26
MM-34675 Improve UI for system console toggle.

crspeller:
#### Summary
Improved UI for system console toggle. Went ahead and used the updated UI instead of the one specified in the ticket. 

New UI:
image

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34675


webapp/src/system_console_enabled_teams.tsx
```diff
@@ -53,18 +54,53 @@ const TeamSelector: FC<TeamSelectorProps> = (props: TeamSelectorProps) => {
         }
     };
 
+    const radioPressed = (e: React.ChangeEvent<HTMLInputElement>) => {
+        if (e.target.value === 'enabled') {
+            setEnabled(true);
+        } else {
+            props.onTeamsSelected([]);
```
lieut-data, line 61: Does this result in forcibly clearing the selected teams? Seems slightly counterintuitive for how I expect dependent fields to behave: i.e. if I disable LDAP, I don't lose all my LDAP configuration.

webapp/src/system_console_enabled_teams.tsx
```diff
@@ -53,18 +54,53 @@ const TeamSelector: FC<TeamSelectorProps> = (props: TeamSelectorProps) => {
...
+                </RadioLabel>
+                <RadioLabel>
+                    <RadioInput
+                        type='radio'
+                        name='enabled'
+                        value='enabled'
+                        checked={enabled}
+                        onChange={radioPressed}
+                    />
+                    {'Incident Collaboration enabled on slected teams.'}
```
lieut-data, line 87:
```suggestion
                    {'Incident Collaboration enabled on selected teams.'}
```

lieut-data (approved):
A few comments, but don't need to block on a re-review from me.

lieut-data:
@agarciamontoro, we're considering this for cherry-picking to 1.9 -- can you prioritize review?

webapp/src/system_console_enabled_teams.tsx
```diff
@@ -20,6 +20,7 @@ type GetTeamType = (teamID: string) => Team
 const TeamSelector: FC<TeamSelectorProps> = (props: TeamSelectorProps) => {
     const selectableTeams = useSelector<GlobalState, Team[]>(getMyTeams);
     const getTeamFromID = useSelector<GlobalState, GetTeamType>((state) => (teamId) => getTeam(state, teamId) || {display_name: 'Unknown Team', id: teamId});
+    const [enabled, setEnabled] = useState(Boolean(props.teamsSlected) && props.teamsSlected.length !== 0);
```
lieut-data, line 23: Typo here as well: "teamSelected".

itao:
Can we change the toggle text to: `Enable for every team` and `Enable for selected teams only`. Repeating Incident Collaboration here feels redundant

agarciamontoro (approved):
Apart from Jesse's comments, it looks good!

webapp/src/system_console_enabled_teams.tsx
crspeller: Yup. Given the UX, I don't think it's worth the complexity to save the list of teams selected even when you disable it. I don't really see a use case for keeping the list of teams around.

lieut-data (approved):

lieut-data:
Huh, I didn't even realize `mattermod` would try in this repo! Guess I should have setup the release branch in advance :p

crspeller:
Nice try @mattermod 😆

PR #534, opened 2021-04-27
[MM-34647] Clearable fields in playbook automation

JoelRummel:
#### Summary

This ticket makes the Announcement Channel and the Assign Commander fields, located under Playbook Edit/Automation, clearable. Previously, there was no way through the user interface to empty these fields entirely, short of disabling the features (and even then, re-enabling them would simply bring back the previously entered value). This was accomplished by making the react-select components clearable, and by giving them a custom ClearIndicator icon (with styles derived from the clear-input-x icon given to search bars in the main web app).

Users can now empty these fields by clicking on the x icon, or by pressing backspace when the input element has focus.

MM-34647 also mentions placing the text cursor at the end of the input field when re-focusing on it. Regrettably, I could not find a solution for this short of replacing the react-select component entirely, as the underlying issue stems from react-select itself as described [here](https://github.com/JedWatson/react-select/issues/164).

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/17351
Jira ticket: https://mattermost.atlassian.net/browse/MM-34647


agarciamontoro:
Thank you very much for this, @JoelRummel, we'll take a look as soon as we can!

crspeller (approved):

lieut-data (approved):

agarciamontoro:
@lieut-data, do we need QA here?

lieut-data:
Yes, I'd love to get @prapti's eyes on this before we merge. Thanks for the poke!

prapti (approved):
This is great improvement! Thank you @JoelRummel!

@lieut-data One question after testing is that both fields -- Assign commander & Announce in another channel -- can be cleared and still be enabled in the playbook. This does not cause any negative impact really, but should this be handled more clearly to avoid any confusion?

lieut-data:
@prapti, I'll merge as-is and we can iterate on the specific of the UI to perhaps treat "empty" fields as disabled. Thanks!

PR #535, opened 2021-04-27
MM-35051 Experimental features flag

crspeller:
#### Summary
Adding experimental features flag for users to turn on and off in the system console.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35051

### DIFF
--- a/server/config/configuration.go
+++ b/server/config/configuration.go
@@ -18,6 +18,9 @@ type Configuration struct {
 	// PlaybookCreatorsUserIds is a list of users that can edit playbooks
 	PlaybookCreatorsUserIds []string
 
+	// PlaybookCreatorsUserIds is a list of users that can edit playbooks
+	EnableExperimentalFeatures bool
+
 	// ** The following are NOT stored on the server
 	// AdminUserIDs contains a list of user IDs that are allowed
 	// to administer plugin functions, even if not Mattermost sysadmins.
--- a/server/config/configuration.go
+++ b/server/config/configuration.go
@@ -18,7 +18,7 @@ type Configuration struct {
 	// PlaybookCreatorsUserIds is a list of users that can edit playbooks
 	PlaybookCreatorsUserIds []string
 
-	// PlaybookCreatorsUserIds is a list of users that can edit playbooks
+	// EnableExperimentalFeatures determines if experimental features are enabled.
 	EnableExperimentalFeatures bool
 
 	// ** The following are NOT stored on the server
--- a/webapp/src/browser_routing.ts
+++ b/webapp/src/browser_routing.ts
@@ -13,11 +13,7 @@ export const navigateToUrl = (urlPath: string) => {
 };
 
 export const teamPluginUrl = (teamName: string, urlPath: string) => {
-    let cleanPath = urlPath;
-    while (cleanPath.startsWith('/')) {
-        cleanPath = cleanPath.substr(1);
-    }
-    return `/${teamName}/${pluginId}/` + cleanPath;
+    return `/${teamName}/${pluginId}` + urlPath;
 };
 
 export const navigateToTeamPluginUrl = (teamName: string, urlPath: string) => {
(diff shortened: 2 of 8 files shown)


webapp/src/browser_routing.ts
```diff
@@ -13,11 +13,7 @@ export const navigateToUrl = (urlPath: string) => {
 };
 
 export const teamPluginUrl = (teamName: string, urlPath: string) => {
-    let cleanPath = urlPath;
-    while (cleanPath.startsWith('/')) {
-        cleanPath = cleanPath.substr(1);
-    }
-    return `/${teamName}/${pluginId}/` + cleanPath;
```
cpoile, line 20: Why removed?

webapp/src/hooks.ts
```diff
@@ -144,6 +144,10 @@ export function useCanCreatePlaybooks() {
     });
 }
 
+export function useExperimentalFeaturesEnabled() {
+    return useSelector<GlobalState, boolean>((state: GlobalState) => Boolean(globalSettings(state)?.enable_experimental_features));
```
cpoile, line 148: extract into a constant function (so it won't rerender on every global state change)?

cpoile (changes requested):

webapp/src/hooks.ts
crspeller: This is quite annoying. At it to the list of reasons why JS sucks and the reasons why redux sucks. 
Fixed the one above it too.

webapp/src/hooks.ts
```diff
@@ -134,14 +134,22 @@ export function useProfilesInCurrentChannel() {
...
 export function useCanCreatePlaybooks() {
-    return useSelector<GlobalState, boolean>((state: GlobalState) => {
-        const playbookCreators = globalSettings(state)?.playbook_creators_user_ids;
-        if (!playbookCreators || playbookCreators.length === 0) {
-            return true;
-        }
-        return playbookCreators.includes(getCurrentUserId(state));
-    });
+    return useSelector<GlobalState, boolean>(selectCanCreatePlaybooks);
+}
```
cpoile, line 147: Can this be removed now completely? (use the selector instead?)

cpoile (approved):

webapp/src/hooks.ts
crspeller: I like the renaming. It make it more clear. See: https://kyleshevlin.com/use-encapsulation

cpoile: Except that blog is about encapsulating useEffects, which makes sense. Above you're encapsulating a single selector in a hook, why bother?  Up to you, but seems to be adding lines of code for no reason to me.

server/config/configuration.go
```diff
@@ -18,6 +18,9 @@ type Configuration struct {
 	// PlaybookCreatorsUserIds is a list of users that can edit playbooks
 	PlaybookCreatorsUserIds []string
 
+	// PlaybookCreatorsUserIds is a list of users that can edit playbooks
```
lieut-data, line 21:
```suggestion
	// EnableExperimentalFeatures determines if experimental features are enabled.
```

webapp/src/hooks.ts
lieut-data: The pain point here seems to be the `<GlobalState, boolean>` -- how hard does TypeScript complain without this? (i.e. if you could write `useSelector(selectCanCreatePlaybooks)`, the hook doesn't seem justified.)

I'm guessing it's our use of redux-thunk that makes `GlobalState` typing a requirement, hence the otherwise unnecessary verbosity.

lieut-data (approved):
Looks good -- just a few comments below.

webapp/src/hooks.ts
crspeller: It actually can infer the type here. 
But I see your point of it kinda just looking like a selector. I modified how the function was written to be more, custom hook like. Makes the behavior way more obvious.
Thanks for the nudge!

crspeller: Feel free to debate this if you still don't like my solution. I am up for a refactor PR :)

PR #526, opened 2021-04-21
MM-32419 Auto send an outgoing webhook on incident archive

hakansa:
#### Summary

Implementation of automated outgoing webhook when an incident archived.

Many thanks to @agarciamontoro for exhaustive help and support.  I've read a lot the PR https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/503  when I was working on this feature.

#### Changes in the Backstage

A new "When an incident archived" Section and Webhook component added.

Screen Shot 2021-04-21 at 22 04 50

Screen Shot 2021-04-21 at 22 05 23

#### Changes in the DB
- New columns in `IR_Playbook`:
    * `WebhookOnArchiveURL`, which contains the, ehem, webhook URL. 
    * `WebhookOnArchiveEnabled`, that stores the on/off state of the setting.
- New column in IR_Incident:
    * `WebhookOnArchiveURL`, which stores a copy of the data in the playbook's column when the incident is created. If the playbook's `WebhookOnArchiveEnabled` contains false, the data is not copied, so there's no need to add `WebhookOnArchiveEnabled` column to the `IR_Incident` table.

#### Tests
I tested that the webhook is sent and that the payload is correct in `server/incident/service_test.go`.
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32419

PR #536, opened 2021-04-28
MM-32420 - Send a message when new incident member joins

cpoile:
#### Summary
- When a user joins an incident channel the message is sent 10 seconds after (if it's sent right away it doesn't work).
- Heheh, I didn't read the ticket details and I called it `MessageOnJoin` -- let me know if anyone has a problem with that. :) (I'm a bad contributor)

#### Todo
- update migration and fix conflicts when #526 is merged

#### UX review
- @abhijit-singh to review the automation tab style.  

When disabled:
image

When enabled and focused:
image

With more text than can be contained by the 300px max height:
image

Ephemeral message shown for everyone (even reporter/commander when they get shown the channel automatically after starting it):
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32420

### DIFF
--- a/server/sqlstore/incident.go
+++ b/server/sqlstore/incident.go
@@ -211,16 +211,16 @@ func (s *incidentStore) GetIncidents(requesterInfo permissions.RequesterInfo, op
 	}, nil
 }
 
-// CreateIncident creates a new incident.
+// CreateIncident creates a new incident. If newIncident has an ID, that ID will be used.
 func (s *incidentStore) CreateIncident(newIncident *incident.Incident) (out *incident.Incident, err error) {
 	if newIncident == nil {
 		return nil, errors.New("incident is nil")
 	}
-	if newIncident.ID != "" {
-		return nil, errors.New("ID should not be set")
-	}
 	incidentCopy := newIncident.Clone()
-	incidentCopy.ID = model.NewId()
+
+	if incidentCopy.ID == "" {
+		incidentCopy.ID = model.NewId()
+	}
 
 	rawIncident, err := toSQLIncident(*incidentCopy)
 	if err != nil {
--- a/server/sqlstore/incident.go
+++ b/server/sqlstore/incident.go
@@ -53,7 +53,7 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 			"i.CreateAt", "i.EndAt", "i.DeleteAt", "i.PostID", "i.PlaybookID", "i.ReporterUserID", "i.CurrentStatus",
 			"i.ChecklistsJSON", "COALESCE(i.ReminderPostID, '') ReminderPostID", "i.PreviousReminder", "i.BroadcastChannelID",
 			"COALESCE(ReminderMessageTemplate, '') ReminderMessageTemplate", "ConcatenatedInvitedUserIDs", "ConcatenatedInvitedGroupIDs", "DefaultCommanderID",
-			"AnnouncementChannelID", "WebhookOnCreationURL").
+			"AnnouncementChannelID", "WebhookOnCreationURL", "MessageOnJoin").
 		From("IR_Incident AS i").
 		Join("Channels AS c ON (c.Id = i.ChannelId)")
 
@@ -253,6 +253,7 @@ func (s *incidentStore) CreateIncident(newIncident *incident.Incident) (out *inc
 			"DefaultCommanderID":          rawIncident.DefaultCommanderID,
 			"AnnouncementChannelID":       rawIncident.AnnouncementChannelID,
 			"WebhookOnCreationURL":        rawIncident.WebhookOnCreationURL,
+			"MessageOnJoin":               rawIncident.MessageOnJoin,
 			// Preserved for backwards compatibility with v1.2
 			"ActiveStage":      0,
 			"ActiveStageTitle": "",
@@ -298,6 +299,7 @@ func (s *incidentStore) UpdateIncident(newIncident *incident.Incident) error {
 			"DefaultCommanderID":          rawIncident.DefaultCommanderID,
 			"AnnouncementChannelID":       rawIncident.AnnouncementChannelID,
 			"WebhookOnCreationURL":        rawIncident.WebhookOnCreationURL,
+			"MessageOnJoin":               rawIncident.MessageOnJoin,
 		}).
 		Where(sq.Eq{"ID": rawIncident.ID}))
 
--- a/server/sqlstore/incident.go
+++ b/server/sqlstore/incident.go
@@ -57,7 +57,7 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 			"i.CreateAt", "i.EndAt", "i.DeleteAt", "i.PostID", "i.PlaybookID", "i.ReporterUserID", "i.CurrentStatus",
 			"i.ChecklistsJSON", "COALESCE(i.ReminderPostID, '') ReminderPostID", "i.PreviousReminder", "i.BroadcastChannelID",
 			"COALESCE(ReminderMessageTemplate, '') ReminderMessageTemplate", "ConcatenatedInvitedUserIDs", "ConcatenatedInvitedGroupIDs", "DefaultCommanderID",
-			"AnnouncementChannelID", "WebhookOnCreationURL", "MessageOnJoin").
+			"AnnouncementChannelID", "WebhookOnCreationURL", "Retrospective", "MessageOnJoin").
 		From("IR_Incident AS i").
 		Join("Channels AS c ON (c.Id = i.ChannelId)")
 
@@ -257,6 +257,7 @@ func (s *incidentStore) CreateIncident(newIncident *incident.Incident) (out *inc
 			"DefaultCommanderID":          rawIncident.DefaultCommanderID,
 			"AnnouncementChannelID":       rawIncident.AnnouncementChannelID,
 			"WebhookOnCreationURL":        rawIncident.WebhookOnCreationURL,
+			"Retrospective":               rawIncident.Retrospective,
 			"MessageOnJoin":               rawIncident.MessageOnJoin,
 			// Preserved for backwards compatibility with v1.2
 			"ActiveStage":      0,
@@ -303,6 +304,7 @@ func (s *incidentStore) UpdateIncident(newIncident *incident.Incident) error {
 			"DefaultCommanderID":          rawIncident.DefaultCommanderID,
 			"AnnouncementChannelID":       rawIncident.AnnouncementChannelID,
 			"WebhookOnCreationURL":        rawIncident.WebhookOnCreationURL,
+			"Retrospective":               rawIncident.Retrospective,
 			"MessageOnJoin":               rawIncident.MessageOnJoin,
 		}).
 		Where(sq.Eq{"ID": rawIncident.ID}))
--- a/server/sqlstore/incident.go
+++ b/server/sqlstore/incident.go
@@ -604,6 +604,39 @@ func (s *incidentStore) ChangeCreationDate(incidentID string, creationTimestamp 
 	return nil
 }
 
+// HasViewed returns true if userID has viewed channelID
+func (s *incidentStore) HasViewedChannel(userID, channelID string) bool {
+	query := s.queryBuilder.
+		Select("COUNT(*)").
+		From("IR_ViewedChannel vc").
+		Where(sq.And{sq.Eq{"vc.ChannelID": channelID}, sq.Eq{"vc.UserID": userID}})
+
+	var count int
+	err := s.store.getBuilder(s.store.db, &count, query)
+	if err != nil {
+		return false
+	}
+
+	return count > 0
+}
+
+// SetViewed records that userID has viewed channelID. NOTE: does not check if there is already a
+// record of that userID/channelID (i.e., will create duplicate rows)
+func (s *incidentStore) SetViewedChannel(userID, channelID string) error {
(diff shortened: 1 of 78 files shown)


cpoile:
requirements changed, needs more work

cpoile:
@crspeller @agarciamontoro Ready for review. Still waiting on final copy for the default welcome message from @justinegeffen.

webapp/src/types/playbook.ts
```diff
@@ -165,3 +171,10 @@ export function isChecklistItem(arg: any): arg is ChecklistItem {
         typeof arg.command === 'string' &&
         typeof arg.command_last_run === 'number';
 }
+
+export const defaultMessageOnJoin = 'Welcome. This channel was automatically created by an Incident Collaboration playbook. To view information about this incident, such as the commander\'s name, and list of tasks, select the shield icon in the channel header. You can also use the `/incident info` slash command.\n' +
```
justinegeffen, line 175:
```suggestion
export const defaultMessageOnJoin = 'Welcome. This channel was automatically created by an Incident Collaboration playbook. To view information about this incident, such as the commander\'s name and list of tasks, select the shield icon in the channel header. You can also use the `/incident info` slash command.\n' +
```

justinegeffen (commented):
Just a minor suggestion. But I have a question: Where is the help text that leads into this as an example? Similar to the content we provide for the incident status updates.

cpoile:
@justinegeffen Yep, we can add that, but then that will actually be in the welcome message (if they choose to keep it as the default). Is that fine?

justinegeffen:
Thanks for clarifying! No that definitely won't work. :) Let's leave it as it is for now then?

justinegeffen:
@cpoile, re-reading that default message, as this is what is displayed when someone joins, should we make those URLs valid and provide a Mattermost docs/IC channel link instead of a placeholder?

cpoile:
@justinegeffen Yes please!

webapp/src/types/playbook.ts
```diff
@@ -165,3 +171,10 @@ export function isChecklistItem(arg: any): arg is ChecklistItem {
         typeof arg.command === 'string' &&
         typeof arg.command_last_run === 'number';
 }
+
+export const defaultMessageOnJoin = 'Welcome. This channel was automatically created by an Incident Collaboration playbook. To view information about this incident, such as the commander\'s name and list of tasks, select the shield icon in the channel header. You can also use the `/incident info` slash command.\n' +
+    '\n' +
+    'You may find the following resources helpful:\n' +
+    '\n' +
+    '`[Link to list of useful contacts](your-url-here)`\n' +
```
justinegeffen, line 179:
```suggestion
    '`[Mattermost Incident Collaboration channel](https://community.mattermost.com/core/channels/ee-incident-response)`\n' +
```

webapp/src/types/playbook.ts
```diff
@@ -165,3 +171,10 @@ export function isChecklistItem(arg: any): arg is ChecklistItem {
...
+    '`[Link to list of processes](your-url-here)`';
```
justinegeffen, line 180:
```suggestion
    '`[Incident Collaboration documentation](https://docs.mattermost.com/administration/devops-command-center.html)`';
```

justinegeffen:
@cpoile, done. I added them as comments for your approval and just in case I messed up any formatting inadvertently. :)

server/sqlstore/migrations.go
```diff
@@ -639,6 +639,72 @@ var migrations = []Migration{
...
+				if _, err := e.Exec("UPDATE IR_Incident SET MessageOnJoin = '' WHERE MessageOnJoin IS NULL"); err != nil {
+					return errors.Wrapf(err, "failed setting default value in column MessageOnJoin of table IR_Incident")
+				}
+
+				if _, err := e.Exec(`
+					CREATE TABLE IF NOT EXISTS IR_ViewedChannel
+					(
+						ChannelID VARCHAR(26) NOT NULL,
+						UserID    VARCHAR(26) NOT NULL,
+						INDEX     IR_ViewedChannel_ChannelID (ChannelID)
```
crspeller, line 675: Maybe use a combined index on ChannelID and UserID since that's the main query?

webapp/src/websocket_events.ts
```diff
@@ -191,3 +194,22 @@ export const handleWebsocketChannelUpdated = (getState: GetStateFunc, dispatch:
         }
     };
 };
+
+export const handleWebsocketChannelViewed = (getState: GetStateFunc, dispatch: Dispatch) => {
```
crspeller, line 198: Does it maybe make more sense to have this logic in a redux store listener on channel change rather than listening for a websocket event? Given this behavior can be driven by the client. I am thinking this approach maybe makes less intuitive sense and adds latency.

crspeller (changes requested):

justinegeffen (approved):

webapp/src/websocket_events.ts
cpoile: @crspeller Having a bit of a delay is a good thing, in this case -- you want to make sure the user is seeing the channel and /then/ the ephemeral post is posted. But I'm fine with listening to the store, too. Do you think there would be any downsides to using the `currentChannelId` to detect a change?

server/sqlstore/incident.go
```diff
@@ -602,6 +604,39 @@ func (s *incidentStore) ChangeCreationDate(incidentID string, creationTimestamp
...
+	err := s.store.getBuilder(s.store.db, &count, query)
+	if err != nil {
+		return false
+	}
+
+	return count > 0
+}
+
+// SetViewed records that userID has viewed channelID. NOTE: does not check if there is already a
+// record of that userID/channelID (i.e., will create duplicate rows)
```
lieut-data, line 624: Can we avoid the duplicate rows? The query to check existence can be much faster if we assume uniqueness.

server/sqlstore/incident.go
```diff
@@ -602,6 +604,39 @@ func (s *incidentStore) ChangeCreationDate(incidentID string, creationTimestamp
 	return nil
 }
 
+// HasViewed returns true if userID has viewed channelID
+func (s *incidentStore) HasViewedChannel(userID, channelID string) bool {
+	query := s.queryBuilder.
+		Select("COUNT(*)").
```
lieut-data, line 610: Propose using `EXISTS` and a `UNIQUE` index that constrains this to only ever one row per channel and user.

server/sqlstore/migrations.go
lieut-data: I think the index will basically be the same size as the table, but in creating it, you trade storage space for the knowledge that the database intends to query just a single row. See comments above, but otherwise agree with @crspeller on using a combined index.

webapp/src/websocket_events.ts
crspeller: I think the current channel ID in redux should work.

server/sqlstore/incident.go
cpoile: Sounds good, done.

webapp/src/websocket_events.ts
cpoile: @crspeller Now I'm not so sure I want to worry about the complications of listening to the redux store (#552). I think the websocket event is simpler.

lieut-data: I LOVE the idea of a redux store listener... mumble, mumble, saga ... but defer to what's pragmatic here :)

server/sqlstore/incident.go
```diff
@@ -606,23 +606,28 @@ func (s *incidentStore) ChangeCreationDate(incidentID string, creationTimestamp
...
 func (s *incidentStore) HasViewedChannel(userID, channelID string) bool {
-	query := s.queryBuilder.
-		Select("COUNT(*)").
-		From("IR_ViewedChannel vc").
-		Where(sq.And{sq.Eq{"vc.ChannelID": channelID}, sq.Eq{"vc.UserID": userID}})
-
-	var count int
-	err := s.store.getBuilder(s.store.db, &count, query)
+	query := sq.Expr(
+		`SELECT EXISTS(SELECT *
```
lieut-data, line 610: Reading this now, I realize it might be easier to just `SELECT TRUE FROM ... ` to get the same result without the indentation. Probably need a pointer for `exists`, though. 0/5.

server/sqlstore/incident.go
```diff
@@ -606,23 +606,28 @@ func (s *incidentStore) ChangeCreationDate(incidentID string, creationTimestamp
...
 
-	return count > 0
+	return exists
 }
 
-// SetViewed records that userID has viewed channelID. NOTE: does not check if there is already a
-// record of that userID/channelID (i.e., will create duplicate rows)
+// SetViewed records that userID has viewed channelID.
 func (s *incidentStore) SetViewedChannel(userID, channelID string) error {
+	if s.HasViewedChannel(userID, channelID) {
```
lieut-data, line 627: nit: This will catch many of the duplicate inserts, but not all of them. Another option is to `Insert` with a condition using the same query as `HasViewedChannel`. But even that won't guarantee anything, and one usually has to explicitly handle the unique constraint exception. 0/5 here, just commenting.

lieut-data (approved):
Nice! A few nits, but nothing blocking. (Note that I just reviewed the delta since my comments, not the whole review. If you want a more holistic look, let me know!)

server/sqlstore/incident.go
cpoile: @lieut-data Could you clarify why it wouldn't catch duplicate inserts? (Other than the obvious race condition possibility.) But regardless, we don't even technically need that check (or a conditional statement) -- a duplicate entry should return an error. (Correct me if I'm wrong, but that's what I've seen from the tests.)
I've made a commit so that returning an error (for a duplicate entry) is handled correctly.

server/sqlstore/incident.go
cpoile: I just took this pattern from what a docs page suggested, but since you know more about this I'll defer to your expertise. Since you're 0/5 on it I'll assume you don't mind either way. :) So I'll keep it as it is.

server/sqlstore/incident.go
```diff
@@ -636,6 +636,15 @@ func (s *incidentStore) SetViewedChannel(userID, channelID string) error {
 		}))
 
 	if err != nil {
+		duplicateMsg := "duplicate key value"
+		if s.store.db.DriverName() == model.DATABASE_DRIVER_MYSQL {
+			duplicateMsg = "Duplicate entry"
+		}
+
+		if strings.Contains(err.Error(), duplicateMsg) {
```
lieut-data, line 644: Can we do this by checking the error code in question, vs. string matching? For example, on MySQL see https://stackoverflow.com/questions/38493985/how-to-parse-mysql-error-golang. I think this is factored out as a helper function on the server code.

cpoile: Good idea, done!

webapp/src/types/playbook.ts
```diff
@@ -142,7 +146,9 @@ export function isPlaybook(arg: any): arg is Playbook {
         typeof arg.announcement_channel_id === 'string' &&
         typeof arg.announcement_channel_enabled === 'boolean' &&
         typeof arg.webhook_on_creation_url === 'string' &&
-        typeof arg.webhook_on_creation_enabled === 'boolean';
+        typeof arg.webhook_on_creation_enabled === 'boolean' &&
+        typeof arg.message_on_join === 'string' &&
+        typeof arg.messageOnJoinEnabled === 'boolean';
```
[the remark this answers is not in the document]
cpoile: Yes! Good catch!

webapp/src/types/actions.ts
```diff
@@ -25,6 +25,7 @@ export const SET_RHS_STATE = pluginId + '_set_rhs_state';
 export const SET_RHS_TAB_STATE = pluginId + '_set_rhs_tab_state';
 export const SET_RHS_EVENTS_FILTER = pluginId + '_set_rhs_events_filter';
 export const RECEIVED_GLOBAL_SETTINGS = pluginId + '_received_global_settings';
+export const SET_HAS_VIEWED_Channel = pluginId + '_set_has_viewed';
```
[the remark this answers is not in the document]
cpoile: Nice catch (again)!

webapp/src/types/actions.ts
```diff
@@ -108,3 +109,9 @@ export interface ReceivedGlobalSettings {
     type: typeof RECEIVED_GLOBAL_SETTINGS;
     settings: GlobalSettings;
 }
+
+export interface SetHasViewedChannel {
+    type: typeof SET_HAS_VIEWED_Channel;
+    channelId: string;
+    hasViewed: boolean;
```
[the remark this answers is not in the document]
cpoile: I just wanted to keep it flexible. I'm 0/5 if you're stronger I'll change it.

cpoile:
@agarciamontoro 
> P.S; You're a good contributor, I like the naming you came up with better than the one I suggested.

😂

webapp/src/reducer.ts
```diff
@@ -227,6 +233,18 @@ const globalSettings = (state: GlobalSettings | null = null, action: ReceivedGlo
     }
 };
 
+const hasViewedByChannel = (state: Record<string, boolean> = {}, action: SetHasViewedChannel) => {
```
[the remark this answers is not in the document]
agarciamontoro: Oh no :( Thanks for the info!

webapp/src/types/actions.ts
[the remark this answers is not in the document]
agarciamontoro: Up to you completely! It's just that I cannot imagine a situation where we want that boolean to be false, but if you do, I'm ok with it.

agarciamontoro (approved):
Thank you! :tada:

PR #538, opened 2021-04-29
MM-35296 - Fix description bugs

cpoile:
#### Summary
- We don't ask for the description when starting an incident, it is inherited from the playbook. So:
  - Don't put the description in the new incident broadcast message
  - Don't put the description in the channel header

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-35296

### DIFF
--- a/server/incident/incident.go
+++ b/server/incident/incident.go
@@ -265,7 +265,7 @@ type Service interface {
 	GetIncidents(requesterInfo permissions.RequesterInfo, options FilterOptions) (*GetIncidentsResults, error)
 
 	// CreateIncident creates a new incident. userID is the user who initiated the CreateIncident.
-	CreateIncident(incdnt *Incident, userID string, public bool) (*Incident, error)
+	CreateIncident(incdnt *Incident, playbook *playbook.Playbook, userID string, public bool) (*Incident, error)
 
 	// OpenCreateIncidentDialog opens an interactive dialog to start a new incident.
 	OpenCreateIncidentDialog(teamID, commanderID, triggerID, postID, clientID string, playbooks []playbook.Playbook, isMobileApp bool) error
@@ -373,7 +373,7 @@ type Store interface {
 	// GetIncidents returns filtered incidents and the total count before paging.
 	GetIncidents(requesterInfo permissions.RequesterInfo, options FilterOptions) (*GetIncidentsResults, error)
 
-	// CreateIncident creates a new incident.
+	// CreateIncident creates a new incident. If newIncident has an ID, that ID will be used.
 	CreateIncident(incdnt *Incident) (*Incident, error)
 
 	// UpdateIncident updates an incident.
--- a/server/incident/incident.go
+++ b/server/incident/incident.go
@@ -373,7 +373,7 @@ type Store interface {
 	// GetIncidents returns filtered incidents and the total count before paging.
 	GetIncidents(requesterInfo permissions.RequesterInfo, options FilterOptions) (*GetIncidentsResults, error)
 
-	// CreateIncident creates a new incident. If newIncident has an ID, that ID will be used.
+	// CreateIncident creates a new incident. If incdnt has an ID, that ID will be used.
 	CreateIncident(incdnt *Incident) (*Incident, error)
 
 	// UpdateIncident updates an incident.
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -108,9 +108,6 @@ func (s *ServiceImpl) broadcastIncidentCreation(theIncident *Incident, commander
 
 	announcementMsg := fmt.Sprintf("#### New Incident: ~%s\n", incidentChannel.Name)
 	announcementMsg += fmt.Sprintf("**Commander**: @%s\n", commander.Username)
-	if theIncident.Description != "" {
-		announcementMsg += fmt.Sprintf("**Description**: %s\n", theIncident.Description)
-	}
 
 	if _, err := s.poster.PostMessage(theIncident.AnnouncementChannelID, announcementMsg); err != nil {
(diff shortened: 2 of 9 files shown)


server/incident/service.go
```diff
@@ -1332,7 +1323,7 @@ func (s *ServiceImpl) createIncidentChannel(incdnt *Incident, public bool) (*mod
 		Type:        channelType,
 		DisplayName: incdnt.Name,
 		Name:        cleanChannelName(incdnt.Name),
-		Header:      channelHeader,
+		Header:      "The channel was created by the Incident Collaboration plugin.",
```
crspeller, line 1326: Do we really want to set that as the header? Kinda seems like noise @itao ?

crspeller (approved):

agarciamontoro (approved):
Makes sense. Thanks!

server/incident/service.go
agarciamontoro: I guess that's just the previous behaviour, right?

cpoile: Previous behavior. @itao can you decide if you want to keep this behavior?

itao: I think there is benefit to providing some context here, although a link to the playbook and overview would be more helpful. I'd recommend something like:

`This channel was created as part of the [<insert_playbook_name>][#]. See [the overview page](#) for more info.`

justinegeffen: If we can use the fields that @itao suggested then I'd go with that. If not, something like:

This is an incident channel. To view more information, select the shield icon then select **Tasks** or **Overview**.

cpoile: @justinegeffen Yep, we can insert those fields because it's built in. In that case, I'll assume that text is good to go (correct me if I'm wrong). Thanks!

justinegeffen: Yep - I am happy with: This channel was created as part of the [<insert_playbook_name>](#) playbook. Visit [the overview page](#) for more information.

Thank you!

justinegeffen (approved):

cpoile:
@crspeller @agarciamontoro Probably needs a re-review. Not too much new code, but it's significant -- we need to have the incident ID before creating the incident in the DB, so required some changes.

crspeller (approved):
That escalated quickly. 😆

server/incident/incident.go
```diff
@@ -373,7 +373,7 @@ type Store interface {
 	// GetIncidents returns filtered incidents and the total count before paging.
 	GetIncidents(requesterInfo permissions.RequesterInfo, options FilterOptions) (*GetIncidentsResults, error)
 
-	// CreateIncident creates a new incident.
+	// CreateIncident creates a new incident. If newIncident has an ID, that ID will be used.
```
agarciamontoro, line 376: What's `newIncident` here? Is it the `incdnt` argument?

server/incident/service.go
```diff
@@ -195,9 +192,30 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
...
+	}
+
+	siteURL := model.SERVICE_SETTINGS_DEFAULT_SITE_URL
+	if s.pluginAPI.Configuration.GetConfig().ServiceSettings.SiteURL != nil {
+		siteURL = *s.pluginAPI.Configuration.GetConfig().ServiceSettings.SiteURL
+	}
+	overviewURL := ""
+	playbookURL := ""
+
+	header := "This is an incident channel. To view more information, select the shield icon then select *Tasks* or *Overview*."
```
agarciamontoro, line 209: Worth a quick re-review by Justine?

server/incident/service.go
```diff
@@ -195,9 +192,30 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
...
+	if siteURL != "" && pb != nil {
+		overviewURL = fmt.Sprintf("%s/%s/%s/incidents/%s", siteURL, team.Name, s.configService.GetManifest().Id, incdnt.ID)
+		playbookURL = fmt.Sprintf("%s/%s/%s/playbooks/%s", siteURL, team.Name, s.configService.GetManifest().Id, pb.ID)
+		header = fmt.Sprintf("This channel was created as part of the [%s](%s) playbook. Visit [the overview page](%s) for more information.",
```
agarciamontoro, line 213: Cool, this is actually useful info. Again, maybe a quick re-review by Justine?

server/incident/service.go
```diff
@@ -195,9 +192,30 @@ func (s *ServiceImpl) CreateIncident(incdnt *Incident, userID string, public boo
 	}
 
 	incdnt.ReporterUserID = userID
+	incdnt.ID = model.NewId()
```
agarciamontoro, line 195: Should we fail here if `incdnt` has an ID? 0/5. Probably not, but that's the behaviour we had before in the store, so maybe it was actually needed? If not, please ignore this comment!

agarciamontoro (approved):
This actually adds really useful info to the header, thanks! Non-blocking comments, just some comments typos and minor suggestions, so approving :)

server/incident/incident.go
cpoile: Yes, good eye, fixed!

server/incident/service.go
cpoile: This is her text (from discussion above)

server/incident/service.go
cpoile: I'm going to assume no -- until we need to handle the case where the caller wants/needs a specific ID, we can assume it doesn't matter whether we give the ID earlier (here) or later (in the store). Just for simplicity. Sound good @agarciamontoro ?

PR #539, opened 2021-04-29
MM-35291 - Fix incident backstage rendering

cpoile:
#### Summary
- Fixed the background rendering (good eye!)
- Discovered that the playbook edit screen is doing something very cool: the navbar (with the save and back buttons) scrolls then sticks to the top, hiding the Incident/Playbooks/Settings, like so:

image
image

- So changing the Incidents page to do the same:

image
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-35291


agarciamontoro (approved):
:tada:

PR #541, opened 2021-04-30
Fix permissions server-side

cpoile:
#### Summary
- Found that we had a permissions function that wasn't doing what it said it was doing.
- It was named incorrectly. Luckily, it was only being used incorrectly in one place. But to avoid this mistake in the future I removed it and replaced all its calls with the correct permissions function.

#### Ticket Link
- none


agarciamontoro (approved):
Oh god. Great catch!

cpoile:
@crspeller ready (and fixes e2e tests for all the other branches)

crspeller (approved):

PR #542, opened 2021-05-03
MM32419: Auto send outgoing webhook on status update

madhavhugar:
#### Summary

See [issue](https://github.com/mattermost/mattermost-server/issues/17317) for details

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-server/issues/17310

  JIRA: https://mattermost.atlassian.net/browse/MM-32419

Screenshot 2021-05-03 at 20 53 11


agarciamontoro:
Thank you for this, @madhavhugar, we'll take a look as soon as we can!

madhavhugar:
@agarciamontoro looks like the e2e tests are failing because of lack of permissions, anything that can be done from my end to fix that?

agarciamontoro:
That's an error I'd never seen before, weird. I tried to re-run it manually, but no luck. I think it may be for some changes we are doing in the pipelines infrastructure. I'll get back to you when I know something else!

agarciamontoro:
It looks like we were using an old docker image that was removed right after you opened the PR, but that's already fixed on the main branch, so just merging master should do the trick. Thanks for the heads-up!

crspeller (approved):
Looking good!

server/api/playbooks.go
```diff
@@ -91,6 +91,20 @@ func (h *PlaybookHandler) createPlaybook(w http.ResponseWriter, r *http.Request)
 		}
 	}
 
+	if pbook.WebhookOnStatusUpdateURL != "" {
```
agarciamontoro, line 94: We should change this to the following, so we can actually save playbooks that have the setting disabled, even if the URL is wrong.

```suggestion
	if pbook.WebhookOnStatusUpdateEnabled {
```

server/incident/service.go
```diff
@@ -123,6 +123,10 @@ func (s *ServiceImpl) broadcastIncidentCreation(theIncident *Incident, commander
 // It blocks until a response is received.
 func (s *ServiceImpl) sendWebhookOnCreation(theIncident *Incident) error {
 	siteURL := s.pluginAPI.Configuration.GetConfig().ServiceSettings.SiteURL
+	if siteURL == nil {
+		s.pluginAPI.Log.Warn("cannot send webhook on creation, please set siteURL")
+		return errors.New("Could not send webhook, please set siteURL")
+	}
```
agarciamontoro, line 129: Good catch!

server/incident/service.go
```diff
@@ -589,6 +584,71 @@ func (s *ServiceImpl) broadcastStatusUpdate(statusUpdate string, theIncident *In
...
+	}{
+		Incident:   *theIncident,
+		ChannelURL: channelURL,
+		DetailsURL: detailsURL,
+		StatusUpdate: StatusUpdateOptions{
+			Status:      theIncident.CurrentStatus,
+			Description: theIncident.Description,
+			Message:     theIncident.ReminderMessageTemplate,
+			Reminder:    theIncident.PreviousReminder,
+		},
```
agarciamontoro, line 624: Do we need to add this as a separate payload? Because it's already in `theIncident`, so I guess we're duplicating it here, right?

server/incident/service.go
```diff
@@ -589,6 +584,71 @@ func (s *ServiceImpl) broadcastStatusUpdate(statusUpdate string, theIncident *In
 	return nil
 }
 
+// sendWebhookOnUpdateStatus sends a POST request to the status update webhook URL.
+// It blocks until a response is received.
+func (s *ServiceImpl) sendWebhookOnUpdateStatus(theIncident *Incident) error {
```
agarciamontoro, line 589: There's no need to pass a pointer here, I think. What do you think about passing it by value?

server/incident/service.go
```diff
@@ -1636,6 +1705,23 @@ func getUserDisplayName(user *model.User) string {
...
+}
+
+func getDetailsURL(siteURL string, teamName string, manifestID string, incidentID string) string {
+	return fmt.Sprintf("%s/%s/%s/incidents/%s",
+		siteURL,
+		teamName,
+		manifestID,
+		incidentID,
+	)
+}
```
agarciamontoro, line 1723: Cool refactor, thank you!

server/sqlstore/incident.go
```diff
@@ -345,7 +347,7 @@ func (s *incidentStore) UpdateStatus(statusPost *incident.SQLStatusPost) error {
 	return nil
 }
 
-// UpdateTimelineEvent updates (or inserts) the timeline event
+// CreateTimelineEvent creates the timeline event
```
agarciamontoro, line 350: Thanks!

server/sqlstore/incident.go
```diff
@@ -381,6 +383,7 @@ func (s *incidentStore) CreateTimelineEvent(event *incident.TimelineEvent) (*inc
 	return event, nil
 }
 
+// UpdateTimelineEvent updates (or inserts) the timeline event
```
agarciamontoro, line 386: And thanks again!

server/sqlstore/migrations.go
```diff
@@ -639,6 +639,45 @@ var migrations = []Migration{
 				}
 			}
 
+			return nil
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.14.0"),
+		toVersion:   semver.MustParse("0.15.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
```
agarciamontoro, line 648: This looks good, but https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/526/files will probably be merged sooner, so you will need to move this to a migration from 0.15 to 0.16. I'll ping you when that's needed, no worries!

server/sqlstore/playbook.go
```diff
@@ -48,7 +48,7 @@ func NewPlaybookStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 			"ConcatenatedInvitedUserIDs", "ConcatenatedInvitedGroupIDs", "InviteUsersEnabled",
 			"DefaultCommanderID", "DefaultCommanderEnabled",
 			"AnnouncementChannelID", "AnnouncementChannelEnabled",
-			"WebhookOnCreationURL", "WebhookOnCreationEnabled").
+			"WebhookOnCreationURL", "WebhookOnCreationEnabled", "WebhookOnStatusUpdateURL", "WebhookOnStatusUpdateEnabled").
```
agarciamontoro, line 51: Can you move this to another line?
```suggestion
			"WebhookOnCreationURL", "WebhookOnCreationEnabled",
			"WebhookOnStatusUpdateURL", "WebhookOnStatusUpdateEnabled").
```

webapp/src/components/backstage/automation/webhook.tsx
```diff
@@ -15,7 +15,7 @@ interface Props {
     onChange: (url: string) => void;
 }
 
-export const WebhookOnCreation: FC<Props> = (props: Props) => (
+export const Webhook: FC<Props> = (props: Props) => (
```
agarciamontoro, line 18: This will also conflict with https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/526/files#diff-905c0f515dd19d445500217b98ac49fb422cbeade4d21f18427335b895825aba. I'll also ping you when needed!

agarciamontoro (changes requested):
This is really really cool! Thank you for your contribution, @madhavhugar, it will be really useful :tada:

I left a couple of minor comments below.

server/incident/service.go
madhavhugar: Yes, we are duplicating it. It was to indicate explicitly that the status has been updated. Also, in a way to differentiate between a created and an updated payload. 

Should we avoid the duplication?

server/incident/service.go
madhavhugar: agreed 💯

server/incident/service.go
agarciamontoro: I think we should avoid the duplication, and tackle the problem of differentiating the different webhooks independently. Maybe we could add an "event" field or something similar, but we can tackle that in a separate PR.

agarciamontoro (changes requested):
Thank you for keeping this in sync, @madhavhugar! I left a couple more comments, and we still need to remove the duplication of the fields in the struct sent with the webhook.

Thank you :tada:

server/incident/service.go
madhavhugar: makes sense ➕1
I've updated the payload to avoid duplication

agarciamontoro (approved):
Thank you for all your work on this, @madhavhugar :tada: It looks good to me!

Now it's time for QA to review it!

A piece of feedback for next PRs: we usually do not force-push anything to an already open PR, but instead just keep adding commits. When merging, we squash all commits into one, so there is no need to worry for having a long list of commits with several changes. This way makes it easier for the reviewers to see the incremental changes.

prapti (approved):
This PR is looking great. Tested with valid and invalid outgoing webhook URLs during incident channel update (including reopening the incident), and things are working as expected.
Approving.
Thank you for this contribution @madhavhugar!

agarciamontoro:
Took another quick look and everything looks great. Thank you very much, @madhavhugar, great job!

PR #543, opened 2021-05-04
MM-35244: updated telemetry

lieut-data:
#### Summary
I combed through the actions and properties of our current telemetry and the in-product experience, adding the following:
* remove timeline event
* ChecklistItem properties
* updated Incident and Playbook properties
* timeline_tab_filter_selected event to track additional timeline interactions

Proposing including this in v1.9 -- cc @prapti.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35244

### DIFF
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -541,7 +541,7 @@ func (s *ServiceImpl) AddPostToTimeline(incidentID, userID, postID, summary stri
 }
 
 // RemoveTimelineEvent removes the timeline event (sets the DeleteAt to the current time).
-func (s *ServiceImpl) RemoveTimelineEvent(incidentID, eventID string) error {
+func (s *ServiceImpl) RemoveTimelineEvent(incidentID, userID, eventID string) error {
 	event, err := s.store.GetTimelineEvent(incidentID, eventID)
 	if err != nil {
 		return err
@@ -551,6 +551,13 @@ func (s *ServiceImpl) RemoveTimelineEvent(incidentID, eventID string) error {
 	if err = s.store.UpdateTimelineEvent(event); err != nil {
 		return err
 	}
+
+	incidentModified, err := s.store.GetIncident(incidentID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve incident")
+	}
+
+	s.telemetry.RemoveTimelineEvent(incidentModified, userID)
 
 	if err = s.sendIncidentToClient(incidentID); err != nil {
 		return err
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -820,8 +820,6 @@ func (s *ServiceImpl) ModifyCheckedState(incidentID, userID, newState string, ch
 
 	// Send modification message before the actual modification because we need the postID
 	// from the notification message.
-	s.telemetry.ModifyCheckedState(incidentID, userID, newState, incidentToModify.CommanderUserID == userID, itemToCheck.AssigneeID == userID)
-
 	mainChannelID := incidentToModify.ChannelID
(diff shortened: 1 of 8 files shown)


server/incident/service.go
```diff
@@ -834,6 +839,8 @@ func (s *ServiceImpl) ModifyCheckedState(incidentID, userID, newState string, ch
 		return errors.Wrapf(err, "failed to update incident, is now in inconsistent state")
 	}
 
+	s.telemetry.ModifyCheckedState(incidentID, userID, itemToCheck, incidentToModify.CommanderUserID == userID)
```
lieut-data, line 842: Moved this out from underneath a comment that wasn't relevant, until after the change is actually effected.

server/incident/service.go
```diff
@@ -930,6 +937,8 @@ func (s *ServiceImpl) SetAssignee(incidentID, userID, assigneeID string, checkli
 		return errors.Wrapf(err, "failed to update incident; it is now in an inconsistent state")
 	}
 
+	s.telemetry.SetAssignee(incidentID, userID, itemToCheck)
```
lieut-data, line 940: Move this up before the timeline events (that might fail).

server/incident/service.go
```diff
@@ -993,6 +1000,8 @@ func (s *ServiceImpl) RunChecklistItemSlashCommand(incidentID, userID string, ch
 		return "", errors.Wrapf(err, "failed to update incident recording run of slash command")
 	}
 
+	s.telemetry.RunTaskSlashCommand(incidentID, userID, itemToRun)
```
lieut-data, line 1003: Again, up before timeline events.

server/telemetry/rudder.go
```diff
@@ -107,14 +109,22 @@ func incidentProperties(incdnt *incident.Incident, userID string) map[string]int
 	return map[string]interface{}{
```
lieut-data, line 109: I know we have plans to annotate the structs here, but I didn't want to add too much complexity to this first pass.

server/telemetry/rudder.go
```diff
@@ -247,15 +270,27 @@ func playbookProperties(pbook playbook.Playbook, userID string) map[string]inter
...
 		"UsesReminderMessageTemplate": pbook.ReminderMessageTemplate != "",
+		"ReminderTimerDefaultSeconds": pbook.ReminderTimerDefaultSeconds,
+		"NumInvitedUserIDs":           len(pbook.InvitedUserIDs),
+		"NumInvitedGroupIDs":          len(pbook.InvitedGroupIDs),
+		"InviteUsersEnabled":          pbook.InviteUsersEnabled,
+		"DefaultCommanderID":          pbook.DefaultCommanderID,
+		"DefaultCommanderEnabled":     pbook.DefaultCommanderEnabled,
+		"AnnouncementChannelID":       pbook.AnnouncementChannelID,
+		"AnnouncementChannelEnabled":  pbook.AnnouncementChannelEnabled,
+		"NumWebhookOnCreationURLs":    len(strings.Split(pbook.WebhookOnCreationURL, ",")),
```
lieut-data, line 292: I'm assuming we'll add support for multiple webhooks, so splitting on commas.

crspeller (approved):

server/telemetry/rudder.go
cpoile: just wondering if commas are allowed in urls -- quick search says technically they are (in filenames?). Don't know if we want to use a space instead?

cpoile (approved):
One non-blocking question, nice!

server/telemetry/rudder.go
lieut-data: Ah, good catch -- I think the server /actually/ uses newlines, so I'll split on that instead.

PR #544, opened 2021-05-05
MM-35041 MM-35042 Basic retrospective tab and infrastructure

crspeller:
#### Summary
Infrastructure for the backstage retrospective tab. This is all behind the experimental features flag for now. All this is is the tab and and editable field in the backstage. Pressing the save button will save a "draft" that you can't do anything with for now. 

Thanks to @cpoile for the client framework and timeline which is the basis for this PR. 

FYI and feedback appreciated @itao @abhijit-singh 

#### Where we are
- **Placeholder Elements**
- **Basic saving of retrospective** <-- We are here
- Ability to publish retrospective
- Reminder to fill out retrospective
- Templating retrospective report
- Configurable persistant reminders
- Polish

#### Retrospective tab:
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35041
https://mattermost.atlassian.net/browse/MM-35042

### DIFF
--- a/server/incident/incident.go
+++ b/server/incident/incident.go
@@ -466,10 +466,10 @@ type Telemetry interface {
 	RunTaskSlashCommand(incidentID, userID string)
 
 	// UpdateRetrospective event
-	UpdateRetrospective(incidentID, userID string)
+	UpdateRetrospective(incident *Incident, userID string)
 
 	// PublishRetrospective event
-	PublishRetrospective(incidentID, userID string)
+	PublishRetrospective(incident *Incident, userID string)
 }
 
 type JobOnceScheduler interface {
--- a/server/incident/incident.go
+++ b/server/incident/incident.go
@@ -52,6 +52,7 @@ type Incident struct {
 	DefaultCommanderID      string               `json:"default_commander_id"`
 	AnnouncementChannelID   string               `json:"announcement_channel_id"`
 	WebhookOnCreationURL    string               `json:"webhook_on_creation_url"`
+	Retrospective           string               `json:"retrospective"`
 }
 
 func (i *Incident) Clone() *Incident {
@@ -366,6 +367,12 @@ type Service interface {
 	// UserHasLeftChannel is called when userID has left channelID. If actorID is not blank, userID
 	// was removed from the channel by actorID.
 	UserHasLeftChannel(userID, channelID, actorID string)
+
+	// UpdateRetrospective updates the retrospective for the given incident.
+	UpdateRetrospective(incidentID, userID, newRetrospective string) error
+
+	// PublishRetrospective publishes the retrospective.
+	PublishRetrospective(incidentID, userID string) error
 }
 
 // Store defines the methods the ServiceImpl needs from the interfaceStore.
@@ -457,6 +464,12 @@ type Telemetry interface {
 	// RunTaskSlashCommand tracks the execution of a slash command attached to
 	// a checklist item.
 	RunTaskSlashCommand(incidentID, userID string)
+
+	// UpdateRetrospective event
+	UpdateRetrospective(incidentID, userID string)
+
+	// PublishRetrospective event
+	PublishRetrospective(incidentID, userID string)
 }
 
 type JobOnceScheduler interface {
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -1624,6 +1624,31 @@ func (s *ServiceImpl) sendIncidentToClient(incidentID string) error {
 	return nil
 }
 
+func (s *ServiceImpl) UpdateRetrospective(incidentID, updaterID, newRetrospective string) error {
+	incidentToModify, err := s.store.GetIncident(incidentID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve incident")
+	}
+
+	incidentToModify.Retrospective = newRetrospective
+
+	if err = s.store.UpdateIncident(incidentToModify); err != nil {
+		return errors.Wrap(err, "failed to update incident")
+	}
+
+	s.poster.PublishWebsocketEventToChannel(incidentUpdatedWSEvent, incidentToModify, incidentToModify.ChannelID)
+	s.telemetry.UpdateRetrospective(incidentID, updaterID)
+
+	return nil
+}
+
+func (s *ServiceImpl) PublishRetrospective(incidentID, publisherID string) error {
+	//TODO: Publish the retrospective
+	s.telemetry.PublishRetrospective(incidentID, publisherID)
+
+	return nil
+}
+
 func getUserDisplayName(user *model.User) string {
 	if user == nil {
 		return ""
--- a/server/incident/service.go
+++ b/server/incident/service.go
@@ -1637,14 +1637,19 @@ func (s *ServiceImpl) UpdateRetrospective(incidentID, updaterID, newRetrospectiv
 	}
 
 	s.poster.PublishWebsocketEventToChannel(incidentUpdatedWSEvent, incidentToModify, incidentToModify.ChannelID)
-	s.telemetry.UpdateRetrospective(incidentID, updaterID)
+	s.telemetry.UpdateRetrospective(incidentToModify, updaterID)
 
 	return nil
 }
 
 func (s *ServiceImpl) PublishRetrospective(incidentID, publisherID string) error {
+	incidentToPublish, err := s.store.GetIncident(incidentID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve incident")
+	}
+
 	//TODO: Publish the retrospective
-	s.telemetry.PublishRetrospective(incidentID, publisherID)
+	s.telemetry.PublishRetrospective(incidentToPublish, publisherID)
 
 	return nil
 }
--- a/server/telemetry/rudder.go
+++ b/server/telemetry/rudder.go
@@ -235,14 +235,14 @@ func (t *RudderTelemetry) RunTaskSlashCommand(incidentID, userID string) {
 	t.track(eventTasks, properties)
 }
 
-func (t *RudderTelemetry) UpdateRetrospective(incidentID, userID string) {
-	properties := taskProperties(incidentID, userID)
+func (t *RudderTelemetry) UpdateRetrospective(incident *incident.Incident, userID string) {
+	properties := incidentProperties(incident, userID)
 	properties["Action"] = actionUpdateRetrospective
 	t.track(eventTasks, properties)
 }
 
-func (t *RudderTelemetry) PublishRetrospective(incidentID, userID string) {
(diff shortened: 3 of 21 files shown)


server/incident/incident.go
```diff
@@ -52,6 +52,7 @@ type Incident struct {
 	DefaultCommanderID      string               `json:"default_commander_id"`
 	AnnouncementChannelID   string               `json:"announcement_channel_id"`
 	WebhookOnCreationURL    string               `json:"webhook_on_creation_url"`
+	Retrospective           string               `json:"retrospective"`
```
lieut-data, line 55: I'd love to see this modelled in a separate table. It's not super obvious that I'd want to query the retrospective for an incident quite like this, and I think there's a potential that retrospectives might not always be tied to an incident.

server/telemetry/rudder.go
```diff
@@ -232,6 +235,18 @@ func (t *RudderTelemetry) RunTaskSlashCommand(incidentID, userID string) {
 	t.track(eventTasks, properties)
 }
 
+func (t *RudderTelemetry) UpdateRetrospective(incidentID, userID string) {
+	properties := taskProperties(incidentID, userID)
```
lieut-data, line 239: Why `taskProperties` and not `incidentProperties`?

webapp/src/components/backstage/incidents/incident_backstage/retrospective/profile_vertical.tsx
```diff
@@ -0,0 +1,73 @@
...
+        if (!user) {
+            dispatch(fetchUser(props.userId));
+        }
+    }, [props.userId]);
+
+    let name = null;
+    let profileUri = null;
+    if (user) {
+        name = displayUsername(user, teamnameNameDisplaySetting);
+        profileUri = Client4.getProfilePictureUrl(props.userId, user.last_picture_update);
```
lieut-data, line 60: nit: while `uri` is technically more correct, I fear it's extra cognitive load to just using `url`.

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline.tsx
```diff
@@ -0,0 +1,76 @@
...
+        return (
+            <NoEventsNotice>
+                {'Timeline events are displayed here as they occur. Hover over an event to remove it.'}
+            </NoEventsNotice>
+        );
+    }
+
+    return (
+        <TimelineLine data-testid='timeline-view'>
+            {
```
lieut-data, line 61: nit: worth merging this to the previous to avoid unnecessary indents?

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+const PrimaryButtonNotRight = styled(PrimaryButtonRight)`
+    margin-left: 20px;
+`;
+
+type IdToUserFn = (userId: string) => UserProfile;
+
+const TimelineRetro = (props: { incident: Incident }) => {
+    const dispatch = useDispatch();
+    const displayPreference = useSelector<GlobalState, string | undefined>(getTeammateNameDisplaySetting) || 'username';
```
lieut-data, line 74: Ugh, I despise the constant need to resolve the display name setting. Would love to "fix" this in Redux at some point :)

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const [allEvents, setAllEvents] = useState<TimelineEvent[]>([]);
+    const [filteredEvents, setFilteredEvents] = useState<TimelineEvent[]>([]);
+    const eventsFilter = useSelector<GlobalState, TimelineEventsFilter>((state) => rhsEventsFilterForChannel(state, props.incident.channel_id));
+    const getStateFn = useStore().getState;
```
lieut-data, line 78: Something smells "off" here -- it shouldn't normally be necessary to access the state like this. Can you elaborate?

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
```diff
@@ -0,0 +1,203 @@
...
+        setFilteredEvents(allEvents.filter((e) => showEvent(e.event_type, eventsFilter)));
+    }, [eventsFilter, allEvents]);
+
+    useEffect(() => {
+        Promise.all(props.incident.timeline_events.map(async (e) => {
+            let user = selectUser(e.subject_user_id) as UserProfile | undefined;
+
+            if (!user) {
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                const ret = await getUserFn(e.subject_user_id) as { data?: UserProfile, error?: any };
```
lieut-data, line 92: Is there an opportunity to fetch users in bulk vs. firing off a request per user?

lieut-data (changes requested):
Nice! Love that we can iterate "safely" under the experimental flag. A few comments below.

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
crspeller: Yes I hate this so much.

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
crspeller: I don't understand this. Since this code was copy pasted by ChrisP from the RHS I am hesatent to start modifying it here. Ticket to resolve this and the following comment: https://mattermost.atlassian.net/browse/MM-35475

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
crspeller: https://mattermost.atlassian.net/browse/MM-35475

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline.tsx
crspeller: Yup.

webapp/src/components/backstage/incidents/incident_backstage/retrospective/profile_vertical.tsx
crspeller: Yup. Also fixed the logic a bit. Added a comment to make sure people know this is a stub component.

server/telemetry/rudder.go
crspeller: Because... I copy pasted that.

server/incident/incident.go
crspeller: I am hesitant to add the complexity of a new table and another join. Right now I have seen no plans for multiple retrospectives. 
Are you OK with this for now and seeing how the design evolves? Or do you have a strong feeling we are going to need a table? A whole separate object?

webapp/src/components/backstage/incidents/incident_backstage/retrospective/report.tsx
```diff
@@ -0,0 +1,98 @@
...
+    flex-grow: 1;
+`;
+
+const SaveButton = (props: SaveButtonProps) => {
+    const [saved, setSaved] = useState(false);
+
+    const doSave = () => {
+        props.onSave();
+        setSaved(true);
+        setTimeout(() => setSaved(false), 1000);
```
agarciamontoro, line 78: This is for changing the text from `Saved!` back to `Save Draft` in one second? Why do we need that?

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
agarciamontoro: Hating this is indeed a general feeling.

agarciamontoro (changes requested):
Looks good! I really like that the experimental features setting allow us to ship in-progress stuff <3 Btw, I had the setting disabled, so I enabled it, but then I went to the incident overview page and the Retrospective tab did not show up until I refreshed the page.

Also, a minor UI thing with dark themes, maybe worth fixing it here:
image

webapp/src/components/backstage/incidents/incident_backstage/retrospective/report.tsx
agarciamontoro: I realized that this may be one of the in-progress features.

server/incident/incident.go
lieut-data: I'm less concerned about the table than if fetching `/incidents/blah` will also return a big markdown blob. I /feel/ like this is more of a "document" than a property of the incident.

server/incident/service.go
```diff
@@ -1624,6 +1624,36 @@ func (s *ServiceImpl) sendIncidentToClient(incidentID string) error {
...
+}
+
+func (s *ServiceImpl) PublishRetrospective(incidentID, publisherID string) error {
+	incidentToPublish, err := s.store.GetIncident(incidentID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve incident")
+	}
+
+	//TODO: Publish the retrospective
+	s.telemetry.PublishRetrospective(incidentToPublish, publisherID)
```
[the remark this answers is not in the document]
crspeller: I don't think it matters to just leave it there. Next PR will be implementing that.

server/telemetry/rudder.go
```diff
@@ -40,6 +40,9 @@ const (
 	actionSetAssigneeForTask  = "set_assignee_for_task"
 	actionRunTaskSlashCommand = "run_task_slash_command"
 
+	actionUpdateRetrospective  = "update_retrospective"
+	actionPublishRetrospective = "publish_retrospective"
+
```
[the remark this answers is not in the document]
crspeller: Ok.

webapp/src/components/backstage/incidents/incident_backstage/retrospective/report.tsx
crspeller: Yes the UX will be improved here.

server/incident/incident.go
crspeller: I feel like doing that would be a bit of over-optimization at this point. I agree that it's a better overall design, I just don't know where retrospectives are going at this point. What if suddenly they need to be in the RHS? Then the optimization of only fetching them sometimes was pointless. 

Maybe this is an item for the new Monday meeting?

lieut-data (approved):

PR #545, opened 2021-05-05
Fixed broken link in incident response image

iamrajiv:


agarciamontoro (approved):
Thank you for this, I had not noticed that the link was broken!

lieut-data (approved):

PR #546, opened 2021-05-05
MM-34665: Pricing differentiation features

agarciamontoro:
#### Summary
This PR finishes the first batch of pricing differentiation features. Namely:

1. It opens up the API and UI to all self-managed servers with any of E0, E10 or E20 licenses.
  - Cloud servers are untouched, as the flow to upgrade and the packaging features are still to be decided. So the behaviour should not change at all in cloud servers.
  - Team Edition servers have the UI still restricted (not the API, because differentiating between E0 and TE happens not to be possible as of today. 
2. It keeps three restrictions for the E0 users, who are not able to:
  - create more than one playbook per team,
  - access the Timeline tab in the RHS nor
  - add a post to the timeline through the post menu "Add to timeline" entry.
3. When interacting with these restricted features, the flow for admins and normal users is different:
  - As an admin, you get a modal (or, in the case of the RHS timeline, a placeholder image) from where you can start a trial.
  - As a normal user, you get a modal (or, in the case of the RHS timeline, a placeholder image) from where you can notify the administrator. This sends a DM to all the admins in the server, from where they can start the trial.

I encourage you to review this commit by commit. Each comment should be self-explanatory on its own, and the commit message gives some context on what it is doing. Also, all commits are independent (except for the couple ones that touch the `PricingPlanDifferentiationEnabled` flags, which are added in one commit and modified later), so you should not worry about reviewing edited code.

#### As a normal user

https://user-images.githubusercontent.com/3924815/117140450-b73b5b00-adad-11eb-9d7a-f099f82f8d92.mp4

https://user-images.githubusercontent.com/3924815/117140457-ba364b80-adad-11eb-9e25-1f1eba166441.mp4

https://user-images.githubusercontent.com/3924815/117140467-bc98a580-adad-11eb-8fc6-95ccc8acb14f.mp4

#### As an admin:

https://user-images.githubusercontent.com/3924815/117140269-80654500-adad-11eb-81e4-f5a77a3c868e.mp4

https://user-images.githubusercontent.com/3924815/117140288-865b2600-adad-11eb-8796-fc9159aeb2b3.mp4

https://user-images.githubusercontent.com/3924815/117140302-89eead00-adad-11eb-9022-0e9a23275037.mp4

#### As an admin with notifications initiated by a normal user

https://user-images.githubusercontent.com/3924815/117140411-ac80c600-adad-11eb-9be4-d173e9c4e270.mp4

The notifications initiated from the Timeline tab and from the post menu entry are the same, so I'm showing just one here:

https://user-images.githubusercontent.com/3924815/117140432-b1de1080-adad-11eb-9378-047f1d04e9e6.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34665
https://mattermost.atlassian.net/browse/MM-34666
https://mattermost.atlassian.net/browse/MM-34667


justinegeffen:
Thanks, @agarciamontoro! I haven't started reviewing but this is an awesome PR! Looking forward to going through it.

abhijit-singh (changes requested):
Great work @agarciamontoro!

A few small changes:

1.  In the normal user flow where they have an option to Notify System Admin, can you update the illustration and text on the success modal to match https://www.figma.com/file/1M0Kg4QpQ64fhHslLNeXK7/MM-34664-Pricing-Gate-UI?node-id=48%3A233031?
2. In the normal user flow modals, there does seem to be some extra space above and below the buttons - 
extra-space
The space between the text and the buttons is correct in the Start trial modal - 
correct-spacing

In the design, I had 32px between the text and the buttons, & 48 px below the buttons. 


3. In the timeline tab on the RHS, is it easily possible to hide the `Overview` & `Update status` buttons? They're conflicting with `Start trial` button because of proximity and aren't tightly related to the timeline feature.

webapp/src/components/backstage/upgrade_modal.tsx
```diff
@@ -0,0 +1,343 @@
...
+            return UserStartModal;
+        case ModalActionState.Success:
+            return UserSuccessModal;
+        default:
+            return UserStartModal;
+        }
+    }
+};
+
+const UpgradeModal: FC<Props> = (props: Props) => {
```
crspeller, line 161: This is a complicated modal!

webapp/src/hooks.ts
```diff
@@ -214,3 +217,45 @@ export function useLatestUpdate(incident: Incident) {
...
 }
+
+export function useNumPlaybooksInCurrentTeam() {
+    const dispatch = useDispatch();
+    const team = useSelector(getCurrentTeam);
+    const numPlaybooks = useSelector(currentTeamNumPlaybooks);
+
+    useEffect(() => {
+        const fetch = async () => {
+            const response = await clientFetchPlaybooksCount(team.id);
```
crspeller, line 228: This is a lot of extra requests to grab the number of playbooks.

webapp/src/hooks.ts
```diff
@@ -214,3 +217,45 @@ export function useLatestUpdate(incident: Incident) {
...
+
+        fetch();
+    }, [team.id]);
+
+    return numPlaybooks;
+}
+
+function useIsAllowedInE0() {
+    const numPlaybooks = useNumPlaybooksInCurrentTeam();
+    const isLicensed = useSelector(isE10LicensedOrDevelopment);
```
crspeller, line 240: `useIsAllowedInE0` -> `isE10Licenced` Confusing naming.

webapp/src/hooks.ts
```diff
@@ -214,3 +217,45 @@ export function useLatestUpdate(incident: Incident) {
...
+// useAllowPlaybookCreationInCurrentTeam returns whether a user can create
+// a playbook in the current team
+export function useAllowPlaybookCreationInCurrentTeam() {
+    return useIsAllowedInE0();
+}
+
+// useAllowTimelineViewInCurrentTeam returns whether a user can view the RHS
+// timeline in the current team
+export function useAllowTimelineViewInCurrentTeam() {
+    return useIsAllowedInE0();
```
crspeller, line 254: So we allow timeline in E0? Why are we checking then?

webapp/src/make_update_main_menu.ts
```diff
@@ -9,16 +9,20 @@ import {getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
...
 export function makeUpdateMainMenu(registry: PluginRegistry, store: Store<GlobalState>): () => Promise<void> {
     let mainMenuActionId: string | null;
 
     return async () => {
-        const disable = isDisabledOnCurrentTeam(store.getState());
-        const show = !disable && !isMobile() && isE20LicensedOrDevelopment(store.getState());
+        let show = !isMobile();
+
+        if (!isPricingPlanDifferentiationEnabled(store.getState())) {
+            const disable = isDisabledOnCurrentTeam(store.getState());
```
crspeller, line 23: The disabling on the current team thing shouldn't be contingent on pricing plan differentiation being enabled. Right? Still confused by the `isPricingPlanDifferentiationEnabled` stuff.

crspeller (changes requested):
Lots of complicated stuff!

I am really confused by the `PricingPlanDifferentiationEnabled` stuff. Maybe because I don't understand what behavior that is trying to model?

webapp/src/components/backstage/upgrade_modal.tsx
cpoile: @crspeller Would it be better to have the different modals as separate components? (even if it meant some duplication)

webapp/src/components/assets/buttons.tsx
```diff
@@ -103,3 +105,20 @@ export const GrayTertiaryButton = styled.button`
...
 
+export type UpgradeButtonProps = React.ComponentProps<typeof PrimaryButton>;
+
+export const UpgradeButton: FC<UpgradeButtonProps> = (props: UpgradeButtonProps) => {
+    const {children, ...rest} = props;
+    return (
+        <PrimaryButton {...rest}>
+            {children}
+            <PositionedUpgradeBadge/>
+        </PrimaryButton>
```
cpoile, line 116: This is cool!

cpoile (changes requested):
Awesome work @agarciamontoro ! This is a huge (and complex) feature, nicely done.

justinegeffen (commented):
This is awesome, @agarciamontoro! I've made some suggestions to terminology and also suggested changing "Administrator" to "System Admin" for consistency with our docs. Other than that, it looks good to me!

agarciamontoro:
@abhijit-singh 

> In the normal user flow where they have an option to Notify System Admin, can you update the illustration and text on the success modal to match https://www.figma.com/file/1M0Kg4QpQ64fhHslLNeXK7/MM-34664-Pricing-Gate-UI?node-id=48%3A233031?

Done:

image


> In the normal user flow modals, there does seem to be some extra space above and below the buttons

Fixed:

image

> In the timeline tab on the RHS, is it easily possible to hide the Overview & Update status buttons? They're conflicting with Start trial button because of proximity and aren't tightly related to the timeline feature.

Removed:

image

I also fixed what we discussed in the meeting last time:

1. I removed the purchase license button

image

2. I added a badge to the templates header

image

3. I modified the behaviour of the interactive message sent to the admin to change only the text of the button while the license is being downloaded:


https://user-images.githubusercontent.com/3924815/117676211-9d36b980-b1ad-11eb-9fe7-1f66172c2001.mp4

server/api/api.go
```diff
@@ -36,11 +36,13 @@ func NewHandler(pluginAPI *pluginapi.Client, config config.Service) *Handler {
 
 	handler.APIRouter = api
 	handler.root = root
-
-	e20Middleware := E20LicenseRequired{config}
+	handler.config = config
 
 	settingsRouter := handler.APIRouter.PathPrefix("/settings").Subrouter()
-	settingsRouter.Use(e20Middleware.Middleware)
+	if !config.IsPricingPlanDifferentiationEnabled() {
```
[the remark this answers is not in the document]
agarciamontoro: > I don't get this. If we are on cloud... which is what this function checks. We are defiantly licensed why do we do a check here?

The `IsPricingPlanDifferentiationEnabled` is actually a feature flag. The pricing plan differentiation is enabled only on on-prem servers (on the webapp side, it is further restricted to only on-prem servers **with at least E0**), because the upgrade flow for cloud is not yet finished. For now, all cloud servers are licensed, but this will change when 5.36 is released, which is the deadline for the completion of this feature (this PR and the one coming).

But now that I think about it, I can actually get rid of the whole thing on the server side, and simply have every endpoint open for all servers, licensed or not (when the feature is complete, this will be the case, so we can just do it now). The frontend will still be restricted while I implement the cloud upgrade flow. What do you think, @crspeller, @cpoile?

> I also think the IsPricingPlanDifferentiationEnabled name is confusing because why would you want to disable it? It seems pretty fundamental.

I agree that it is not my best naming decision :joy: We want to disable it in cloud because it is not yet implemented. Again, we can simply keep the flag in the client, and simply remove this one.

> Also it seems like this could be handled centrally by using a use on the api subrouter.

I did not want to add it as a subrouter because of the nature of the function being a feature flag. But yup, getting rid of it in the API would solve all of these cases.

webapp/src/components/backstage/upgrade_modal.tsx
agarciamontoro: > This is a complicated modal!

I agree.

> Would it be better to have the different modals as separate components? (even if it meant some duplication)

I will try something similar to this, I'll ping you when it's ready.

webapp/src/hooks.ts
agarciamontoro: We should keep the use because it is actually a hook (https://reactjs.org/docs/hooks-custom.html#using-a-custom-hook), but I agree the name is confusing, but it does not only check if the server is licensed, but if the creation of playbooks is allowed. What about being super explicit and `useIsPlaybookCreationAllowed`?

webapp/src/hooks.ts
agarciamontoro: It is actually incorrect. This is the restriction for the playbooks, but the timeline cannot be viewed in E0 in any case. Thanks for catching this, I'm fixing it!

webapp/src/make_update_main_menu.ts
agarciamontoro: Oops, good catch! I will fix this logic.

server/bot/poster.go
```diff
@@ -82,6 +83,81 @@ func (b *Bot) PublishWebsocketEventToUser(event string, payload interface{}, use
...
+func (b *Bot) NotifyAdmins(messageType, authorUserID string) error {
+	author, err := b.pluginAPI.User.Get(authorUserID)
+	if err != nil {
+		return errors.Wrap(err, "unable to find author user")
+	}
+
+	admins, err := b.pluginAPI.User.List(&model.UserGetOptions{
+		Role:    string(model.SYSTEM_ADMIN_ROLE_ID),
+		Page:    0,
+		PerPage: 100,
```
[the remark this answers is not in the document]
agarciamontoro: I was paranoid too. We can bump it to 1000 no problem (what if there are more?!).

server/command/command.go
```diff
@@ -1398,7 +1398,7 @@ func (r *Runner) Execute() error {
 		return err
 	}
 
-	if !r.configService.IsLicensed() {
+	if !config.PricingPlanDifferentiationEnabled && !r.configService.IsLicensed() {
```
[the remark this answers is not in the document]
agarciamontoro: If it is enabled, everything should be open. So when I get rid of the `PricingPlanDifferentiationEnabled`, I will simply get rid of this `if` altogether.

webapp/src/hooks.ts
```diff
@@ -214,3 +217,45 @@ export function useLatestUpdate(incident: Incident) {
...
+            dispatch(receivedTeamNumPlaybooks(team.id, response.count));
+        };
+
+        fetch();
+    }, [team.id]);
+
+    return numPlaybooks;
+}
+
+function useIsAllowedInE0() {
```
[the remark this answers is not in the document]
agarciamontoro: It is a hook, so according to https://reactjs.org/docs/hooks-custom.html#using-a-custom-hook it should be prefixed with `use`, right?

webapp/src/hooks.ts
```diff
@@ -214,3 +217,45 @@ export function useLatestUpdate(incident: Incident) {
...
+// useAllowTimelineViewInCurrentTeam returns whether a user can view the RHS
+// timeline in the current team
+export function useAllowTimelineViewInCurrentTeam() {
+    return useIsAllowedInE0();
+}
+
+// useAllowAddMessageToTimelineInCurrentTeam returns whether a user can add a
+// post to the timeline in the current team
+export function useAllowAddMessageToTimelineInCurrentTeam() {
+    return useIsAllowedInE0();
```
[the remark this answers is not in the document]
agarciamontoro: Yup, you're right, thank you!

webapp/src/components/backstage/upgrade_modal.tsx
```diff
@@ -0,0 +1,343 @@
...
+                {' to avoid any disruption.'}
+            </span>
+        ),
+        confirmButtonText: 'Purchase license now',
+        handleConfirm: () => window.open('https://customers.mattermost.com/signup', '_blank'),
+    });
+
+    const UserStartModal = Object.assign({...CommonModal}, {
+        helpText: helpText + ' Notify your administrator to upgrade.',
+        confirmButtonText: 'Notify Administrator',
```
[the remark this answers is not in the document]
agarciamontoro: @justinegeffen, should this be "Notify System Admin" as you suggested [for the RHS placehodler](https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/files#r629328206)? Or change that one to be simply "Notify", too. Otherwise, it will not be consistent.

webapp/src/components/rhs/rhs_timeline_placeholder.tsx
```diff
@@ -0,0 +1,185 @@
...
+        if (response.error) {
+            setActionState(ActionState.Error);
+        } else {
+            setActionState(ActionState.Success);
+        }
+    };
+
+    let helpText = 'Make retros easy. Your timeline includes all the events in your incident, separated by type, and downloadable for offline review.';
+    if (actionState === ActionState.Success) {
+        helpText = 'A notification has been sent to your administrator.';
```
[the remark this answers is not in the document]
agarciamontoro: @justinegeffen, same question here as before

webapp/src/hooks.ts
agarciamontoro: This will need to change when we go to a workspace-scoped product, so it will be easier to change this there. I created https://mattermost.atlassian.net/browse/MM-35609 to track it.

server/api/api.go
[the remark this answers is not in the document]
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/6b8b0026c53e51cb318d8948071f97559f3695a8

webapp/src/components/backstage/upgrade_modal.tsx
agarciamontoro: I ended up splitting it into smaller, simpler pieces, and I think it is now a bit more easy to digest. See https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/a34f311f2e6919d8c22c50bc3adae78be118a7d7. What do you think?

webapp/src/hooks.ts
agarciamontoro: Simplified this in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/5b43013d93fc63d34c548df8ea03409b0fe85bd4

webapp/src/hooks.ts
agarciamontoro: Fixed in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/5b43013d93fc63d34c548df8ea03409b0fe85bd4

server/bot/poster.go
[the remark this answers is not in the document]
agarciamontoro: Admin level over 1000 https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/db0d22888c780085795067d189b5987af9ea6102

webapp/src/hooks.ts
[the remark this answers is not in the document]
agarciamontoro: Fixed in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/5b43013d93fc63d34c548df8ea03409b0fe85bd4

abhijit-singh (approved):

webapp/src/components/backstage/upgrade_modal.tsx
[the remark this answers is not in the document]
justinegeffen: Good question! Notify System Admin works for me, thanks. :)

justinegeffen (approved):

webapp/src/components/backstage/upgrade_modal.tsx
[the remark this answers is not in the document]
agarciamontoro: @justinegeffen, for both the "Add more to your timeline" and the "Playbook limit reached" modals, yes.

server/config/service.go
```diff
@@ -183,12 +183,12 @@ func (c *ServiceImpl) IsCloud() bool {
...
-func (c *ServiceImpl) IsLicensed() bool {
+// IsE20Licensed returns true when the server is appropriately licensed to run this plugin.
+func (c *ServiceImpl) IsE20Licensed() bool {
 	return pluginapi.IsE20LicensedOrDevelopment(c.api.Configuration.GetConfig(), c.api.System.GetLicense())
 }
 
-func (c *ServiceImpl) IsPricingPlanDifferentiationEnabled() bool {
-	// Disable for all cloud servers
-	return !c.IsCloud()
+// IsE10Licensed returns true when the server is appropriately licensed to run this plugin.
```
cpoile, line 191: These two comments could be better I think -- reading just the `IsE20Licensed` comment makes me think we need E20 to run the plugin, but then this comment makes me think we need E10, so I'm left thinking "what licence do we need"?

webapp/src/hooks.ts
```diff
@@ -235,27 +235,41 @@ export function useNumPlaybooksInCurrentTeam() {
...
 }
 
 // useAllowAddMessageToTimelineInCurrentTeam returns whether a user can add a
 // post to the timeline in the current team
 export function useAllowAddMessageToTimelineInCurrentTeam() {
-    return useIsAllowedInE0();
+    return useSelector(isE10LicensedOrDevelopment);
+}
+
+export function useEnsureProfiles(userIds: string[]) {
```
cpoile, line 259: Is this duplicating on purpose, or a mistake with github?

cpoile (approved):
Cool, looks much nicer, thanks @agarciamontoro! That upgrade modal is kinda amazing. :) Nothing blocking, good work.

webapp/src/hooks.ts
agarciamontoro: Mmm not sure what you mean. Is something duplicated here? I don't see it :thinking:

server/config/service.go
agarciamontoro: You're right! Fixed in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/c9b7f9b4016e96ba39777775a9544a66453801b4

server/api/bot.go
```diff
@@ -0,0 +1,131 @@
...
+	"net/http"
+
+	"github.com/gorilla/mux"
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
+	"github.com/mattermost/mattermost-plugin-incident-collaboration/server/bot"
+	"github.com/mattermost/mattermost-plugin-incident-collaboration/server/config"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+type BotHandler struct {
```
crspeller, line 14: Why is this called bot handler? How about LicensingHandler?

server/config/config.go
```diff
@@ -31,6 +31,9 @@ type Service interface {
 	// `EnableTesting` configuration settings enabled.
 	IsConfiguredForDevelopmentAndTesting() bool
 
-	// IsLicensed returns true when the server is appropriately licensed to run this plugin.
-	IsLicensed() bool
+	// IsE20Licensed returns true when the server is appropriately licensed to run this plugin.
+	IsE20Licensed() bool
+
+	// IsE10Licensed returns true when the server is appropriately licensed to run this plugin.
+	IsE10Licensed() bool
```
crspeller, line 38: I assume this will also succeed if the server has an E20 license correct? Maybe we should make this explicit in the name. `IsAtLeastE10Licensed`? At least change the comment to clarify.

webapp/src/components/backstage/upgrade_modal.tsx
crspeller: Much better!

crspeller (approved):
Couple of non-blocking suggestions.

server/api/bot.go
agarciamontoro: The root route is `/bot/notify-admins`, and I thought it we could use it for more bot-related activities, if needed?

server/config/config.go
agarciamontoro: You're right, changed the comment in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546/commits/c06ed2e4745379b03a13182fbeea2c09f674cdbd. Thanks!

agarciamontoro:
As discussed offline, QA will test this directly on master. Merging now! :tada:

PR #552, opened 2021-05-10
MM-35551 Fix menu visibility issues on enable/disable

crspeller:
#### Summary
Adds unregistration step for the two listeners we added to the redux store. 

When the plugin was enabled/disabled on a running server with an active webapp, the previous redux listeners would not be removed automatically (since it doesn't run though the plugin system). So phantom RHS opening and main menu items could appear under certain conditions. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35551

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React from 'react';
-import {Store} from 'redux';
+import {Store, Unsubscribe} from 'redux';
 import {debounce} from 'debounce';
 
 import {GlobalState} from 'mattermost-redux/types/store';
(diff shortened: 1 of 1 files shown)


webapp/src/index.tsx
```diff
@@ -120,6 +126,15 @@ export default class Plugin {
         registry.registerWebSocketEventHandler(WebsocketEvents.LICENSE_CHANGED, checkRegistrations);
         checkRegistrations();
     }
+
+    public uninitialize() {
```
cpoile, line 130: Are we sure this will be called? What happens if it doesn't?

cpoile (approved):
I'm approving, but I'm not sure if I understand all the ways this might get into a weird state. Are there other edge cases we haven't encountered yet (and which will be difficult to debug)?

webapp/src/index.tsx
lieut-data: Yeah, `uninitialize` is buggy -- it gets called if you explicitly disable a plugin, but generally doesn't get called if you upload a replacement.

webapp/src/index.tsx
```diff
@@ -120,6 +126,15 @@ export default class Plugin {
...
+        if (this.removeMainMenuSub) {
+            this.removeMainMenuSub();
```
lieut-data, line 132: for completeness, should we set `removeMainMenuSub` to `null`?

lieut-data (approved):

PR #553, opened 2021-05-12
Set defaults for playbooks to be open. Access to all and create public channel.

crspeller:
#### Summary
- Default is now for playbooks to be accessible by everyone. 
~~- Default is now to create a public channel.~~

### DIFF
--- a/webapp/src/components/backstage/playbook_edit.tsx
+++ b/webapp/src/components/backstage/playbook_edit.tsx
@@ -158,7 +158,6 @@ const PlaybookEdit: FC<Props> = (props: Props) => {
     const [playbook, setPlaybook] = useState<Playbook>({
         ...emptyPlaybook(),
         team_id: props.currentTeam.id,
-        member_ids: [currentUserId],
     });
     const [changesMade, setChangesMade] = useState(false);
     const [confirmOpen, setConfirmOpen] = useState(false);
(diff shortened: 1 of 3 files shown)


lieut-data (approved):

itao (approved):

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -158,7 +158,6 @@ const PlaybookEdit: FC<Props> = (props: Props) => {
     const [playbook, setPlaybook] = useState<Playbook>({
         ...emptyPlaybook(),
         team_id: props.currentTeam.id,
-        member_ids: [currentUserId],
```
lieut-data, line 161: Looking back to my first review, I recognize now that I did not then understand the nuance of `member_ids: []` (or is it `member_ids: undefined`?). In terms of "function continuity", it seems surprising that going from 1 member to 0 members means all members vs. "only system admins". 

Not required for this PR, but I'd love to see us distinguish whether a playbook is public or private from who should have access to the playbook to avoid unexpected API semantics.

(Or am I misinterpreting all of this?)

lieut-data (approved):

webapp/src/components/backstage/playbook_edit.tsx
crspeller: No, you are correct. I used `[]` as everyone has access since `[]` doesn't make any sense otherwise.

cpoile (approved):

PR #554, opened 2021-05-12
MM-35600 - Fix JS error on change commander

cpoile:
#### Summary
- the endpoint needed something to send back, because it was a `doPost` -> `doFetchWithResponse` call on the client.
- while investigating, found that the errors logged through logrus are not great--they don't conform to the way the server logs errors, and so they aren't coded correctly. The second commit fixes that.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-35600

#### Checklist

[ ] ~~Telemetry updated~~
[x] Unit tests updated

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -4,10 +4,11 @@ import (
 	"encoding/json"
 	"net/http"
 
-	"github.com/gorilla/mux"
-	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"github.com/mattermost/mattermost-plugin-incident-collaboration/server/bot"
+
+	"github.com/gorilla/mux"
 	pluginapi "github.com/mattermost/mattermost-plugin-api"
 	"github.com/mattermost/mattermost-plugin-incident-collaboration/server/config"
 	"github.com/mattermost/mattermost-plugin-incident-collaboration/server/permissions"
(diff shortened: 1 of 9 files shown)


server/api/api.go
```diff
@@ -121,14 +114,7 @@ func HandleErrorWithCode(w http.ResponseWriter, code int, publicErrorMsg string,
...
 
-	loggedMsg, _ := json.Marshal(struct {
-		Message string `json:"message"` // A public facing message providing details about the error.
-		Details string `json:"details"` // More details, potentially sensitive, about the error.
-	}{
-		Message: publicErrorMsg,
-		Details: details,
-	})
-	logrus.Warn(string(loggedMsg))
+	logger.Warnf("public error message: %v; internal details: %v", publicErrorMsg, details)
```
cpoile, line 117: This is the main change. In order to use `logger` we need to pass it, but I didn't want to have all the `HandleErrorWithCode` calls require the `logger` parameter, so I made copies of the calls as part of each service. The actual logging code is still focused here, so we're not duplicating the code that "does the work."

server/api/api.go
```diff
@@ -138,6 +124,23 @@ func HandleErrorWithCode(w http.ResponseWriter, code int, publicErrorMsg string,
 	_, _ = w.Write(responseMsg)
 }
 
+// ReturnJSON writes the given pointer to object as json with a success response
```
agarciamontoro, line 127: Nit: not entirely true, as you could pass any response code as the last argument, not only a success one.

agarciamontoro (approved):
Looks great, thank you!

I'm curious: what's the difference in the way the errors are logged now?

server/api/api.go
cpoile: Correct -- it was from the previous author. :) But fixed.

cpoile:
@agarciamontoro The logs are posted with the bot logger (through the pluginAPI) rather than through logrus

server/api/incidents.go
```diff
@@ -104,23 +104,34 @@ func NewIncidentHandler(router *mux.Router, incidentService incident.Service, pl
 	return handler
 }
 
+// HandleError logs the internal error and sends a generic error as JSON in a 500 response.
+func (h *IncidentHandler) HandleError(w http.ResponseWriter, internalErr error) {
```
crspeller, line 108: Maybe instead of duplicating this, use an embedded struct? 0/5

crspeller (approved):
Looks good. Just one 0/5 suggestion.

server/api/incidents.go
cpoile: @crspeller you're right, this is much cleaner. Look good?

crspeller: Sweet! 🎉

PR #562, opened 2021-05-19
MM-35650 - Long slash commands don't ellipsis

cpoile:
#### Summary
- Show a tooltip, but only for commands that ellipsis

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-35650

#### Checklist

[ ] ~~Telemetry updated~~
[ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/checklist_item.tsx
+++ b/webapp/src/components/checklist_item.tsx
@@ -11,9 +11,7 @@ import {Overlay, Popover, PopoverProps} from 'react-bootstrap';
 import Scrollbars from 'react-custom-scrollbars';
 import {useDispatch, useSelector} from 'react-redux';
 import {components, ControlProps} from 'react-select';
-
 import styled from 'styled-components';
-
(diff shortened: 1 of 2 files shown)


webapp/src/components/checklist_item.tsx
```diff
@@ -85,18 +85,20 @@ const ExtrasRow = styled.div`
...
-    >div {
+
+    > div {
         margin: 0 5px;
         border: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
         border-radius: 13px;
         padding: 2px 8px;
         background: rgba(var(--center-channel-color-rgb), 0.08);
         display: flex;
+        max-width: calc(100%);
```
lieut-data, line 96: curious: does `max-width: 100%` work without the `calc`? Usually `calc` is employed only when there's some external input, like `vh` or a CSS variable.

lieut-data (approved):
Love that this is "intelligent", not showing the tooltip when it's not needed.

PR #566, opened 2021-05-20
MM-35046 MM-35043 Reminders to fill in the retrospective

crspeller:
#### Summary
When an incident is resolved, a reminder will be sent to fill in the retrospective report

#### Where we are
- **Placeholder Elements**
- **Basic saving of retrospective** 
- **Ability to publish retrospective** 
- **Reminder to fill out retrospective**
- **Configurable persistent reminders**  <-- We are here
- Templating retrospective report
- _NEW:_ Replace reminder message with custom message for interactivity.
- _NEW:_ Add ability to disable retrospectives at playbook level.
- Polish

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35043
https://mattermost.atlassian.net/browse/MM-35046

#### Checklist
~~[ ] Telemetry updated~~
[x] Gated by experimental feature flag
[x] Unit tests updated


server/incident/reminder.go
```diff
@@ -15,11 +16,48 @@ type Reminder struct {
 	IncidentID string `json:"incident_id"`
 }
 
+const RetrospectivePrefix = "retro_"
```
crspeller, line 19: @cpoile Interested in your opinion on how I ended up using the reminder functionality.

server/incident/reminder.go
```diff
@@ -15,11 +16,48 @@ type Reminder struct {
...
+	if incidentToRemind.RetrospectivePublishedAt != 0 {
+		return
+	}
+
+	if err = s.postRetrospectiveReminder(incidentToRemind); err != nil {
+		s.logger.Errorf(errors.Wrapf(err, "couldn't post incident reminder").Error())
+		return
+	}
+
+	// Jobs can't be rescheduled within themselves with the same key. As a temporary workaround do it in a delayed goroutine
```
crspeller, line 47: This isn't ideal. This is done because rescheduling a reminder within the callback doesn't work as the reminder is removed after the callback is invoked. 
@cpoile Does a ticket make sense to fix this, or am I missing something?

server/playbook/playbook_test.go
```diff
@@ -34,7 +34,7 @@ func TestPlaybook_MarshalJSON(t *testing.T) {
 				InvitedUserIDs:              nil,
 				InvitedGroupIDs:             nil,
 			},
-			expected: []byte(`{"id":"playbookid","title":"the playbook title","description":"the playbook's description","team_id":"theteamid","create_public_incident":true,"create_at":4503134,"delete_at":0,"num_stages":0,"num_steps":0,"checklists":[],"member_ids":[],"broadcast_channel_id":"channelid","reminder_message_template":"This is a message","reminder_timer_default_seconds":0,"invited_user_ids":[],"invited_group_ids":[],"invite_users_enabled":false,"default_commander_id":"","default_commander_enabled":false,"announcement_channel_id":"","announcement_channel_enabled":false,"webhook_on_creation_url":"","webhook_on_creation_enabled":false,"message_on_join":"","message_on_join_enabled":false}`),
+			expected: []byte(`"checklists":[]`),
```
cpoile, line 37: A little confused why this changed so much (and only checklists was left)?

cpoile (changes requested):
Nice!

server/incident/service.go
```diff
@@ -1761,6 +1814,49 @@ func (s *ServiceImpl) PublishRetrospective(incidentID, text, publisherID string)
...
+	incidentToCancel, err := s.store.GetIncident(incidentID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve incident")
+	}
+
+	now := model.GetMillis()
+
+	// Update the text to keep syncronized
+	incidentToCancel.Retrospective = "No retrospective for this incident."
+	incidentToCancel.RetrospectivePublishedAt = now
```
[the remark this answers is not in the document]
crspeller: I could have a separate field that indicates that the retrospective was canceled.
This way it's like you are publishing the fact that it's canceled. 
Hmm... but what if you wanted to filter by which ones are canceled? Sounds like I need a field. Will add one.

server/playbook/playbook_test.go
crspeller: Changed the test so it doesn't take so much maintenance. Just checks for contains.

server/incident/reminder.go
crspeller: The problem with that is then I need to keep track of attempt_1 attempt_2 etc.
But if the suffixing is the solution, maybe we can just build that in. It also would be nice to have a separate data field so the key didn't have to be the data too. 
Made a ticket for that here: https://mattermost.atlassian.net/browse/MM-35938

cpoile: Why do you need to keep track of attempt_1, attempt_2, etc? You only have to send in the parsed attempt number and set the next attempt (attempt + 1) and then return from this one. No need to keep track. Or if you don't want to parse, you could use a uuid after the `retro_incidentid_` bit.

For https://mattermost.atlassian.net/browse/MM-35938 -- personally, I think we should keep that kind of complexity out of the jobOnce scheduler, and let consumers decide how they want to handle it.

server/incident/incident.go
```diff
@@ -25,36 +25,38 @@ const (
...
+	InvitedGroupIDs                      []string             `json:"invited_group_ids"`
+	TimelineEvents                       []TimelineEvent      `json:"timeline_events"`
+	DefaultCommanderID                   string               `json:"default_commander_id"`
+	AnnouncementChannelID                string               `json:"announcement_channel_id"`
+	WebhookOnCreationURL                 string               `json:"webhook_on_creation_url"`
+	Retrospective                        string               `json:"retrospective"`
+	RetrospectivePublishedAt             int64                `json:"retrospective_published_at"` // The last time a retrospective was published. 0 if never published.
+	RetrospectiveWasCanceled             bool                 `json:"retrospective_was_canceled"`
+	MessageOnJoin                        string               `json:"message_on_join"`
+	RetrospectiveReminderIntervalSeconds int64                `json:"retrospective_reminder_interval_seconds"`
```
agarciamontoro, line 59: Super nit: Can we group all retrospective fields together?
```suggestion
	RetrospectiveWasCanceled             bool                 `json:"retrospective_was_canceled"`
	RetrospectiveReminderIntervalSeconds int64                `json:"retrospective_reminder_interval_seconds"`
	MessageOnJoin                        string               `json:"message_on_join"`
```

server/incident/reminder.go
agarciamontoro: I think I'm with @cpoile here: unless I'm missing something obvious, it looks easy to just schedule a new one.

server/incident/service.go
```diff
@@ -658,6 +658,24 @@ func (s *ServiceImpl) UpdateStatus(incidentID, userID string, options StatusUpda
 		s.pluginAPI.Log.Warn("failed to broadcast the status update to channel", "ChannelID", incidentToModify.BroadcastChannelID)
 	}
 
+	// If we are resolving the incident, send the reminder to fill out the retrospective
+	// Also start the recurring reminder if enabled.
+	if incidentToModify.RetrospectivePublishedAt == 0 &&
+		options.Status == StatusResolved &&
+		previousStatus != StatusArchived &&
+		previousStatus != StatusResolved &&
```
agarciamontoro, line 666: This is good. Although it makes me wonder whether updating the status (from Active to Resolved, for example) and posting a status update (as in writing what's going on) should be two clearly separated processes, so it would not be possible to update the status from Resolved to Resolved. Definitely for the future, not this PR, but I think it's worth giving it a second thought.

server/incident/service.go
```diff
@@ -658,6 +658,24 @@ func (s *ServiceImpl) UpdateStatus(incidentID, userID string, options StatusUpda
...
+		s.configService.GetConfiguration().EnableExperimentalFeatures {
+		if err = s.postRetrospectiveReminder(incidentToModify); err != nil {
+			return errors.Wrap(err, "couldn't post retrospective reminder")
+		}
+		s.scheduler.Cancel(RetrospectivePrefix + incidentID)
```
agarciamontoro, line 671: Why do we need this `Cancel` here?

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -526,6 +538,29 @@ const PlaybookEdit: FC<Props> = (props: Props) => {
...
+                                            <BackstageSubheaderDescription>
+                                                {'Reminds the channel at a specified interval to fill out the retrospective.'}
+                                            </BackstageSubheaderDescription>
+                                        </BackstageSubheader>
+                                        <StyledSelect
+                                            value={retrospectiveReminderOptions.find((option) => option.value === playbook.retrospective_reminder_interval_seconds)}
+                                            onChange={(option: { label: string, value: number }) => {
+                                                setPlaybook({
+                                                    ...playbook,
+                                                    retrospective_reminder_interval_seconds: option ? option.value : option,
```
agarciamontoro, line 554: Is this different than the following?
```suggestion
                                                    retrospective_reminder_interval_seconds: option?.value,
```

webapp/src/types/incident.ts
```diff
@@ -28,6 +28,7 @@ export interface Incident {
     timeline_events: TimelineEvent[];
     retrospective: string;
     retrospective_published_at: number;
+    retrospective_was_canceled: boolean;
```
agarciamontoro, line 31: Is this interface missing the `retrospective_reminder_interval_seconds` field?

agarciamontoro (approved):
Great work! Only non-blocking comments below. Thank you :)

server/incident/service.go
cpoile: hmm, good question, didn't notice that.

server/incident/service.go
crspeller: Ya. I think the idea is actually to remove all the extra statuses. So decoupling will make sense then.

server/incident/service.go
crspeller: Because you can't reschedule without canceling the existing one.

webapp/src/types/incident.ts
crspeller: Yup, but we never use that on the client.

server/incident/reminder.go
crspeller: I could schedule a new one. But now I need to know this magic number in order to cancel it. From the clients point of view it only has one thing it cares about, but if we don't implement this on the job scheduler side, suddenly it has more than one.

cpoile (approved):
Ok, I understand what's going on now. I misunderstood where the reminder was being set and why it was being cancelled. Makes sense now, and yes, that should be fixed! :)

PR #569, opened 2021-05-24
[MM-35825] Message on every keyword

iomodo:
#### Summary
This PR takes a first stab at keyword signals. User is able to specify comma separated keywords which will trigger the ephemeral message with the list of relevant playbooks. 

https://user-images.githubusercontent.com/2340127/119346824-c1b18c00-bcab-11eb-8bbf-866b8a5d1c56.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-35825
  Fixes: https://mattermost.atlassian.net/browse/MM-36007

#### Checklist

~~[ ] Telemetry updated~~
~~[ ] Gated by experimental feature flag~~
[✅] Unit tests updated


server/plugin.go
```diff
@@ -231,3 +235,53 @@ func (p *Plugin) UserHasLeftChannel(c *plugin.Context, channelMember *model.Chan
...
+			return true
+		}
+	}
+	return false
+}
+
+func (p *Plugin) suggestPlaybooksToTheUser(playbooks []*playbook.Playbook, userID, channelID string) {
+	message := "May be you want to run one of these playbooks \n"
+	for _, playbook := range playbooks {
+		message += playbook.Title + "\n"
```
agarciamontoro, line 284: These could be links to the playbooks themselves, maybe.

server/plugin.go
```diff
@@ -231,3 +235,53 @@ func (p *Plugin) UserHasLeftChannel(c *plugin.Context, channelMember *model.Chan
...
+			continue
+		}
+
+		if !isPlaybookTriggeredByMessage(&playbooks[i], post.Message) {
+			continue
+		}
+
+		if err := permissions.PlaybookAccess(post.UserId, playbooks[i], p.pluginAPI); err != nil {
+			continue
+		}
```
agarciamontoro, line 264: In order to get rid of the `PlaybookAccess` call everytime a playbook matches, we could maybe ask for only the playbooks available for this user, instead of all the playbooks with `GetPlaybooks`. We'd need to add a new `GetPlaybooksForUser` (or something similar) to the store, but maybe it would be better? I'd say so, but difficult to know  without measuring, actually. This is just me being concerned with the overall performance impact of the feature.

server/playbook/playbook.go
```diff
@@ -36,6 +36,8 @@ type Playbook struct {
 	WebhookOnCreationEnabled    bool        `json:"webhook_on_creation_enabled"`
 	MessageOnJoin               string      `json:"message_on_join"`
 	MessageOnJoinEnabled        bool        `json:"message_on_join_enabled"`
+	SignalAnyKeywords           string      `json:"signal_any_keywords"`
```
agarciamontoro, line 39: What if we make this a `[]string` and split the string stored in the database when we perform the select, instead of splitting with every message?

webapp/src/components/backstage/automation/patterned_input.tsx
```diff
@@ -11,31 +11,36 @@ import {Toggle} from 'src/components/backstage/automation/toggle';
...
+    placeholderText: string;
+    errorText: string;
+    input: string;
+    type: string;
+    pattern: string;
+    onChange: (updatedInput: string) => void;
 }
 
-export const WebhookOnCreation: FC<Props> = (props: Props) => (
+export const PatternedInput: FC<Props> = (props: Props) => (
```
agarciamontoro, line 23: Nice refactor :ok_hand:

agarciamontoro (changes requested):
Great work, @iomodo, thank you :tada:

I'm a bit concerned with the performance impact of the whole feature (not actually of your specific implementation, which seems quite sensible). I'm not sure whether the ratio between the improvement to the discoverability of playbooks and the impact on performance is worth it, but I think we will know when we ship it.

iomodo:
Heads up @agarciamontoro @crspeller 
Implemented optimization - playbooks are being cached and retrieved from DB only if they've been updated. Still on every user message we have to get a channel and make a query to DB to know if playbooks have been updated.

Do not pay attention to the ephemeral message format. I will submit a different PR to make it more presentable/usable.

server/sqlstore/migrations.go
```diff
@@ -744,4 +745,53 @@ var migrations = []Migration{
...
+				}
+
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "UpdatedAt", "BIGINT NOT NULL DEFAULT 0"); err != nil {
+					return errors.Wrapf(err, "failed adding column UpdatedAt to table IR_Playbook")
+				}
+				if _, err := e.Exec("UPDATE IR_Playbook SET UpdatedAt = CreateAt"); err != nil {
+					return errors.Wrapf(err, "failed setting default value in column UpdatedAt of table IR_Playbook")
+				}
+				if _, err := e.Exec(`ALTER TABLE IR_Playbook ADD INDEX IR_Playbook_UpdatedAt (UpdatedAt)`); err != nil {
+					if !strings.Contains(err.Error(), "Duplicate key name") { // Instead of IF NOT EXISTS which is not supported
```
iomodo, line 771: Please pay attention to this. When you name the index mysql is throwing an `duplicate key name` error if you try to execute twice. Another way to implement this is to create something like `IF NOT EXISTS` by querying the index first. I thought error throwing was less effort :)

server/playbook/service.go
```diff
@@ -88,3 +97,76 @@ func (s *service) Delete(playbook Playbook, userID string) error {
...
+		s.api.Log.Error("can't get channel", "err", channelErr.Error())
+		return triggeredPlaybooks
+	}
+	teamID := channel.TeamId
+
+	if err := s.cacher.UpdatePlaybooksIfNeeded(); err != nil {
+		s.api.Log.Error("can't update playbooks", "err", err.Error())
+		return triggeredPlaybooks
+	}
+	playbooks := s.cacher.GetPlaybooks()
```
agarciamontoro, line 115: Is there any case where we want to call `cacher.GetPlaybooks` without calling `UpdatePlaybooksIfNeeded`? If not, maybe we can embed everything in `GetPlaybooks`. What do you think?

webapp/src/types/playbook.ts
```diff
@@ -138,8 +142,8 @@ export function isPlaybook(arg: any): arg is Playbook {
         typeof arg.broadcast_channel_id === 'string' &&
         typeof arg.reminder_message_template == 'string' &&
         typeof arg.reminder_timer_default_seconds == 'number' &&
-        arg.invited_user_ids && Array.isArray(arg.invited_user_ids) && arg.checklists.every((id: any) => typeof id === 'string') &&
-        arg.invited_group_ids && Array.isArray(arg.invited_group_ids) && arg.checklists.every((id: any) => typeof id === 'string') &&
+        arg.invited_user_ids && Array.isArray(arg.invited_user_ids) && arg.invited_user_ids.every((id: any) => typeof id === 'string') &&
+        arg.invited_group_ids && Array.isArray(arg.invited_group_ids) && arg.invited_group_ids.every((id: any) => typeof id === 'string') &&
```
agarciamontoro, line 146: Whoops, good catch!

agarciamontoro (changes requested):
This is great, I really liked reading through it :) Thank you!

I left some comments, but nothing major.

server/playbook/service_test.go
```diff
@@ -0,0 +1,393 @@
...
+		telemetryService := &telemetry.NoopTelemetry{}
+		s := playbook.NewService(store, poster, telemetryService, client)
+
+		channelID := model.NewId()
+		teamID := model.NewId()
+		channel := model.Channel{Id: channelID, TeamId: teamID}
+		pluginAPI.On("GetChannel", channelID).Return(&channel, nil)
+
+		firstCall := store.EXPECT().GetTimeLastUpdated().Return(int64(1000), nil)
+		store.EXPECT().GetTimeLastUpdated().Return(int64(1100), nil).After(firstCall)
```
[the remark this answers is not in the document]
agarciamontoro: Oh, but it got updated with the max time in the playbooks, which is 1100. Is that right?

server/playbook/service_test.go
```diff
@@ -0,0 +1,393 @@
...
+				UpdatedAt:         1100,
+				TeamID:            teamID,
+				SignalAnyKeywords: []string{"bla", "something"},
+			},
+			{
+				ID:                model.NewId(),
+				Title:             "playbook 3",
+				UpdatedAt:         900,
+				TeamID:            teamID,
+				SignalAnyKeywords: []string{" some", "other"},
```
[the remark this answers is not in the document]
iomodo: It is. Generally a keyword can be any string not containing a comma.

server/playbook/service_test.go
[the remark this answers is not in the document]
iomodo: Correct, time should be greater than `1100` to trigger an update from the DB

server/sqlstore/migrations.go
```diff
@@ -744,4 +745,53 @@ var migrations = []Migration{
...
+				}
+
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "SignalAnyKeywordsEnabled", "BOOLEAN DEFAULT FALSE"); err != nil {
+					return errors.Wrapf(err, "failed adding column SignalAnyKeywordsEnabled to table IR_Playbook")
+				}
+
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "UpdatedAt", "BIGINT NOT NULL DEFAULT 0"); err != nil {
+					return errors.Wrapf(err, "failed adding column UpdatedAt to table IR_Playbook")
+				}
+				if _, err := e.Exec("UPDATE IR_Playbook SET UpdatedAt = CreateAt"); err != nil {
```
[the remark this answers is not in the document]
iomodo: Here it is https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/7f144404dcdeec0f6be715679f13c318ae807a47/server/sqlstore/migrations.go#L787

server/sqlstore/playbook_test.go
```diff
@@ -1194,6 +1212,382 @@ func TestDeletePlaybook(t *testing.T) {
...
+	expected := int64(600)
+	for _, driverName := range driverNames {
+		db := setupTestDB(t, driverName)
+		playbookStore := setupPlaybookStore(t, db)
+
+		t.Run("zero playbooks", func(t *testing.T) {
+			result, err := playbookStore.GetPlaybooks()
+			require.NoError(t, err)
+			require.ElementsMatch(t, []playbook.Playbook{}, result)
+		})
```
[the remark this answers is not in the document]
iomodo: Great catch!

server/sqlstore/migrations.go
```diff
@@ -744,4 +745,53 @@ var migrations = []Migration{
...
+				}
+				if _, err := e.Exec("UPDATE IR_Playbook SET ConcatenatedSignalAnyKeywords = '' WHERE ConcatenatedSignalAnyKeywords IS NULL"); err != nil {
+					return errors.Wrapf(err, "failed setting default value in column ConcatenatedSignalAnyKeywords of table IR_Playbook")
+				}
+
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "SignalAnyKeywordsEnabled", "BOOLEAN DEFAULT FALSE"); err != nil {
+					return errors.Wrapf(err, "failed adding column SignalAnyKeywordsEnabled to table IR_Playbook")
+				}
+
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "UpdatedAt", "BIGINT NOT NULL DEFAULT 0"); err != nil {
```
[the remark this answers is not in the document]
iomodo: Yep you are right!

server/playbook/service.go
iomodo: Good point

server/sqlstore/migrations.go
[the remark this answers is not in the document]
agarciamontoro: I meant something like `time.Now()`, instead of the `CreateAt` timestamp.

agarciamontoro (approved):
It looks good! The only thing is the `CreateAt` / `time.Now()` issue (which I defer to your opinion, but wanted to clarify what I meant!)

server/sqlstore/migrations.go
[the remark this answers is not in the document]
iomodo: Sorry, totally misread your comment.

server/plugin.go
crspeller: Feature seems pretty useless without this. I am guessing the plan is to be able to start an incident from here?

server/playbook/cache.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type CacherImpl struct {
+	updateAt  int64
+	playbooks []*CachedPlaybook
+	store     Store
+	logger    pluginapi.LogService
+}
+
+type Cacher interface {
```
crspeller, line 22: The naming here is confusingly generic for something that solves a very specific problem.

server/sqlstore/playbook.go
```diff
@@ -285,6 +289,89 @@ func (p *playbookStore) GetNumPlaybooksForTeam(teamID string) (int, error) {
...
+		}
+		playbooks = append(playbooks, out)
+	}
+	return playbooks, nil
+}
+
+// GetTimeLastUpdated retrieves time last playbook was updated at.
+// Passed argument determins whether to include playbooks with
+// SignalAnyKeywordsEnabled flag or not.
+func (p *playbookStore) GetTimeLastUpdated(onlyPlaybooksWithKeywordsEnabled bool) (int64, error) {
```
crspeller, line 326: This has to be run for every cache request. Do we know that this query is actually significantly faster than the above? Otherwise there is a lot of complexity here that might be avoided.

server/playbook/playbook.go
```diff
@@ -185,6 +198,17 @@ type Store interface {
 	// GetNumPlaybooksForTeam retrieves the number of playbooks in a given team
 	GetNumPlaybooksForTeam(teamID string) (int, error)
 
+	// GetPlaybooksWithKeywords retrieves all playbooks with keywords enabled
+	GetPlaybooksWithKeywords(opts Options) ([]Playbook, error)
+
+	// GetTimeLastUpdated retrieves time last playbook was updated at.
+	// Passed argument determins whether to include playbooks with
+	// SignalAnyKeywordsEnabled flag or not.
+	GetTimeLastUpdated(onlyPlaybooksWithKeywordsEnabled bool) (int64, error)
```
crspeller, line 207: Why is there a boolean option here? I don't see anywhere we use this with `false`.

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -372,6 +372,53 @@ const PlaybookEdit: FC<Props> = (props: Props) => {
...
+            final_keywords = [keywords];
+        } else {
+            const typingKeyword = keywords.substring(index + 1);
+            const alreadyTypedKeywords = keywords.substring(0, index);
+            const split_keywords = alreadyTypedKeywords.split(',').filter((keyword) => keyword);
+            const unique_keywords = [...new Set(split_keywords)];
+            final_keywords = [...unique_keywords, typingKeyword];
+        }
+
+        if (!hasSameElements(playbook.signal_any_keywords, final_keywords)) {
```
crspeller, line 388: Why bother checking?

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -372,6 +372,53 @@ const PlaybookEdit: FC<Props> = (props: Props) => {
         setChangesMade(true);
     };
 
+    const handleSignalAnyKeywordsChange = (keywords: string) => {
+        const index = keywords.lastIndexOf(',');
+        let final_keywords: string[];
+        if (index === -1) {
+            final_keywords = [keywords];
+        } else {
+            const typingKeyword = keywords.substring(index + 1);
```
crspeller, line 381: Why bother doing any of this? Why not just split on `,`?

server/sqlstore/migrations.go
```diff
@@ -744,4 +745,54 @@ var migrations = []Migration{
...
+					return errors.Wrapf(err, "failed adding column SignalAnyKeywordsEnabled to table IR_Playbook")
+				}
+
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "UpdateAt", "BIGINT NOT NULL DEFAULT 0"); err != nil {
+					return errors.Wrapf(err, "failed adding column UpdateAt to table IR_Playbook")
+				}
+				if _, err := e.Exec("UPDATE IR_Playbook SET UpdateAt = CreateAt"); err != nil {
+					return errors.Wrapf(err, "failed setting default value in column UpdateAt of table IR_Playbook")
+				}
+				if _, err := e.Exec(`ALTER TABLE IR_Playbook ADD INDEX IR_Playbook_UpdateAt (UpdateAt)`); err != nil {
```
crspeller, line 770: Why an index on UpdateAt?

crspeller (changes requested):

server/sqlstore/playbook.go
iomodo: If number of playbooks is small time difference will be negligent. But we decided to go this route in case of many playbooks.

server/playbook/playbook.go
iomodo: Correct, `false` is not used. I thought creating a method like `GetTimeLastUpdatedForPlaybooksWithKeywordsEnabled` would be mouthful. Maybe we will need to get last updated playbook in the future, the method will be there.

webapp/src/components/backstage/playbook_edit.tsx
iomodo: To use `setPlaybook` only when there is an actual update.

webapp/src/components/backstage/playbook_edit.tsx
iomodo: There is a difference between representation and data storage. For example
`one,two,three` is saved as a list `["one", "two", "three"]` and later represented as `one,two,three`
`one,two,three,` is saved as a list `["one", "two", "three", ""]` and later represented as `one,two,three,`
If we just split on`,` we will lose the last element in the array and store `one,two,three,` as `["one", "two", "three"]` and later represent as `one,two,three`. In fact we will not be able to even represent that because the commas will be lost.
`one,` will be saved as `["one"]` and represented as `one` later.

server/sqlstore/migrations.go
iomodo: We have an index here to perform the `MAX` on `UpdateAt` field, which will be logarithmic complexity with the `B-Tree`. Exactly this makes the caching strategy faster for the many playbooks case.

iomodo:
> Feature seems pretty useless without this. I am guessing the plan is to be able to start an incident from here?

Correct, my next PR will do exactly that.

server/playbook/cache.go
iomodo: Will rename it to `KeywordsCacher`

server/sqlstore/playbook.go
crspeller: I am not convinced this is worth the cost. I wonder how many playbooks we would need to see a difference?

webapp/src/components/backstage/playbook_edit.tsx
crspeller: How are you supposed to actually cause this to happen? Any edit would cause this, even if it reverts later.

webapp/src/components/backstage/playbook_edit.tsx
crspeller: Split trims the last element? 
Also I think if I type something like `re, result` The second re will be deleted before I finish result? 
This appears to be too much control over input. I would suggest maybe just using a text field, otherwise you can get strangeness. Like what happens if I try to add a comma right after another one in the middle?

server/sqlstore/playbook.go
iomodo: I did some benchmarking here are some data points `getTimeLastUpdated` vs `GetPlaybooksForKeywords`.

10-30 playbooks - same
50-100 playbooks - 20% time decrease
200-500 playbooks - 40% time decrease
600-1000 playbooks - 50% time decrease - twice as fast. In absolute terms `3ms` vs `6ms`.

When measured difference is not that big. Maybe it's better to avoid all the caching?

P.S. I tested this before but I compared `GetLastUpdate` with `GetPlaybooks`(which already existed) and there was a big difference for 10-30 playbooks twice as fast, for 100-200 5 times as fast.

Thoughts? cc @agarciamontoro @crspeller

webapp/src/components/backstage/playbook_edit.tsx
iomodo: Yes, that is correct, I'll remove the check

webapp/src/components/backstage/playbook_edit.tsx
iomodo: `re, result` will work fine. 
If you repeat the keyword it will get deleted. That is if you'll type `re,re,` you will get `re,`. Do you think we should remove this feature?

server/sqlstore/playbook.go
agarciamontoro: I'm not sure whether any normal server would get to 600 playbooks :thinking: But I'd say that even a 20% performance improvement is worth it, given that this happens in every message posted in the whole server.

webapp/src/components/backstage/playbook_edit.tsx
crspeller: I would defer to @abhijit-singh for an answer to that.

server/sqlstore/playbook.go
crspeller: Alright. I think we can keep it then. I would happen a lot on a production server.

webapp/src/components/backstage/playbook_edit.tsx
iomodo: Talked to @abhijit-singh offline, removed extra control over the input. Duplicate keywords will still be removed on save.

cc @crspeller ready for your review.

crspeller (approved):

PR #571, opened 2021-05-25
MM-35951 - Fix default message on join

cpoile:
#### Summary
- Fix how we insert the default message on join, that was the problem.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-35951

#### Checklist
[ ] ~~Telemetry updated~~
[ ] ~~Unit tests updated~~


prapti:
Sorry, got a bit late in commenting as I was giving it a quick pass. But everything looked good here and no objection (just for the record :) )

PR #574, opened 2021-05-25
Fix channel selector selector

lieut-data:
#### Summary
The channel selector did not respond to Redux changes to the available channels. While this is rarely a problem in normal usage, when the set of channels aren't likely to change in the backstage, it did cause issues during e2e tests when the selector raced with the messaging application's initial load of available channels.

Switch from `AsyncSelect` to just a controlled `Select` component that directly renders from the selected channels. While we're in here, exclude DMs and GMs from the list of selectable channels.

#### Ticket Link
None

#### Checklist
[ ] ~~Telemetry updated~~
[ ] ~~Unit tests updated~~


cpoile (approved):

agarciamontoro (approved):
Makes sense, thank you! :tada:

PR #579, opened 2021-06-03
MM-35881 Better retrospective reminder message.

crspeller:
#### Summary
Replaces the reminder message for retrospective with a custom one. The old style is used as fallback for mobile. 

First reminder when reminder is set to happen only once:
image

First reminder but is configured for 1h:
image

Subsequent reminders (ignore the <1m it's a test value):
image

Dark theme (@abhijit-singh Do we want the colours in the SVG to change based on theme? If so can you provide an SVG with the appropriate values?):
image

@justinegeffen Adding you for final check on the copy

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35881

#### Checklist
~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


abhijit-singh:
@crspeller This is awesome! Here's the link to the SVG with the updated colours, should work better with all themes -
https://www.figma.com/file/1u5HZZMamACauYfR8121Uj/MM-35038-Retrospective-Report?node-id=302%3A349885

Not attaching the image directly cause I'm not sure if you exported it with or without the background circle.

Also, small nit, can we increase the left-spacing between text and the box to 16px? Was exactly like what you have here in the design, but seems a little tight now that I look at it. Thanks!
image

webapp/src/components/retrospective_reminder_posts.tsx
```diff
@@ -0,0 +1,526 @@
...
+
+const ReminderCommon = (props: ReminderCommonProps) => {
+    const incident = useSelector(currentIncident);
+    const reminderDuration = incident?.retrospective_reminder_interval_seconds;
+    const currentTeam = useSelector(getCurrentTeam);
+
+    let reminderText = (
+        <>{'You will not be reminded again.'}</>
+    );
+    if (reminderDuration !== 0) {
```
agarciamontoro, line 79: Not very important, but this can be undefined, which is different than 0.

agarciamontoro (approved):
This is so cool, I did not even know that the registerPostTypeComponent existed! Thanks :tada:

crspeller:
@abhijit-singh Here is the updated margin and SVG (I exported with the circle and combined):
image

iomodo (approved):
Looks Great @crspeller !
Just a question did you check how it renders on the mobile? Should we be checking?

abhijit-singh:
Okay, not really sure how to solve it then. I'm using the theme variables in Figma and the tool we use to test in there is giving me the correct results. 

@agarciamontoro You had colour-changing SVGs in the pricing modal that worked fine, can you help find the solution here?

image
image

crspeller:
@iomodo I did. It renders the fallback, so as it did before.

crspeller:
@abhijit-singh When I click the export button in figma the resulting SVG has hard coded color values. Is there a way to export with the appropriate variables? Can you try exporting and see what you get?

abhijit-singh:
Yep Figma exports SVGs with hardcoded color values. They probably need to be manually swapped with the color variables we have on the dev side.

crspeller:
@abhijit-singh Ah ok. Then I just need the mapping of the colour values to the colour variables.  Can you provide that?

webapp/src/components/retrospective_reminder_posts.tsx
```diff
@@ -0,0 +1,576 @@
...
+            <Image/>
+        </MessageContainer>
+    );
+};
+
+export const RetrospectiveFirstReminder = () => {
+    return (
+        <ReminderCommon
+            header={'Would you like to fill out the retrospective report?'}
+            primary={'Yes, fill retrospective'}
```
justinegeffen, line 121:
```suggestion
            primary={'Yes, start retrospective'}
```

webapp/src/components/retrospective_reminder_posts.tsx
```diff
@@ -0,0 +1,576 @@
...
+            secondary={'No, skip retrospective'}
+        />
+    );
+};
+
+export const RetrospectiveReminder = () => {
+    return (
+        <ReminderCommon
+            header={'Reminder to fill out the retrospective'}
+            primary={'Fill retrospective'}
```
justinegeffen, line 131:
```suggestion
            primary={'Start retrospective'}
```

justinegeffen (commented):
Just two suggestions, thank you! :)

crspeller:
@abhijit-singh Took a guess. Seems to look good. 
image

crspeller:
Since the e2e passed before, the failures are nothing to do with this PR. Merging.

PR #580, opened 2021-06-03
Add hot module loading.

crspeller:
#### Summary
Adds the webpack dev server and hot module loading as an option for development. This allows near instant updates for client changes without a refresh. Perfect for tweaking CSS or fiddling with JSX. 

#### Usage
To use run `make dev` then visit http://localhost:9005 instead of the usual 8065. 

#### Known issues
- Works normally while looking at the main view, however in the backstage it requires you to click on your browser to show changes. Not sure why this is.


Makefile
```diff
@@ -134,6 +134,11 @@ else
 	cd webapp && $(NPM) run debug:watch
 endif
 
+## Builds and installs the plugin to a server, then starts the webpack dev server on 9005
+.PHONY: watch
+dev: apply server bundle
+	cd webapp && $(NPM) run dev-server
```
cpoile, line 140: Will this conflict with the watch above? (line 129)
And what is different from `run debug:watch`? If this is better, then maybe just replace debug:watch (and remove it) with this `run dev-server`?

cpoile (changes requested):
sweet! Just a question about the previous watch command we had

webapp/webpack.config.js
```diff
@@ -94,3 +101,39 @@ module.exports = {
     mode,
     plugins,
 };
+
+if (targetIsDevServer) {
+    config = {
+        ...config,
+        devtool: 'eval-cheap-module-source-map',
```
lieut-data, line 108: I love this for speed, but it's annoying when I need more data. This also forces the `devtool` vs. letting it get defined above when we first set `devtool`. Should we at least move this up there, and perhaps later iterate on a way to "build fast" vs. "debug slow"?

Makefile
crspeller: If you run them both at the same time? Probably. 
I was hoping we would try it out and see how we liked it. If it is exclusively better we would remove watch. 
Or maybe you are confused because the PHONY is wrong... fixed!

lieut-data (approved):
Neat!

agarciamontoro (approved):
I quickly tested it and it's much faster than `make watch`, I like it! I was also going to ask whether it would be nice to remove `watch`, but I agree we can test it for a few weeks before making a decision. Thanks, @crspeller!!

iomodo (approved):
wonderful!

Makefile
cpoile: Indeed. Thanks! :)

cpoile (approved):

PR #581, opened 2021-06-04
MM-35695 - First pass at new playbook landing page

cpoile:
#### Summary
- First pass. Backend is (should be) complete and correct. New stats, was able to turn all stats calls into single queries (thanks @lieut-data for the pointer to `sum(case`).
- For the frontend I duplicated the incident list component from `BackstageIncidentList` because there were enough changes that making it generic would be a pain, and I'm still not sure if we're making them both look exactly the same. If we do end up wanting that, I think I can make `BackstageIncidentList` use the new IncidentList component.
- I wanted to get this work under review, rather than waiting for everything to be perfect. So the frontend is not tweaked yet -- there are a few little things like navigation and the changes to IncidentList, but otherwise it's close to the designs. 

Where we are:
- [x] Backend stats, rough frontend. <-- We are here
- [ ] New navigation model (between list, landing, edit, messaging)
- [ ] UI tweaks to bring into 100% in line with designs
- [ ] license checked and upgrade bannered

Looks like:
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35695

#### Checklist

- [ ] Telemetry updated
- [x] Unit tests updated


lieut-data:
This is stunning, @cpoile! Looking forward to finishing this review :)

server/sqlstore/stats.go
```diff
@@ -23,11 +27,21 @@ func NewStatsStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLStore
...
-	TeamID string
+	TeamID     string
+	PlaybookID string
 }
 
 func applyFilters(query sq.SelectBuilder, filters *StatsFilters) sq.SelectBuilder {
-	return query.Where(sq.Eq{"i.TeamId": filters.TeamID})
+	ret := query
+
+	if filters.TeamID != "" {
```
lieut-data, line 37: There's a subtle change here, in that an empty team id no longer constrains the query at all (so presumably "no results") -- just noting that explicitly in case it has unexpected access side effects.

server/sqlstore/stats.go
```diff
@@ -75,7 +106,30 @@ func (s *StatsStore) TotalActiveParticipants(filters *StatsFilters) int {
...
 	if err := s.store.getBuilder(s.store.db, &total, query); err != nil {
-		s.log.Warnf("Error retriving stat total active participants %w", err)
+		s.log.Warnf("Error retrieving stat total active participants %w", err)
+		return -1
+	}
+
+	return total
+}
+
+// RunsFinishedBetweenDays are calculated from startDay to endDay (inclusive), where "days"
```
lieut-data, line 116: What's the implicit timezone in these calculations?

server/sqlstore/stats.go
```diff
@@ -75,7 +106,30 @@ func (s *StatsStore) TotalActiveParticipants(filters *StatsFilters) int {
...
+// would be 0 (days ago) (up until now).
+func (s *StatsStore) RunsFinishedBetweenDays(filters *StatsFilters, startDay, endDay int) int {
+	dayInMS := int64(86400000)
+	startInMS := beginningOfTodayMillis() - int64(startDay)*dayInMS
+	endInMS := endOfTodayMillis() - int64(endDay)*dayInMS
+
+	query := s.store.builder.
+		Select("COUNT(i.Id) as Count").
+		From("IR_Incident as i").
+		Where(sq.And{sq.Expr("i.EndAt > ?", startInMS), sq.Expr("i.EndAt < ?", endInMS)})
```
lieut-data, line 127: nit: I usually expect to see at least one include range, i.e. `>` and `<=`, otherwise events can "go missing" in consecutive queries. Probably not an issue.

server/sqlstore/stats.go
```diff
@@ -129,6 +183,159 @@ func (s *StatsStore) MovingWindowQueryActive(query sq.SelectBuilder, numDays int
...
+func (s *StatsStore) RunsStartedPerWeekLastXWeeks(x int, filters *StatsFilters) ([]int, []string) {
+	day := int64(86400000)
+	week := day * 7
+	startOfWeek := beginningOfTodayMillis() - week
+	endOfWeek := endOfTodayMillis()
+	var selectStatements []string
+	var weeksAsStrings []string
+
+	for i := 0; i < x; i++ {
+		selectStatements = append(selectStatements,
```
lieut-data, line 195: nit: instead of maintaining an array slice, is it possible to "start" the builder and just keep calling `q.Select(...)` to let that do its thing instead?

server/sqlstore/stats.go
```diff
@@ -129,6 +183,159 @@ func (s *StatsStore) MovingWindowQueryActive(query sq.SelectBuilder, numDays int
...
+			fmt.Sprintf("SUM(CASE WHEN i.createat < %d AND "+
+				"i.createat > %d THEN 1 ELSE 0 END) week%d", endOfWeek, startOfWeek, i))
+
+		// use the middle of the day to get the date, just in case
```
lieut-data, line 199: I guess this would be implicitly UTC+12, or Kamchatka time? I'd be inclined to stick with UTC.

server/sqlstore/stats.go
```diff
@@ -129,6 +183,159 @@ func (s *StatsStore) MovingWindowQueryActive(query sq.SelectBuilder, numDays int
...
+		// COUNT( DISTINCT( CASE: the CASE will return the userId if the row satisfies the conditions,
+		// therefore COUNT( DISTINCT will return the number of unique userIds
+		//
+		// first two lines of the WHEN: an incident was active if it was created before the
+		// end of the day and ended after the start of the day (or still active)
+		//
+		// second two lines: a user was active in the same way--if they joined before the
+		// end of the day and left after the start of the day (or are still in the channel)
+		selectStatements = append(selectStatements,
+			fmt.Sprintf("COUNT( DISTINCT (CASE "+
```
lieut-data, line 276: Instead of string concatenation, can we just start a raw literal that spans multiple lines?

server/api/stats.go
```diff
@@ -46,6 +52,19 @@ type Stats struct {
 	AverageStartToResolved []int `json:"average_start_to_resolved"`
 }
 
+type PlaybookStats struct {
+	RunsInProgress                 int      `json:"runs_in_progress"`
+	ParticipantsActive             int      `json:"participants_active"`
+	RunsFinishedPrev30Days         int      `json:"runs_finished_prev_30_days"`
+	RunsFinishedPercentageChange   int      `json:"runs_finished_percentage_change"`
+	RunsStartedPerWeek             []int    `json:"runs_started_per_week"`
+	RunsStartedPerWeekLabels       []string `json:"runs_started_per_week_labels"`
```
lieut-data, line 61: I'm not a huge fan of building and returning the labels -- feels like a client-side concern IMO, and I guess it also requires pairing up the label with the value manually. Minimally, should we send down a `map[string]int` instead? Or maybe just a `map[int]int` and let the client render the date from the milliseconds as desired?

lieut-data (changes requested):
A handful of comments around the API format and SQL queries, but looking good!

server/api/stats.go
```diff
@@ -1,34 +1,40 @@
...
-		statsStore:   statsStore,
+		ErrorHandler:    &ErrorHandler{log: log},
+		pluginAPI:       api,
+		log:             log,
+		statsStore:      statsStore,
+		playbookService: playbookService,
 	}
 
 	statsRouter := router.PathPrefix("/stats").Subrouter()
+	statsRouter.HandleFunc("/playbook", handler.playbookStats).Methods(http.MethodGet)
```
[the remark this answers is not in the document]
cpoile: I thought about that-- i agree `/playbooks/id/stats` reads better. But the existing stats functions (some of which I reuse), the stats package itself, the stats service/store are all accessed from the `/stats` route handler, so I decided it would be more natural to put it there.

server/sqlstore/stats.go
cpoile: I tried to think of a case where this would break, but I'm pretty sure it's okay. Any access to the stats that assume a `filters.TeamId` value goes through the `parseStatsFilters`, and will reject requests without a `teamId`. So I think we're okay here.

server/sqlstore/stats.go
cpoile: The server's local timezone. Maybe this was a bad assumption? Should we be sending the client's timezone with the request? Working with time is so annoying. :)

server/sqlstore/stats.go
cpoile: Yep, good point just to be safe. I changed it to be inclusive on one end of the range for all the selects (the end of the day).

server/sqlstore/stats.go
cpoile: I tried, and afaict it looks like you can only call the select once. You can add where's and so forth as you like, but the select returns a SelectBuilder, which won't let you add more selects.

server/sqlstore/stats.go
cpoile: Nope, server localtime. This +halfday was purely defensive on my part. I just didn't want to deal with wierdness, it's only to find the date itself.

server/sqlstore/stats.go
cpoile: 👍 
I wasn't sure how to properly indent it (is there one "proper" way?), so I used the way intellij indents its SQL.

server/api/stats.go
cpoile: @lieut-data I suppose we could return the millis for each date, and let the client convert it. But that might introduce some wierdness with the labelling -- the client may display a timestamp as the wrong day from what the server was actually measuring.. I'm not sure how to solve this perfectly, so I decided on having the server declare the truth, and send that to the client. 

Open to other approaches though. What do you think? If we send the client's timezone, then we could send back millis. That's one option.

server/api/stats.go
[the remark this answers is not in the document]
crspeller: The idea of separating it is that `/playbooks` and `/incidents` deal with specificly those, but `/stats` can be across the whole system.

server/sqlstore/stats.go
crspeller: The perils of writing our own stats :P

server/api/stats.go
crspeller: I like the idea of including the TZ in the request. Not sure how to get the correct day boundaries from the client's perspective otherwise.

crspeller (approved):
Looking good!

server/api/stats.go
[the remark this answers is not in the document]
lieut-data: I think the plan is to remove `/stats`, at least as a landing page. When we re-introduce a home page, is it likely that we'd want a `/home`, or would it be better to compose widgets that query dedicated `/playbooks` and `/incidents` pages? Is there an example of a stat endpoint that /should/ return something for both playbooks and incidents at the same time?

Ultimately, it's immaterial to our main user experience, but I think it speaks to whether we want `stats` as a first class /API/ concept or not. Maybe we promote it later when we have a concrete use case (and demote it now?)

server/sqlstore/stats.go
lieut-data: Do we test that querying with an empty `TeamId` at the SQL layer works "as expected"?

server/sqlstore/stats.go
lieut-data: I'd vote for forcing UTC for everyone.

server/sqlstore/stats.go
lieut-data: What about https://pkg.go.dev/github.com/Masterminds/squirrel?utm_source=godoc#SelectBuilder.Column? Alternatively, I believe `Select` also accepts a spread directly.

server/sqlstore/stats.go
lieut-data: I'm a fan of indenting like this: https://github.com/mattermost/mattermost-server/blob/12e439a98ddc6062591cc36da3b845e72f599b30/store/sqlstore/post_store.go#L381-L405

Of course, I brought that style with me... :)

lieut-data: (Basically, just start the tick, then indent like any other code.)

server/api/stats.go
lieut-data: Propose deferring client-defined time. This has the added complexity of having different team members seem different graphs. UTC seems fine, IMO.

server/api/stats.go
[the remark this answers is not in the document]
cpoile: >  Is there an example of a stat endpoint that /should/ return something for both playbooks and incidents at the same time?

For the stats that I added, there's no reason you couldn't use them for all incidents in a team, we just happen to be adding a `where i.playbookid = id` to the select. That's kind of why I think I like have a separate stats endpoint and package--there's no reason the stats can't apply team-wide (or org wide, if you have no `where` clause).

server/sqlstore/stats.go
cpoile: Yes, we do now. :)

server/sqlstore/stats.go
cpoile: Sounds good, done.

server/sqlstore/stats.go
cpoile: Yep, `Column` did it. Thanks, didn't know about that!

server/sqlstore/stats.go
cpoile: Yes UTC+12 -- since we're using the exact nanosecond of 12:00:00.000 am, I'm worried that we're basically asking for a weird edge case. Adding some time removes any possibility of an edge case. It could be an hour, or a minute, or whatever.

server/api/stats.go
cpoile: Settled on UTC.  👍

server/sqlstore/stats.go
```diff
@@ -124,7 +123,7 @@ func (s *StatsStore) RunsFinishedBetweenDays(filters *StatsFilters, startDay, en
 	query := s.store.builder.
 		Select("COUNT(i.Id) as Count").
 		From("IR_Incident as i").
-		Where(sq.And{sq.Expr("i.EndAt > ?", startInMS), sq.Expr("i.EndAt < ?", endInMS)})
+		Where(sq.And{sq.Expr("i.EndAt > ?", startInMS), sq.Expr("i.EndAt <= ?", endInMS)})
```
lieut-data, line 126: nit: I find it helpful to write this as:
```suggestion
		Where(sq.And{
		    sq.Expr("i.EndAt > ?", startInMS), 
		    sq.Expr("i.EndAt <= ?", endInMS),
		})
```

lieut-data (approved):
Nice!

PR #584, opened 2021-06-09
MM-36262 Disable buttons on older or published retrospective reminders.

crspeller:
#### Summary
* When a new reminder happens, the buttons on the previous reminder message are disabled. So at any given time, there is only one reminder in-channel for retros that has active buttons.
* When someone clicks no retro, it disables the buttons on the original message and creates an entry in the timeline.
* When someone clicks start retro, it takes them to that page and nothing happens to the message.
* The reminder will repeat at the configured interval according to the playbook until either retro is published or someone chooses to skip it.
* When the retro is published, the buttons on the previous reminder messages are disabled.

#### Disabled messages above and active message below (ignore the <1m)
image

@abhijit-singh I used the disabled styling from the existing buttons. The hover effect still fires when the buttons are disabled (consistent with other buttons around Playbooks) does that make sense?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36262

#### Checklist
~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


abhijit-singh:
> I used the disabled styling from the existing buttons. The hover effect still fires when the buttons are disabled (consistent with other buttons around Playbooks) does that make sense?

@crspeller Disabled buttons should be greyed out according to the [design system](https://zeroheight.com/29be2c109/p/40e456-buttons/t/35949a). I've also seen them work that way in other parts of the product, not with that exact styling always though. And ideally, there shouldn't be a hover effect apart from triggering a tooltip if any.

image

---
I remember the UI platform team working on a [common `button` component](https://mattermost.github.io/compass-components/?path=/story/components-button--default-story&args=disabled:true). Can we start using this?

agarciamontoro (approved):
Looks good, thanks!

iomodo (approved):
Looks great!

crspeller:
Fixed so it is grey and no hover effect. Common components are not ready yet. I asked UI Platform and they say they are close.


image

abhijit-singh (approved):
Awesome, thank you!

PR #587, opened 2021-06-09
s/commander/owner/

lieut-data:
#### Summary
In preparing this change, I first considered just changing the user facing text. This turned out to be extremely complicated to do correctly, with the words "commander" and "owner" often appearing in the same line of code, in strings, but having different intents. This also had the downside of leaving the API with the word "commander" instead of "owner".

This change replaces all instances of commander, except:
* database column names and values: I opted to preserve backwards compatibility here.
* telemetry properties: Avoiding churn on the looker side here.

One other change is included: namely the use of `Role changes` instead of `Commander changes` or `Owner changes`, as per the reviewed terminology spec. I've commented everywhere the changes aren't just a simple string replacement.

I'd love feedback on this approach, as I intend to replicate it for the remaining (and larger!) changes, namely: preserve backwards compatibility for database and telemetry, but otherwise exhaustively replace the old nomenclature. Changing `incident` will be most challenging, but likely also the most rewarding.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36296

#### Checklist
- [x] Telemetry updated
- [ ]  ~~Gated by experimental feature flag~~
- [x] Unit tests updated


server/sqlstore/incident.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
```
lieut-data, line 26: Remembering `commander_changed` here so we can translate in SQL queries below.

server/sqlstore/incident.go
```diff
@@ -53,10 +57,10 @@ type incidentStatusPosts []struct {
 func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLStore) incident.Store {
 	// When adding an Incident column #1: add to this select
 	incidentSelect := sqlStore.builder.
-		Select("i.ID", "c.DisplayName AS Name", "i.Description", "i.CommanderUserID", "i.TeamID", "i.ChannelID",
+		Select("i.ID", "c.DisplayName AS Name", "i.Description", "i.CommanderUserID AS OwnerUserID", "i.TeamID", "i.ChannelID",
```
lieut-data, line 60: Using aliases to preserve existing column names.

server/sqlstore/incident.go
```diff
@@ -53,10 +57,10 @@ type incidentStatusPosts []struct {
...
 			"i.CreateAt", "i.EndAt", "i.DeleteAt", "i.PostID", "i.PlaybookID", "i.ReporterUserID", "i.CurrentStatus",
 			"i.ChecklistsJSON", "COALESCE(i.ReminderPostID, '') ReminderPostID", "i.PreviousReminder", "i.BroadcastChannelID",
-			"COALESCE(ReminderMessageTemplate, '') ReminderMessageTemplate", "ConcatenatedInvitedUserIDs", "ConcatenatedInvitedGroupIDs", "DefaultCommanderID",
+			"COALESCE(ReminderMessageTemplate, '') ReminderMessageTemplate", "ConcatenatedInvitedUserIDs", "ConcatenatedInvitedGroupIDs", "DefaultCommanderID AS DefaultOwnerID",
```
lieut-data, line 63: Using aliases to preserve existing column names.

server/sqlstore/incident.go
```diff
@@ -68,9 +72,30 @@ func NewIncidentStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
...
-			"te.EventType", "te.Summary", "te.Details", "te.PostID", "te.SubjectUserID",
-			"te.CreatorUserID").
+		Select(
+			"te.ID",
+			"te.IncidentID",
+			"te.CreateAt",
+			"te.DeleteAt",
+			"te.EventAt",
+		).
+		// Map "commander_changed" to "owner_changed", preserving database compatibility
```
lieut-data, line 82: Using a `CASE` statement to translate `commander_changed` to `owner_changed`, simplifying client code.

server/sqlstore/incident.go
```diff
@@ -380,6 +405,11 @@ func (s *incidentStore) CreateTimelineEvent(event *incident.TimelineEvent) (*inc
 	}
 	event.ID = model.NewId()
 
+	eventType := string(event.EventType)
```
lieut-data, line 408: Translating /to/ `commander_changed` when writing.

server/sqlstore/incident.go
```diff
@@ -415,14 +445,19 @@ func (s *incidentStore) UpdateTimelineEvent(event *incident.TimelineEvent) error
 		return errors.New("needs event type")
 	}
 
+	eventType := string(event.EventType)
```
lieut-data, line 448: Translating /to/ `commander_changed` when writing.

server/sqlstore/playbook.go
```diff
@@ -46,7 +46,7 @@ func NewPlaybookStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 			"DeleteAt", "NumStages", "NumSteps", "BroadcastChannelID",
 			"COALESCE(ReminderMessageTemplate, '') ReminderMessageTemplate", "ReminderTimerDefaultSeconds",
 			"ConcatenatedInvitedUserIDs", "ConcatenatedInvitedGroupIDs", "InviteUsersEnabled",
-			"DefaultCommanderID", "DefaultCommanderEnabled",
+			"DefaultCommanderID AS DefaultOwnerID", "DefaultCommanderEnabled AS DefaultOwnerEnabled",
```
lieut-data, line 49: Alias to preserve column name.

server/telemetry/rudder.go
```diff
@@ -28,7 +28,7 @@ const (
 	actionCreate                   = "create"
 	actionEnd                      = "end"
 	actionRestart                  = "restart"
-	actionChangeCommander          = "change_commander"
+	actionChangeOwner              = "change_commander"
```
lieut-data, line 31: Telemetry /value/ doesn't change.

server/telemetry/rudder.go
```diff
@@ -126,7 +126,7 @@ func incidentProperties(incdnt *incident.Incident, userID string) map[string]int
 		"UserActualID":        userID,
 		"IncidentID":          incdnt.ID,
 		"HasDescription":      incdnt.Description != "",
-		"CommanderUserID":     incdnt.CommanderUserID,
+		"CommanderUserID":     incdnt.OwnerUserID,
```
lieut-data, line 129: Telemetry /property/ doesn't change.

webapp/src/components/backstage/incidents/incident_backstage/retrospective/timeline_retro.tsx
```diff
@@ -126,9 +126,9 @@ const TimelineRetro = (props: { incident: Incident }) => {
             value: 'divider',
         } as CheckboxOption,
         {
-            display: 'Commander changes',
-            value: TimelineEventType.CommanderChanged,
-            selected: eventsFilter.commander_changed,
+            display: 'Role changes',
```
lieut-data, line 129: Showing `Role changes` here, not just `Owner changes`. I don't think the underlying logic supports multiple event filtering, but there are no other role changes at present anyway.

webapp/src/components/rhs/rhs_about.tsx
```diff
@@ -21,6 +21,8 @@ import {useLatestUpdate, useProfilesInCurrentChannel} from 'src/hooks';
 import PostText from 'src/components/post_text';
 import {updateStatus} from 'src/actions';
 
+import Duration from '../duration';
```
lieut-data, line 24: I think this was done by my editor on applying linting changes.

webapp/src/components/rhs/rhs_about.tsx
```diff
@@ -89,10 +91,10 @@ const RHSAbout = (props: Props) => {
                 <Row>
                     <div className='side-by-side'>
                         <div className='inner-container first-container'>
-                            <div className='first-title'>{'Commander'}</div>
+                            <div className='first-title'>{'Owner'}</div>
                             <ProfileSelector
-                                selectedUserId={props.incident.commander_user_id}
-                                placeholder={'Assign Commander'}
+                                selectedUserId={props.incident.owner_user_id}
+                                placeholder={'Assign the owner role'}
```
lieut-data, line 97: I can't actually get this placeholder text to show up -- any ideas?

webapp/src/components/rhs/rhs_list_incident.tsx
```diff
@@ -14,9 +14,11 @@ import Profile from 'src/components/profile/profile';
 import Duration from 'src/components/duration';
 import {Incident, incidentCurrentStatus} from 'src/types/incident';
 import {lastUpdatedByIncidentId} from 'src/selectors';
-import StatusBadge from '../backstage/incidents/status_badge';
+
 import {SmallerProfile} from 'src/components/rhs/rhs_shared';
 
+import StatusBadge from '../backstage/incidents/status_badge';
```
lieut-data, line 20: Another automagic editor change.

webapp/src/components/rhs/rhs_timeline.tsx
```diff
@@ -159,9 +159,9 @@ const RHSTimeline = (props: Props) => {
             value: 'divider',
         } as CheckboxOption,
         {
-            display: 'Commander changes',
-            value: TimelineEventType.CommanderChanged,
-            selected: eventsFilter.commander_changed,
+            display: 'Role changes',
```
lieut-data, line 162: Again, `Role changes`, not `Owner changes`. I guess this component is duplicated?

webapp/src/components/backstage/automation/auto_assign_owner.tsx
```diff
@@ -32,12 +32,12 @@ export const AutoAssignCommander = (props: Props) => {
                         isChecked={props.enabled}
                         onChange={props.onToggle}
                     />
-                    <div>{'Assign commander'}</div>
+                    <div>{'Assign the owner role'}</div>
```
lieut-data, line 35: Another minor tweak here, as per spec.

lieut-data:
Adding the whole gang, not to all review, but to provide feedback on the specific approach documented above.

server/sqlstore/incident.go
agarciamontoro: I like this approach!

server/sqlstore/incident.go
agarciamontoro: Maybe worth adding a comment here, as the one for the `CASE` statement?

server/sqlstore/incident.go
agarciamontoro: Same here

webapp/src/components/rhs/rhs_about.tsx
agarciamontoro: No idea how to make it show. Neither here nor in the other places where ProfileSelector is used. That component has been waiting for a refactor for a while.

agarciamontoro (approved):
Looks good! I like the approach taken, particularly in the DB. Good luck with the change for "incident"!

webapp/src/components/rhs/rhs_timeline.tsx
crspeller: https://mattermost.atlassian.net/browse/MM-35475

crspeller (approved):
Approach makes sense. No need to cause telemetry churn. 

You could duplicate the column to do the migration. That way we still get reversion capabilities (loosing new data). 0/5

lieut-data:
> You could duplicate the column to do the migration. That way we still get reversion capabilities (loosing new data). 0/5

Good observation, @crspeller! There's even part of me that stilll wants to break the schema altogether -- what would you say to keeping track of these various items and formally breaking with the v2.0 release later this year? I'll probably leave this PR as-is for now, but would include the column name changes in that list.

crspeller:
@lieut-data I almost suggested that too :P
Let's make that list. A 2.0 break makes sense.

PR #590, opened 2021-06-10
MM-36359 - Fix backstage stats not working on MySql

cpoile:
#### Summary
- This was weird. MySql's sum returns an int (but not always). Then, squirrel interpreted it as a `[]uint8` (!?). This fixcasts it to something we expect.
- Changing the type check to fail and give us some info, in case this happens again
- Remember that nil dereference we got from community a few days ago? Turns out it was because we were using an uninitialized bot to report errors. Fixed that.
- I think I should write tests for the time-related stats, but these pages are changing so fast I don't think it's worth the time -- propose we wait till the features are a little more stable.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36359

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -111,6 +111,7 @@ func (p *Plugin) OnActivate() error {
 	p.config.RegisterConfigChangeListener(toggleTelemetry)
 
 	apiClient := sqlstore.NewClient(pluginAPIClient)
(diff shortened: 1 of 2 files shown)


lieut-data (approved):

crspeller (approved):
Fun!

PR #592, opened 2021-06-11
MM-35626: Adapt upgrade flow to Cloud

agarciamontoro:
#### Summary
This PR adapts the upgrade flow to Cloud, finishing the core work for the Pricing Plan Differentiation epic (there are a few more minor details to finish; namely: generalizing the terminology for removing specific mentions to Professional or Enterprise, and improving the license checks).

The end user experience is not changed at all. The only difference is for System Admins when the server is on Cloud:
- The Call to Action buttons (in the placeholder for the RHS, and in the modals shown when trying to interact with any locked feature) are slightly different: the button reads `Upgrade now`, and it opens the Cloud modal to upgrade the subscription. This modal is the one exported by https://github.com/mattermost/mattermost-webapp/pull/8228. Although we can assume that it is always present in Cloud, the logic to mount it and show it is encapsulated with some checks to try not to break everything if they are not exported.
- When the end user clicks on the `Notify Admin` button in the modals, the System Admins receive a custom post. This is needed because interactive messages cannot open up modals directly, so we need a custom component there to trigger the openModal action. In mobile, this defaults to a link to customers.mattermost.com. Right now there is no direct path to directly upgrade, but they can log in there, see their cloud plan and have the option to upgrade (this need is tracked by https://mattermost.atlassian.net/browse/MM-36350, assigned to the Growth team). 

A few notes:
- The Cloud modal is currently not themed, so it does not look good on dark themes. This is tracked by https://mattermost.atlassian.net/browse/MM-34899, again assigned to the Growth team.
- If you try to test this locally, you won't be able to see anything, as https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/578 hid the UI. Ping me if you need to test it locally, I have a patch on my machine to do so, and you'll need a Cloud license.
- As usual, feel free to review this commit by commit, I tried to make them independent enough.

##### Modal in channel

https://user-images.githubusercontent.com/3924815/121714676-5f93bc00-cade-11eb-85fe-fe391f192266.mp4

##### Modal from RHS

https://user-images.githubusercontent.com/3924815/121714698-64f10680-cade-11eb-8ce6-fac449f50ecc.mp4

##### Modal in backstage

https://user-images.githubusercontent.com/3924815/121714715-68848d80-cade-11eb-8bcd-b061d15ba193.mp4

##### Notification sent to System Admin

https://user-images.githubusercontent.com/3924815/121714728-6b7f7e00-cade-11eb-8edc-9f049ba792eb.mp4

Mobile view:

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35626

#### Checklist

- [ ] ~~Telemetry updated~~
- [X] ~~Gated by experimental feature flag~~ Hidden in the UI until 5.37 is released.
- [ ] ~~Unit tests updated~~


abhijit-singh:
Looks good @agarciamontoro!

A couple points:
- In the third video, `Modal in backstage`, when is the modal getting triggered? Is it when the user clicks on the `Edit playbook` button?
- On mobile, the title (in the shared example - "Create multiple playbooks in Incident Collab...") seems to be smaller than the body text. Can we make the following changes?
  - Have the same font-size for the title as the body-text and make the font-weight **bold**.
  - Add a line break between `Upgrade now` and the body text.

agarciamontoro:
@abhijit-singh, thanks for taking a look!

>  In the third video, `Modal in backstage`, when is the modal getting triggered? Is it when the user clicks on the `Edit playbook` button?

Argh, the framerate of the videos was terrible, sorry, I had not noticed. I re-recorded them and edited the PR description, please take another look. Answering your question: it is triggered from the setting to restrict access to certain users. I just wanted to showcase that the modal is shown wherever the user is, no matter how "deep" in the backstage.

> On mobile, the title (in the shared example - "Create multiple playbooks in Incident Collab...") seems to be smaller than the body text. Can we make the following changes?

Actually, we cannot if we stick to interactive messages. With these type of messages, we send a data structure from the server and it is the client's responsibility (in this case, the mobile app) to render them. I don't know why the app shows it like that, but changing that code there is a bit involved, as it literally changes all interactive messages sent by anybody to any channel; it would require a more involved process.
If it's too bad, I can try to use markdown in the message itself, effectively getting rid of the interactive message for the Cloud case. I guess that would render it better. Thoughts?

agarciamontoro:
Also, all Cloud E2E tests are broken, so I need to review the checks, which are not doing their job. The errors will disappear as soon as Cloud pushes a new image, too.

abhijit-singh:
> I just wanted to showcase that the modal is shown wherever the user is, no matter how "deep" in the backstage.

Got it, thanks for re-recording, looks awesome!

> Actually, we cannot if we stick to interactive messages. 

I was checking other examples of what I think are interactive messages. Can we add line breaks like how Matterpoll does here? I'm not stuck up on exactly how it should look, and upgrading on mobile is surely not a major use case, but just wanted to make the content a little more consumable.  Mobile screenshot:
image

cpoile (approved):
This looks amazing, thanks @agarciamontoro!

webapp/src/hooks.ts
```diff
@@ -285,3 +285,30 @@ export function useEnsureProfiles(userIds: string[]) {
...
+
+    // @ts-ignore
+    const {openModal, ModalIdentifiers} = window.WebappUtils.modals;
+
+    // @ts-ignore
+    const PurchaseModal = window.Components.PurchaseModal;
+
+    if (!openModal || !ModalIdentifiers?.CLOUD_PURCHASE || !PurchaseModal) {
+        // eslint-disable-next-line no-process-env
+        if (process.env.NODE_ENV !== 'production') {
```
lieut-data, line 300: Why limited to developer builds?

lieut-data (approved):
Looks good!

agarciamontoro:
@abhijit-singh, what about something like this?
image
Without the top separator, it looks like this:
image

webapp/src/hooks.ts
agarciamontoro: Removed in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/592/commits/a6275f0418cc212da41bd132f317b2b66fad9d5f

abhijit-singh:
@agarciamontoro With the top separator looks good to me. Thank you!

abhijit-singh (approved):

PR #593, opened 2021-06-11
s/UserIDtoIsAdmin/IsAdmin

lieut-data:
#### Summary
Simplify the `playbook.RequestInfo` struct to just use an `IsAdmin` bit instead of the old `UserIDtoIsAdmin` map which could track multiple users. This is part of the clean up necessary to collapse the `playbook`, `incident` and `permissions` packages into an `app` package, and ultimately streamline the renaming of `incident` to `run` (or `playbookRun`).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35597

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/api/playbooks.go
+++ b/server/api/playbooks.go
@@ -310,9 +310,9 @@ func (h *PlaybookHandler) getPlaybooks(w http.ResponseWriter, r *http.Request) {
 	}
 
 	requesterInfo := playbook.RequesterInfo{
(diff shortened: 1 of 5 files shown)


cpoile (approved):
Yess!

crspeller (approved):

PR #594, opened 2021-06-12
Collapse into app package

lieut-data:
#### Summary
Another biggish change in anticipation of renaming incident: collapsing `permissions`, `playbook` and `incident` into a single `app` package. Both `permissions` and `playbook` have remained fairly small, with the bulk of this layer of the business logic going into `incident`. But more painful has been the creative ways in which we've had to avoid using variables named `incident` or `playbook` to avoid the corresponding package shadowing. The seizure that is `incdnt` is no more with this change, as we group this under an `app` package to house this instead, following the lead of a long-ago decision around the mattermost-server repository.

This also fits well with Dave Cheney's advice on [package naming](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common):
> For example, the net/http package does not have client and server packages, instead it has client.go and server.go files, each holding their respective types. transport.go holds for the common message transport code used by both HTTP clients and servers.

> Name your packages after what they provide, not what they contain.

The `app` package provides the key part of the application business logic, with the `api` and `sqlstore` remaining separate to provide their own contributions.

I've already synced with most of you on these changes, so hopefully they don't come as a surprise, but feedback welcome as usual!

#### Ticket Link
N/A

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


server/app/incident.go
```diff
@@ -506,3 +483,85 @@ type JobOnceScheduler interface {
...
+	// not returned in relevance order).
+	SearchTerm string `url:"search_term,omitempty"`
+
+	// PlaybookID filters incidents that are derived from this playbook id.
+	// Defaults to blank (no filter).
+	PlaybookID string `url:"playbook_id,omitempty"`
+}
+
+// IsValid either makes the given filter options valid, or returns an error if invalid.
+func (options *IncidentFilterOptions) IsValid() error {
```
agarciamontoro, line 528: I think it would be worth changing this name to something that makes it explicit that `options` is actually being modified. Something like `MakeValid`, `ApplyValidDefaults` or `Correct`, but better :)

server/app/sort.go
```diff
@@ -0,0 +1,51 @@
...
+	SortByOwnerUserID = "owner_user_id"
+
+	// SortByTeamID sorts by the team id of an incident or playbook.
+	SortByTeamID = "team_id"
+
+	// SortByEndAt sorts by the end time of an incident.
+	SortByEndAt = "end_at"
+
+	// SortByStatus sorts by the status of an incident.
+	SortByStatus = "status"
```
agarciamontoro, line 35: Not sure I like having the incident and playbook options in the same place: some are valid only for incidents, other only for playbooks, and some are valid for both, right? Is this what we want? I don't know if there's a better way of doing it, just a data point.

agarciamontoro (commented):
I really like the general idea, and seeing `incdnt` and `pbook` disappear is so satisfying, thank you! :tada:

A couple of considerations:
- What about what we discussed on namespacing `Incident` and `Playbook` à la `mattermost-plugin-api`, so we have `app.Incident.Store`, `app.Playbook.Telemetry` and such? Is that feasible?
- I saw that some of the error checking in the sqlstore is gone. Are we effectively moving the error checking out of the sqlstore into the app layer?
- The first commit reads "Medium term, the plan will also be to remove the interfaces between the sqlstore, app and api packages and just consume each other directly. This would result in all unit tests hitting the database, eliminating most mocking." Could you elaborate a bit more on the rationale of removing the interfaces?

lieut-data:
> - What about what we discussed on namespacing `Incident` and `Playbook` à la `mattermost-plugin-api`, so we have `app.Incident.Store`, `app.Playbook.Telemetry` and such? Is that feasible?

Yes, it's definitely feasible, though I was hoping to defer any actual reorganization of the service code (which involves a bunch more unit test changes). This change is /mostly/ about enabling the subsequent `s/incident/playbookRun/`.

> - I saw that some of the error checking in the sqlstore is gone. Are we effectively moving the error checking out of the sqlstore into the app layer?

Indeed, in parsing through the code, I recall observing that ownership of the validity of inputs was spread across the package layers. My hope with the (relatively few?) changes here was to concentrate the "business logic" validity into the `app` layer, and the "syntactical validity" into the `sqlStore` layer. It would be neat if the `sqlStore` logic could be used apart from business logic, say, to implement admin-only operations, or help with migrations and such, but this is challenging when the two are mixed together.

> - The first commit reads "Medium term, the plan will also be to remove the interfaces between the sqlstore, app and api packages and just consume each other directly. This would result in all unit tests hitting the database, eliminating most mocking." Could you elaborate a bit more on the rationale of removing the interfaces?

This was something that came out of an end-of-day discussion with @crspeller -- and something I'd like to discuss more with the team before doing anything. I'll queue something up for our architecture meeting, but in short: our interfaces really only exist to enable our mocks, and otherwise seem superfluous. I was a big fan of the (earlier) mattermost-server approach of skipping the mocks and just having integration tests (i.e. `api` tests hit the database directly). This does make running unit tests slower, but ultimately far more satisfying since they are demonstrably more correct.

server/app/incident.go
lieut-data: Ah, yes, good observation -- perhaps a `Validate() (options IncidentFilterOptions, error)` to avoid the mutation altogether.

server/app/sort.go
lieut-data: Indeed, there is some overlap here. Of course, there's the runtime check to accept only the ones that make sense, but we do lose the build-time check that we're only sorting by fields that make sense. I'm really 0/5 here, with the only material tradeoff probably being the need to be more verbose: `SortByIncidentOwnerUserID`, `SortByPlaybookTeamId`, etc. I could fold this into the `client` package too.

server/app/incident.go
agarciamontoro: I like that, too :)

server/app/sort.go
agarciamontoro: Yeah, I'd lean again towards something like `Incident.SortByOwnerUserID`, `Playbook.SortByTeamID`.  In any case, maybe better to address in the next PRs?

agarciamontoro (approved):
Sounds good! About the permissions scattered across all layers, it's something we've wanted to tackle for a while, I like that we are moving towards solving it :)

server/app/sort.go
lieut-data: Done, including beefed up unit tests for these methods after I found some issues.

server/app/incident.go
lieut-data: Done, including beefed up unit tests for these methods after I found some issues.

server/app/sort.go
lieut-data: I'll leave this one unresolved for now and we can iterate, unless folks feel strongly.

iomodo: And maybe add a functional options as well?

iomodo (approved):
Wonderful!

server/api/playbooks.go
```diff
@@ -329,29 +328,32 @@ func (h *PlaybookHandler) getPlaybooksAutoComplete(w http.ResponseWriter, r *htt
...
 		UserID:  userID,
 		TeamID:  teamID,
-		IsAdmin: permissions.IsAdmin(userID, h.pluginAPI),
+		IsAdmin: app.IsAdmin(userID, h.pluginAPI),
 	}
 
-	playbooksResult, err := h.playbookService.GetPlaybooksForTeam(requesterInfo, teamID, playbook.Options{})
+	playbooksResult, err := h.playbookService.GetPlaybooksForTeam(requesterInfo, teamID, app.PlaybookFilterOptions{
+		Page:    0,
+		PerPage: 15,
```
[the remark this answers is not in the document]
lieut-data: Ooh, good idea! Done!

PR #595, opened 2021-06-14
[MM-32418] Export channel on incident archive

arvinDarmawan:
Summary
This code will implement export channel on incident archive

Ticket Link
Jira card: https://mattermost.atlassian.net/browse/MM-32418
Github issue: https://github.com/mattermost/mattermost-server/issues/17311

```release-note
Release Note
- create new column ExportChannelOnArchiveEnabled on IR_Playbook and IR_Incident table
- update select queries in each playbook and incident
- update api createIncident to set property ExportChannelOnArchiveEnabled from playbook
- update function UpdateStatus to export the channel
```


agarciamontoro:
Thank you, @arvinDarmawan, this looks really good! We will try to review it as soon as we can :) I see that you created this PR as a draft. Should we wait for more changes?

arvinDarmawan:
> Thank you, @arvinDarmawan, this looks really good! We will try to review it as soon as we can :) I see that you created this PR as a draft. Should we wait for more changes?

 It's already done for  the ExportChannelOnArchiveEnabled ui and create/update column ExportChannelOnArchiveEnabled. Only  left export  the channel  for this issue. I will check it.

arvinDarmawan:
Hi @agarciamontoro I already successfully install the channel export plugin in my local according to documentation.

Screen Shot 2021-06-21 at 19 30 29

But after I go to Incident page, I can't export the channel even though the incident status is "Archived".
Screen Shot 2021-06-21 at 19 30 59

Screen Shot 2021-06-21 at 19 36 19

The error  on server log is same as before too.
```
{"level":"error","ts":1624276400.731646,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, muxBroker couldn't Accept request body connection","plugin_id":"com.mattermost.plugin-incident-management","error":"timeout waiting for accept"}
{"level":"error","ts":1624276400.731712,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, muxBroker couldn't accept connection","plugin_id":"com.mattermost.plugin-incident-management","serve_http_stream_id":7,"error":"timeout waiting for accept"}
```

What am I missing ?  Is it because of the license type or  error in my code ?

agarciamontoro:
I'm reviewing the Channel Export code and it looks like we don't check for the `EnableTesting` and `EnableDeveloper` settings, as we do in Incident Collaboration. Could you try to request a trial license? You can do so by going to the System Console and clicking on the Start trial button. That should unlock you for now; in the meanwhile, I'm gonna fix that in the Channel Export plugin. Thank you for reporting!

agarciamontoro:
If you don't see the Start trial button, you'll see a similar button with `Upgrade to Enterprise Edition`. You can use that button if you're developing from a Linux machine to upgrade to the E0 edition, and then you'll see the Star trial button. Otherwise, let me fix that in the Channel Export plugin so I can unblock you there. Sorry for the inconvenience!

arvinDarmawan:
> If you don't see the Start trial button, you'll see a similar button with `Upgrade to Enterprise Edition`. You can use that button if you're developing from a Linux machine to upgrade to the E0 edition, and then you'll see the Star trial button. Otherwise, let me fix that in the Channel Export plugin so I can unblock you there. Sorry for the inconvenience!

Ok, thank you @agarciamontoro

arvinDarmawan:
Hi @agarciamontoro , this is the expected result, right ?
It already running in my local and send direct message with the url link (it will download the csv).

I testing it by comment out the license checking in my local channel-export plugin and deploy it to my local mattermost.
I already push it to  this PR.

Screen Shot 2021-06-21 at 23 06 39

agarciamontoro:
@arvinDarmawan:

> send direct message with the url link (it will download the csv).

Oh I had not thought of that solution, it's actually really straightforward! The thing is that I took a quick look at your code and you're making the request but then discarding the response, right? What I was imagining is that the code makes the request and use the returned file (which is the body of the response) to attach it to the DM. To attach a file to a post, you first need to upload it (take a look at the code of [this function](https://github.com/mattermost/mattermost-plugin-channel-export/blob/5fa64f135a7c0f5d61383d1c2ef79eb90a4306e9/server/command_hooks.go#L147-L156)) and then create the post with the `FileIds` field set to the file ID returned by the uploading process (see [this line](https://github.com/mattermost/mattermost-plugin-channel-export/blob/5fa64f135a7c0f5d61383d1c2ef79eb90a4306e9/server/command_hooks.go#L133)).

> testing it by comment out the license checking in my local channel-export plugin and deploy it to my local mattermost.
I already push it to this PR.

That's a brilliant solution, thanks!

arvinDarmawan:
> To attach a file to a post, you first need to upload it (take a look at the code of this function) and then create the post with the FileIds field set to the file ID returned by the uploading process (see this line)

hi @agarciamontoro , sorry I'm still new in these plugin things. I have another question.
 
For uploading and create the post with the FileIds, I can't call  these two functions directly in channel-exports plugin  from mattermost-plugin-incident-collaboration since  I only have access to channel-exports plugin  from  api  v1/exports in channel-exports plugin, right ?.

So I need to call these function from  plugin_api.go  in mattermost-server ??

arvinDarmawan:
> > hi @agarciamontoro , sorry I'm still new in these plugin things. I have another question.
> 
> Don't be sorry, you're doing a great job, getting started with these codebases is never an easy thing.
> 
> > For uploading and create the post with the FileIds, I can't call these two functions directly in channel-exports plugin from mattermost-plugin-incident-collaboration since I only have access to channel-exports plugin from api v1/exports in channel-exports plugin, right ?
> > So I need to call these function from plugin_api.go in mattermost-server ??
> 
> Not exactly. The code I linked in my previous comment was just to illustrate what we need to do. We'd need to use the same code inside the upload function to upload the contents returned by the `/api/v1/export` request to Mattermost, and then use the returned ID in the creation of the post. It would be something like the following (I have not tested this at all, so you will probably need to tweak a lot of things):
> 
> ```go
> // set url and query string
> exportPluginURL := fmt.Sprintf("plugins/com.mattermost.plugin-channel-export/api/v1/export?format=csv&channel_id=%s", incidentToModify.ChannelID)
> 
>  // The request has no body, so we can set the last param to nil
> req, err := http.NewRequest(http.MethodGet, exportPluginURL, nil)
> req.Header.Add("Mattermost-User-ID", incidentToModify.OwnerUserID)
> if err != nil {
>     s.pluginAPI.Log.Warn("failed to create request for exporting channel", "plugin", "channel-export", "error", err)
> }
> 
> res := s.pluginAPI.Plugin.HTTP(req)
> defer res.Body.Close() // <- always good to call Close as soon as possible, but deferring it so it is actually done at the end of the function
> if res.StatusCode == http.StatusOK {
>     // Instead of using ReadAll and discarding the results, pass the body to the Upload function. This returns a file struct, which contains the ID we need!
>     file, err := s.PluginAPI.File.Upload(res.Body, "filename, maybe the channel name", incidentToModify.ChannelID)
>     if err != nil {
>         s.PluginAPI.Log.Error("unable to upload the exported file to the channel",
> 	        "Channel ID", channelID, "Error", err)
>         return nil, errors.Wrap(err, "unable to upload the exported file")
>     }
> 
>     // We don't need these anymore
>     // url := fmt.Sprintf("/%s", exportPluginURL)
>     // msg := fmt.Sprintf("[%s](%s)", incidentToModify.Name, url) 
> 
>     // Now, use something like what I linked before, instead of s.poster.DM, which is not flexible enough.
>     // You'll need to add the PostDM function to the Poster interface and to the bot/poster.go file, which would just
>     // get a post and override the channel ID with the ID of the DM between the user and the bot, and the user ID with
>     // the bot ID , similar as to what the DM function does
>     err = s.poster.PostDM(incidentToModify.OwnerUserID, &model.Post{
> 	    Message:   fmt.Sprintf("Incident %s exported", incidentToModify.Name),
>             FileIds:[]string{file.Id},
>     })
> 
>     if err = s.poster.DM(incidentToModify.OwnerUserID, "Data successfully exported %s", msg); err != nil {
>         return errors.Wrap(err, "failed to send exported channel result to incident's commander")
>     }
> }
> ```
> 
> If you want to, we can have a quick Zoom chat and go through the code together. Just ping me at https://community.mattermost.com/core/messages/@alejandro.garcia and we can schedule a quick call.

Done @agarciamontoro .
More or less like this the result, right ?

Screen Shot 2021-06-22 at 19 09 26

server/app/incident_service.go
```diff
@@ -218,6 +218,9 @@ func (s *IncidentServiceImpl) CreateIncident(incident *Incident, pb *Playbook, u
 	incident.CreateAt = now
 	incident.LastStatusUpdateAt = now
 	incident.CurrentStatus = StatusReported
+	if pb != nil {
+		incident.ExportChannelOnArchiveEnabled = pb.ExportChannelOnArchiveEnabled
+	}
```
agarciamontoro, line 223: Can we move this to the `createIncident` function in the `server/api/incidents.go`? All the assignments from the playbook to the incident happen there.

server/app/incident_service.go
```diff
@@ -778,6 +781,34 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
...
 	}
 
+	if options.Status == StatusArchived && incidentToModify.ExportChannelOnArchiveEnabled {
+		// set url and query string
+		exportPluginURL := fmt.Sprintf("plugins/com.mattermost.plugin-channel-export/api/v1/export?format=csv&channel_id=%s", incidentToModify.ChannelID)
+
+		req, err := http.NewRequest(http.MethodGet, exportPluginURL, bytes.NewBufferString(""))
+		req.Header.Add("Mattermost-User-ID", incidentToModify.OwnerUserID)
+		if err != nil {
+			s.pluginAPI.Log.Warn("failed to create request for exporting channel", "plugin", "channel-export", "error", err)
```
agarciamontoro, line 791: If the request failed to create, we should not continue this process. What about moving everything that is inside the `if options.Status == StatusArchived && incidentToModify.ExportChannelOnArchiveEnabled` except for the last PostDM call into a helper function that can return with an error if there is one?
That way, we can just call that function from here and handle the error as we want (for example, we should post a message in the channel saying that there was an error, similar to what you can see just above this whole block, when the webhook on status update fails). If it succeeds, we make the call to PostDM (the helper function should return the file ID if successful).

server/app/incident_service.go
```diff
@@ -778,6 +781,34 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
...
+		}
+
+		res := s.pluginAPI.Plugin.HTTP(req)
+		defer res.Body.Close()
+		if res.StatusCode == http.StatusOK {
+			file, err := s.pluginAPI.File.Upload(res.Body, fmt.Sprintf("%s.csv", incidentToModify.Name), incidentToModify.ChannelID)
+			if err != nil {
+				s.pluginAPI.Log.Error("unable to upload the exported file to the channel",
+					"Channel ID", incidentToModify.ChannelID, "Error", err)
+				return errors.Wrap(err, "unable to upload the exported file")
```
agarciamontoro, line 801: These returns would be ok inside the helper function!

server/bot/poster.go
```diff
@@ -52,6 +52,11 @@ func (b *Bot) PostCustomMessageWithAttachments(channelID, customType string, att
 	return post, nil
 }
 
+// Post custom DM to the specified user
+func (b *Bot) PostDM(userID string, post *model.Post) error {
+	return b.dm(userID, post)
+}
+
```
agarciamontoro, line 59: What about just making DM an exported function and call that one where we need to? We can just rename `dm` to `PostDM`.

webapp/src/components/backstage/automation/export_channel_on_archive.tsx
```diff
@@ -0,0 +1,25 @@
...
+        <AutomationTitle>
+            <Toggle
+                isChecked={props.enabled}
+                onChange={props.onToggle}
+            />
+            <div>{'Export channel'}</div>
+        </AutomationTitle>
+
+    </AutomationHeader>
+);
```
agarciamontoro, line 25: Can you add a new line at the end of the file? That will prevent future automatic changes to this file that only add that line.

agarciamontoro (changes requested):
This looks fantastic, @arvinDarmawan, thank you for your contribution! I left some suggestions to make the code a bit more clear, but the overall solution is great :tada:

webapp/src/components/backstage/automation/export_channel_on_archive.tsx
arvinDarmawan: Done

server/app/incident_service.go
arvinDarmawan: Done

server/app/incident_service.go
arvinDarmawan: Done @agarciamontoro

server/app/incident_service.go
arvinDarmawan: Done @agarciamontoro , kindly check it to make sure what I have done already correct.

server/bot/poster.go
arvinDarmawan: Could you elaborate more @agarciamontoro  ? 

Do you mean, I just need to switch the name of function ?


```
DM(userID string, post *model.Post) error
PostDM(userID, format string, args ...interface{}) error
```


And I call DM function from incident_service.go ??

agarciamontoro: Yeah, what I want to avoid is to have this `PostDM` wrapper around `dm`. Right now, it just passes down the exact same arguments to `dm`, so why not use directly `dm`? The thing is that `dm` is not publicly exposed, so we'd need to rename it. The name `DM` is already taken, so we should maybe just rename it `PostDM`? Any other solution that prevents this wrapper is also good, mine was just a suggestion!

server/app/incident_service.go
agarciamontoro: Perfect!

server/app/incident_service.go
agarciamontoro: Thanks!

webapp/src/components/backstage/automation/export_channel_on_archive.tsx
agarciamontoro: Perfect!

server/app/incident_service.go
agarciamontoro: Perfect, just a couple more suggestions I'm adding.

server/app/incident_service.go
```diff
@@ -778,9 +779,54 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
...
+		if err = s.poster.PostDM(incidentToModify.OwnerUserID, &model.Post{Message: fmt.Sprintf("Incident %s exported succesfully", incidentToModify.Name), FileIds: []string{fileID}}); err != nil {
+			return errors.Wrap(err, "failed to send exported channel result to incident's commander")
+		}
+
+	}
+
 	return nil
 }
 
+func (s *IncidentServiceImpl) exportFileToChannel(incidentName string, ownerUserID string, channelID string) (string, error) {
```
agarciamontoro, line 799: I think a more clear name for this function would be `exportChannelToFile`.

server/app/incident_service.go
```diff
@@ -778,9 +779,54 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
...
+	if options.Status == StatusArchived && incidentToModify.ExportChannelOnArchiveEnabled {
+
+		fileID, err := s.exportFileToChannel(incidentToModify.Name, incidentToModify.OwnerUserID, incidentToModify.ChannelID)
+		if err != nil {
+			_, _ = s.poster.PostMessage(incidentToModify.ChannelID, "Incident Collaboration failed to export channel information. Contact your System Admin for more information.")
+			return errors.Wrap(err, "failed to export file to a channel")
+		}
+
+		if err = s.poster.PostDM(incidentToModify.OwnerUserID, &model.Post{Message: fmt.Sprintf("Incident %s exported succesfully", incidentToModify.Name), FileIds: []string{fileID}}); err != nil {
+			return errors.Wrap(err, "failed to send exported channel result to incident's commander")
```
agarciamontoro, line 791: We recently changed the terminology from commander to owner:
```suggestion
			return errors.Wrap(err, "failed to send exported channel result to incident's owner")
```

server/app/incident_service.go
```diff
@@ -778,9 +779,54 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
 		}()
 	}
 
+	if options.Status == StatusArchived && incidentToModify.ExportChannelOnArchiveEnabled {
+
+		fileID, err := s.exportFileToChannel(incidentToModify.Name, incidentToModify.OwnerUserID, incidentToModify.ChannelID)
+		if err != nil {
+			_, _ = s.poster.PostMessage(incidentToModify.ChannelID, "Incident Collaboration failed to export channel information. Contact your System Admin for more information.")
```
agarciamontoro, line 786: I think we can remove the `information` word in this message.
```suggestion
			_, _ = s.poster.PostMessage(incidentToModify.ChannelID, "Incident Collaboration failed to export channel. Contact your System Admin for more information.")
```

server/app/incident_service.go
```diff
@@ -778,9 +779,54 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
...
+			return errors.Wrap(err, "failed to export file to a channel")
```
agarciamontoro, line 787: Maybe reword this:
```suggestion
			return errors.Wrap(err, "failed to export channel")
```

server/app/incident_service.go
```diff
@@ -778,9 +779,54 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
...
+			errMessage := "unable to upload the exported file to a channel"
+			s.pluginAPI.Log.Error(errMessage, "Channel ID", channelID, "Error", err)
+
+			return "", errors.Wrap(err, errMessage)
+		}
+
+		return file.Id, nil
+	}
+
+	return "", errors.New(fmt.Sprintf("There is an error when make a request to upload file with status code %s", strconv.Itoa(res.StatusCode)))
```
agarciamontoro, line 826: Maybe return early if there is an error? Something like the following:

```suggestion
	if res.StatusCode != http.StatusOK {
		return "", errors.New(fmt.Sprintf("There is an error when make a request to upload file with status code %s", strconv.Itoa(res.StatusCode)))
	}
	
	file, err := s.pluginAPI.File.Upload(res.Body, fmt.Sprintf("%s.csv", incidentName), channelID)
	if err != nil {
		errMessage := "unable to upload the exported file to a channel"
		s.pluginAPI.Log.Error(errMessage, "Channel ID", channelID, "Error", err)

		return "", errors.Wrap(err, errMessage)
	}

	return file.Id, nil
```

agarciamontoro (changes requested):
Some more minor things, thanks for the patience! I also clarified in the previous review what I meant with renaming `b.dm`.

server/app/incident_service.go
arvinDarmawan: Done

server/app/incident_service.go
arvinDarmawan: Done

server/app/incident_service.go
arvinDarmawan: Done

server/app/incident_service.go
arvinDarmawan: Done

server/app/incident_service.go
arvinDarmawan: Done @agarciamontoro

server/app/incident_service.go
```diff
@@ -821,10 +821,10 @@ func (s *IncidentServiceImpl) exportFileToChannel(incidentName string, ownerUser
 		}
 
 		return file.Id, nil
-	} else {
-		return "", errors.New(fmt.Sprintf("There is an error when make a request to upload file with status code %s", strconv.Itoa(res.StatusCode)))
 	}
 
+	return "", errors.New(fmt.Sprintf("There is an error when make a request to upload file with status code %s", strconv.Itoa(res.StatusCode)))
```
justinegeffen:
```suggestion
	return "", errors.New(fmt.Sprintf("There was an error when making a request to upload file with status code %s", strconv.Itoa(res.StatusCode)))
```

server/app/incident_service.go
```diff
@@ -781,13 +781,13 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
...
-		fileId, err := s.exportFileToChannel(incidentToModify.Name, incidentToModify.OwnerUserID, incidentToModify.ChannelID)
+		fileID, err := s.exportFileToChannel(incidentToModify.Name, incidentToModify.OwnerUserID, incidentToModify.ChannelID)
 		if err != nil {
 			_, _ = s.poster.PostMessage(incidentToModify.ChannelID, "Incident Collaboration failed to export channel information. Contact your System Admin for more information.")
 			return errors.Wrap(err, "failed to export file to a channel")
 		}
 
-		if err = s.poster.PostDM(incidentToModify.OwnerUserID, &model.Post{Message: fmt.Sprintf("Incident %s exported succesfully", incidentToModify.Name), FileIds: []string{fileId}}); err != nil {
+		if err = s.poster.PostDM(incidentToModify.OwnerUserID, &model.Post{Message: fmt.Sprintf("Incident %s exported succesfully", incidentToModify.Name), FileIds: []string{fileID}}); err != nil {
 			return errors.Wrap(err, "failed to send exported channel result to incident's commander")
```
justinegeffen:
```suggestion
			return errors.Wrap(err, "failed to send exported channel result to playbook owner")
```

server/app/incident_service.go
```diff
@@ -781,13 +781,13 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
 
 	if options.Status == StatusArchived && incidentToModify.ExportChannelOnArchiveEnabled {
 
-		fileId, err := s.exportFileToChannel(incidentToModify.Name, incidentToModify.OwnerUserID, incidentToModify.ChannelID)
+		fileID, err := s.exportFileToChannel(incidentToModify.Name, incidentToModify.OwnerUserID, incidentToModify.ChannelID)
 		if err != nil {
 			_, _ = s.poster.PostMessage(incidentToModify.ChannelID, "Incident Collaboration failed to export channel information. Contact your System Admin for more information.")
```
justinegeffen:
```suggestion
			_, _ = s.poster.PostMessage(incidentToModify.ChannelID, "Mattermost Playbooks failed to export channel information. Contact your System Admin for more information.")
```

justinegeffen:
@agarciamontoro, I've taken a look at this and see some of the wording needs to be updated for the UI messaging. Is there a separate PR for this work or should I review this PR's files for that? I'm 0/5 on how to do it. :)

server/bot/poster.go
arvinDarmawan: Hi @agarciamontoro , how about to make this code become as generic as possible ?
So in server/bot/poster.go, I will create this function :

```
// Post custom DM to the specified user
func (b *Bot) DM(userID string, post *model.Post) error {
	channel, err := b.pluginAPI.Channel.GetDirect(userID, b.botUserID)
	if err != nil {
		b.pluginAPI.Log.Info("Couldn't get bot's DM channel", "user_id", userID)
		return err
	}
	post.ChannelId = channel.Id
	post.UserId = b.botUserID
	if err := b.pluginAPI.Post.CreatePost(post); err != nil {
		return err
	}
	return nil
}
```

So, 
1. if I want to post message only, i will call :
`s.poster.DM(incidentToModify.OwnerUserID, &model.Post{Message: "Test message"})`

2. if I want to post message with file , i will call :
`s.poster.DM(incidentToModify.OwnerUserID, &model.Post{Message: fmt.Sprintf("Incident %s exported succesfully", incidentToModify.Name), FileIds: []string{fileID}})`

3. if I want to post message with attachment , i will call :
`s.poster.DM(incidentToModify.OwnerUserID, &model.Post{Message: "Test with attachment", []*model.SlackAttachment{attachment}})`

So, in the future, if there is new case, we didn't need to create new function to handle that ( just use DM(userID string, post *model.Post) )

and I will delete these functions :
```
DM(userID, format string, args ...interface{}) error
DMWithAttachments(userID string, attachments ...*model.SlackAttachment) error
dm(userID string, post *model.Post) error
dmAdmins(format string, args ...interface{}) error
```

but of course it will have different code style with EphemeralPost since it have:
```
EphemeralPost(userID, channelID string, post *model.Post)
EphemeralPostWithAttachments(userID, channelID, postID string, attachments []*model.SlackAttachment, format string, args ...interface{})
```
what do you think, @agarciamontoro ?

server/app/incident_service.go
arvinDarmawan: Hi @justinegeffen  I already merge my branch from master. I already rename the message too.

server/app/incident_service.go
arvinDarmawan: I already rename this message too.

server/app/incident_service.go
arvinDarmawan: I already rename this message too.

server/app/incident_service.go
agarciamontoro: Perfect!

server/app/incident_service.go
agarciamontoro: :tada:

server/app/incident_service.go
agarciamontoro: :tada:

server/app/incident_service.go
agarciamontoro: :tada:

server/app/incident_service.go
agarciamontoro: Perfect!

server/app/incident_service.go
agarciamontoro: @justinegeffen, should this be "playbook owner" or "run owner"?

agarciamontoro:
@justinegeffen, what UI messages do you mean? In any case, you can review them here, no problem!

server/bot/poster.go
agarciamontoro: @arvinDarmawan,  think that makes way more sense. Are you willing to change all the places where the DM and DMWithAttachments are used? That'd be awesome :_

arvinDarmawan: Oke, I will do it. How about EphemeralPost and EphemeralPostWithAttachments, @agarciamontoro  ?

webapp/src/components/backstage/automation/settings.tsx
```diff
@@ -151,6 +154,17 @@ export const AutomationSettings = (props: Props) => {
                 </Setting>
 
             </Section>
+            <Section>
+                <SectionTitle>
+                    {'When an incident is archived'}
```
justinegeffen, line 159:
```suggestion
                    {'When a run is archived'}
```

justinegeffen: @lieut-data, I cross-checked this with the generalized terminology doc and wanted to confirm that this is correct.

justinegeffen (commented):
Just one suggestion and a question for Jesse.

justinegeffen:
> @justinegeffen, what UI messages do you mean? In any case, you can review them here, no problem!

Sorry - I was referring to messaging/modals. Looks like there weren't as many as I anticipated. :)

server/app/playbook_run_service.go
```diff
@@ -806,9 +807,54 @@ func (s *PlaybookRunServiceImpl) UpdateStatus(playbookRunID, userID string, opti
 		}()
 	}
 
+	if options.Status == StatusArchived && playbookRunToModify.ExportChannelOnArchiveEnabled {
```
crspeller, line 810: I think we need to check that the previous status here is not archived, since we only want to do this once? There isn't actually a restriction on updating the status of an incident after archival.

crspeller (changes requested):
Nice work! One behavior thing.

server/bot/poster.go
arvinDarmawan: Hi @agarciamontoro , I already refactor  function DM, and I already delete old related DM functions **except dmAdmins**  because it is used by logger.go. 

Please check it . Thank youu!!

server/app/playbook_run_service.go
arvinDarmawan: Thank you for the feedback! Already push it!!

server/bot/poster.go
```diff
@@ -52,19 +52,19 @@ func (b *Bot) PostCustomMessageWithAttachments(channelID, customType string, att
...
-	})
-}
-
-// DMWithAttachments posts a Direct Message that contains Slack attachments.
-// Often used to include post actions.
-func (b *Bot) DMWithAttachments(userID string, attachments ...*model.SlackAttachment) error {
-	post := model.Post{}
-	model.ParseSlackAttachment(&post, attachments)
-	return b.dm(userID, &post)
+// Post DM to the specified user
```
agarciamontoro, line 55: Just to make this explicit enough:
```suggestion
// Post DM from the plugin bot to the specified user
```

agarciamontoro (approved):
Thank you once again, @arvinDarmawan, this is great :tada:

I left one minor suggestion, but I'm already approving this, it looks fantastic :) Thank you very much for all your work on this!

crspeller (approved):
Great work!

agarciamontoro:
Removing the Editor review label as Justine is out on vacation and everything was actually solved!

@prapti, all yours to test :)

This should be fairly straightforward to test:
1. Create a playbook with the `Export channel` setting, under the `When a run is archived` section, enabled.
2. Run that playbook.
3. Update the status, changing the state to Archived.
4. Check that the owner receives a DM from the Playbooks bot with the CSV file of the exported channel.

agarciamontoro:
@arvinDarmawan, I have tested this (Prapti is out these days, so I'm doing it myself) and it works great! I tested the feature with admins and end users, with public and private channels, and with the user archiving the run being different than the owner, and it worked in all cases :muscle: 

I did find a couple of minor things, which I fixed in the two latest commits (I did not want to disturb you anymore with this!):
- https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/595/commits/6dd446c6970a12fbc6a43506f956b888d834f36a: When the channel export fails, I found that returning an error results in the modal showing an error, which looks like the status was not updated, although it was. I fixed this by simply returning nil here.
- https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/595/commits/20d805cd8537c37ce158b0a69e1748b3874db2ec: This is just a quality of life improvement: I added a link to the playbook run channel in the message sent to the owner.

We're discussing a last thing in the team, as the channel export feature is an E20 feature, while all actions are available in all plans. I'll go and merge this PR as soon as tests pass, though, as it works flawlessly. If we need to tweak something in the license side, we'll do it in a separate PR.

Thank you very much for all your work on this, it was awesome! :tada:

PR #597, opened 2021-06-15
MM-36167: Use SKU short name with feature check as fallback

agarciamontoro:
#### Summary
This PR updates our license checks to use the proposed new way of detecting them. The details are [in this document](https://docs.google.com/document/d/1PhlIiGXbAl-sTRFXLRT5R71eSzUTA0HXAVNXPu1aAVU/edit), but the gist of it is that we are now using the SKU shortname instead of specific features. As the old names will be in use for at least two more years, we need to check for both the old ones (E10, E20) and the new ones (professional, enterprise). The possible names are the keys in this map: https://github.com/mattermost/customer-web-server/blob/master/licenses/licenses.go#L98

As a fallback for legacy licenses with potentially unknown SKU shortnames, we default to the old way of looking at a specific feature that we know is in the license we are checking. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36167
#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/license.ts
```diff
@@ -1,31 +1,65 @@
 import {GlobalState} from 'mattermost-redux/types/store';
 import {getLicense, getConfig} from 'mattermost-redux/selectors/entities/general';
 
-// isE20LicensedOrDevelopment returns true when the server is licensed with a Mattermost
-// Enterprise E20 License, or has `EnableDeveloper` and `EnableTesting` configuration settings
-// enabled, signaling a non-production, developer mode.
+const e10 = 'E10';
```
lieut-data, line 4: I wonder if there's a universe where this can live server-side (on our plugin) with a quick test vs. replicating the logic? (non-blocking comment)

lieut-data (approved):
Looks good, but can we add unit tests to ensure coverage?

iomodo (approved):
LGTM

agarciamontoro:
Feel free to take a look at the tests, @lieut-data :)

webapp/src/license.ts
agarciamontoro: We could maybe have some sort of feature-specific structure in the server and grab it when needed (on startup and when the license changes?). Let's leave this one as it is and explore it in the future.

lieut-data (approved):
Spiffy!

PR #601, opened 2021-06-17
MM-36243 - Playbook dashboard UI polish #1

cpoile:
#### Summary

(Should be able to review by commit, join the wip with the next commit)

- Incident list matches design specs (https://www.figma.com/file/vhVlHvJVNTxW6g9PLLnzI0/Playbook-Details?node-id=2%3A3949)
- Added `LastUpdateAt` to sql, allows sorting on that (for the incident list `Last updated` column, which is now the default)
- When change in stats boxes is 0, or div/0, don't show change pill
- Chart axes are cleaned up
- Chart tooltips are informational
- Icon and subtitle under playbook title indicating access levels
- Added four parameters to incident filter options, allows us to search on ranges of time for `Created at` and `Active at`
- Clicking on graphs will update the incident list for that day/week
- Playbook icon and name of playbook as link underneath incident overview incident name

Where we are:
- [x] UI tweaks to bring into 100% in line with designs <-- We are here
- [ ] New: Add "Run playbook" button on playbook dashboard screen
- [ ] New: license checked and upgrade bannered

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36243

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


webapp/src/components/backstage/playbooks/playbook_backstage.tsx
```diff
@@ -279,42 +311,99 @@ const PlaybookBackstage = () => {
...
+                                    let nextFetchParamsTime = DefaultFetchIncidentsParamsTime;
+                                    if (index >= 0) {
+                                        nextFetchParamsTime = {
+                                            ...DefaultFetchIncidentsParamsTime,
+                                            active_gte: stats.active_runs_per_day_times[index][0],
+                                            active_lt: stats.active_runs_per_day_times[index][1],
+                                        };
+                                    }
+
+                                    if (!fetchParamsTimeEqual(fetchParamsTime, nextFetchParamsTime)) {
```
cpoile, line 365: So that we don't make a request on every click on the graphs.

server/api/incidents.go
```diff
@@ -1327,6 +1327,42 @@ func parseIncidentsFilterOptions(u *url.URL) (*app.IncidentFilterOptions, error)
 
 	playbookID := u.Query().Get("playbook_id")
 
+	activeGTEParam := u.Query().Get("active_gte")
+	if activeGTEParam == "" {
+		activeGTEParam = "0"
+	}
+	activeGTE, err := strconv.ParseInt(activeGTEParam, 10, 64)
+	if err != nil {
```
lieut-data, line 1335: Given we default to `0` anyway, should we simplify each block to the following, and effectively just ignore the error of a garbage value?
```
activeGTE, _ := strconv.ParseInt(u.Query().Get("active_gte"), 10, 64)
```

server/app/incident_service.go
```diff
@@ -693,6 +695,7 @@ func (s *IncidentServiceImpl) UpdateStatus(incidentID, userID string, options St
 
 	incidentToModify.PreviousReminder = options.Reminder
 	incidentToModify.Description = options.Description
+	incidentToModify.LastUpdateAt = post.CreateAt
```
lieut-data, line 698: Given we only track create and status update, should we call this `LastStatusUpdateAt`?

server/sqlstore/incident.go
```diff
@@ -876,3 +902,22 @@ func addTimelineEventsToIncidents(timelineEvents []app.TimelineEvent, incidents
 		incidents[i].TimelineEvents = iToTe[incident.ID]
 	}
 }
+
+func queryActiveBetweenTimes(query sq.SelectBuilder, start int64, end int64) sq.SelectBuilder {
```
lieut-data, line 906: Instead of having conditional logic to pass in the current time if one endpoint is undefined, is it feasible to programmatically decide what parts to query here?

server/sqlstore/migrations.go
```diff
@@ -830,6 +830,52 @@ var migrations = []Migration{
...
+				From("IR_Incident as i").
+				LeftJoin("IR_StatusPosts as sp on i.Id = sp.IncidentId").
+				LeftJoin("Posts as p on sp.PostId = p.Id").
+				GroupBy("i.Id")
+
+			if err := sqlStore.selectBuilder(e, &lastUpdateAts, lastUpdateAtSelect); err != nil {
+				return errors.Wrapf(err, "failed getting incidents to update their LastUpdatedAt")
+			}
+
+			for _, row := range lastUpdateAts {
```
lieut-data, line 868: Can we do this all inline as one big SQL query? Might require a bit of MySQL-ing/Postgresql-ing, but would avoid fetching and writing back:
```sql
UPDATE IR_Incident i SET LastUpdateAt = COALESCE(...) FROM IR_StatusPosts sp, Posts p WHERE (i.Id = sp.IncidentId) AND (sp.PostId = p.Id)
```

server/sqlstore/stats.go
```diff
@@ -571,3 +592,21 @@ func endOfTodayMillis() int64 {
 	bod := time.Date(year, month, day, 0, 0, 0, 0, time.UTC)
 	return bod.UnixNano() / int64(time.Millisecond)
 }
+
+func reverseInts(vals []int) {
```
lieut-data, line 596: There seems to be some witchcraft optional at https://stackoverflow.com/questions/28058278/how-do-i-reverse-a-slice-in-go to do this "generally". 0/5

lieut-data (approved):
Nice! Just a few random thoughts.

server/app/incident_service.go
cpoile: Ah, that's a good point -- a year from now we'll think that means every update.  (or, really, a couple months from now :) )

server/sqlstore/incident.go
cpoile: I moved the conditional logic into the function, so it simplifies the caller. Good idea!

I could have also used a generic `queryBetweenTimes` function and asked the caller to pass a typed value (or string) to change whether we're testing for `active` or `started` (or even pass in the relevant function), but that seemed like code golfing and less clear for the future.

server/sqlstore/migrations.go
cpoile: We probably could, I just didn't want to worry about a mistake in the sql (we're doing a groupBy and a max, and I didn't want to worry about testing it) -- since we're only running this once on a migration, I figured simpler is better? But open to changing it if you feel strongly? @lieut-data

server/sqlstore/stats.go
cpoile: That's pure reflect magic, thanks!!

webapp/src/components/backstage/playbooks/incident_list/incident_list.tsx
```diff
@@ -97,20 +94,11 @@ const IncidentList = (props: Props) => {
...
-            };
-        });
-    }, [props.fetchParamsTime]);
-
     useEffect(() => {
         let isCanceled = false;
 
         async function fetchIncidentsAsync() {
-            const incidentsReturn = await fetchIncidents(fetchParams);
+            const incidentsReturn = await fetchIncidents({...fetchParams, ...props.fetchParamsTime});
```
cpoile, line 101: This (and the `DefaultFetchIncidentsParamsTime` below) means we won't be sending four params set to `0` with every call.

server/sqlstore/migrations.go
lieut-data: No objections -- thanks, @cpoile :)

lieut-data (approved):

agarciamontoro (approved):
Great job, looks fantastic! I was not able to find anything else than some typos!

server/sqlstore/migrations.go
lieut-data: Yeah, it should be "safe", just "expensive" in terms of copying data back and forth. Definitely not a high-priority issue :)

webapp/src/components/backstage/playbooks/incident_list/row.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    completed={completedTasks}
+                    total={totalTasks}
+                />
+            </div>
+        </div>
+    );
+};
+
+const participantsText = (participants: UserProfile[]) => {
+    const num = participants.length - 1;
```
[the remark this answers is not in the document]
cpoile: Actually it was to not count the owner as a "participant" -- I didn't even think about the bot. :) 
Propose to put this into a tech debt/PM decision -- we have to go through all our participant lists (and even the owner assignment drop down) to remove the incident bot, and be consistent about participant count.

webapp/src/components/backstage/playbooks/line_graph.tsx
```diff
@@ -34,6 +34,9 @@ const LineGraph = (props: LineGraphProps) => {
                     scales: {
                         yAxes: [{
                             ticks: {
+                                callback: (val: any) => {
+                                    return (val % 1 === 0) ? val : null;
```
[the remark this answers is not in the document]
cpoile: This is so it only shows whole numbers (it was showing 0.5, 1.5, etc.).

PR #602, opened 2021-06-17
[MM-35623] signal keywords feature

iomodo:
#### Summary
This is the full signal keywords feature as was reviewed in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/588

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-35623
#### Checklist

- [x] Telemetry updated
- [x] Unit tests updated


iomodo:
Added 3 more things:
1. After choosing the playbook ephemeral post will be updated
2. After ignoring the thread ephemeral message will be deleted
3. Changed the button styles from `default` to `primary`

cc @lieut-data @cpoile

iomodo:
Here is the settings screen
Screen Shot 2021-06-17 at 21 34 04 (2)

Here is the updated UI
https://user-images.githubusercontent.com/2340127/122447458-d2a4a300-cfb4-11eb-8ea7-1b6307119788.mp4

cc @itao @abhijit-singh

lieut-data (approved):

lieut-data:
Let's aim to merge this before https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/603, and I'll handle the merge conflicts myself.

itao:
- it'd be helpful to visually separate the delimited keywords to make it clear to users what they are verbatim. For example I wasn't sure if the space character was required during the demo. One idea could be displaying the text in a tag pill when user types comma and backspace deletes the whole tag - open to suggestions.
- This setting should be visually separate from the rest in some way because user is configuring `trigger` with fixed action whereas the rest is configuring `actions` with fixed trigger. Looking for help from @abhijit-singh

itao:
Follow up question: are the keywords case-sensitive?

crspeller:
@iomodo Maybe I missed something but why is the UI so squished in your screenshot? It's not actually like that is it?

crspeller (approved):
Approving. Only reviewed the last two commits since everything else has been reviewed.
Also contingent on the UI not actually being squished like that.

iomodo:
@itao should the tag pill be included in this iteration of the feature? Same question regarding the setting separation.

> Follow up question: are the keywords case-sensitive?

No, they are not. keywords are just comma separated list of the tokens, spaces and punctuation might be the part of the keyword(except comma).

itao:
> should the tag pill be included in this iteration of the feature? Same question regarding the setting separation.

Yes, the feature should solve the intended problem well - however narrow the scope may be. These usability improvements would elevate what we have from a technical proof-of-concept to a shippable MVP.

Also, bump on @crspeller's comment. Could you confirm that the UI is not actually squished like the screenshot?

iomodo:
@itao squishing problem is on `v1.13` here is the ticket for it https://mattermost.atlassian.net/browse/MM-36558. After refresh it goes away

lieut-data:
> spaces and punctuation might be the part of the keyword(except comma)

@iomodo, this might be surprising, given the way we implement keyword notifications in the product. I wonder if we shouldn't just trim these?

lieut-data:
I'm going to go ahead and merge this, and we can iterate as needed in follow-up PRs.

PR #607, opened 2021-06-23
Update mattermost-plugin-api to use the DB RPC API

agnivade:
We use the new NewClient constructor to let the DB RPC API work.
No further changes are needed.


agnivade:
@lieut-data / @crspeller - One thing to consider would be that the new changes aren't backward compatible with older versions of MM. How do we handle situations like these?

lieut-data (approved):
I've pinged @itao, but given we're focussing on the versions shipping pre-packaged, I don't /think/ this will be a problem. Can you help bump `min_server_version` in our `plugin.json` @agnivade to make this dependency explicit?

agnivade:
Done. :+1:

agnivade:
The Cypress tests are failing with:

```
Failed: failed to upload plugin bundle: : Unable to restart plugin on upgrade., 

Usage:
    pluginctl deploy <plugin id> <bundle path>
    pluginctl disable <plugin id>
    pluginctl enable <plugin id>
    pluginctl reset <plugin id>
make: *** [Makefile:126: deploy] Error 1
```
Is that due to the bump of `min_server_version` or just flaky? I cannot see any more detailed logs.

lieut-data:
I /think/ that's because the server is still `CurrentSchemaVersion = Version5360`. I'm not sure how that version is being bumped anymore -- I would have expected it to have been bumped by now. @cpanato / @metanerd, do you know when the `master` and `cloud` branches will have their versions formally bumped?

cpanato:
humm, i dont have this answer, maybe @amyblais

amyblais:
Carlos / Elisabeth normally bump the version when the self-managed release branch is cut. It can probably done earlier as needed.

agnivade:
Maybe I'll just revert the min version bump and get the commit in, and when 5.37 is released, bump the version up. There isn't a big deal either way from what I am surmising.

metanerd:
We should own this. I will add a job to bump `model/version.go` in the master branch after the latest stable release got released.

PR #608, opened 2021-06-24
Add better UI for keywords

iomodo:
#### Summary
This PR implements better UI for keywords
I struggled a lot with css, so please take a good look on it. Suggestions are appreciated ^_^

https://user-images.githubusercontent.com/2340127/123255368-cfec0580-d500-11eb-8595-5c9b1644ae56.mp4

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/automation/input_keywords.tsx
+++ b/webapp/src/components/backstage/automation/input_keywords.tsx
@@ -1,0 +1,122 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+import styled from 'styled-components';
+
+import {Chip} from 'src/components/backstage/automation/chip';
+import {AutomationHeader, AutomationTitle, SelectorWrapper} from 'src/components/backstage/automation/styles';
+import {Toggle} from 'src/components/backstage/automation/toggle';
+
+interface Props {
+    enabled: boolean;
+    textOnToggle: string;
+    onToggle: () => void;
+    placeholderText: string;
+    keywords: string[];
+    onAppendKeyword: (keyword: string) => void;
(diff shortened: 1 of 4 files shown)


itao (approved):
Very slick! I'll approve the PR assuming @abhijit-singh will be able to give more specific styling guidance such as padding.

abhijit-singh (changes requested):
@iomodo This is great! A few points:
1. Can we update the padding inside the tag as per the below markings 
image
2. When the tags grow beyond the width of the input box, can we increase the height? Just like it works in the `Add members to channel` modal
3. Is the field keyboard accessible? Ideally, it should work the same as the `Add members to channel` modal. (Attaching a gif here, it shows the keys pressed as well)

webapp/src/components/backstage/automation/input_keywords.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <AutomationHeader>
+            <AutomationTitle>
+                <Toggle
+                    isChecked={props.enabled}
+                    onChange={props.onToggle}
+                />
+                <div>{props.textOnToggle}</div>
+            </AutomationTitle>
+            <SelectorWrapper>
+                <StyledInputKeywords
```
crspeller, line 47: We actually already use a different component for this style of input for our system console component. Maybe consider just using that? https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/master/webapp/src/system_console_enabled_teams.tsx#L91

crspeller (changes requested):

agarciamontoro (approved):
This is great, thank you! :tada:

lieut-data:
@iomodo, can you help share next steps for this PR? Are there outstanding questions to be resolved?

PR #610, opened 2021-06-26
MM-36244 - Generic upgrade banner/flow; Upgrade banner/flow for playbook dashboard

cpoile:
#### Summary
- For ease of review I split the commits into mostly independent chunks.
- Genericized the upgrade banner with upgrade flow (derived from the rhs timeline banner)
- Converted the rhs timeline banner to use this, and using it for the playbook dashboard
- Tested cloud admin, cloud user, self-hosted user: success and failure; could only test self-hosted admin failure
- Test self-hosted admin success only once, because after the first licence it wouldn't give me another. Will work on that.
- @justinegeffen I'll need copy for the notify admin message (I'll highlight that below), and a review of the copy in the screencaps below
- @abhijit-singh in order to make the upgrade banner generic, I chose to follow the way the rhs timeline presented feedback to the user on success and failure--not a modal, but in the banner itself. 
- @abhijit-singh @itao The trial request success modal (for this and all the other upgrade flows) might be out of scope for this epic (we're running into time constraints).

Where we are:
- [x] UI tweaks to bring into 100% in line with designs
- [x] New: license checked and upgrade bannered  <-- We are here
- [ ] New: Add "Run playbook" button on playbook dashboard screen
- [ ] New: Trial request success modal for all upgrade paths

Cloud admin pre:
image

Cloud admin post (success):
image

Cloud user / Self-hosted user pre:
image

Cloud user / Self-hosted user post (success):
image

Cloud user / Self-hosted user post (failure):
image

Self-hosted admin pre:
image

Self-hosted admin post (success): 
TODO: Need to test. But, it should immediately change into working stats view. 

Self-hosted admin post (failure):
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36244

#### Checklist
- [x] Telemetry updated
- [ ] ~~Unit tests updated~~


server/bot/poster.go
```diff
@@ -159,6 +159,10 @@ func (b *Bot) NotifyAdmins(messageType, authorUserID string, isTeamEdition bool)
 		message = fmt.Sprintf("@%s requested access to the timeline in Incident Collaboration.", author.Username)
 		title = "Keep a complete record of the playbook run timeline"
 		text = "The playbook run timeline automatically tracks key events and messages in chronological order so that they can be traced and reviewed afterwards. Teams use timeline to perform retrospectives and extract lessons for the next time that they run the playbook."
+	case "start_trial_to_access_playbook_dashboard":
+		message = fmt.Sprintf("@%s requested access to the playbook dashboard in Incident Collaboration.", author.Username)
+		title = "View playbook statistics in the Playbook Dashboard"
+		text = "The Playbook Dashboard shows trends for total runs, active runs, and participants involved in each type of playbook."
```
cpoile, line 165: @justinegeffen This is the copy used to notify the sys admin

webapp/src/components/upgrade_banner.tsx
```diff
@@ -0,0 +1,229 @@
...
+type HandlerType = undefined | (() => (Promise<void> | void));
+
+interface Props {
+    illustration: JSX.Element;
+    successIllustration: JSX.Element;
+    errorIllustration: JSX.Element;
+    titleText: string;
+    helpText: string;
+    notificationType: AdminNotificationType;
+    upgradeWrapperSC: StyledComponent<'div', any>;
```
crspeller, line 36: Do we have to be specific with these types? Or just use JSX.Element?

webapp/src/components/upgrade_banner.tsx
```diff
@@ -0,0 +1,229 @@
...
+enum ActionState {
+    Uninitialized,
+    Loading,
+    Error,
+    Success,
+}
+
+type HandlerType = undefined | (() => (Promise<void> | void));
+
+interface Props {
```
crspeller, line 29: Also is this level of customization on the individual areas necessary? Are they that different? Should they be?

crspeller (approved):
Non blocking questions. Thanks for doing this!

webapp/src/components/upgrade_banner.tsx
cpoile: If we use JSX.Element, we can't treat it like a StyledComponent -- try it and you'll see the error. Unless there's an easy way to use it just like a SC? (Open to whatever works best)

webapp/src/components/upgrade_banner.tsx
cpoile: Yep, as long as you want control over the styling of the elements (which I did for the backstage banner and text). I only genericized what I had to. I bet we could make the designs a bit more standard, but that would need more work and back and forth with Abhijit.

server/bot/poster.go
```diff
@@ -159,6 +159,10 @@ func (b *Bot) NotifyAdmins(messageType, authorUserID string, isTeamEdition bool)
 		message = fmt.Sprintf("@%s requested access to the timeline in Incident Collaboration.", author.Username)
 		title = "Keep a complete record of the playbook run timeline"
 		text = "The playbook run timeline automatically tracks key events and messages in chronological order so that they can be traced and reviewed afterwards. Teams use timeline to perform retrospectives and extract lessons for the next time that they run the playbook."
+	case "start_trial_to_access_playbook_dashboard":
+		message = fmt.Sprintf("@%s requested access to the playbook dashboard in Incident Collaboration.", author.Username)
```
agarciamontoro, line 163: We are removing the "in Incident Collaboration" part. See https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/611/files
```suggestion
		message = fmt.Sprintf("@%s requested access to the playbook dashboard.", author.Username)
```

server/bot/poster.go
agarciamontoro: @cpoile, Justine is on vacation, so we should either ask Carrie for a review or just ship this and review later.

agarciamontoro (approved):
This is fantastic! Thank you for the work on the common component. The only missing piece is the telemetry for this. Take a look to [this commit](https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/616/commits/bc7f9694a44060b076550d6f441f02d43f29fe0e) to see the needed changes.

Approving now to unblock the PR :tada:

abhijit-singh (approved):
Looks good @cpoile!

One small thing that I just noticed - can we make the background color of the page white (center-channel-bg)?

webapp/src/components/upgrade_banner.tsx
crspeller: Talked about this and above offline. I pushed a commit fixing both of these concerns and making it easier to add this to more places in the future.

webapp/src/components/upgrade_banner.tsx
```diff
@@ -188,11 +229,11 @@ const Button = (props: ButtonProps) => {
     if (props.actionState === ActionState.Error) {
         if (props.isCurrentUserAdmin) {
             return (
-                <ButtonSC
+                <PrimaryButton
```
cpoile, line 232: @crspeller @abhijit-singh Wants this to be a `Secondary` button for the playbook dashboard banner. I'm not sure what you need on yours. Can we assume vertical is a `Primary` button, non-vertical is `Secondary`?

crspeller: @abhijit-singh Are you sure? I don't see why we wouldn't be consistent.

cpoile: @crspeller I see two options: pass in a buttonSC, or a `secondaryButton` flag. Can you think of another? Any preference?

crspeller: @cpoile I would go with the flag. 2/5

cpoile: done!

PR #614, opened 2021-06-28
Require mattermost-server 5.37

lieut-data:
#### Summary
This is a follow-up to @agnivade's changes to use the plugin-api for managing the DB connection pool instead of just hard-coding two. Once deployed, the plugin will only be live on `-daily` until `community` is upgraded on July 16th. Marketplace users will not be prompted to upgrade until they are running at least v5.37.

#### Ticket Link
None.

#### Checklist
- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -5,7 +5,7 @@
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-incident-collaboration/",
     "support_url": "https://github.com/mattermost/mattermost-plugin-incident-collaboration/issues",
     "icon_path": "assets/incident_plugin_icon.svg",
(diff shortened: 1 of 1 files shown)


cpoile (approved):

crspeller (approved):

lieut-data:
I've tested this with the latest master where it seems to build OK, but I know we only /just/ bumped the `ServerVersion`, so the Docker images may be a while before things are stable. Either way, this should get merged before 1.14.

lieut-data:
Build passed on master as expected. Cloud will fail for a bit until it catches up.

PR #615, opened 2021-06-28
MM-36345 - Run directly from playbook dashboard

cpoile:
#### Summary
- Add a run button on the playbook backstage
- Added a new `/playbook start-playbook <playbookID> <clientID>` slash command. It's not meant to be used by the user directly, so I didn't add it to any of the help texts or autocomplete.
- @calebroseland you can use the same `startPlaybookRunById` client action for the playbook run buttons in the RHS list view

Where we are:
- [x] UI tweaks to bring into 100% in line with designs
- [x] New: license checked and upgrade bannered
- [x] New: Add "Run playbook" button on playbook dashboard screen   <-- We are here
- [ ] New: Trial request success modal for all upgrade paths

Looks like:
image

And the playbook is auto selected (and the only option available):
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36345

#### Checklist

- [x] Telemetry updated
- [ ] ~~Unit tests updated~~


server/command/command.go
```diff
@@ -236,6 +236,65 @@ func (r *Runner) actionStart(args []string) {
 	}
 }
 
+// actionStartPlaybook is intended for scripting use, not use by the end user (they would have
+// to type in the correct playbookID).
```
agarciamontoro, line 240: This can be kind of solved with the autocomplete suggestions (in case we want to make this available to end users, I'm fine keeping it as is now).

webapp/src/components/backstage/playbooks/playbook_backstage.tsx
```diff
@@ -82,6 +86,14 @@ const PlaybookBackstage = () => {
         navigateToUrl(location.pathname + '/edit');
     };
 
+    const runPlaybook = () => {
+        navigateToUrl(`/${currentTeam.name}`);
+
+        if (playbook?.id) {
+            dispatch(startPlaybookRunById(playbook.id));
+        }
```
agarciamontoro, line 94: Is this navigation to the team time-sensitive in any way?

agarciamontoro (approved):
I really like this solution, but have we considered creating a brand new dialog that only asks for the run name, and that already has the playbook ID?

server/command/command.go
cpoile: Yah, I figured it's wasted engineering to add autocomplete because it's not intended for end users.

webapp/src/components/backstage/playbooks/playbook_backstage.tsx
cpoile: We added it a long time ago for the case where you go to the run list in the backstage and you have no runs so it gives you a `Run playbook` button. It hasn't seemed to break yet. :)

cpoile:
Apologies for force pushing -- had to rebase after merging the branch this was based on.

@agarciamontoro Good question -- we did, but decided that for now the standard interactive modal is good enough. We're going to revisit that option soon in another epic. Here we're only giving one playbook as the option, and it's auto-picked for the user, and they can't unselect it. So almost the same UX.

cpoile:
Hey @calebroseland, could you take a look at this this morning so it can go into the new release? Thanks!

cpoile:
@itao added telemetry. This will appear under the `frontend` event, as the action: `playbook_dashboard_run_clicked`

calebroseland (approved):
LGTM! Only question on https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/MM-36345-run-from-dashboard/webapp/src/slash_command.ts#L17 —does this need adjustment now that there's `start-playbook` that overlaps with the `startsWith` condition?

itao:
The icon on `Run` doesn't quite resonate as starting something, maybe finishing given the checkmark. What do you think about some version of a play icon, as in play/pause?

calebroseland:
CleanShot 2021-06-30 at 15 58 14@2x
maybe?

cpoile:
@calebroseland Good eye! But `/playbook start-playbook` isn't intended to be typed in by the user, so we don't have to worry about this.

cpoile:
@itao @calebroseland Good catch, thank you! Here's the correct icon:
image

PR #617, opened 2021-06-28
Adminify settings

lieut-data:
#### Summary
Always allow system administrators to configure who can create playbooks. I took the liberty of refactoring the settings handler and extending the client API accordingly. (Feel free to review by commit!)

As the admin user `test`, I can still configure this:

image

As the non-admin user `test2`, I can't:

image

When there are no restrictions, the non-admin user `test2` can't (but `test` can!):

image

#### Ticket Link
None.

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/api/settings.go
+++ b/server/api/settings.go
@@ -57,7 +57,7 @@ func (h *SettingsHandler) setSettings(w http.ResponseWriter, r *http.Request) {
 	}
 
 	isAdmin := app.IsAdmin(userID, h.pluginAPI)
-	if err := app.ModifySettings(userID, isAdmin, h.config); err != nil {
+	if err := app.ModifyPlaybookCreators(userID, isAdmin, h.config); err != nil {
 		h.HandleErrorWithCode(w, http.StatusForbidden, "Not authorized", err)
 		return
 	}
--- a/server/api/settings.go
+++ b/server/api/settings.go
@@ -56,7 +56,8 @@ func (h *SettingsHandler) setSettings(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := app.ModifySettings(userID, h.config); err != nil {
+	isAdmin := app.IsAdmin(userID, h.pluginAPI)
+	if err := app.ModifySettings(userID, isAdmin, h.config); err != nil {
 		h.HandleErrorWithCode(w, http.StatusForbidden, "Not authorized", err)
 		return
 	}
--- a/server/api/settings.go
+++ b/server/api/settings.go
@@ -1,0 +1,78 @@
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-plugin-incident-collaboration/client"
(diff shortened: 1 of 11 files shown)


server/app/permissions.go
```diff
@@ -342,21 +342,24 @@ func PlaybookModify(userID string, playbook, oldPlaybook Playbook, cfgService co
 	return nil
 }
 
-func ModifySettings(userID string, config config.Service) error {
+func ModifySettings(userID string, isAdmin bool, config config.Service) error {
```
crspeller, line 345: Maybe we should rename this to be explicit given that this behavior is specifically for that one setting. 
```suggestion
func ModifyPlaybookCreators(userID string, isAdmin bool, config config.Service) error {
```

server/api/settings.go
```diff
@@ -0,0 +1,79 @@
...
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	var settings client.GlobalSettings
+	if err := json.NewDecoder(r.Body).Decode(&settings); err != nil {
+		h.HandleErrorWithCode(w, http.StatusBadRequest, "unable to decode settings", err)
+		return
+	}
+
+	isAdmin := app.IsAdmin(userID, h.pluginAPI)
+	if err := app.ModifySettings(userID, isAdmin, h.config); err != nil {
```
crspeller, line 60: This tripped me for a second since we dropped the permissions package. Maybe (as a separate PR) we should rename all the permissions functions with a permissions prefix? Like `PermissionTo...`

server/app/permissions.go
```diff
@@ -342,21 +342,24 @@ func PlaybookModify(userID string, playbook, oldPlaybook Playbook, cfgService co
...
-			if candidateUserID == userID {
-				found = true
-				break
-			}
-		}
 
-		if !found {
-			return errors.Wrap(ErrNoPermissions, "not a playbook creator")
+	// Only admins are allowed to initially modify the settings.
+	if len(cfg.PlaybookCreatorsUserIds) == 0 && !isAdmin {
```
crspeller, line 354: We actually already know that the user is not an admin since we return early for admins.

webapp/src/client.ts
```diff
@@ -306,7 +306,7 @@ export async function telemetryEventForPlaybookRun(playbookRunID: string, action
 
 export async function setGlobalSettings(settings: GlobalSettings) {
     await doFetchWithoutResponse(`${apiUrl}/settings`, {
-        method: 'POST',
+        method: 'PUT',
```
crspeller, line 309: 0/5 but I think the behavior of setting the config on the server might go beyond the scope of a simple PUT in terms of side effects. Doesn't really matter though.

crspeller (changes requested):
Nice!

lieut-data:
Thanks for the quick and thorough review, @crspeller -- changes applied, and I fully agree with a sweep on the permissions nomenclature.

crspeller (approved):

agarciamontoro (approved):
Perfect, thank you! :tada:

PR #619, opened 2021-06-29
Support prompting for feedback

lieut-data:
#### Summary
If the NPS plugin is running, show a "Give Feedback" button in the backstage, redirecting to the @surveybot DM and "faking" a prompt from that bot for feedback:

image

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36566

#### Checklist
- ~~[] Telemetry updated~~
- [ ] Unit tests updated


cpoile (approved):
Cool!

I'm a little concerned we're going back to classnames (vs styled-components), but this looks like less code and cleaner that a styled-components version.

lieut-data:
> I'm a little concerned we're going back to classnames (vs styled-components), but this looks like less code and cleaner that a styled-components version.

Yeah, that's just me ~~being lazy~~ trying to keep the PR as small as possible. I briefly contemplated fixing the existing cases I was copying, but figured we should probably just do a full sweep here instead.

calebroseland (approved):
LGTM!

PR #620, opened 2021-06-29
MM-36540 - Improve graph searching & polish

cpoile:
#### Summary
- Can review based on commit
- When clicking on the graph, add a pill to indicate that the runs list is being filtered
- Cursor changed to pointer (only on graphs that are clickable)
- Weeks start on Sunday for stats, and end a millisecond before midnight on the last day of the range. Hopefully this will fix the wonky graphs on community.
- Only show whole numbers on the y axis for the line graphs
- Designs: (see the topleftmost design) https://www.figma.com/file/vhVlHvJVNTxW6g9PLLnzI0/Playbook-Details?node-id=2%3A3949

E.g., after clicking on the `Total Runs` graph:
image

E.g., after clicking on the `Active Runs` graph:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36540

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/assets/icons/clear_icon.tsx
+++ b/webapp/src/components/assets/icons/clear_icon.tsx
@@ -1,13 +1,8 @@
 // Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
 // See License for license information.
 
-import React from 'react';
 import styled from 'styled-components';
 
-const ClearIcon = (props: React.PropsWithoutRef<JSX.IntrinsicElements['i']>): JSX.Element => (
-    <i className={`icon icon-close-circle ${props.className}`}/>
-);
-
-export default styled(ClearIcon)`
+export default styled.i.attrs(() => ({className: 'icon icon-close-circle'}))`
     color: rgba(var(--center-channel-color-rgb), 0.56);
 `;
--- a/webapp/src/components/backstage/playbooks/stats_view.tsx
+++ b/webapp/src/components/backstage/playbooks/stats_view.tsx
@@ -3,6 +3,7 @@
 
 import React from 'react';
 import styled from 'styled-components';
+import moment from 'moment';
 
 import LineGraph from 'src/components/backstage/playbooks/line_graph';
 import {
@@ -18,11 +19,13 @@ import {PlaybookStats} from 'src/types/stats';
 import {useAllowPlaybookStatsView} from 'src/hooks';
 import UpgradePlaybookPlaceholder
     from 'src/components/backstage/playbooks/upgrade_playbook_placeholder';
+import Pill from 'src/components/widgets/pill';
 
 interface Props {
     stats: PlaybookStats
     fetchParamsTime: FetchPlaybookRunsParamsTime
     setFetchParamsTime: (params: FetchPlaybookRunsParamsTime) => void
+    setFilterPill: (pill: JSX.Element | null) => void
 }
 
 const StatsView = (props: Props) => {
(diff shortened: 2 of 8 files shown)


webapp/src/components/assets/icons/clear_icon.tsx
```diff
@@ -1,13 +1,8 @@
...
 
-import React from 'react';
 import styled from 'styled-components';
 
-const ClearIcon = (props: React.PropsWithoutRef<JSX.IntrinsicElements['i']>): JSX.Element => (
-    <i className={`icon icon-close-circle ${props.className}`}/>
-);
-
-export default styled(ClearIcon)`
+export default styled.i.attrs(() => ({className: 'icon icon-close-circle'}))`
```
cpoile, line 6: I think we need to change our compass icon elements into something like this, or else we can't attach onClick handlers (and many others, I assume).

webapp/src/components/backstage/playbooks/stats_view.tsx
```diff
@@ -36,6 +39,64 @@ const StatsView = (props: Props) => {
...
+        const ended = props.stats.active_runs_per_day_times[index][1];
+        const nextFetchParamsTime = {
+            active_gte: started,
+            active_lt: ended,
+        };
+
+        if (!fetchParamsTimeEqual(props.fetchParamsTime, nextFetchParamsTime)) {
+            const text = 'Runs active on ' + moment.utc(started).format('D MMM');
+
+            props.setFilterPill(pill(text));
```
lieut-data, line 81: nit: It feels weird to be passing around a JSX component like this, especially persisted in state and passed up to the parent, though it does have the advantage of coupling the text with the onClick handler without having the intermediates "know" about it. No immediate suggestion for improvement, though.

I wonder

lieut-data (approved):
Looks good! Just a handful of thoughts.

webapp/src/components/assets/icons/clear_icon.tsx
cpoile: Yep, we could do that too! I kind of wanted to keep it as close to the original `i` element as possible, to let the caller wrap it in a `span` or whatever if they wanted to. But i'm 0/5 on which approach. Or maybe 1/5 to keep it close to `i`. :)

webapp/src/components/backstage/playbooks/stats_view.tsx
cpoile: I see what you mean. We could pass around the text and have the parent, or the sibling create the pill when there's text available. On the other hand, this way the parent and sibling don't care, they just pass around and display an element (even if it's null). All logic and pill creation is isolated in the stats component. 
But yah, it's not the cleanest.

lieut-data: I imagine the "right" thing to do is actually to use shared state here, ala Redux. Not necessary now!

agarciamontoro (changes requested):
Really nice feature, and I did not find anything remarkable in the code :tada:

I have found a bug while trying it out, though: after you click on any of the graphs to filter the run list, if you click on the Owner filter, it crashes. The console shows this:

image
image

agarciamontoro (approved):
A `make clean && make deploy` (after updating nvm to 16.4) fixed that error only I could reproduce :man_shrugging: Sorry for the noise!

PR #627, opened 2021-07-08
MM-36766 - Remove RHS tabs

cpoile:
#### Summary
- First step in the RHS redesign.
- Removes the tabs
- The `RHSRunDetails` is the parent of two components: `RHSAbout` and `RHSChecklists`
- @agarciamontoro please rename `RHSAbout` if you want to, of course!
- @agarciamontoro I did not touch the styling in `RHSAbout`. It's mostly the old style scss, so I'm not sure what you want to do with that.
- Not gating with a flag ~~because I'm assuming we're just flipping the switch? (Plus I ripped out a bunch of code that the old way needed...)~~ edit: because we're merging to a feature branch

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36766

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/rhs/rhs_run_details.tsx
```diff
@@ -0,0 +1,48 @@
...
+    renderThumbVertical, renderView,
+    RHSContainer,
+    RHSContent,
+} from 'src/components/rhs/rhs_shared';
+import {currentPlaybookRun} from 'src/selectors';
+import RHSAbout from 'src/components/rhs/rhs_about';
+import RHSChecklists from 'src/components/rhs/rhs_checklists';
+
+const RHSRunDetails = () => {
+    const playbookRun = useSelector<GlobalState, PlaybookRun | undefined>(currentPlaybookRun);
```
agarciamontoro, line 22: We can remove the type here, but I don't know if you actually want it there for being super explicit? I think I prefer readability over specificity here, but 1/5.

```suggestion
    const playbookRun = useSelector(currentPlaybookRun);
```

tests-e2e/cypress/integration/frontstage/rhs/timeline-spec.js
```diff
@@ -1,450 +1,452 @@
...
-        cy.findByTestId('dropdownmenu').within(() => {
-            // # Click on desired filter
-            cy.findByText(option).click();
-        });
-
-        // # Hide menu
-        cy.findByTestId('timeline').click();
-    });
-};
+//import {TINY} from '../../../fixtures/timeouts';
```
agarciamontoro, line 4: Should we just remove it instead of commenting?

agarciamontoro (approved):
Yay, looks perfect! I left a couple of non-blocking comments. Thanks!

About the feature branch: I'm actually ok merging this to master. Any reason not to do it? It does not look actually that bad, the only weird thing is the spacing between the about and the checklists.
image

tests-e2e/cypress/integration/frontstage/rhs/timeline-spec.js
cpoile: @agarciamontoro Do you mean this one import, or the whole file? If you mean the whole file: I wanted to keep it, because for sure we're going to want to bring it back and use it for the timeline in the backstage. Or we could remove it and remember where it is in the git history (I wish there was a better way to stash code across branches and through time. do you know of any?)

webapp/src/components/rhs/rhs_run_details.tsx
cpoile: Awesome, done. Agree, I prefer implicit types for that.

cpoile:
@agarciamontoro The reason I was thinking to use a branch is the case where we're halfway through the feature and we have to worry about the next release. I think having it half done, or having that unrelated release create pressure to commit an in-progress PR just so that the product isn't in a broken state -- I think we'd want to avoid that.

What does everyone else think?

tests-e2e/cypress/integration/frontstage/rhs/timeline-spec.js
agarciamontoro: I mean the whole file. Regarding your question, I don't know of any way, maybe just create a branch with what you want? You'll still need to remember which branch you used, though.

cpoile: Personally, I would prefer to comment it out, so that we know exactly where it is and don't have to waste time finding it again. That file is also an annoyance, reminding us that we have to put it back in. :)

agarciamontoro:
Sorry I was not clear enough before: I'm also ok with the feature branch. It's just that even if we would cut a new release only with this PR, it would not be that bad, as the RHS looks good. And then I think it will be one isolated PR from each of us for each section.
But I agree it could add some pressure when the release comes, so I'm ok using the branch for this specific feature (and I think using a feature flag here is not even an option, it looks like more work than just doing the whole feature).

lieut-data (approved):
Just skimmed this one for now -- looks good! Confused about the branch discussion, though -- is the issue that tasks are missing somehow, or did we just render them below the current status, making it untidy but functional?

Prefer merging to `master` if at all possible, unless definitely broken.

cpoile:
@lieut-data At the moment there isn't a way to update status, other than slash commands. So howabout we merge this to the branch until we at least have a fully functional RHS? That should be @agarciamontoro 's first commit. Then we'll merge the branch and develop the rest on master. Sound good?

lieut-data:
Works for me! But I'm curious why the `Update Status` button would have disappeared from the remaining tab?

cpoile:
@lieut-data Ah, because there's no more footer -- the update status button moved into the top component, and I didn't move it there. I was working under the assumption that this was a quick PR to get @agarciamontoro unblocked and then the next PR would make it functional.

lieut-data:
Works for me, @cpoile! 👍🏼

PR #628, opened 2021-07-09
MM-36876 - Fix: Participant count in playbook dashboard incident list is not fetching correctly

cpoile:
#### Summary
- Send the participant list so the user's client does not try to fetch the users list for a channel for which they are not a member (doing so returned a 403, and the code would try again on an infinite loop)
- On the plus side, this speeds up the front end noticeably -- the client needs far fewer requests when loading the a playbook's dashboard, or when viewing a run's overview page.
- See the ticket for more details, a link to the thread, and repro steps

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36876

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


server/sqlstore/playbook_run.go
```diff
@@ -110,6 +111,29 @@ func applyPlaybookRunFilterOptionsSort(builder sq.SelectBuilder, options app.Pla
 
 // NewPlaybookRunStore creates a new store for playbook run ServiceImpl.
 func NewPlaybookRunStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLStore) app.PlaybookRunStore {
+	// construct the participants list so that the frontend doesn't have to query the server, bc if
+	// the user is not a member of the channel they won't have permissions to get the user list
+	participantsCol := `
+        COALESCE(
```
cpoile, line 117: I'm not super happy about this, but gosec won't let me concat or fmt.Sprintf in the different agg function, so we're left with this.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
```diff
@@ -44,30 +41,35 @@ interface Props {
     playbookRun: PlaybookRun;
 }
 
-const Participants = (props: Props) => {
-    const profilesInChannel = useProfilesInChannel(props.playbookRun.channel_id);
+const Participants = (
```
cpoile, line 44: Trying something new here. I think it really cleans up the body, at the expense of a multiline param definition. But since we spend more time reading the body, we should value clarity in the body over terseness in the fn definition (imo). Tell me what you think.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
```diff
@@ -44,30 +41,35 @@ interface Props {
...
+    {
+        playbookRun: run,
+        playbookRun: {participant_ids: ids},
```
cpoile, line 47: `ids` could be `participantIds` if that makes the rest clearer. 0/5

lieut-data:
Does this mean channels associated with playbook runs now have knowledge of their membership gated by channel membership /or/ playbook membership? I know we were about to embark on exposing this anyway (i.e. status updates and all), but just want to make sure we've got buy-in here.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
crspeller: I have done this both ways. I think I like the `props.` pattern better because it's immediately clear what is a prop. Verses this destructuring which makes them look the same as locals.

lieut-data (dismissed):
Let's sync up next week on the plan here to make sure we're actually ok with "leaking" the participant list regardless of channel membership, and to do it holistically. (I haven't actually reviewed any code yet.)

cpoile:
@lieut-data Note for when you get back: Tao and Abijit confirmed that it is okay to be sending the participant list regardless of channel membership, since you are allowed to see the participants list in the Run overview screen anyway (even if you're not a member of the channel). 

So, given that, I'm merging now, considering it's a bug that we should fix (we're currently calling the users endpoint on an infinite loop).

server/sqlstore/support_for_test.go
```diff
@@ -405,6 +407,34 @@ func setupPostsTable(t testing.TB, db *sqlx.DB) {
 	require.NoError(t, err)
 }
 
+func setupBotsTable(t testing.TB, db *sqlx.DB) {
+	t.Helper()
+
+	// This is completely handmade
```
agarciamontoro, line 413: What does handmade mean here? Is this not the definition of the table in the server?

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
agarciamontoro: I also like using `props.` because you don't have to remember what's a prop and what isn't. 2/5

agarciamontoro (approved):
It looks good! Great solution :) :tada:

crspeller (approved):

server/sqlstore/support_for_test.go
cpoile: correct -- it only has the fields we need

PR #629, opened 2021-07-10
MM-36347 - Fix for spamming users endpoint

cpoile:
#### Summary
- can review by commit
- change how we use `useProfilesInCurrentChannel` so that only the parent calls it; the children just need to use the selector
- fixed the infinite loop just in case
- Note: I tried to have `useProfilesInCurrentChannel` just `useCurrentChannel` and then delegate to the `useProfilesInChannel` hook, but that interacts in a weird way with how we're using react-beautiful-dnd: because we're hiding the drag handle in the hovermenu of our checklist items, react-beautiful-dnd complains it is `unable to find drag handle`. This spits out about 100 console errors normally. When changing `useProfilesInCurrentChannel` to delegate to the `useProfilesInChannel` hook, you get ~1000 console errors. Filed a new bug for the main issue: https://mattermost.atlassian.net/browse/MM-36948 .  When that's fixed, I will simplify these hooks.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36347

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data (approved):
This looks pretty clean (especially reviewing by commit). @crspeller, any risks that we should consider before merging for v1.14?

agarciamontoro (approved):
Elegant solution!

webapp/src/hooks.ts
```diff
@@ -186,14 +186,18 @@ export function useExperimentalFeaturesEnabled() {
...
+    const [fetched, setFetched] = useState(false);
 
     useEffect(() => {
         if (profilesInChannel.length > 0) {
             return;
         }
 
-        dispatch(getProfilesInChannel(channelId, 0, PROFILE_CHUNK_SIZE));
-    }, [channelId, profilesInChannel]);
+        if (!fetched) {
```
crspeller, line 196: Won't this not fetch for new channels now?

crspeller (changes requested):

webapp/src/hooks.ts
lieut-data: I ended up resolving this in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/632 by just `profilesInChannel` from the dependency list for `useEffect`. The main underlying problem is that `getProfileSetForChannel` returns a new object every time it's called.

cpoile: @crspeller Ahh, maybe you're right. Need to add `if (channelId !== lastChannelId) { setFetched(false) }` above. Does that make sense? Sorry but do you think you can add that? (I'm not on my work machine)

cpoile: @lieut-data Would the above problem still be there? (when no remounting, and changing to a new channel unseen channel, `fetched` might be set to true)

lieut-data: See linked PR, but by just having `channelId` as a dependency to `useEffect`, it only runs when that channel changes, fetching profiles at most once for a given channel id per component.

PR #630, opened 2021-07-10
MM-36266 - Fix: Incident commander can be changed to the incident bot

cpoile:
#### Summary
- Remove bot from drop down menus, and disallow from command
- Chose to make changing owner or assignee through API be a no-op, to follow how we handle self-assignment

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36266

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -295,7 +295,7 @@ type PlaybookRunService interface {
 	IsOwner(playbookRunID string, userID string) bool
 
 	// ChangeOwner processes a request from userID to change the owner for playbookRunID
-	// to ownerID. Changing to the same ownerID is a no-op.
+	// to ownerID. Changing to the same ownerID or to a bot is a no-op.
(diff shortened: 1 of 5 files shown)


lieut-data:
I'm not actually sure we want to disallow this. In fact, I could totally see assigning a run to a bot as a signal for that bot to drive interactions. It could even be an interesting way to build a "plugin" for playbooks.

iomodo (approved):
Looks great! Thanks @cpoile

lieut-data (changes requested):
Let's discuss if we actually want to do this. It might be "useless" right now, but it doesn't seem like something we should explicitly disallow.

cpoile:
@lieut-data So, optional and not going to merge? Or merge because at least it's done? :)

lieut-data:
Propose we err on the side of "less functionality" and leave things as-is until we have a compelling reason to do otherwise.

PR #632, opened 2021-07-13
MM-36876: useProfilesInChannel should try once

lieut-data:
#### Summary
Change `useProfilesInChannel` to only try getting the profiles in the channel once (or, at least as long as the given `channelId` never changes). Previously, the `profilesInChannel` selector would /always/ result in a new value, triggering the effect hook indefinitely.

Relates-to: https://mattermost.atlassian.net/browse/MM-36876

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36876

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/hooks.ts
+++ b/webapp/src/hooks.ts
@@ -193,7 +193,7 @@ export function useProfilesInChannel(channelId: string) {
         }
 
         dispatch(getProfilesInChannel(channelId, 0, PROFILE_CHUNK_SIZE));
-    }, [channelId, profilesInChannel]);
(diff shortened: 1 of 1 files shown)


agarciamontoro (approved):
Oh, good catch!

crspeller (approved):

lieut-data:
Failing test seems to be upstream -- I've seen it on other PRs. Will investigate, but not blocking.

PR #635, opened 2021-07-15
stop random scroll on edit playbook

lieut-data:
#### Summary
When editing a playbook, the unconditional `autoFocus=true` on the last visible description forced the page to scroll to that element. Only do this is the description is empty, making it easy to start typing a description when adding one but not scrolling when first loading.

#### Ticket Link
None.

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/checklist_item_input.tsx
+++ b/webapp/src/components/checklist_item_input.tsx
@@ -95,7 +95,7 @@ export const ChecklistItemDescription = (props: ChecklistItemDescriptionProps) =
         descriptionBox = (
             <>
                 <StyledTextarea
-                    autoFocus={true}
+                    autoFocus={!description}
                     value={description}
                     onBlur={save}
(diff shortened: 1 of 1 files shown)


lieut-data:
I'm separately looking at failing Cypress tests (among other Cypress improvements), but the failure above appears to be unrelated.

agarciamontoro (approved):
Approving to unblock, but what happens when the last visible description is empty? Isn't it the same as before?

lieut-data:
> Approving to unblock, but what happens when the last visible description is empty? Isn't it the same as before?

If the description is empty, it won't actually render the component, preventing the whole business. I think something similar happens with tasks, except we pre-fill them with data instead of allowing them to stay empty.

crspeller (approved):

PR #636, opened 2021-07-15
make playbook navbar sticky

lieut-data:
#### Summary
Make the playbook navbar sticky. It was technically already sticky, but had a `top: 0px` that prevented it from staying below the sticky header. This makes `Save` front and centre, instead of `X`; it also helps fix a few Cypress unit tests with the (yet to be merged) upgraded Cypress 7.

image

#### Ticket Link
None.

#### Checklist
- ~~[ ] Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/backstage.tsx
+++ b/webapp/src/components/backstage/backstage.tsx
@@ -34,6 +34,7 @@ import CloudModal from 'src/components/cloud_modal';
 
 import StatsView from './stats';
 import SettingsView from './settings';
+import {BackstageNavbar, BackstageNavbarIcon} from './backstage_navbar';
 
 const BackstageContainer = styled.div`
     background: var(--center-channel-bg);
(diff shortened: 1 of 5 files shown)


webapp/src/components/backstage/backstage.tsx
```diff
@@ -43,45 +44,6 @@ const BackstageContainer = styled.div`
...
-    color: var(--center-channel-color-56);
-
-    &:hover {
-        background: var(--button-bg-08);
-        text-decoration: unset;
-        color: var(--button-bg);
-    }
-`;
-
-export const BackstageNavbar = styled.div`
```
lieut-data, line 66: Something about using overriding `BackstageNavbar` as `PlaybookNavbar` caused webpack to fail to account for the circular dependency between these two files. Splitting it out solved the issue.

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -182,6 +182,10 @@ const retrospectiveReminderOptions = [
 // @ts-ignore
 const WebappUtils = window.WebappUtils;
 
+const PlaybookNavbar = styled(BackstageNavbar)`
+    top: 80px;
```
lieut-data, line 186: This is the only real change here.

calebroseland (approved):

iomodo (approved):

PR #637, opened 2021-07-15
properly initialize fixture users

lieut-data:
#### Summary
`user-2` as a username doesn't actually exist -- its `samuel.tucker` in the test data -- and we weren't correctly initializing preferences for the welcome modal for this users, used in a few tests. I presume the failure itself was upstream from incoming changes around the welcome modal.

#### Ticket Link
None.

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/support/index.js
+++ b/tests-e2e/cypress/support/index.js
@@ -13,11 +13,13 @@ import './api/preference';
 import './api/user';
 import './test_commands';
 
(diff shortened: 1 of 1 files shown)


lieut-data:
This should fix at least one of the flaky unit tests we've been encountering.

crspeller (approved):

PR #641, opened 2021-07-19
[MM-35264] Workspace wide scope - playbooks

iomodo:
#### Summary
This PR adds playbooks to the workspace-wide context.

https://user-images.githubusercontent.com/2340127/126183656-bf7e6a20-9006-4e49-b02e-61913c52c119.mp4

https://user-images.githubusercontent.com/2340127/126183707-d974eab8-b97d-4b6d-8d0e-f10a88179e58.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-35264

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+import {PrimaryButton} from 'src/components/assets/buttons';
+
+import {useClientRect} from 'src/hooks';
+
+import TeamWithIcon from './team_with_icon';
+
+interface Option {
+    value: string;
+    label: JSX.Element | string;
```
crspeller, line 18:
```suggestion
    label: React.ElementType;
```

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+    const [position, setPosition] = useState({x: 0, y: 0});
+    const toggleOpen = (x: number, y: number) => {
+        if (!isOpen) {
+            updateTeamOptions();
+        }
+        setPosition({x, y});
+        setOpen(!isOpen);
+    };
+
+    const [teamOptions, setTeamOptions] = useState<Option[]>([]);
```
crspeller, line 50: I am confused. Why is this state? It seems to be derived directly from props?

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+                teamId: team.id,
+            } as Option);
+        });
+
+        setTeamOptions(optionList);
+    }
+
+    // Fill in the userOptions on mount.
+    useEffect(() => {
+        updateTeamOptions();
```
crspeller, line 71: Why?

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+    }, []);
+
+    function getTeam(teamId: string) {
+        return props.teams.filter((team) => team.id === teamId)[0];
```
crspeller, line 75: Are you looking for `find`?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+const dropdownYShift = 27;
+
+interface Position {
+    x: number;
+    y: number
+}
+
+export default function CreatePlaybookTeamSelector(props: Props) {
+    const [isOpen, setOpen] = useState(false);
```
crspeller, line 40: This open and coordinate state might look nicer as a one-off custom hook.

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        const innerHeight = window.innerHeight;
+        const numTeamsShown = Math.min(6, teamOptions.length);
+        const spacePerProfile = 48;
+        const dropdownReqSpace = 80;
+        const extraSpace = 10;
+        const dropdownBottom = rect.top + dropdownYShift + dropdownReqSpace + (numTeamsShown * spacePerProfile) + extraSpace;
+        setMoveUp(Math.max(0, dropdownBottom - innerHeight));
+    }, [rect, teamOptions.length]);
```
crspeller, line 102: rect won't change except on mount, and teamOptions appears to be derived from props so I think this can be cleaned up when teamOptions is fixed.

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+    let target;
+    if (props.withButton) {
+        target = (
+            <PrimaryButton
```
crspeller, line 106: Maybe leave the styling to the children? Then we don't have to so the `withButton` thing and we just watch for clicks on the container div like below.

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+            ...provided,
+            backgroundColor: state.isSelected ? hoverColor : bgHover,
+            color: 'unset',
+        };
+    },
+};
+
+// styled components
+interface DropdownProps {
+    children: JSX.Element;
```
crspeller, line 183: I think the correct type for children is `React.ReactNode`

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,251 @@
...
+        if (props.dependsOnPosition) {
+            return props.position ? props.position.x : 0;
+        }
+        return 0;
+    }}px;
+    position: ${(props) => (props.dependsOnPosition ? 'fixed' : 'absolute')};
+    z-index: 2;
+`;
+
+const Dropdown = ({children, isOpen, showOnRight, moveUp, target, onClose, dependsOnMousePosition, position}: DropdownProps) => {
```
crspeller, line 229: Why is this copy pasted?

crspeller (changes requested):
Excited for these changes! I think that's good for now. Will do a more thorough review after the first set of changes.

abhijit-singh (commented):
Awesome work on the team selector popup for templates opening up wherever the user clicks! 🎉

A few small points:
1. The search in the team selector popup seems to have an`arrow icon`, like in a dropdown. It doesn't really do anything, right? It isn't there in other places and I don't think it was there in the last PR. Can we remove it?
2. Related to the above point, there's also a `x` cross icon that appears if there's a selected team, but clicking on it doesn't do anything. It is also there in the `Owner` filter. Ideally, we should remove that too, but doesn't have to be a part of this effort. 
3. The key icon should be in the same color as the `status - online indicator`(#06D6A0). 
4. Wherever a button is triggering a dropdown, it'll be ideal if the dropdown is placed just below the button instead of overlapped on it. Screenshot attached - 
image

iomodo:
Thanks for the wonderful suggestion @crspeller ! I think I addressed all of them.
Will address @abhijit-singh 's suggestions tomorrow.

iomodo:
Addressed @abhijit-singh 's suggestions.

iomodo:
Some screenshots for @abhijit-singh 
Screen Shot 2021-07-21 at 22 30 02
Screen Shot 2021-07-21 at 22 30 19

abhijit-singh:
> Some screenshots for @abhijit-singh
> Screen Shot 2021-07-21 at 22 30 19

@iomodo Awesome, the styling changes look good! 

One more thing from the screenshot above - if all teams have a playbook each, can we show the key icon on the `Create playbook` button like we do today? That was the behaviour suggested in [this conversation](https://community-daily.mattermost.com/core/pl/buwj7zmajbyc78wq8tyxy13phh): 

> 3. Two playbooks created, one in each team: There is a key icon on the Create playbook button and clicking on it triggers a modal.


`

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+interface DropdownPosition {
+    x: number;
+    y: number;
+    isOpen: boolean;
+}
+
+export default function CreatePlaybookTeamSelector(props: Props) {
+    const teamOptions = getTeamOptionList();
+    const [dropdownPosition, setDropdownPosition] = useState({x: 0, y: 0, isOpen: false});
```
crspeller, line 40: Better! If you made this into a custom hook it would look really nice.

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,226 @@
...
+                components={{DropdownIndicator: null, IndicatorSeparator: null}}
+                backspaceRemovesValue={false}
+                controlShouldRenderValue={false}
+                hideSelectedOptions={false}
+                menuIsOpen={true}
+                options={teamOptions}
+                placeholder={'Search'}
+                styles={selectStyles}
+                tabSelectsValue={false}
+                onChange={(option) => onSelectedChange(option as Option)}
```
crspeller, line 134: So we need this extra arrow function just to do the type conversion? Or can we just change the type the function accepts to match and do it within there.

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    function getTeam(teamId: string) {
+        return props.teams.find((team) => team.id === teamId);
+    }
+
+    const onSelectedChange = async (value: Option | undefined) => {
+        toggleOpen(0, 0);
+
+        const team = getTeam(value?.teamId || '');
```
crspeller, line 69: Could we not just attach the team object to the option list to avoid this?

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,226 @@
...
+        if (team) {
+            props.onSelectedChange(team);
+        }
+    };
+
+    // Decide where to open the team selector
+    const [rect, ref] = useClientRect();
+    const [moveUp, setMoveUp] = useState(0);
+
+    useEffect(() => {
```
crspeller, line 79: Why have this useEffect and state to calculate this? It's not a lot of calculation.

webapp/src/components/team/team_with_icon.tsx
```diff
@@ -8,6 +8,7 @@ import {Client4} from 'mattermost-redux/client';
 import styled from 'styled-components';
 interface Props {
     team: Team;
+    allowed: boolean;
```
crspeller, line 11: It's more like not licensed right?

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,226 @@
...
+        if (props.dependsOnPosition) {
+            return props.position ? props.position.x : 0;
+        }
+        return 0;
+    }}px;
+    position: ${(props) => (props.dependsOnPosition ? 'fixed' : 'absolute')};
+    z-index: 2;
+`;
+
+const Dropdown = ({children, showOnRight, moveUp, target, onClose, dependsOnMousePosition, position}: DropdownProps) => {
```
crspeller, line 204: Why is this copy pasted form profile selector? Could we not share some code here?

crspeller (changes requested):

webapp/src/components/backstage/new_playbook.tsx
```diff
@@ -13,14 +16,23 @@ interface Props {
 
 export const NewPlaybook = (props: Props) => {
     const allowPlaybookCreation = useAllowPlaybookCreationInCurrentTeam();
+    const teams = useSelector<GlobalState, Team[]>(getMyTeams);
```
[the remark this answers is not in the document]
iomodo: Missed that. Do you mean using `getMyTeams` without the `useSelector`? Won't we need a global state object for that? It seems an overkill to introduce `store` object in `new_playbook`. Am I missing something?

webapp/src/components/team/team_with_icon.tsx
iomodo: Depends on the context actually. It is used in `Team` filter as well which has nothing to do with licensing. Maybe `showIcon` is a better name here.

iomodo: Or maybe `showNotLicensedIcon`?

webapp/src/components/team/create_playbook_team_selector.tsx
iomodo: `Dropdown` is in 4 different places copy/pasted (with slight changes). I'd suggest to address this in a different PR for the whole repo.

webapp/src/components/backstage/template_selector.tsx
```diff
@@ -124,51 +125,84 @@ const TemplateItemDiv = styled.div`
...
+const TemplateSelector = ({templates = PresetTemplates, onSelect, teams, allowPlaybookCreationInTeams, showUpgradeModal}: Props) => {
+    const allowPlaybookCreation = isPlaybookCreationAllowed(allowPlaybookCreationInTeams);
     return (
         <BackgroundColorContainer>
             <RootContainer>
                 <InnerContainer>
                     <Title>
                         {'Create a playbook'}
-                        {!allowPlaybookCreation && <PositionedUpgradeBadge/>}
+                        {!allowPlaybookCreation && <NotAllowedIcon className='icon icon-key-variant-circle'/>}
```
agarciamontoro, line 153: Same comment for reusing (the refactored version of) UpgradeBadge, just not to duplicate the icon definition.

webapp/src/components/team/create_playbook_team_selector.tsx
```diff
@@ -0,0 +1,207 @@
...
+        </Dropdown>
+    );
+}
+
+// styles for the select component
+const selectStyles: StylesConfig = {
+    control: (provided) => ({...provided, minWidth: 240, margin: 8}),
+    menu: () => ({boxShadow: 'none'}),
+    option: (provided, state) => {
+        const hoverColor = 'rgba(20, 93, 191, 0.08)';
```
agarciamontoro, line 128: Is this theme-independent?

webapp/src/components/backstage/playbook_list.tsx
```diff
@@ -207,18 +212,22 @@ const PlaybookList = () => {
...
+                    allowPlaybookCreationInTeams={allowPlaybookCreationInTeams}
+                    showUpgradeModal={showUpgradeModal}
                 />
             }
             {
                 (playbooks?.length === 0) &&
                 <>
                     <NoContentPage
-                        onNewPlaybook={newPlaybook}
+                        onNewPlaybook={(team: Team) => newPlaybook(team)}
```
agarciamontoro, line 227: Does Typescript complain here? It should not, right?

agarciamontoro (approved):
Great job here, @iomodo, I guess these PRs are being tricky! I left a couple of non-blocking comments, approving right away :)

webapp/src/components/backstage/new_playbook.tsx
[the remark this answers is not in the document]
crspeller: No I mean why are we selecting a bunch of teams, then picking the one we want? Why not just use the get team selector directly?

webapp/src/components/team/create_playbook_team_selector.tsx
crspeller: Alright. Could you make a ticket for it?

crspeller (approved):
Awesome! The rest is non-blocking.

webapp/src/components/team/create_playbook_team_selector.tsx
iomodo: Yes. Same behaviour as in profile select.

webapp/src/components/team/team_with_icon.tsx
```diff
@@ -48,6 +49,7 @@ const TeamWithIcon = (props: Props) => {
             <NameWrapper>
                 {props.team.display_name}
             </NameWrapper>
+            {props.showNotLicensedIcon ? <NotAllowedIcon className='icon icon-key-variant-circle'/> : ''}
```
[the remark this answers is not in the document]
iomodo: I created separate ticket for refactoring `UpgradeBadge` and `NotAllowedIcon`

iomodo:
Screenshots for @abhijit-singh 

https://user-images.githubusercontent.com/2340127/127495330-1725b0f5-c646-41a3-99c9-b98e9f73a592.mp4


https://user-images.githubusercontent.com/2340127/127495345-54631875-b0f3-4be9-bd5f-19cd705e332d.mp4

abhijit-singh (approved):
Thanks @iomodo!

A couple of non-blocking things so adding them here and approving:
1. Ideally the key icon should have a white background, like @agarciamontoro had in his implementation that can be seen here https://github.com/mattermost/mattermost-plugin-playbooks/pull/592
2. We should try and explain to the user why creating playbooks for certain teams is locked while for others it isn't. The idea is to change the upgrade modal text to explain that the limit for that particular team has been hit. 
Suggestion: `This team already has a playbook. Upgrade to create multiple playbooks to prescribe different workflows for each team.`

^ This will mean we'll need to show different text when the user clicks on a team and sees the modal vs when the user clicks on `+ Create playbook` and sees the modal. 
In an attempt to unify those, maybe we can use: `In your current plan, each team can have just 1 playbook. Upgrade to create multiple playbooks to prescribe different workflows for each team.`

Looking to @justinegeffen for feedback & review .

justinegeffen:
Thanks, @abhijit-singh! I really like your suggestion of a generic option and your wording makes sense. I've reworded it a little, LMK if it works for you: 

"Your subscription allows one playbook per team. Upgrade your subscription and create multiple playbooks with unique workflows for each team."

PR #646, opened 2021-07-22
MM-36854: First pass at RHS top section

agarciamontoro:
#### Summary

This builds the general layout and first features of the top section of the new RHS. The Post update button is filling the whole section, but it will soon make room for the update timer. Also, I removed the second button in the on hover menu, that will be added in future PRs. And, as you know, still no inline editing experience.

I think a video will be much easy to understand the current state than a thorough description, so there it is:

https://user-images.githubusercontent.com/3924815/126681323-501da9a7-230f-44fe-9795-d785d6280620.mp4

Please note that this PR is being merged mainly to get the RHS to a functional state, that's why I removed some of the features. @abhijit-singh, I assigned you as reviewer to let you take an early look and note any improvements I need to make, but if you're ok with it, we will merge this as is (unless there's something completely horrendous) and I'll polish everything in the next PRs.

Adding also @itao for visibility.

Where we are:
- [X] Tabs removed (@cpoile's work).
- [X] Checklist redesign layout with no additional features (@cpoile's work).
- [ ] Top section redesign with no additional features. **<-- We are here**
- [ ] Next status update timestamp (https://mattermost.atlassian.net/browse/MM-37325).
- [ ] Title and description inline editing (https://mattermost.atlassian.net/browse/MM-37326).
- [ ] Save collapsed checklist state per user (https://mattermost.atlassian.net/browse/MM-36913).
- [ ] Filter tasks with default (https://mattermost.atlassian.net/browse/MM-36914)
- [ ] Collapse/expand all checklists (https://mattermost.atlassian.net/browse/MM-36915)
- [ ] Finish Run button (https://mattermost.atlassian.net/browse/MM-37323)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36854

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


agarciamontoro:
Of course all E2E tests are failing. On it :)

agarciamontoro:
Also, please don't review commit by commit, they're quite a mess :)

cpoile (approved):
Awesome!

webapp/src/components/rhs/rhs_about.tsx
```diff
@@ -51,59 +78,150 @@ const RHSAbout = (props: Props) => {
...
+                title={collapsed ? 'Expand' : 'Collapse'}
+                className={(collapsed ? 'icon-arrow-expand' : 'icon-arrow-collapse') + ' icon-16 btn-icon'}
+                onClick={toggleCollapsed}
+            />
+        </ButtonsRow>
+    );
+};
+
+const ButtonsRow = styled(HoverMenu)`
+    position: absolute;
```
cpoile, line 154: nit: doesn't need absolute (HoverMenu has it)

webapp/src/components/rhs/rhs_about.tsx
```diff
@@ -25,6 +28,30 @@ const RHSAbout = (props: Props) => {
...
+        return (
+            <Container
+                onMouseEnter={() => setHovered(true)}
+                onMouseLeave={() => setHovered(false)}
+            >
+                {true &&
+                <Buttons
+                    collapsed={collapsed}
+                    toggleCollapsed={toggleCollapsed}
+                /> }
```
calebroseland, line 46: Remove condition?

calebroseland (commented):
CleanShot 2021-07-23 at 10 45 35

Something seems broken with the participants on hover, and only the right-most one shows while not hovering.

And while minor, the whole block gray bg on hover feels a little over stimulating to me, is there maybe a more subtle alternative? If kept, perhaps the upper margin/padding could be changed to remove the small sliver that stays white.

webapp/src/components/rhs/rhs_about.tsx
agarciamontoro: Actually, I have to use the right condition! Fixed in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/646/commits/fb7bc124a0a50faac14decd1a2e24685253cfa7d Good catch :)

webapp/src/components/rhs/rhs_about.tsx
agarciamontoro: Thank you! Fixed in https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/646/commits/5d2d4c1f717c8555cbb0ba68b67da85a79b7c391

webapp/src/hooks/general.ts
```diff
@@ -364,11 +364,7 @@ export function useOpenCloudModal() {
     }
 
     // @ts-ignore
-    if (
-        !window.WebappUtils?.modals?.openModal ||
-        !window.WebappUtils?.modals?.ModalIdentifiers?.CLOUD_PURCHASE ||
-        !window.Components?.PurchaseModal
-    ) {
+    if (!window.WebappUtils?.modals?.openModal || !window.WebappUtils?.modals?.ModalIdentifiers?.CLOUD_PURCHASE || !window.Components?.PurchaseModal) {
```
calebroseland, line 367: I actually prefer the prev. 😅 More readable IMHO—less horizontal scanning/scrolling and each condition gets its own line

agarciamontoro: @calebroseland I agree, but then the linter complaints on the second and third line about WebappUtils not existing, so I have to add `// @ts-ignore` in each of them. We should actually fix that properly (I'll rely on your Typescript knowledge), but we can tackle it in a future PR if that's ok

calebroseland (commented):
Could we add keyboard accessibility for these elements?
CleanShot 2021-07-27 at 10 08 51@2x

webapp/src/hooks/general.ts
calebroseland: Totally fine—it's a 1/5 preference for me. Re: `// @ts-ignore`. I renamed a file to `src/webapp_globals.ts`, this does a single ts-ignore for each global-destructure, and is then re-exported to create a centralized place we could import webapp stuff. Maybe the `Components` and `WebappUtils` stuff could be added there?

agarciamontoro:
@calebroseland, done! I learned a bunch doing this, thanks for the review! Please take a look :)

The framerate in the video is all messed up at the end, please take a look locally, both Space and Enter should behave as clicking on the elements.

https://user-images.githubusercontent.com/3924815/127214265-86015f40-88aa-4331-b5ba-96967dd97ebf.mp4

abhijit-singh (approved):
@agarciamontoro This is awesome. And great suggestion on keyboard accessibility by @calebroseland!

A couple of non-blocking good to haves if they don't take too much time: 
1. In the RHS header, can the entire text + icon be clickable instead of just the icon? 
image

2. It'll be awesome for the owner selector input to have a hover state.  Just changing the background to `16% Center Channel Text` instead of the 8% should do. 
image

agarciamontoro:
@abhijit-singh, done! Check out the video below. The top and bottom padding is less than the one in the designs because we're restricted to 24px height containers inside the RHS title, which is a constraint that comes from the webapp itself. 

https://user-images.githubusercontent.com/3924815/127290086-0b4cbbe3-bb00-4c1b-a219-971f3211f29e.mp4

webapp/src/hooks/general.ts
agarciamontoro: Sounds good. I'll do it in the next PR, once this is merged. Thanks for the suggestion!

abhijit-singh:
@agarciamontoro That's perfect, thanks!

calebroseland (approved):

PR #647, opened 2021-07-23
MM-36915 - Collapse/expand all checklists (save state per user)

cpoile:
#### Summary
- Does what it says in the title, pretty straightforward.
- Collapsed icon changes to expand even when user manually closes each checklist
- Used a `setAllChecklistsCollapsedState` action which takes the number of checklists. This is one more param to pass, but otherwise there's no way for the reducer to know how many object entries to set as `true` or `false`. 

- With only one checklist collapsed:
image
- With all checklists collapsed:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-36915

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/selectors.ts
```diff
@@ -62,6 +62,14 @@ export const currentPlaybookRun = createSelector(
     },
 );
 
+export const currentChecklistCollapsedState = createSelector(
+    getCurrentChannelId,
+    pluginState,
+    (channelId, plugin) => {
+        return plugin.checklistCollapsedState[channelId] ?? {} as Record<number, boolean>;
```
agarciamontoro, line 69: Oh, I did not know about `??`, it's cool.

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -14,32 +14,71 @@ import {
...
     const checklists = props.playbookRun.checklists || [];
 
+    const allCollapsed = (() => {
+        for (let i = 0; i < checklists.length; i++) {
+            if (!checklistsState[i]) {
+                return false;
+            }
+        }
+        return true;
+    })();
```
agarciamontoro, line 56: Why the immediately called function?

webapp/src/components/rhs/collapsible_checklist.tsx
```diff
@@ -12,32 +12,36 @@ import TextWithTooltipWhenEllipsis from 'src/components/widgets/text_with_toolti
 export interface Props {
     title: string;
     index: number;
+    collapsed: boolean;
+    setCollapsed: (now: boolean) => void;
```
agarciamontoro, line 16: Why is this parameter called `now`?

agarciamontoro (approved):
It looks great! I was going to suggest using a `channelId => array of booleans` as the checklists state, but I guess we cannot do that because we don't know the length of the checklists a priori, right?

I left a a couple of questions below, none blocking.

webapp/src/selectors.ts
cpoile: It's a bit safer -- `||` will give you the right for any falsey value.  Doesn't matter much for use since an empty collapsed state would be `{}` which is not falsey. But maybe it's a good habit.

webapp/src/components/rhs/rhs_checklists.tsx
cpoile: No need to check everytime we call it, just needs to be computed once on render.  But anyway I moved this to a selector, so it's a moot point. :)

webapp/src/components/rhs/collapsible_checklist.tsx
cpoile: I just meant the new state. `now` is a bad name, will change, thanks!

webapp/src/selectors.ts
calebroseland: Will this thrash a bit with returning a new object? Maybe hoist the empty `{}` outside so its a constant? (0/5; probably a premature optimization, ignore as needed)

calebroseland (approved):
LGTM! Just one suggestion re: the react-beautiful-dnd warning, and a few non-blocker nits on whitespace and perf

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -14,32 +14,63 @@ import {
...
 import {setChecklistItemState, clientReorderChecklist} from 'src/client';
 import {ChecklistItemDetails} from 'src/components/checklist_item';
 import {isMobile} from 'src/mobile';
 import CollapsibleChecklist from 'src/components/rhs/collapsible_checklist';
+import {HoverMenu, HoverMenuButton} from 'src/components/rhs/rhs_shared';
+import {currentChecklistAllCollapsed, currentChecklistCollapsedState} from 'src/selectors';
+
+// disable all react-beautiful-dnd development warnings
+// @ts-ignore
+window['__react-beautiful-dnd-disable-dev-warnings'] = true;
```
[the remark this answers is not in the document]
cpoile: Thank you! However I tried it and it doesn't fix the errors. Getting hundreds still. I think it's because the drag handles aren't attached to the dom and it doesn't like that. 
This might be fixed by a polish PR I'll be doing later (making the hover hidden instead of unattached).

webapp/src/components/rhs/collapsible_checklist.tsx
```diff
@@ -12,32 +12,36 @@ import TextWithTooltipWhenEllipsis from 'src/components/widgets/text_with_toolti
...
-const CollapsibleChecklist = ({title, index, items, children}: Props) => {
+const CollapsibleChecklist = (
+    {
+        title,
+        index,
+        collapsed,
+        setCollapsed,
+        items,
+        children,
+    }: Props) => {
```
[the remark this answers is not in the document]
cpoile: ahh, that's so much nicer. I had to find my editor setting to format like that, but now I can. Thank you!

PR #651, opened 2021-07-27
MM-35598 - Finish run button

cpoile:
#### Summary
- Adds a default status option to the `/playbook update` slash command, including autocomplete (and autocomplete parsing). But this will be outdated as soon as we do the next epic, so it's only temporary.
- Adds `Finish run` button and confirm modal.
  - Button is tertiary until all checklist items are checked, then it turns primary.
  - Button disappears when run is not in an active state (I assumed this is what we want to do)
  - Confirm modal text changes its text and plurality based on number of tasks unchecked. 
  - Number of tasks left unchecked is ephasized as in the screencaps below (following the designs)

#### Editor review needed
@justinegeffen The text will be the following, please suggest edits from these:

- 0 tasks unchecked
`Are you sure you want to finish the run? You will not be able to add or change tasks and checklists once you finish the run.`
- 1 task unchecked
`Are you sure you want to finish the run? There is 1 outstanding task and it will be automatically marked as skipped. You will not be able to add or change tasks and checklists once you finish the run.`
- 2 or more tasks unchecked
`Are you sure you want to finish the run? There are 2 outstanding tasks and they will be automatically marked as skipped. You will not be able to add or change tasks and checklists once you finish the run.`

#### UX review
@abhijit-singh I followed the designs, but you /are/ able to check and edit tasks even when the run is `Resolved` or `Archived` -- do we change the text then?

Looks like:
image

image

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-37323

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


server/command/command.go
```diff
@@ -656,7 +663,14 @@ func (r *Runner) actionUpdate() {
...
 
-	err = r.playbookRunService.OpenUpdateStatusDialog(playbookRunID, r.args.TriggerId)
+	defaultStatus := ""
+	if len(args) == 1 {
+		defaultStatus = args[0]
+	} else if len(args) == 2 && args[0] == "--next-status" {
+		defaultStatus = args[1]
+	}
+
+	err = r.playbookRunService.OpenUpdateStatusDialog(playbookRunID, r.args.TriggerId, defaultStatus)
```
agarciamontoro, line 673: Should we check here what the value of `defaultStatus` is before sending the parameter down to the service function?

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -137,4 +174,35 @@ const ChecklistContainer = styled.div`
...
 
+const StyledTertiaryButton = styled(TertiaryButton)`
+    display: inline-block;
+    margin: 12px 0;
+`;
+
+const StyledPrimaryButton = styled(PrimaryButton)`
+    display: inline-block;
+    margin: 12px 0;
+`;
```
agarciamontoro, line 185: Nit: I ended up coding more or less the same thing (with different styles) and this is what I did. Not saying that you should refactor the styles out to a commons css, but just in case it works for you. 0/5 because in your case you're only repeating the styles twice (in my case, I had it three times, so it seemed more convenient).

```ts
const PostUpdateButtonCommon = css<CollapsedProps>`
    justify-content: center;
    flex: 1;
    ${(props) => props.collapsed && css`
        height: 32px;
        font-size: 12px;
        font-height: 9.5px;
    `}
`;

const PostUpdatePrimaryButton = styled(PrimaryButton)<CollapsedProps>`
    ${PostUpdateButtonCommon}
`;

const PostUpdateTertiaryButton = styled(TertiaryButton)<CollapsedProps>`
    ${PostUpdateButtonCommon}
`;

const PostUpdateDestructiveButton = styled(DestructiveButton)<CollapsedProps>`
    ${PostUpdateButtonCommon}
`;
```

agarciamontoro (approved):
Looks fantastic, @chris-overton! I left a couple of comments, but non-blocking at all, feel free to ignore them :) Thanks!

justinegeffen:
Thanks, @cpoile! Here are my suggestions:


*0 tasks unchecked*
Are you sure you want to finish the run? You can't add or change tasks and checklists when the run is finished.

*1 task unchecked*
Are you sure you want to finish the run? There is 1 outstanding task which will automatically be marked as **Skipped**. You can't add or change tasks and checklists when the run is finished.

*2 or more tasks unchecked*
Are you sure you want to finish the run? There are 2 outstanding tasks and they'll automatically be marked as skipped. You can't add or change tasks and checklists when the run is finished.

I'm torn between using "when the run has finished" and "when the run is finished". I think the latter is more fitting but just wanted to throw that out there and see what you think.

Otherwise I've tried to shorten it a bit but on the whole just minor edits. :)

justinegeffen (approved):

server/command/command.go
cpoile: I just almost went through with a validator, but then I realized if we do that then we should validate the field in the API as well, and then we have to return help text or a proper error and then we should probably do that for all the parameters in our command system... That seems like a lot of work for something that's going to change pretty soon anyway (custom statuses, or the new fewer statuses change Tao is thinking of). What we do have is autocomplete for the valid options, and the consequence for manually writing in an invalid option is simply no default is chosen in the dropdown. 

TLDR: a lot of work for something that's going to change soon anyway. :)

webapp/src/components/rhs/rhs_checklists.tsx
cpoile: Ohh, cool, that's neat. I didn't know you could just inline an entire styled component like that. 
I think you're right though--for my case twice (and only two lines of css) seems fine (it would be more code to pull it out). I'll keep that in mind though, thanks!

cpoile:
Thanks @justinegeffen! In the end we're going to remove that modal anyway. :( But we'll put it back when we have the new status update system (next release?).

calebroseland (approved):

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -22,34 +22,17 @@ import {ChecklistItemDetails} from 'src/components/checklist_item';
...
-            <span>{'Are you sure you want to finish the run?'}</span>
-            {
-                (tasks !== 0) &&
-                <>
-                    <span>{` There ${(tasks === 1) ? 'is' : 'are'} `}</span>
-                    <EmphasisText>{`${tasks} outstanding task${(tasks === 1) ? '' : 's'}`}</EmphasisText>
-                    <span>{` and ${(tasks === 1) ? 'it' : 'they'} will be automatically marked as skipped.`}</span>
-                </>
-            }
-            <span>{' You will not be able to add or change tasks and checklists once you finish the run.'}</span>
```
calebroseland: For whenever this is reimplemented, we have access to `intl`/`formatMessage` now, which I think would be well suited for these conditionals (even without the translations)

PR #652, opened 2021-07-27
Updated playbook templates

lieut-data:
#### Summary
As per @itao's [proposal](https://docs.google.com/document/d/1pJscJCqiL3CDemsBN-V_3NRvOeapB3nbjE7eEKTwmGY/edit), replace the existing templates:
* Blank (unchanged)
* Product Release
* Customer Onboarding
* Service Reliability Incident (refreshed)
* Feature Swimlane (basically empty)

The existing template selector scrolls on demand, but to give a slightly better experience on a typical MacBook Pro, I reduced the required margin to avoid scrolling in this case:

image

On smaller screens, scrolling still works:

image

And @calebroseland's updated RHS works beautifully with the new templates (though it seems like `num_actions` is hard-coded at the moment to `0`).

image

Telemetry is updated to track clicks on the template icons in the backstage, as well as the `Use template` option in the new RHS experience.

#### Ticket Link
None.

#### Checklist
- [x] Telemetry updated
- [x] Unit tests updated

### DIFF
--- a/server/sqlstore/playbook.go
+++ b/server/sqlstore/playbook.go
@@ -105,6 +105,16 @@ func NewPlaybookStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 			"DeleteAt",
 			"NumStages",
 			"NumSteps",
+			`(
+				CASE WHEN InviteUsersEnabled THEN 1 ELSE 0 END +
+				CASE WHEN DefaultCommanderEnabled THEN 1 ELSE 0 END +
+				CASE WHEN AnnouncementChannelEnabled THEN 1 ELSE 0 END +
+				CASE WHEN WebhookOnCreationEnabled THEN 1 ELSE 0 END +
+				CASE WHEN MessageOnJoinEnabled THEN 1 ELSE 0 END +
+				CASE WHEN WebhookOnStatusUpdateEnabled THEN 1 ELSE 0 END +
+				CASE WHEN SignalAnyKeywordsEnabled THEN 1 ELSE 0 END +
+				CASE WHEN CategorizeChannelEnabled THEN 1 ELSE 0 END
+			) AS NumActions`,
 			"BroadcastChannelID",
 			"COALESCE(ReminderMessageTemplate, '') ReminderMessageTemplate",
 			"ReminderTimerDefaultSeconds",
@@ -290,7 +300,8 @@ func (p *playbookStore) GetPlaybooks() ([]app.Playbook, error) {
 				CASE WHEN p.WebhookOnCreationEnabled THEN 1 ELSE 0 END +
 				CASE WHEN p.MessageOnJoinEnabled THEN 1 ELSE 0 END +
 				CASE WHEN p.WebhookOnStatusUpdateEnabled THEN 1 ELSE 0 END +
-				CASE WHEN p.SignalAnyKeywordsEnabled THEN 1 ELSE 0 END
+				CASE WHEN p.SignalAnyKeywordsEnabled THEN 1 ELSE 0 END +
+				CASE WHEN p.CategorizeChannelEnabled THEN 1 ELSE 0 END
 			) AS NumActions`,
 		).
 		From("IR_Playbook AS p").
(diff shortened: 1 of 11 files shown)


webapp/src/components/backstage/template_selector.tsx
```diff
@@ -17,61 +19,285 @@ export interface PresetTemplate {
...
+        icon: (
+            <TemplateIcon
+                path={mdiRocketLaunchOutline}
+                size={2.5}
+            />
+        ),
+        template: {
+            ...emptyPlaybook(),
+            title: 'Product Release',
+            num_stages: 4,
```
calebroseland, line 43: @lieut-data I think `num_actions` may need to be set manually alongside `num_stages`

lieut-data: Ah, of course! Thank you :)

crspeller (approved):

lieut-data:
@calebroseland, `num_actions` calculations updated (including to count a recently added action). I feel like there's some duplication in the backstage around this property that we could iterate on in the future -- maybe, for example, we just always send down all the data for playbooks all the time, vs. trying to compute fields like `num_stages`, `num_actions`, etc.

calebroseland:
> @calebroseland, `num_actions` calculations updated (including to count a recently added action). I feel like there's some duplication in the backstage around this property that we could iterate on in the future -- maybe, for example, we just always send down all the data for playbooks all the time, vs. trying to compute fields like `num_stages`, `num_actions`, etc.

100% agree with that—and thanks for adding updating those calculations.

server/sqlstore/playbook.go
```diff
@@ -105,6 +105,16 @@ func NewPlaybookStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
...
+			`(
+				CASE WHEN InviteUsersEnabled THEN 1 ELSE 0 END +
+				CASE WHEN DefaultCommanderEnabled THEN 1 ELSE 0 END +
+				CASE WHEN AnnouncementChannelEnabled THEN 1 ELSE 0 END +
+				CASE WHEN WebhookOnCreationEnabled THEN 1 ELSE 0 END +
+				CASE WHEN MessageOnJoinEnabled THEN 1 ELSE 0 END +
+				CASE WHEN WebhookOnStatusUpdateEnabled THEN 1 ELSE 0 END +
+				CASE WHEN SignalAnyKeywordsEnabled THEN 1 ELSE 0 END +
+				CASE WHEN CategorizeChannelEnabled THEN 1 ELSE 0 END
+			) AS NumActions`,
```
calebroseland, line 117: `ExportChannelOnArchiveEnabled`
```suggestion
			`(
				CASE WHEN InviteUsersEnabled THEN 1 ELSE 0 END +
				CASE WHEN DefaultCommanderEnabled THEN 1 ELSE 0 END +
				CASE WHEN AnnouncementChannelEnabled THEN 1 ELSE 0 END +
				CASE WHEN WebhookOnCreationEnabled THEN 1 ELSE 0 END +
				CASE WHEN MessageOnJoinEnabled THEN 1 ELSE 0 END +
				CASE WHEN WebhookOnStatusUpdateEnabled THEN 1 ELSE 0 END +
				CASE WHEN SignalAnyKeywordsEnabled THEN 1 ELSE 0 END +
				CASE WHEN CategorizeChannelEnabled THEN 1 ELSE 0 END + 
				CASE WHEN ExportChannelOnArchiveEnabled THEN 1 ELSE 0 END
			) AS NumActions`,
```

server/sqlstore/playbook.go
```diff
@@ -290,7 +300,8 @@ func (p *playbookStore) GetPlaybooks() ([]app.Playbook, error) {
 				CASE WHEN p.WebhookOnCreationEnabled THEN 1 ELSE 0 END +
 				CASE WHEN p.MessageOnJoinEnabled THEN 1 ELSE 0 END +
 				CASE WHEN p.WebhookOnStatusUpdateEnabled THEN 1 ELSE 0 END +
-				CASE WHEN p.SignalAnyKeywordsEnabled THEN 1 ELSE 0 END
+				CASE WHEN p.SignalAnyKeywordsEnabled THEN 1 ELSE 0 END +
+				CASE WHEN p.CategorizeChannelEnabled THEN 1 ELSE 0 END
```
calebroseland, line 304: `ExportChannelOnArchiveEnabled`
```suggestion
				CASE WHEN p.CategorizeChannelEnabled THEN 1 ELSE 0 END + 
				CASE WHEN p.ExportChannelOnArchiveEnabled THEN 1 ELSE 0 END
```

server/sqlstore/playbook.go
```diff
@@ -352,7 +363,8 @@ func (p *playbookStore) GetPlaybooksForTeam(requesterInfo app.RequesterInfo, tea
...
```
calebroseland, line 367: `ExportChannelOnArchiveEnabled`
```suggestion
				CASE WHEN p.CategorizeChannelEnabled THEN 1 ELSE 0 END + 
				CASE WHEN ExportChannelOnArchiveEnabled THEN 1 ELSE 0 END
```

calebroseland (changes requested):
I think the action `ExportChannelOnArchiveEnabled` that was also recently added, otherwise LGTM!

lieut-data:
Excellent! Thank you, @calebroseland :)

calebroseland (approved):

PR #657, opened 2021-07-29
[MM-37314] Remove channel from prev category

iomodo:
#### Summary
Remove channel from previous category and fix the concurrency bug

#### Ticket Link

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -13,7 +13,6 @@ import (
 	"github.com/pkg/errors"
 	stripmd "github.com/writeas/go-strip-markdown"
 
-	"github.com/mattermost/mattermost-plugin-api/cluster"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/bot"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/config"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/timeutils"
@@ -1570,13 +1569,6 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
 // createOrUpdatePlaybookRunSidebarCategory creates or updates a "Playbook Runs" sidebar category if
 // it does not already exist and adds the channel within the sidebar category
 func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID string) error {
-	m, err := cluster.NewMutex(s.api, "playbook_run_categories")
-	if err != nil {
-		return err
-	}
-	m.Lock()
-	defer m.Unlock()
-
 	sidebar, err := s.pluginAPI.Channel.GetSidebarCategories(userID, teamID)
 	if err != nil {
 		return err
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -13,10 +13,12 @@ import (
 	"github.com/pkg/errors"
 	stripmd "github.com/writeas/go-strip-markdown"
 
+	"github.com/mattermost/mattermost-plugin-api/cluster"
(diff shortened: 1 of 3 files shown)


server/app/playbook_run_service.go
```diff
@@ -1564,16 +1570,26 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
 // createOrUpdatePlaybookRunSidebarCategory creates or updates a "Playbook Runs" sidebar category if
 // it does not already exist and adds the channel within the sidebar category
 func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID string) error {
+	m, err := cluster.NewMutex(s.api, "playbook_run_categories")
```
iomodo, line 1573: wouldn't it be better to have `s.pluginAPI.Cluster.NewMutex("key")` instead of this?

server/app/playbook_run_service.go
```diff
@@ -1564,16 +1570,26 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
...
 		return err
 	}
 
 	var categoryID string
 	for _, category := range sidebar.Categories {
-		if strings.ToLower(category.DisplayName) == "playbook runs" {
+		if category.DisplayName == playbookRunsCategoryName {
 			categoryID = category.Id
-			category.Channels = append(category.Channels, channelID)
+			if !sliceContains(category.Channels, channelID) {
```
agarciamontoro, line 1589: Good safeguard here!

agarciamontoro (approved):
Looks good! Great investigation work :)

server/app/playbook_run_service.go
agarciamontoro: +1

lieut-data: I think a `Cluster.*` could increase discoverability, but on the flip side it encourages baking in "complex" logic into the plugin API, and I personally found it elegant to isolate the `cluster` package away from the rest of the API.

server/app/playbook_run_service.go
```diff
@@ -1564,16 +1570,26 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
 // createOrUpdatePlaybookRunSidebarCategory creates or updates a "Playbook Runs" sidebar category if
 // it does not already exist and adds the channel within the sidebar category
 func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID string) error {
+	m, err := cluster.NewMutex(s.api, "playbook_run_categories")
+	if err != nil {
+		return err
+	}
+	m.Lock()
```
lieut-data, line 1577: @iomodo and I discussed offline, and I don't think we have conclusive evidence of a race condition here to justify use of the mutex yet.

server/app/playbook_run_service.go
```diff
@@ -1594,6 +1610,19 @@ func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID
 		return nil
 	}
 
+	// remove channel from previous category
```
lieut-data, line 1613: This /feels/ like a complication that we might want to consider solving for users in the pluginapi at some point. @hmhealey, can you also confirm the intended use of the category APIs in this case?

lieut-data (changes requested):

lieut-data (approved):

server/app/playbook_run_service.go
hmhealey: I thought we did that in the DB layer, but maybe that was never tested thoroughly. I'm pretty sure the web app makes a request to update both categories at once.

PR #656, opened 2021-07-28
MM-37325: RHS update timer

agarciamontoro:
#### Summary

This PR adds the status update timer and the dropdown menu in the on-hover buttons. There is also some style polishing.

#### Videos

General walkthrough: a playbook with no default reminder is created (hence, the Post update button occupies the whole width at first). Then, an update is made with no reminder for the next one, so the button is in tertiary mode and the copy reads "Last update in...". Then an update is made with a reminder for the next one, so the button is in primary mode and the copy reads "Update due in...". Finally, the action of the three-dots button is demoed.

https://user-images.githubusercontent.com/3924815/127398181-4db7e087-3a11-4814-8328-bc0ed81beae3.mp4

Overdue updates and behaviour of the interactive post button Dismiss:

https://user-images.githubusercontent.com/3924815/127398247-99c105a2-1cb6-4d0d-b22c-c5c602dd4ffb.mp4

Overdue updates and behaviour of the interactive post button Update status:

https://user-images.githubusercontent.com/3924815/127398253-d7f9b072-31b6-499a-8438-e8cdbb5ce4ff.mp4

Quick view of the dark theme (I think the red here is a bit weird, but still readable):

https://user-images.githubusercontent.com/3924815/127398271-e4a6f090-b6a3-4354-8c65-0807c9070771.mp4

##### New run

| New run, no default reminder 	| New run, default reminder 	|
|---	|---	|
| image 	| image 	|

##### Scheduled
| Days 	| One hour 	| Minutes 	| 
|---	|---	|---	|
| image 	| image 	| image 	|

One minute 	| This minute 	| Now 	|
|---	|---	|---	|
 image 	| image 	| image 	|

##### Overdue
| This minute 	| One minute 	| Minutes 	| Hours 	|
|---	|---	|---	|---	|
| image 	| image 	| image 	| image 	|

##### No update scheduled

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37325

#### Checklist

- [ ] Telemetry updated <- We did not talk about this in the meetings, so I'm just skipping and adding next week if needed
- [ ] ~~Gated by experimental feature flag~~
- [ ] ~~Unit tests updated~~

PR #658, opened 2021-07-29
MM-37326: Inline editing

agarciamontoro:
#### Summary
This PR makes both the title and the description editable inline. It also removes the Description field from the interactive dialog, making the inline edition the only way to change the description.

https://user-images.githubusercontent.com/3924815/127485888-f2872e2b-5982-4071-9bef-8278045123ce.mp4

When editing the title, hitting `Enter` closes the edition and saves the new title. For the description, I chose `Ctrl+Enter` for that. Both the title and description are automatically saved whenever the user clicks outside of the input area.

As you can see in the video and in the code, changing the run name is actually a channel name change. It was the most straightforward way to do it given the time constraints.

This branch is based off of #656, I'll change it to master as soon as that one's merged.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37326

#### Checklist

- [ ] ~~Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -918,7 +918,6 @@ func (h *PlaybookRunHandler) updateDescription(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	playbookRun.Description = requestBody.Description
 	if err := h.playbookRunService.UpdateDescription(playbookRunID, requestBody.Description); err != nil {
 		h.HandleErrorWithCode(w, http.StatusInternalServerError, "unable to update description", err)
 		return
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -665,12 +665,6 @@ func (h *PlaybookRunHandler) status(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	options.Description = strings.TrimSpace(options.Description)
-	if options.Description == "" {
-		h.HandleErrorWithCode(w, http.StatusBadRequest, "description must not be empty", errors.New("description field empty"))
-		return
-	}
-
 	options.Message = strings.TrimSpace(options.Message)
 	if options.Message == "" {
 		h.HandleErrorWithCode(w, http.StatusBadRequest, "message must not be empty", errors.New("message field empty"))
@@ -735,15 +729,6 @@ func (h *PlaybookRunHandler) updateStatusDialog(w http.ResponseWriter, r *http.R
 	if options.Message == "" {
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte(fmt.Sprintf(`{"errors": {"%s":"This field is required."}}`, app.DialogFieldMessageKey)))
-		return
-	}
-
-	if description, ok := request.Submission[app.DialogFieldDescriptionKey]; ok {
-		options.Description = strings.TrimSpace(description.(string))
-	}
-	if options.Description == "" {
-		w.WriteHeader(http.StatusOK)
-		_, _ = w.Write([]byte(fmt.Sprintf(`{"errors": {"%s":"This field is required."}}`, app.DialogFieldDescriptionKey)))
 		return
 	}
 
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -71,7 +71,7 @@ func NewPlaybookRunHandler(router *mux.Router, playbookRunService app.PlaybookRu
 	playbookRunRouterAuthorized.HandleFunc("/no-retrospective-button", handler.noRetrospectiveButton).Methods(http.MethodPost)
 	playbookRunRouterAuthorized.HandleFunc("/timeline/{eventID:[A-Za-z0-9]+}", handler.removeTimelineEvent).Methods(http.MethodDelete)
 	playbookRunRouterAuthorized.HandleFunc("/check-and-send-message-on-join/{channel_id:[A-Za-z0-9]+}", handler.checkAndSendMessageOnJoin).Methods(http.MethodGet)
-	playbookRunRouterAuthorized.HandleFunc("/update-description", handler.updateDescription).Methods(http.MethodPatch)
+	playbookRunRouterAuthorized.HandleFunc("/update-description", handler.updateDescription).Methods(http.MethodPut)
 
 	channelRouter := playbookRunsRouter.PathPrefix("/channel").Subrouter()
 	channelRouter.HandleFunc("/{channel_id:[A-Za-z0-9]+}", handler.getPlaybookRunByChannel).Methods(http.MethodGet)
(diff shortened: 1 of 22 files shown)


agarciamontoro:
By the way, the title ellipsizes when it overflows, taking also into account the presence of the top section buttons.


https://user-images.githubusercontent.com/3924815/127487064-2e569fe5-c8fc-4fe5-9107-8eecb5043f6f.mp4

abhijit-singh (commented):
@agarciamontoro Looks awesome! 

A couple of doubts: 
1. What happens if you click on a really long title to edit it, or add more characters beyond the available space while editing?
2. Can you also please add a recording/screenshot of the collapsing functionality with the huge description added? Thanks!

agarciamontoro:
I'm starting to like answering to questions with videos, it's really easy :joy: @abhijit-singh, there you go:

https://user-images.githubusercontent.com/3924815/127502345-8ef04e0d-02b8-41e5-ad73-49b3bd486b8b.mp4

When you click either in the title or the description, the caret is always at the end. But the title does not "scroll" horizontally to see it, I had to hit the `End` key for it to show. That's a bit annoying.

server/api/playbook_runs.go
```diff
@@ -904,6 +890,43 @@ func (h *PlaybookRunHandler) checkAndSendMessageOnJoin(w http.ResponseWriter, r
...
+	playbookRunID := mux.Vars(r)["id"]
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	playbookRun, err := h.playbookRunService.GetPlaybookRun(playbookRunID)
+	if err != nil {
+		h.HandleError(w, err)
+		return
+	}
+
+	requestBody := struct {
```
lieut-data, line 903: nit: I /usually/ see this written as:
```
var requestBody struct {
    ...
}
```

The trailing `}{}` threw me off for a bit? Totally nit picking here...

server/api/playbook_runs.go
```diff
@@ -904,6 +890,43 @@ func (h *PlaybookRunHandler) checkAndSendMessageOnJoin(w http.ResponseWriter, r
...
+		Description string `json:"description"`
+	}{}
+
+	if err2 := json.NewDecoder(r.Body).Decode(&requestBody); err2 != nil {
```
lieut-data, line 907: Why `err2`?

server/api/playbook_runs.go
```diff
@@ -904,6 +890,43 @@ func (h *PlaybookRunHandler) checkAndSendMessageOnJoin(w http.ResponseWriter, r
...
+	if err = app.EditPlaybookRun(userID, playbookRun.ChannelID, h.pluginAPI); err != nil {
+		if errors.Is(err, app.ErrNoPermissions) {
+			ReturnJSON(w, nil, http.StatusForbidden)
+			return
+		}
+		h.HandleErrorWithCode(w, http.StatusInternalServerError, "error getting permissions", err)
+		return
+	}
+
+	playbookRun.Description = requestBody.Description
```
lieut-data, line 921: Is this line needed?

server/app/playbook_run_service.go
```diff
@@ -1642,6 +1641,22 @@ func (s *PlaybookRunServiceImpl) CheckAndSendMessageOnJoin(userID, givenPlaybook
...
 }
 
+func (s *PlaybookRunServiceImpl) UpdateDescription(playbookRunID, description string) error {
+	playbookRun, err := s.store.GetPlaybookRun(playbookRunID)
+	if err != nil {
+		return errors.Wrap(err, "unable to get playbook run")
+	}
+
+	playbookRun.Description = description
+	if err = s.store.UpdatePlaybookRun(playbookRun); err != nil {
```
lieut-data, line 1651: Out of scope for this PR, but I'd love to see solutions in the SQLStore that avoid racy updates. Here, if there were concurrent playbook run updates, we risk losing one.

webapp/src/client.ts
```diff
@@ -407,6 +407,20 @@ export const promptForFeedback = async () => {
...
+export const changeChannelName = async (channelId: string, newName: string) => {
+    await doFetchWithoutResponse(`/api/v4/channels/${channelId}/patch`, {
+        method: 'PUT',
+        body: JSON.stringify({display_name: newName}),
+    });
+};
+
+export const updatePlaybookRunDescription = async (playbookRunId: string, newDescription: string) => {
+    await doFetchWithoutResponse(`${apiUrl}/runs/${playbookRunId}/update-description`, {
+        method: 'PATCH',
```
lieut-data, line 419: nit: `PATCH` on an endpoint that only updates a single field seems weird. I'd have expected `PATCH` against the `/runs/:run_id` that intelligently applies just the description, or a `PUT` here.

webapp/src/hooks/general.ts
```diff
@@ -52,12 +50,20 @@ import {
...
  * Hook that calls handler when targetKey is pressed.
  */
-export function useKeyPress(targetKey: string, handler: () => void) {
+export function useKeyPress(targetKey: string | ((e: KeyboardEvent) => boolean), handler: () => void) {
+    const predicate: (e: KeyboardEvent) => boolean = useMemo(() => {
+        if (typeof targetKey === 'string') {
+            return (e: KeyboardEvent) => e.key === targetKey;
+        }
+
+        return targetKey;
```
lieut-data, line 59: This seems like it will be always true? Do we need to invoke it here?

lieut-data (changes requested):
A handful of nits, and just one point of confusion around the `useKeyPress` hook.

server/api/playbook_runs.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/658/commits/2c70fe98e2d465cfba9e41c35e055f7575369f07

server/api/playbook_runs.go
agarciamontoro: I just changed it because the linter was complaining. I'd like to remove that rule completely, actually, I think it only adds noise.

server/api/playbook_runs.go
agarciamontoro: Of course not, this is from an old implementation that I forgot to remove. Good catch! Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/658/commits/f988bfefeb38d8abb85b26fd9ea3634547ced819

webapp/src/client.ts
agarciamontoro: You're right! Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/658/commits/af00574e8fe7ec5057e18ca560c15a9b6a6865ac

webapp/src/hooks/general.ts
agarciamontoro: It returns the function itself. `targetKey` can be either a string or a function that accepts a keyboard event and return a boolean. In this line, we have already handled the case where it's a string, so it is now a function. Maybe I should use another name for the argument?

server/app/playbook_run_service.go
agarciamontoro: Hmmm that's a good point. Although if there are two concurrent playbook run updates, what do you do? You have to lose one there either way, right?

abhijit-singh (approved):
Thanks @agarciamontoro! Looks good!

server/app/playbook_run_service.go
lieut-data: If the SQL updates are discrete, then one row update will get rolled back and the application can choose to retry (from scratch) or give up.

webapp/src/hooks/general.ts
lieut-data: Oh, that makes sense. I missed that we're returning a function implementation above -- read too quickly I think. 0/5 on renames :)

lieut-data (approved):

cpoile (approved):
Awesome, that was a good solution!

PR #659, opened 2021-07-29
MM-37019: RHS Home Polish

calebroseland:
#### Summary
- Run Detail Swoop/Swoosh
- Refactor-flatten nested CSS
- Fix '..or start with a template'  text color

CleanShot 2021-07-29 at 09 35 59@2x
CleanShot 2021-07-29 at 09 33 39@2x

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37019

#### Checklist

~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


webapp/src/components/rhs/rhs_home.tsx
```diff
@@ -68,67 +95,46 @@ const RunDetail = styled.div<{
             rgba(var(--center-channel-bg-rgb), 0.25) 100%
         ),
         rgba(var(${({exists}) => (exists ? '--button-bg-rgb' : '--center-channel-color-rgb')}), 0.08);
+    mask-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="100%" height="calc(100% - 15px)" viewBox="0 0 400 137" preserveAspectRatio="none"><path d="M0 0H400V122.629C400 122.629 312 137 200 137C101.5 137 0 122.629 0 122.629V0Z"/></svg>');
```
agarciamontoro, line 98: :sunglasses:

webapp/src/components/rhs/rhs_home.tsx
```diff
@@ -68,67 +95,46 @@ const RunDetail = styled.div<{
...
+    margin-right: 2rem;
+    ${({exists}) => (exists ? css`
+        font-size: 14px;
+        line-height: 20px;
+        color: var(--mention-color);
+    ` : css`
+        color: '#6F6F73';
+        font-size: 16px;
+        line-height: 24px;
+    `)}
```
agarciamontoro, line 118: Nit: I always prefer to read these as
```js
font-size: ${({exists}) => exists ? '14px' : '16px'
```
But 0/5 on this.

webapp/src/components/rhs/rhs_home.tsx
agarciamontoro: What do you think about actually defining a constant with the string inside the url (or the whole value) to know what it is? It's impossible to know what the mask is just reading the code.

lieut-data (approved):

agarciamontoro (approved):
A couple of non-blocking nits, but it looks awesome :)

webapp/src/components/rhs/rhs_home.tsx
calebroseland: In this case it's single-use, but I agree it would be better to use an externally.

webapp/src/components/rhs/rhs_home.tsx
calebroseland: 💯 for single-properties. I was just unsure of how that would scale with multiple properties

calebroseland: I'll leave for now but will revisit

PR #663, opened 2021-07-30
MM-37533 - Empty checklists after filtering should omit the body

cpoile:
#### Summary
- Return null if no visible tasks

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-37533

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/rhs/rhs_checklists.tsx
+++ b/webapp/src/components/rhs/rhs_checklists.tsx
@@ -127,10 +127,8 @@ const RHSChecklists = (props: Props) => {
                     collapsed={Boolean(checklistsState[checklistIndex])}
                     setCollapsed={(newState) => dispatch(setChecklistCollapsedState(channelId, checklistIndex, newState))}
                 >
-                    <ChecklistContainer
-                        className='checklist'
-                        empty={noVisibleTasks(checklist, checklistItemsFilter, myUser.id)}
-                    >
+                    {visibleTasks(checklist, checklistItemsFilter, myUser.id) &&
+                    <ChecklistContainer className='checklist'>
                         <DragDropContext
                             onDragEnd={(result: DropResult) => {
                                 if (!result.destination) {
@@ -211,6 +209,7 @@ const RHSChecklists = (props: Props) => {
                             </Droppable>
                         </DragDropContext>
                     </ChecklistContainer>
+                    }
                 </CollapsibleChecklist>
             ))}
             {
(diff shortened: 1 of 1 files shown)


webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -127,7 +127,10 @@ const RHSChecklists = (props: Props) => {
                     collapsed={Boolean(checklistsState[checklistIndex])}
                     setCollapsed={(newState) => dispatch(setChecklistCollapsedState(channelId, checklistIndex, newState))}
                 >
-                    <ChecklistContainer className='checklist'>
+                    <ChecklistContainer
+                        className='checklist'
+                        empty={noVisibleTasks(checklist, checklistItemsFilter, myUser.id)}
```
cpoile, line 132: This, vs returning a null. Open to doing it the other way.

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -127,7 +127,10 @@ const RHSChecklists = (props: Props) => {
                     collapsed={Boolean(checklistsState[checklistIndex])}
                     setCollapsed={(newState) => dispatch(setChecklistCollapsedState(channelId, checklistIndex, newState))}
                 >
-                    <ChecklistContainer className='checklist'>
+                    <ChecklistContainer
```
lieut-data, line 130: Would it make sense to omit rendering this altogether?

lieut-data (approved):

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -238,9 +241,13 @@ const MainTitle = styled.div`
...
 `;
 
-const ChecklistContainer = styled.div`
+const ChecklistContainer = styled.div<{empty: boolean}>`
     background-color: var(--center-channel-bg);
     padding: 16px 12px;
+
+    ${(props) => props.empty && css`
+        padding: 0px;
+    `}
```
agarciamontoro, line 250: Super nit: I prefer to embed this into the property itself if possible. Something like:

```ts
padding: ${(props.empty) => props.empty ? '0px' : '16px 12px';
```

Or even more succintly:

```ts
padding: ${({empty}) => empty ? '0' : '16px 12px';
```

agarciamontoro (approved):
Elegant!

webapp/src/components/rhs/rhs_checklists.tsx
cpoile: Decided to return null, so this is gone. Good advice though, will do in the future. :)

webapp/src/components/rhs/rhs_checklists.tsx
cpoile: Yah, I was thinking that. Will do that.

PR #665, opened 2021-08-02
MM-37538: Reset reminder when updating status to Archived

agarciamontoro:
#### Summary
When updating the status to Archived, reset the reminder timer so the RHS does not show a status update still scheduled in the future. Instead, it shows when the last status update was posted and styles the Post update button as tertiary.

https://user-images.githubusercontent.com/3924815/127842007-11cdc79d-3bcb-49ef-a09b-e02ef1dfe396.mp4

@abhijit-singh, do you think this makes sense from a UX viewpoint?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37538

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -852,22 +852,28 @@ func (s *PlaybookRunServiceImpl) UpdateStatus(playbookRunID, userID string, opti
 		}()
 	}
 
-	if previousStatus != StatusArchived && options.Status == StatusArchived && playbookRunToModify.ExportChannelOnArchiveEnabled {
-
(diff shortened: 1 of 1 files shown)


abhijit-singh (approved):
Thanks, @agarciamontoro! This makes sense with the current status options, but once we update them to just `In progress` and `Finished`, clicking on the `Finish run` button should trigger the same behaviour.

iomodo (approved):
LGTM!

lieut-data (approved):

PR #668, opened 2021-08-02
[MM-35248] Make backstage current team agnostic

iomodo:
#### Summary
This PR makes backstage "current team" agnostic.
Since there is a lot of places where we use `currentTeam` I might have missed couple. This PR should be tested thoroughly before merging. 

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-35248

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated


iomodo:
Just an FYI, already tested thoroughly, I think I did not miss anything. When/if you'll be testing pay attention that after refresh all the `NeedsTeam` context is removed from the redux. So there are couple of places where it gets tricky. It's ready to be reviewed. 
cc @lieut-data @cpoile 
P.S. sorry for the long PR :)

webapp/src/components/backstage/backstage.tsx
```diff
@@ -139,7 +149,7 @@ const Backstage = () => {
                     {npsAvailable &&
                         <BackstageTitlebarItem
                             onClick={promptForFeedback}
-                            to={`/${currentTeam.name}/messages/@surveybot`}
+                            to={`/${teams[0].name}/messages/@surveybot`}
```
cpoile, line 152: Would it make sense to send the user to the current team? This might avoid a surprising switch when they might not expect it.

webapp/src/components/backstage/css_utils.tsx
```diff
@@ -0,0 +1,49 @@
+import {
+    changeOpacity,
+} from 'mattermost-redux/utils/theme_utils';
+import cssVars from 'css-vars-ponyfill';
+
+export function applyTheme(theme: any) {
```
cpoile, line 6: I wonder if there's a better way to do this? No worries if there isn't. I just imagine this will break as soon as we change how themes are done.

webapp/src/components/backstage/playbook_edit_wrapper.tsx
```diff
@@ -0,0 +1,32 @@
...
+import {pluginUrl} from 'src/browser_routing';
+
+import PlaybookEdit from './playbook_edit';
+
+interface URLParams {
+    playbookId?: string;
+    tabId?: string;
+}
+
+const PlaybookEditWrapper = () => {
```
cpoile, line 13: I'm not sure why we need this wrapper?

webapp/src/components/backstage/playbook_runs/playbook_run_list/playbook_run_list.tsx
```diff
@@ -293,12 +292,12 @@ const BackstagePlaybookRunList = () => {
...
 
     const goToMattermost = () => {
-        navigateToUrl(`/${currentTeam.name}`);
+        navigateToUrl('');
     };
 
     const newPlaybookRun = () => {
         goToMattermost();
-        dispatch(startPlaybookRun());
+        dispatch(startPlaybookRun(teams[0].id)); //TODO add team selector ?
```
cpoile, line 300: Or take a team parameter? So that `newPlaybookRun(teamId)` can use the current team (wherever the user currently is).

webapp/src/components/backstage/playbooks/playbook_run_list/playbook_run_list.tsx
```diff
@@ -151,7 +141,7 @@ const PlaybookRunList = (props: Props) => {
     }
 
     async function fetchOwners() {
-        const owners = await fetchOwnersInTeam(currentTeam.id);
+        const owners = await fetchOwnersInTeam(props.playbook?.team_id || ''); //TODO verify
```
cpoile, line 144: Verify?

cpoile (changes requested):
Wow, lots of changes, nice work @iomodo!

I mostly have questions, some may prompt a revision (or not). :)

To help test this, is there a list of all the affected things we should check? Or is it almost everything?

server/app/playbook_run_service_test.go
```diff
@@ -331,7 +330,7 @@ func TestCreatePlaybookRun(t *testing.T) {
 				"http://example.com/ad-1/channels/channel-name",
 				payload.ChannelURL)
 			require.Equal(t,
-				"http://example.com/ad-1/com.mattermost.plugin-incident-management/runs/"+createdPlaybookRun.ID,
+				"http://example.com/plug/com.mattermost.plugin-incident-management/runs/"+createdPlaybookRun.ID,
```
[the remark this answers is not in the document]
iomodo: Yes, it is the new route

webapp/src/actions.ts
```diff
@@ -249,3 +249,8 @@ export const setChecklistItemsFilter = (channelId: string, nextState: ChecklistI
     channelId,
     nextState,
 });
+
+export const setCurrentTeamId = (teamId: string): SetCurrentTeamId => ({
+    type: TeamTypes.SELECT_TEAM,
```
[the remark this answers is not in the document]
iomodo: Good catch. It's redundant. Forgot to remove after the refactoring. 
`<TextBox> needs a `currentTeamId` for the slash command autocomplete and there is none in the store, so I'm setting the `id` manually. I found the `selectTeam` action in the redux repo that does the same and am using that here https://github.com/mattermost/mattermost-plugin-playbooks/blob/9115852d90a329e7e317beb5850af0e12a16d2b9/webapp/src/components/backstage/playbook_edit.tsx#L190

webapp/src/components/backstage/backstage.tsx
iomodo: But which is the current team? There is no such thing in the backstage anymore.

webapp/src/components/backstage/css_utils.tsx
iomodo: I've talked about this with @crspeller and this solution is fine for now. We will refactor it after merging the global header if needed.

webapp/src/components/backstage/playbook_runs/playbook_run_list/playbook_run_list.tsx
iomodo: Again the thing is that we do not have the current team in this context

webapp/src/components/backstage/playbooks/playbook_run_list/playbook_run_list.tsx
iomodo: Forgot to remove after testing.

iomodo:
>> To help test this, is there a list of all the affected things we should check? Or is it almost everything?
Unfortunately there is no concrete list. Almost everything has changed. We'll probably need a bug bash to test this in more detail. I've tested everything what I could think of, but I may have missed something

webapp/src/components/backstage/backstage.tsx
lieut-data: There is the concept of the "previous team" in webapp -- it's used to send the user to the last team they viewed when loading the webapp afresh without an explicitly team in the URL.

However it's awkwardly in local storage and not Redux, and it's something that's shared across all tabs. Could we model this ourselves to remember the last team, keying off the existing `TeamTypes.SELECT_TEAM` (and `UserTypes.LOGOUT_SUCCESS`) actions?

iomodo: I'm guessing it's in the local storage because of the refresh consistency. I don't think we should be storing this in memory, the `currentTeam` is already there if do not refresh it will stay there.

How about getting the `currentTeam` and if it's undefined then use to the `team[0]`?

I'd suggest deferring the `previous team` (if we will use that) to the next PR.

lieut-data: :+1: for subsequent PR :)

server/app/playbook_run_service.go
```diff
@@ -2111,13 +2091,8 @@ func (s *PlaybookRunServiceImpl) PublishRetrospective(playbookRunID, text, publi
...
 
-	team, err := s.pluginAPI.Team.Get(playbookRunToPublish.TeamID)
-	if err != nil {
-		return err
-	}
-
-	retrospectiveURL := fmt.Sprintf("/%s/%s/runs/%s/retrospective",
-		team.Name,
+	retrospectiveURL := getRunRetrospectiveURL(
+		"",
```
lieut-data, line 2095: Why empty string?

server/app/playbook_run_service.go
```diff
@@ -1859,7 +1839,7 @@ func (s *PlaybookRunServiceImpl) newPlaybookRunDialog(teamID, ownerID, postID, c
 	}
 	newPlaybookMarkdown := ""
 	if siteURL != "" && !isMobileApp {
-		url := fmt.Sprintf("%s/%s/%s/playbooks/new", siteURL, team.Name, s.configService.GetManifest().Id)
+		url := getPlaybooksNewURL(siteURL, s.configService.GetManifest().Id)
```
lieut-data, line 1842: Nice!

server/app/playbook_service.go
```diff
@@ -198,7 +192,7 @@ func (s *playbookService) getPlaybookSuggestionsSlackAttachment(playbooks []*Cac
 			Style: "primary",
 		}
 
-		url := fmt.Sprintf("%s/%s/edit/automation", playbooksURL, playbooks[0].ID) // automation tab
+		url := fmt.Sprintf("%s/%s/edit/automation?teamId=%s", playbooksURL, playbooks[0].ID, playbooks[0].TeamID) // automation tab
```
lieut-data, line 195: Is it possible for the client to resolve the team on demand? What happens, for example, if this URL is crafted "maliciously", specifying the "wrong" team id?

server/app/urls.go
```diff
@@ -0,0 +1,43 @@
...
+import "fmt"
+
+const pluginRoute = "plug"
+
+func getRunsURL(siteURL string, manifestID string) string {
+	return fmt.Sprintf("%s/%s/%s/runs",
+		siteURL,
+		pluginRoute,
+		manifestID,
+	)
```
lieut-data, line 12: nit: Might we consider formatting as:
```suggestion
	return fmt.Sprintf(
              "%s/%s/%s/runs",
		siteURL,
		pluginRoute,
		manifestID,
	)
```

server/command/command.go
```diff
@@ -1111,15 +1111,9 @@ func (r *Runner) actionTestGeneratePlaybooks(params []string) {
...
-	team, errGetTeam := r.pluginAPI.Team.Get(r.args.TeamId)
-	if errGetTeam != nil {
-		r.warnUserAndLogErrorf("unable to retrieve team: %v", err)
-		return
-	}
-
 	msg := "Playbooks successfully created"
 	for i, playbookID := range playbookIds {
-		url := fmt.Sprintf("/%s/com.mattermost.plugin-incident-management/playbooks/%s", team.Name, playbookID)
+		url := fmt.Sprintf("/%s/%s/playbooks/%s", "plug", r.configService.GetManifest().Id, playbookID)
```
lieut-data, line 1116: Can we inline `"plug"` into the format string here (like `playbooks`)?

webapp/src/types/actions.ts
```diff
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import Integrations from 'mattermost-redux/action_types/integrations';
+import Teams from 'mattermost-redux/action_types/teams';
```
lieut-data, line 5: Is this used?

webapp/src/index.tsx
```diff
@@ -111,6 +111,9 @@ export default class Plugin {
             r.registerNeedsTeamRoute('/error', ErrorPage);
```
lieut-data, line 111: Do we still use the team routes?

webapp/src/components/rhs/rhs_post_update.tsx
```diff
@@ -74,7 +78,7 @@ const RHSPostUpdate = (props: Props) => {
                 collapsed={props.collapsed}
                 isNextUpdateScheduled={isNextUpdateScheduled}
                 updatesExist={props.updatesExist}
-                onClick={() => dispatch(updateStatus())}
+                onClick={() => dispatch(updateStatus(team.id))}
```
lieut-data, line 81: Can we just use `props.playbookRun.team_id` directly here?

webapp/src/components/rhs/rhs_footer.tsx
```diff
@@ -37,13 +38,13 @@ const RHSFooter = (props: Props) => {
...
                 primary={false}
-                onClick={() => navigateToUrl(`/${currentTeam.name}/${pluginId}/runs/${playbookRun?.id}`)}
+                onClick={() => navigateToPluginUrl(`/runs/${playbookRun?.id}`)}
             >
                 {'Overview'}
             </StyledFooterButton>
             <SpacedFooterButton
                 primary={true}
-                onClick={() => dispatch(updateStatus())}
+                onClick={() => dispatch(updateStatus(team.id))}
```
lieut-data, line 47: Can we just use `props.playbookRun.team_id || playbookRun?.team_id || ''` here directly?

webapp/src/components/rhs/rhs_footer.tsx
```diff
@@ -25,8 +26,8 @@ interface Props {
 
 const RHSFooter = (props: Props) => {
     const dispatch = useDispatch();
-    const currentTeam = useSelector(getCurrentTeam);
     const playbookRun = useSelector(currentPlaybookRun);
+    const team = useSelector<GlobalState, Team>((state) => getTeam(state, props.playbookRun.team_id || playbookRun?.team_id || ''));
```
lieut-data, line 30: Is there a reason we can't just use `playbookRun?` or `props.playbookRun` exclusively? Wondering why this component has to be "aware" of two entry points.

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -214,7 +219,7 @@ const RHSChecklists = (props: Props) => {
             ))}
             {
                 active &&
-                <FinishButton onClick={() => dispatch(updateStatus('Resolved'))}>
+                <FinishButton onClick={() => dispatch(updateStatus(team.id, 'Resolved'))}>
```
lieut-data, line 222: Can we use `props.playbookRun.team_id` directly?Can we use `props.playbookRun.team_id` directly?

webapp/src/components/rhs/rhs_checklists.tsx
```diff
@@ -126,6 +130,7 @@ const RHSChecklists = (props: Props) => {
                     index={checklistIndex}
                     collapsed={Boolean(checklistsState[checklistIndex])}
                     setCollapsed={(newState) => dispatch(setChecklistCollapsedState(channelId, checklistIndex, newState))}
+                    teamId={team.id}
```
lieut-data, line 133: Can we use `props.playbookRun.team_id` directly?

webapp/src/components/rhs/post_card.tsx
```diff
@@ -98,7 +91,7 @@ const PostCard = (props: Props) => {
     if (!props.post) {
         return (
             <NoRecentUpdates>
-                {'No recent updates. '}<a onClick={() => dispatch(updateStatus())}>{'Click here'}</a>{' to update status.'}
+                {'No recent updates. '}<a onClick={() => dispatch(updateStatus(props.team.id))}>{'Click here'}</a>{' to update status.'}
```
lieut-data, line 94: If we only need the id here, should we consider only passing that vs. the whole `team`?

webapp/src/components/rhs/collapsible_checklist.tsx
```diff
@@ -57,7 +59,7 @@ const CollapsibleChecklist = ({
                         data-testid={'addNewTask'}
                         onClick={(e) => {
                             e.stopPropagation();
-                            dispatch(addNewTask(index));
+                            dispatch(addNewTask(teamId, index));
```
lieut-data, line 62: It seems weird to pass `teamId` to `addNewTask`, but we don't have to pass `playbookRunId` and such. I'm guessing `addNewTask` is a thunk? Should we just have it look up the current team where it needs it?

webapp/src/components/post_menu.tsx
```diff
@@ -59,13 +62,14 @@ export const AttachToPlaybookRunPostMenu = (props: Props) => {
...
     const post = useSelector<GlobalState, Post>((state) => getPost(state, props.postId));
+    const channel = useSelector<GlobalState, Channel>((state) => getChannel(state, post.channel_id));
     if (!post || isSystemMessage(post)) {
         return null;
     }
 
     const handleClick = () => {
         if (allowMessage) {
-            dispatch(addToTimeline(props.postId));
+            dispatch(addToTimeline(channel.team_id, props.postId));
```
lieut-data, line 72: See below, but thinking we can make `addToTimeline` do the work of resolving the `team_id` instead.

webapp/src/components/backstage/playbooks/playbook_run_list/playbook_run_list.tsx
```diff
@@ -79,11 +77,9 @@ const PlaybookRunList = (props: Props) => {
     const [playbookRuns, setPlaybookRuns] = useState<PlaybookRun[] | null>(null);
     const [totalCount, setTotalCount] = useState(0);
     const selectUser = useSelector<GlobalState>((state) => (userId: string) => getUser(state, userId)) as (userId: string) => UserProfile;
-    const currentTeam = useSelector<GlobalState, Team>(getCurrentTeam);
-
     const [fetchParams, setFetchParams] = useState<FetchPlaybookRunsParams>(
         {
-            team_id: currentTeam.id,
+            team_id: props.playbook?.team_id,
```
lieut-data, line 82: Can we find a way to omit `team_id` altogether here? Presumably the server can resolve it itself?

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/retrospective/timeline.tsx
```diff
@@ -47,7 +47,7 @@ interface Props {
 
 const Timeline = (props: Props) => {
     const channelNamesMap = useSelector<GlobalState, ChannelNamesMap>(getChannelsNameMapInCurrentTeam);
-    const team = useSelector<GlobalState, Team>(getCurrentTeam);
+    const team = useSelector<GlobalState, Team>((state) => getTeam(state, props.playbookRun.team_id));
```
lieut-data, line 50: Can we check that `RHSTimelineEventItem` needs the whole team and not just the `team_id`?

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -254,9 +251,15 @@ const PlaybookEdit = (props: Props) => {
                     setFetchingState(FetchingStateType.notFound);
                 }
             }
+            if (props.teamId) {
+                const fetchedChannels = await fetchMyChannels(props.teamId);
```
lieut-data, line 255: It would be awesome to be able to dispatch a plain `FETCH_TEAM_CHANNELS` Redux action and have the webapp "handle" this for us instead of having to maintain our own (duplicate state) and without having to import an action from the webapp. This is a place where a side effects library could really help.

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -189,24 +185,25 @@ const PlaybookNavbar = styled(BackstageNavbar)`
...
 
+    dispatch(selectTeam(props.teamId));
+
     const [playbook, setPlaybook] = useState<DraftPlaybookWithChecklist | PlaybookWithChecklist>({
         ...emptyPlaybook(),
-        team_id: props.currentTeam.id,
+        team_id: props.teamId,
     });
     const [changesMade, setChangesMade] = useState(false);
+    const [channels, setChannels] = useState<Channel[]>([]);
```
lieut-data, line 197: Wondering if we can find a way to use what's in the webapp Redux store instead? (Kicking it as needed?)

lieut-data (changes requested):
Thanks, @iomodo! A handful of comments below, but a high-level summary:
* I'm worried about having to pass `team_id` in the query string to inform the client of something I feel like it should be able to look up on its own.
* The proliferation of `team` into components /feels/ like it could be optimized, if only to just pass around `team_id` in many cases. (Various comments below.)
* The need to resolve our own set of team channels in various cases "smells wrong", and wondering if we can leverage the webapp's existing functionality here more closely?

webapp/src/actions.ts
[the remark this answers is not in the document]
cpoile: I guess the `getCurrentTeamId` selector might not work, if we didn't first come from the channels side?

webapp/src/components/backstage/backstage.tsx
cpoile: 👍

webapp/src/components/backstage/playbook_edit_wrapper.tsx
cpoile: Still wondering about this, @iomodo?

webapp/src/components/backstage/playbook_runs/playbook_run_list/playbook_run_list.tsx
cpoile: I have a feeling this is going to be bad UX. Maybe we should have a "default" team to start runs in, set by the admin in the plugin settings? Otherwise new runs will start popping up in random teams and once you've started a run we don't let you move them. @itao , @abhijit-singh ?  
(Adding this setting would be for another PR of course, not this one)

cpoile:
I didn't mean to re-request a review from Jesse -- don't know why it did that.

cpoile (approved):
Approving to unblock, but I'm still interested in the answers to the remaining questions above. Nice work again @iomodo, thanks!

webapp/src/actions.ts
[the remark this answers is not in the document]
iomodo: Correct. Or if we refresh.

lieut-data (changes requested):
Re-requesting review (without additional comment).

server/app/playbook_run_service.go
iomodo: It's was a relative path, I left a relative path

webapp/src/index.tsx
iomodo: We don't but is still works. Not sure if we want to delete, maybe customers have some links that will get broken? I'm 0/5

webapp/src/components/rhs/rhs_footer.tsx
iomodo: I don't think we are using this component at all. Changing to rely on props

webapp/src/components/rhs/post_card.tsx
iomodo: We need whole `team` for the `PostText` below in this component

webapp/src/components/rhs/collapsible_checklist.tsx
iomodo: `addNewTask` runs a slash command. And we need a `team_id` to do that. Getting the `currentTeam` might not work after refresh. So everywhere where client needs to execute a slash command we are passing the `team_id` directly.

server/app/playbook_service.go
lieut-data: @iomodo, this was one of the cases where I think I was confused about the use of `teamId`. It seems like we're using this here not for "new playbook with no team picked", but an existing playbook for which we should already know the team?

webapp/src/index.tsx
lieut-data: Let's add a TODO (in the code?) to deprecate in v2.0.

webapp/src/components/rhs/collapsible_checklist.tsx
lieut-data: Once we eliminate the use of interactive dialogs for things like this, I imagine it will get a lot simpler.

server/app/playbook_service.go
iomodo: Yes, yes I'm changing this.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/retrospective/timeline.tsx
iomodo: It needs whole `team` for markdown options to format text.

webapp/src/components/backstage/playbook_edit_wrapper.tsx
iomodo: Removed it

webapp/src/components/backstage/channel_selector.tsx
```diff
@@ -15,11 +19,17 @@ export interface Props {
...
 
+const getMyPublicAndPrivateChannels = (state: GlobalState) => getMyChannels(state).filter((channel) =>
+    channel.type !== General.DM_CHANNEL && channel.type !== General.GM_CHANNEL,
+);
+
 const ChannelSelector = (props: Props & { className?: string }) => {
-    const getChannelFromID = (channelId: string) => props.selectableChannels.find((channel) => channel.id === channelId) || {display_name: 'Unknown Channel', id: channelId};
+    const selectableChannels = useSelector(getMyPublicAndPrivateChannels);
+
+    type GetChannelType = (channelID: string) => Channel
```
lieut-data: Should we define this outside the closure? (i.e. at the file level)

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -126,8 +125,10 @@ const OuterContainer = styled.div`
 `;
 
 interface Props {
-    teamId: string;
     isNew: boolean;
+    teamId?: string;
+}
+interface URLParams {
```
lieut-data: nit:
```suggestion
}

interface URLParams {
```

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -232,16 +231,16 @@ const PlaybookEdit = (props: Props) => {
 
                     setPlaybook({
                         ...template.template,
-                        team_id: props.teamId,
+                        team_id: props.teamId || '',
```
lieut-data, line 234: It feels weird to have a playbook sometimes fall back to an empty `teamId`. I'm guessing this is TypeScript complaining about `teamId` potentially being `undefined`? What would it take to make `teamId` a non-optional property here (with error checking higher up for the cases where we can't know?)

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -484,7 +480,7 @@ const PlaybookEdit = (props: Props) => {
     };
 
     const getUsers = () => {
-        return dispatch(getProfilesInTeam(props.teamId, 0, PROFILE_CHUNK_SIZE, '', {active: true}));
+        return dispatch(getProfilesInTeam(props.teamId || playbook.team_id, 0, PROFILE_CHUNK_SIZE, '', {active: true}));
```
lieut-data, line 483: Would it make sense to use `playbook.team_id` unconditionally for clarity?

lieut-data (approved):
Super nice cleanup! Just a handful of non-blocking thoughts.

webapp/src/components/backstage/playbook_edit.tsx
iomodo: It could be `undefined` when we create playbook we pass the `teamId`, but when we edit playbook we pass `playbookId` and get `teamId` from there.
In this case since it's a new playbook there will be a `teamId` and the ` || ''` is just for TypeScript.

iomodo:
PR has a known issue. When go to the backstage websocket connection is closed. When return to the channels websocket connection may not be open before we dispatch `executeSlashCommand` actions. 
Steps to reproduce the issue:
1. go to the backstage
2. refresh
3. go to any playbook
4. click run
dialogue is not shown, when it should be.

Notice that if you do not refresh everything is working fine since it takes smaller time to load the page.

Discussed with @cpoile  and @crspeller offline(huge thanks to them in helping detecting the issue), we suggest to merge with this known problem.

webapp/src/client.ts
```diff
@@ -101,12 +101,12 @@ export function fetchPlaybookRunChannels(teamID: string, userID: string) {
...
 }
 
-export async function clientExecuteCommand(dispatch: Dispatch<AnyAction>, getState: GetStateFunc, command: string, teamId: string) {
+export async function clientExecuteCommand(dispatch: Dispatch<AnyAction>, getState: GetStateFunc, command: string, teamId: string, isBackstage: boolean) {
     let currentChannel = getCurrentChannel(getState());
 
     // Default to town square if there is no current channel (i.e., if Mattermost has not yet loaded)
     // or in a different team.
-    if (!currentChannel || currentChannel.team_id !== teamId) {
+    if (!currentChannel || currentChannel.team_id !== teamId || isBackstage) {
```
lieut-data, line 109: It's not super obvious what `isBackstage` does here: given the known issue with the websockets, does this bit actually help us?

lieut-data (approved):

crspeller:
@lieut-data We did discuss that. The decision that was made was that was not necessary given the solution coming that will fix the websockets.

crspeller:
@lieut-data I think the idea there was @iomodo noted it worked under certain conditions of what team the playbook was on. I think this makes it even and adds a delay everywhere that wins the race in certain situations. 
I don't think it's necessary given the eventual fix.

PR #682, opened 2021-08-09
MM-37678: Disallow title editing if user has no permissions to change channel name

agarciamontoro:
#### Summary
While we decide how to handle run titles decoupled from channel names, this PR completely disables the ability to edit the title from the RHS when the user has no permissions to change the channel name.

This mimics the feature to change the channel name from the channel header: the menu entry `Rename Channel` is only shown to users that have permissions to actually change it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37678

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/rhs/rhs_about_title.tsx
+++ b/webapp/src/components/rhs/rhs_about_title.tsx
@@ -2,7 +2,14 @@
 // See LICENSE.txt for license information.
 
 import React, {useState, useRef, useEffect} from 'react';
+import {useSelector} from 'react-redux';
 import styled, {StyledComponent} from 'styled-components';
+
+import {GlobalState} from 'mattermost-redux/types/store';
+import General from 'mattermost-redux/constants/general';
+import Permissions from 'mattermost-redux/constants/permissions';
+import {getCurrentChannel} from 'mattermost-redux/selectors/entities/channels';
+import {haveIChannelPermission} from 'mattermost-redux/selectors/entities/roles';
 
 import {useClickOutsideRef, useKeyPress} from 'src/hooks/general';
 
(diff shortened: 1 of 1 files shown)


agarciamontoro:
@abhijit-singh, I think this makes sense from a UX standpoint, but assigning you here to double-check.

lieut-data (approved):
Classy!

abhijit-singh (approved):

webapp/src/components/rhs/rhs_about_title.tsx
```diff
@@ -74,6 +89,18 @@ const RHSAboutTitle = (props: Props) => {
...
+    const channel = getCurrentChannel(state);
+
+    const permission = channel.type === General.OPEN_CHANNEL ? Permissions.MANAGE_PUBLIC_CHANNEL_PROPERTIES : Permissions.MANAGE_PRIVATE_CHANNEL_PROPERTIES;
+
+    return haveIChannelPermission(state, {
+        channel: channel.id,
+        team: channel.team_id,
+        permission,
+    });
+};
```
cpoile, line 102: I'm wondering... Have you viewed the react profiler after this change? Will this cause the whole component to re-render on any store change?

cpoile (changes requested):
Just a quick question, otherwise nice!

webapp/src/components/rhs/rhs_about_title.tsx
agarciamontoro: I have not. Taking a look, thanks!

agarciamontoro: @cpoile, I'm checking with the paint flashing tool and making changes in the store does not make the component re-render. Maybe I'm not testing it as expected, would you have done something different to check your concern?

cpoile: Nope, that's perfect, thanks @agarciamontoro !

cpoile (approved):
Nice!

PR #683, opened 2021-08-11
Mm 37645 customizable auto categorization

Duaard:
#### Summary

See [issue](https://github.com/mattermost/mattermost-server/issues/18062) description for details

#### Ticket Link

Fixes: mattermost/mattermost-server/issues/18062
JIRA: https://mattermost.atlassian.net/browse/MM-37645

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Modify backstage setting to add input box
- [x] Add a database migration to add new column CategoryName
- [x] Modify data model and queries, copy value when playbook is run
- [x] Modify createOrUpdatePlaybookRunSidebarCategory function to use category name
- [x] Add input validation for category name 
- [x] Unit tests updated
- [x] E2E tests


agarciamontoro:
Thank you for this, @Duaard! We'll review it as soon as possible :)

lieut-data:
Due to our licensing situation, pkg.go.dev won't expose this documentation in a useful way: but wondering if we can do something in the `client/` package which we explicitly license (and module) separately so that it's visible on https://pkg.go.dev/github.com/mattermost/mattermost-plugin-playbooks/client?

server/app/playbook_run.go
```diff
@@ -59,7 +59,7 @@ type PlaybookRun struct {
 	MessageOnJoin                        string          `json:"message_on_join"`
 	ExportChannelOnArchiveEnabled        bool            `json:"export_channel_on_archive_enabled"`
 	ParticipantIDs                       []string        `json:"participant_ids"`
-	CategorizeChannelEnabled             bool            `json:"categorize_channel_enabled"`
+	CategoryName                         string          `json:"category_name"`
```
lieut-data, line 62: Nice! I was going to ask why we didn't add `CategoryName` to `CategorizeChannelEnabled`, but I see now the pattern here to omit the `*Enabled` fields from the `playbook` when unnecessary.

server/app/playbook_run_service.go
```diff
@@ -1578,7 +1576,7 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
 
 // createOrUpdatePlaybookRunSidebarCategory creates or updates a "Playbook Runs" sidebar category if
 // it does not already exist and adds the channel within the sidebar category
-func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID string) error {
+func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID, playbookRunsCategoryName string) error {
```
lieut-data, line 1579: nit: let's simplify the variable name to just `categoryName` here -- we don't write `playbookRunChannelID`, etc., when the context is unambiguous.

server/app/playbook_run_service_test.go
```diff
@@ -629,7 +629,7 @@ func TestUserHasJoinedChannel(t *testing.T) {
 		telemetryService := &telemetry.NoopTelemetry{}
 		scheduler := mock_app.NewMockJobOnceScheduler(controller)
 
-		playbookRun := &app.PlaybookRun{CategorizeChannelEnabled: true}
+		playbookRun := &app.PlaybookRun{CategoryName: "Playbook Run"}
```
lieut-data, line 632: For additional coverage, can we try with a "new" `CategoryName`, ala "Playbook Runs 2"`?

server/app/playbook_run_service_test.go
```diff
@@ -578,10 +578,10 @@ func TestUserHasJoinedChannel(t *testing.T) {
...
```
lieut-data, line 581: nit: since this is a category name, we should probably keep the pluralization: `Playbook Runs`, just to make the test more realistic.

server/sqlstore/migrations.go
```diff
@@ -980,4 +980,28 @@ var migrations = []Migration{
...
+			} else {
+				if err := addColumnToPGTable(e, "IR_Playbook", "CategoryName", "TEXT DEFAULT ''"); err != nil {
+					return errors.Wrapf(err, "failed adding column CategoryName to table IR_Playbook")
+				}
+
+				if err := addColumnToPGTable(e, "IR_Incident", "CategoryName", "TEXT DEFAULT ''"); err != nil {
+					return errors.Wrapf(err, "failed adding column CategoryName to table IR_Incident")
+				}
+			}
+			return nil
```
lieut-data, line 1004: Can you add a data migration here that updates this field (when enabled) to be "Playbook Runs", preserving backwards compatibility for any existing configurations?

server/sqlstore/migrations.go
```diff
@@ -980,4 +980,28 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.24.0"),
+		toVersion:   semver.MustParse("0.25.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
```
lieut-data, line 986: @agarciamontoro, thoughts on what to do with the now unused `CategorizeChannelEnabled` on `IR_Incident`?

server/api/playbooks.go
```diff
@@ -196,6 +205,15 @@ func (h *PlaybookHandler) updatePlaybook(w http.ResponseWriter, r *http.Request)
 		}
 	}
 
+	if playbook.CategorizeChannelEnabled {
+		categoryNameLength := len(playbook.CategoryName)
```
lieut-data, line 209: Can we pull this out into a `validateCategoryName` helper?

webapp/src/components/backstage/automation/settings.tsx
```diff
@@ -55,6 +53,8 @@ interface Props {
     signalAnyKeywords: string[];
     categorizePlaybookRun: boolean;
     onToggleCategorizePlaybookRun: () => void;
+    categoryName: string;
+    categoryNameChange: (url: string) => void;
```
lieut-data, line 57: nit: can we fix the type here to call the parameter `categoryName`, not `url`?

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -476,6 +476,16 @@ const PlaybookEdit = (props: Props) => {
         setChangesMade(true);
     };
 
+    const handleCategoryNameChange = (name: string) => {
+        if (playbook.category_name !== name) {
```
lieut-data, line 480: Nice check!

lieut-data (changes requested):
Really tremendous work, @Duaard! Just a handful of comments below :)

server/sqlstore/migrations.go
agarciamontoro: Good question. I guess we can simply "deprecate" it. We don't have a strict process for that; the only similar thing we've done in the codebase is to mark it as deprecated here https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/sqlstore/playbook_run.go#L388-L392. Thoughts?

server/sqlstore/migrations.go
```diff
@@ -980,4 +980,44 @@ var migrations = []Migration{
...
+				if _, err := e.Exec("UPDATE IR_Playbook SET CategoryName = 'Playbook Runs' WHERE CategorizeChannelEnabled=1"); err != nil {
+					return errors.Wrapf(err, "failed setting default value in column CategoryName of table IR_Playbook")
+				}
+
+				if err := addColumnToMySQLTable(e, "IR_Incident", "CategoryName", "TEXT"); err != nil {
+					return errors.Wrapf(err, "failed adding column CategoryName to table IR_Incident")
+				}
+
+				if _, err := e.Exec("UPDATE IR_Incident SET CategoryName = 'Playbook Runs' WHERE CategorizeChannelEnabled=1"); err != nil {
+					return errors.Wrapf(err, "failed setting default value in column CategoryName of table IR_Playbook")
```
agarciamontoro, line 1001: Typo here
```suggestion
					return errors.Wrapf(err, "failed setting default value in column CategoryName of table IR_Incident")
```

server/api/playbooks.go
```diff
@@ -461,3 +475,12 @@ func removeDuplicates(a []string) []string {
 	}
 	return res
 }
+
+func (h *PlaybookHandler) validateCategoryName(categoryName string) error {
+	categoryNameLength := len(categoryName)
+	if categoryNameLength == 0 || categoryNameLength > 22 {
```
agarciamontoro, line 481: Here we should simply check the maximum length. The code already uses the category name only if it's not empty, so we can skip the minimum length check.

server/app/playbook_run_service_test.go
```diff
@@ -671,4 +671,109 @@ func TestUserHasJoinedChannel(t *testing.T) {
...
+
+		s := app.NewPlaybookRunService(client, store, poster, logger, configService, scheduler, telemetryService, pluginAPI)
+
+		userID := "user_id"
+		channelID := "channel_id"
+		actorID := ""
+
+		s.UserHasJoinedChannel(userID, channelID, actorID)
+	})
 }
```
Duaard, line 779: Let me know if this is fine. I just yoinked the existing test and changed the name, but it should work. 😆

server/app/playbook_run_service.go
```diff
@@ -1578,15 +1576,15 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
 
 // createOrUpdatePlaybookRunSidebarCategory creates or updates a "Playbook Runs" sidebar category if
 // it does not already exist and adds the channel within the sidebar category
-func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID string) error {
+func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID, categoryName string) error {
```
agarciamontoro, line 1579: I think we should make sure to lowercase the `categoryName` string inside this function before using it, as `Custom Category` and `custom category` will actually create two different categories in the sidebar, while they render as the same.
image

The problem with this is that we'll break the `Playbook Runs` category whenever old playbooks that had not changed the category name are saved again (as new runs will change from `Playbook Runs` to `playbook runs`). @lieut-data, thoughts on that migration? I'm not too concerned about this, as people can still move the channels beteween categories and renaming and deleting the categories, but it will be slightly annoying.

agarciamontoro (changes requested):
This is amazing, @Duaard, thank you!

I left a couple of minor comments below that we can discuss.

server/api/playbooks.go
Duaard: It will actually still push through for the update and creation of **Playbook**, the check that will only use the Category Name is for **Playbook Runs**. Similar to how the **Webhook** input is being handled, it will just not push through with the update/create as the API will fail at the URL parsing.

I followed the logic of the webhooks while coding this and also observed their behavior. I can change this for the Category Name input but keep in mind to do the same for the webhook. (if it's not intended to let empty inputs be submitted)

server/app/playbook_run_service.go
Duaard: Good catch, yeah that makes sense. I'll wait on @lieut-data for the decision on `Playbook Runs` before working on this.

webapp/src/components/backstage/automation/settings.tsx
```diff
@@ -157,9 +157,16 @@ export const AutomationSettings = (props: Props) => {
                     />
                 </Setting>
                 <Setting id={'user-joins-channel-categorize'}>
-                    <CategorizePlaybookRun
+                    <PatternedInput
                         enabled={props.categorizePlaybookRun}
                         onToggle={props.onToggleCategorizePlaybookRun}
+                        input={props.categoryName}
+                        onChange={props.categoryNameChange}
+                        pattern={'^[\\s\\S]{1,22}$'}
```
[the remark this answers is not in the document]
Duaard: Yeah I agree on this. I think the error feedback from the `PatternedInput` isn't too helpful in this case. Let me work on modifying the `PatternedInput` component.

server/app/playbook_run_service.go
lieut-data: Oops, apologies for missing this notification, @Duaard & @agarciamontoro. On the case issue, I definitely agree that we want to avoid creating duplicate categories with the same name. Is there a way to check all the category names for the user in question and do the lowercase comparison on our side and reusing the one we find if it matches?

lieut-data: (Not sure if that addresses the `Playbook Runs` question or if there is another comment I might be missing? I'll finish my overall review on this today either way :))

server/app/playbook_run_service_test.go
lieut-data: Looks good!

lieut-data (approved):
Nice work on this, @Duaard -- appreciate your patience with my delay in responding, as well as your determination to get this done well :)

server/app/playbook_run_service.go
Duaard: We can definitely do that in the `createOrUpdatePlaybookRunSidebarCategory` function, with a slight edge case where there already exists duplicate categories (created manually). It will then be added to the first one to match the lower case check.

agarciamontoro: > check all the category names for the user in question and do the lowercase comparison on our side and reusing the one we find if it matches?

Yeah, that also solves the issue with the `Playbook Runs` category too, if I'm not mistaken. @duaard, thoughts on the proposed solution?

Duaard: > edge case where there already exists duplicate categories (created manually)
I'll add the check in the `createOrUpdatePlaybookRunSidebarCategory` function, but please do note this edge case

Duaard: Hi @agarciamontoro added in the check, this should prevent this issue. 😄

agarciamontoro (approved):
Thank you again for all your hard work, @Duaard! I tested this with category names that only differed in the casing and it worked perfect :tada: Approving and merging :)

agarciamontoro:
Oh, let me fix the conflict here, I caused that one with #684 :)

agarciamontoro:
Done :) Thanks once again, @Duaard!

server/app/playbook_run_service.go
```diff
@@ -1578,15 +1576,15 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
...
 	sidebar, err := s.pluginAPI.Channel.GetSidebarCategories(userID, teamID)
 	if err != nil {
 		return err
 	}
 
 	var categoryID string
 	for _, category := range sidebar.Categories {
-		if category.DisplayName == playbookRunsCategoryName {
+		if strings.EqualFold(category.DisplayName, categoryName) {
```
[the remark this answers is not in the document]
Duaard: I didn't knew it also, it was the linter who suggested it. 😆

PR #686, opened 2021-08-13
MM-37482 - Simplify status to InProgress and Finished; add Finish command

cpoile:
#### Summary
- I tried to tease this apart into separate commits, but it's just too interconnected -- sorry. :/
- Migrate Archived to Finished, every other status to InProgress
- Remove the status field from status posts (it is not set and not read anymore)
- Remove cruft code that was getting status/active-ness from status posts -- it is now always on the run's `CurrentStatus` field
- I debated wether to make a new `State` field, because the `InProgress` and `Finished` states are basically independent from any custom status we might add later. But I noticed that the `CurrentStatus` field is used throughout to search/sort/filter in the api and on the front end, and (for now at least) we want to still support that with `InProgress` and `Finished`. So I'm still using the `CurrentStatus` field, and we can revisit what to do if/when we introduce custom statuses.
- We now have the discreet action to "Finish" a run, that's the only way to move a run into the `Finished` state.
- After finishing a run, post a simple message (I'll do a separate message in the channel to figure out the best wording for that)
- UI/UX is basically the same, just with the filter menus showing `In Progress` and `Finished`, the new `/playbook finish` slash command, and the badges showing `In Progress` and `Finished`.

Cleanup and design work needed (separate PR) on the following:
- We still have to remove description from the interactive dialog code path, and all calls that use that (in tests too)
- The finished state is weird (post update is still primary and very prominent in the RHS -- disconcerting when the run is supposed to be finished)

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-37482

#### Checklist
- [x] Telemetry updated
- [x] Unit tests updated


server/sqlstore/migrations.go
```diff
@@ -980,4 +980,49 @@ var migrations = []Migration{
...
+			}
+
+			if _, err := e.Exec(`
+					UPDATE IR_Incident
+					SET CurrentStatus =
+							CASE
+								WHEN CurrentStatus = 'Archived'
+									THEN 'Finished'
+								ELSE 'InProgress'
+								END;
```
lieut-data, line 1020: nit: indentation?

webapp/src/components/backstage/playbook_runs/status_badge.tsx
```diff
@@ -24,19 +24,14 @@ const Badge = styled.div<BadgeProps>`
...
-        `;
-        case PlaybookRunStatus.Active:
+        case PlaybookRunStatus.InProgress:
             return css`
                 background-color: var(--dnd-indicator);
         `;
-        case PlaybookRunStatus.Resolved:
+        case PlaybookRunStatus.Finished:
             return css`
                 background-color: var(--online-indicator);
```
lieut-data, line 33: Can we double check the colours with @abhijit-singh? I saw a mock recently that suggested more neutral colours than the screenshots in this PR.

lieut-data (approved):
Nice work -- a tricky set of changes and decisions to work through.

I've left a handful of comments, but not strictly blocking I think :)

webapp/src/components/backstage/playbook_runs/status_badge.tsx
abhijit-singh: Thanks for highlighting this @lieut-data! It'll be great if we can use:
* `Sidebar Active Border` (#5D89EA in the new `Denim` theme) colour variable for `In progress`
* `64% opacity: Center Channel Text` (#3F4350, 64% in the new `Denim` theme) colour variable  for `Finished`

server/app/playbook_run.go
```diff
@@ -754,5 +723,19 @@ func (o PlaybookRunFilterOptions) Validate() (PlaybookRunFilterOptions, error) {
 		options.StartedLT = 0
 	}
 
+	if options.Status != "" && !validStatus(options.Status) {
```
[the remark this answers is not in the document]
cpoile: I thought about that too, but the way we're doing status filtering we have three possible filtering states (`InProgress`, `Finished`, `InProgress & Finished`), so we still need the list of statuses in the filtering options.

server/api/playbook_runs.go
```diff
@@ -674,24 +675,22 @@ func (h *PlaybookRunHandler) status(w http.ResponseWriter, r *http.Request) {
...
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte(`{"status":"OK"}`))
+}
+
+// updateStatusD handles the POST /runs/{id}/finish endpoint, user has edit permissions
+func (h *PlaybookRunHandler) finish(w http.ResponseWriter, r *http.Request) {
+	playbookRunID := mux.Vars(r)["id"]
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	if err := h.playbookRunService.FinishRun(playbookRunID, userID); err != nil {
```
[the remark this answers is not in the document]
cpoile: I thought that since we're checking edit permissions, and we're moving towards using the Bot to post all messages, we can skip that check (it was always a little over-defensive in the first place, imo)

server/api/playbook_runs.go
```diff
@@ -65,6 +65,7 @@ func NewPlaybookRunHandler(router *mux.Router, playbookRunService app.PlaybookRu
 	playbookRunRouterAuthorized.HandleFunc("", handler.updatePlaybookRun).Methods(http.MethodPatch)
 	playbookRunRouterAuthorized.HandleFunc("/owner", handler.changeOwner).Methods(http.MethodPost)
 	playbookRunRouterAuthorized.HandleFunc("/status", handler.status).Methods(http.MethodPost)
+	playbookRunRouterAuthorized.HandleFunc("/finish", handler.finish).Methods(http.MethodPost)
```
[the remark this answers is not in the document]
cpoile: Yep, that makes sense since it is idempotent. Thanks!

server/app/playbook_run_service.go
```diff
@@ -838,37 +821,116 @@ func (s *PlaybookRunServiceImpl) UpdateStatus(playbookRunID, userID string, opti
...
+		s.scheduler.Cancel(RetrospectivePrefix + playbookRunID)
+		if playbookRunToModify.RetrospectiveReminderIntervalSeconds != 0 {
+			if err = s.SetReminder(RetrospectivePrefix+playbookRunID, time.Duration(playbookRunToModify.RetrospectiveReminderIntervalSeconds)*time.Second); err != nil {
+				return errors.Wrap(err, "failed to set the retrospective reminder for playbook run")
 			}
 		}
+	}
+
+	if playbookRunToModify.ExportChannelOnFinishedEnabled {
+		fileID, err2 := s.exportChannelToFile(playbookRunToModify.Name, playbookRunToModify.OwnerUserID, playbookRunToModify.ChannelID)
```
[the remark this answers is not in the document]
cpoile: Good idea. I will start doing that from now on.

server/sqlstore/migrations.go
```diff
@@ -980,4 +980,49 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.24.0"),
+		toVersion:   semver.MustParse("0.25.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DATABASE_DRIVER_MYSQL {
+				if err := renameColumnMySQL(e, "IR_Playbook", "ExportChannelOnArchiveEnabled", "ExportChannelOnFinishedEnabled", "BOOLEAN NOT NULL DEFAULT FALSE"); err != nil {
```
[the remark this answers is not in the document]
cpoile: From the triage discussion, I'm assuming this is okay for 6.0. 👍

server/sqlstore/migrations.go
cpoile: Fixed! (at least in my IDE it is aligned with the other long-form sql queries in the migrations file)

server/api/playbook_runs.go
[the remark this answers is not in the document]
lieut-data: Yes, that makes sense! It's always a bit unnerving not seeing API checks in the body of the method, so wanted to triple check that the API wrappers sufficed.

server/app/playbook_run.go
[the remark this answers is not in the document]
lieut-data: Good point!!

server/sqlstore/migrations.go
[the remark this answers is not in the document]
lieut-data: Yep, break away!

cpoile:
Added the finish run dialog

server/app/playbook_run_service.go
```diff
@@ -647,7 +641,12 @@ func (s *PlaybookRunServiceImpl) RemoveTimelineEvent(playbookRunID, userID, even
 	return nil
 }
 
-func (s *PlaybookRunServiceImpl) broadcastStatusUpdate(statusUpdate string, playbookRun *PlaybookRun, authorID, originalPostID string) error {
+func (s *PlaybookRunServiceImpl) broadcastStatusUpdate(statusUpdate, playbookRunID, authorID, originalPostID string) error {
+	playbookRun, err := s.store.GetPlaybookRun(playbookRunID)
```
iomodo, line 645: Why not pass `PlaybookRun` as an argument to the func? Won't it reduce DB read?

iomodo (approved):
Wonderful! Just a minor non-blocking question.

server/app/playbook_run_service.go
```diff
@@ -885,21 +918,23 @@ func (s *PlaybookRunServiceImpl) FinishRun(playbookRunID, userID string) error {
 		}
 	}
 
+	var fileID string
```
lieut-data, line 921: nit: if `fileID` is only used inside the condition, we can declare inside that scope instead.

server/app/playbook_run_service.go
```diff
@@ -1874,6 +1909,22 @@ func (s *PlaybookRunServiceImpl) addPlaybookRunUsers(playbookRun *PlaybookRun, c
 	return nil
 }
 
+func (s *PlaybookRunServiceImpl) newFinishPlaybookRunDialog(outstanding int) *model.Dialog {
+	message := "Are you sure you want to finish the run?"
+	if outstanding == 1 {
+		message = "There is **1 outstanding task**, are you sure you want to finish the run?"
```
lieut-data, line 1915: This is a run-on sentence at present. Perhaps something like, `"There is **1 outstanding task**. Are you sure you want to finish the run?"` cc @justinegeffen

lieut-data (approved):

abhijit-singh (approved):
This looks good!

server/app/playbook_run_service.go
cpoile: imo, we should be retrieving the playbook run from the DB when we enter these broadcast/websocket update functions -- that way it pulls in all the up-to-date data. E.g., without doing this, the caller would have to manually set the CurrentStatus, but it doesn't--the store method does. So setting it in the caller would be duplicate work, and it's easy to forget to do that. (this just bit me, that's why I made the change)

server/app/playbook_run_service.go
cpoile: ah, you're right, my mistake. Should have done that in the first place.

server/app/playbook_run_service.go
justinegeffen: +1 on that. :)

PR #689, opened 2021-08-16
MM-37891: remove close button in backstage

lieut-data:
#### Summary
Remove the `X` in the backstage now that we have the product switcher:
image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-37891

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):
Two clicks to leave instead of one now 😒 

But the code looks good.

PR #690, opened 2021-08-16
Show no dropdown for a single team

iomodo:
#### Summary
If we have single team, when user clicks playbook creation there will be no dropdown.
Also it was not clear that we wanted to pick a team to create a playbook in. So `placeholder` is changed to `Select a team`. Not sure if it's enough, need guidance from @abhijit-singh. 
P.S. We could take into account the fact that playbooks won't be tied to teams in the future

https://user-images.githubusercontent.com/2340127/129591214-dc9ac69d-bc29-470f-92e3-8b49f40c8a24.mp4

Screen Shot 2021-08-16 at 18 25 01

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


crspeller (approved):

agarciamontoro (approved):
Looks great :tada:

abhijit-singh (approved):

PR #691, opened 2021-08-16
MM-37893: calc global header height in backstage

lieut-data:
#### Summary
Compensate for the 40px consumed by the backstage header, allowing the full content on the backstage to be displayed. Previously, page navigation controls rendered outside the viewport and could not be reached.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-37893

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/backstage.tsx
+++ b/webapp/src/components/backstage/backstage.tsx
@@ -43,7 +43,8 @@ import {applyTheme} from './css_utils';
 
 const BackstageContainer = styled.div`
     background: var(--center-channel-bg);
-    height: 100%;
+    // The container should take up all vertical real estate, less the height of the global header.
+    height: calc(100% - 40px);
     display: flex;
     flex-direction: column;
     overflow-y: auto;


webapp/src/components/backstage/backstage.tsx
```diff
@@ -43,7 +43,8 @@ import {applyTheme} from './css_utils';
 
 const BackstageContainer = styled.div`
     background: var(--center-channel-bg);
-    height: 100%;
+    // The container should take up all vertical real estate, less the height of the global header.
+    height: calc(100% - 40px);
```
lieut-data, line 47: Is this the best approach? I hate hard-coding `40px` here -- even a CSS variable might be wrong.

deanwhillier: @lieut-data, as question I have before answering your is "Should the Global Header even be visible when the backstage is being rendered?"

There's also [this approach](https://github.com/mattermost/mattermost-webapp/pull/8595), though I really don't like the idea of the GH still rendering when it's not visible, so that's another challenge that should probably be solved for.

lieut-data: @deanwhillier, our "backstage" is a misnomer -- carry over terminology from when we thought of this as a plugin vs. a product. (Open to suggestions for a better term :p) But either way, it just refers to our full screen experience outside of Channels which definitely needs the global header visible.

deanwhillier: Just so I have it in my head @lieut-data, does that mean that the integrations view that renders in the `backstage` container should have the Global Header visible?

I saw a comment somewhere near the end of last week indicating that some buttons like the at-mentions or saved posts don't do anything when the GH is visible while integrations is open. If we need the GH to be visible when backstage is in use, we'll need to solve for these kinds of situations.

nevyangelova (approved):
LGTM I saw this might relate to https://github.com/mattermost/mattermost-webapp/pull/8595/files so we have to check how they work together.

webapp/src/components/backstage/backstage.tsx
lieut-data: @deanwhillier, apologies, but I can't speak to Channels' "backstage" experience, which is definitely an "internal" view within Channels. These particular screens are our core product.

deanwhillier: Okay, got it now, thanks for humouring me!

So this is the easiest solution for now. It's already being done elsewhere out of necessity.

A potentially better way to do it would be to have the parent container like so (not being done):
```css
display: flex;
flex-direction: column;
```

Then have the GH indicate it shouldn't collapse/shrink vertically (already being done):
```css
flex: 0; // equivalent of flex-grow: 0, flex-shrink: 0;
```

And then the container immediately following the GH (i.e. playbooks) take as much space as it can get:
```css
flex: 1; // equivalent of flex-grow: 1, flex-shrink: 0;
```

All that said, I think it's okay to keep what you have for now as we will likely be addressing the overall layout in a future 'Layout Overhaul' project following v.6.

deanwhillier (approved):

webapp/src/components/backstage/backstage.tsx
lieut-data: Excellent -- thank you :)

PR #694, opened 2021-08-17
MM-37895 error page rendering

lieut-data:
#### Summary
Fix the error page rendering issues introduced with the global header. Note that the global header currently renders alongside the error page as opposed to going "full white" like the Channels product. This was "easiest" but we can revisit in the future as needed.

image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-37895

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/backstage/backstage.tsx
```diff
@@ -109,62 +109,67 @@ const Backstage = () => {
...
-                                    title='Give Feedback'
-                                    size={1}
-                                />
-                            </span>
-                            {'Give Feedback'}
-                        </BackstageTitlebarItem>
-                    }
-                </div>
-            </BackstageNavbar>
+            <Switch>
```
lieut-data, line 112: I'm just wrapping with a `Switch` and `Route` so as to avoid rendering our own navbar in the error pages.

webapp/src/components/error_page.tsx
```diff
@@ -48,11 +48,11 @@ const ErrorPage = () => {
                 <div className='error__icon'>
                     <WarningIcon/>
                 </div>
-                <h2>
-                    {title}
+                <h2 data-testid='errorMessageTitle'>
```
lieut-data, line 51: This is copy-pasta'd from the webapp's updated styling for same.

calebroseland (changes requested):
LGTM! Just one with suggestion for some styling issues in the product switcher and account menu when you hit the error page.

webapp/src/components/error_page.tsx
```diff
@@ -13,7 +13,7 @@ import {pluginUrl} from 'src/browser_routing';
 
 const ErrorPage = () => {
     useEffect(() => {
-        document.body.setAttribute('class', 'sticky error');
+        document.body.setAttribute('class', 'error');
         return () => {
             document.body.removeAttribute('class');
```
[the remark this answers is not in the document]
lieut-data: Oh, thank you, @calebroseland!

calebroseland (approved):

PR #696, opened 2021-08-17
MM-37088: updated font styling

lieut-data:
#### Summary
Updates to use Metropolis as outlined in the linked Jira ticket. While I was in here, I ripped out some unused styles and files, aiming to simplify things for the final commit that leverages @calebroseland excellent idea for a "mixin".

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-37088

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/styles/headings.ts
```diff
@@ -0,0 +1,11 @@
+import {css} from 'styled-components';
+
+export const RegularHeading = css`
```
crspeller, line 3: Not blocking, but it feels like this might be the wrong way to do this.  Not sure though.
For example this might break @hmhealey's plugin that changes the font to a dyslexic friendly font.
This seems fine to me, but maybe we should see how this fits into the overall strategy?

crspeller (approved):

webapp/src/styles/headings.ts
lieut-data: That's a really interesting point. I presume the challenge you're identifying isn't the css mixin being used, but the application of the font so specifically vs. being applied at a much higher (and generically mutable) level?

@deanwhillier, thoughts on this topic as it relates to how we [introduced Metropolis](https://github.com/mattermost/mattermost-webapp/pull/8402#event-5083938499)?

deanwhillier: Metropolis has been setup in the Webapp to globally apply to all h1-h3 tags, replacing the previous Open Sans font ... unless otherwise overridden by more specific styles as needed.

That is pretty much the extent of the introduction of the new font at this point. The overall typography handling in the webapp is eclectic and not always properly semantic so a proper global implementation of Metropolis is not really on the table at the moment. As a result, there were some tweaks made to various areas of the webapp that favour visual rather than semantic accuracy for now.

Ideally, semantically accurate heading tags would be used for headings to inherit from the global styling, with local overrides to size/weight/colour via something like styled-components as designed/needed.

Another, possibly better approach would be to use a common component that manages semantics and styling of typography via theming ...

hmhealey: I'm wondering how styled components generally do global styles like these. I was actually expecting my plugin to have to overwrite more styles than it did, but since most of our fonts are done by a few top level definitions currently, it was easy to override those.

One option could be a CSS variable or something that's easier to override by plugins. I know some users have wanted different fonts for other reasons, and while we don't support those currently, maybe that'd be another option to expose that ability for others

calebroseland: I think the eventual ideal would be Compass components to be used for official headings, with Themes perhaps growing beyond colors to include things like fonts, weights, etc.

This in-between is somewhat of a hybrid of global inheritance and manual styling and/or CSS mixins in styled components. 

I think it's a step in the right direction, in that the CSS mixins create a usage trail that we can eventually swap out for the official components when they are ready for usage in plugins.

calebroseland (approved):

PR #699, opened 2021-08-19
MM-37634, MM-37483: New Post update dialog with markdown preview

calebroseland:
#### Summary
- Adds new custom dialog for Update Status (/"Post update")
  - Behind `EnableExperimentalFeatures`
  - Adds markdown preview  
- Reuses `Textbox` from webapp
- Adds markdown preview to Edit Task dialog
- Existing Interactive Dialog remains for backwards-compat

CleanShot 2021-08-20 at 17 40 27
image
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-37634
  - https://mattermost.atlassian.net/browse/MM-37483

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


webapp/src/slash_command.ts
```diff
@@ -1,31 +1,53 @@
...
 
-            messageTrimmed = `/playbook run ${clientId}`;
-
-            return Promise.resolve({message: messageTrimmed, args});
+            return {message: `/playbook run ${clientId}`, args};
         }
 
-        if (messageTrimmed && messageTrimmed.startsWith('/playbook info')) {
-            const state = store.getState();
+        if (experimentalFeaturesEnabled && message?.startsWith('/playbook update') && isInPlaybookRunChannel) {
```
calebroseland, line 42: There's one more permission that I need to bring forward from https://github.com/mattermost/mattermost-plugin-playbooks/blob/MM-37634/server/command/command.go#L657

webapp/src/components/checklist_item.tsx
```diff
@@ -511,14 +513,16 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
...
+    show: boolean;
+    onDone: () => void;
+    checklistNum: number;
+    playbookRunId: string;
+
+    channelId: string;
+    itemNum: number;
+    taskTitle: string;
+    taskDescription: string;
+    taskCommand: string;
```
agarciamontoro, line 525: Should we remove the blank line?
```suggestion
    show: boolean;
    onDone: () => void;
    checklistNum: number;
    playbookRunId: string;
    channelId: string;
    itemNum: number;
    taskTitle: string;
    taskDescription: string;
    taskCommand: string;
```

webapp/src/components/markdown_textbox.tsx
```diff
@@ -0,0 +1,221 @@
...
+import {Link} from 'react-router-dom';
+import styled from 'styled-components';
+
+import {getConfig} from 'mattermost-redux/selectors/entities/general';
+
+import {Textbox} from 'src/webapp_globals';
+
+const DEFAULT_CHAR_LIMIT = 4000;
+
+type MarkdownTextboxProps = {
```
agarciamontoro, line 18: Maybe just `Props`?

webapp/src/components/markdown_textbox.tsx
```diff
@@ -0,0 +1,221 @@
...
+    id: string;
+    className?: string
+}
+
+const MarkdownTextbox = ({
+    value,
+    setValue,
+    autocompleteChannelId,
+    className,
+    ...props
```
agarciamontoro, line 32: I'm still 3/5 on not destructuring the props to actually know what's a prop and what is a value declared inside the component. I queued an item to discuss in our dev meeting, so we can leave this as is and discuss there.

webapp/src/components/markdown_textbox.tsx
```diff
@@ -0,0 +1,221 @@
...
+}: MarkdownTextboxProps) => {
+    const [showPreview, setShowPreview] = useState(false);
+    const config = useSelector(getConfig);
+
+    // @ts-expect-error
+    const charLimit = parseInt(config.MaxPostSize || '', 10) || DEFAULT_CHAR_LIMIT;
+
+    return (
+        <div className={className}>
```
agarciamontoro, line 41: Why are we using a `className` here and not a styled component, maybe provided by the caller (with a default one exported in this file, if needed)

webapp/src/components/markdown_textbox.tsx
```diff
@@ -0,0 +1,221 @@
...
+function TextboxLinks({
+    message = '',
+    characterLimit,
+    showPreview,
+    className,
+    updatePreview,
+}: TextboxLinksProps) {
+    const togglePreview = (e: MouseEvent) => {
+        e.preventDefault();
+        updatePreview?.(!showPreview);
```
agarciamontoro, line 104: Why is `updatePreview` optional? Aren't we always setting it above? I had to read this line twice to understand what was going on, and if `updatePreview` was not optional, it would have been easier.

webapp/src/components/markdown_textbox.tsx
```diff
@@ -0,0 +1,221 @@
...
+    };
+
+    const hasText = message && message.length > 0;
```
agarciamontoro, line 107: Can we do this here?

```suggestion
    const hasText = message?.length > 0;
```

webapp/src/components/markdown_textbox.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    const helpText = (
+        <Helpers>
+            <b>
+                {'**bold**'}
+            </b>
```
agarciamontoro, line 113: Super nit, feel free to ignore: can we write this (and the next ones) in one line? (or maybe the linter does not like it?)
```suggestion
            <b>{'**bold**'}</b>
```

webapp/src/components/markdown_textbox.tsx
```diff
@@ -0,0 +1,221 @@
...
+    );
+
+    return (
+        <div
+            className={classNames(className, {
+                hidden: message?.length > characterLimit,
+            })}
+        >
+            <div
+                style={{visibility: hasText ? 'visible' : 'hidden', opacity: hasText ? '' : '0'}}
```
agarciamontoro, line 143: Nit: Maybe I'm too used to styled components now, but I'd write this as a styled component with hasText as prop and write the CSS logic there. Thoughts?

agarciamontoro (changes requested):
Amazing work, @calebroseland! :tada: I always learn a lot about Typescript and web development in general reviewing your PRs, I really enjoy it :)

I left a handful of comments below, mostly questions and nits.

webapp/src/components/markdown_textbox.tsx
calebroseland: TextboxLinks was copy-pasted, with some refactoring afterwards to strip it down a bit and make it work with our styling. I'll look inter making the props type more deterministic.

webapp/src/components/markdown_textbox.tsx
calebroseland: These non-destructured props need to go to `Textbox` to be able to use or override those properly. The destructuring creates a clear contract that distinguishes between _controlled_ props (that are explicitly handled here in this wrapper component) and _non-controlled_ props (which are implicitly handled via routing directly to `Textbox`)

webapp/src/components/markdown_textbox.tsx
calebroseland: Initially because this started as a React.Fragment wrapper, but then I needed to distinguish, and reached for the closest thing that did that, but indeed that would be better—will adjust.

webapp/src/components/markdown_textbox.tsx
calebroseland: Just a result of copy-paste, if easily switched, I'll convert it.

webapp/src/components/markdown_textbox.tsx
calebroseland: I'm in a bit of a bind re: TS; I wanted to do something with `ComponentProps<typeof Textbox>`, but we don't have the infrastructure in place to import types from mattermost-webapp, so it's technically just `any` here.

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+import {pluginId} from 'src/manifest';
+import GenericModal, {Description, Label} from 'src/components/widgets/generic_modal';
+import {PlaybookRun} from 'src/types/playbook_run';
+
+import {usePlaybook} from 'src/hooks';
+import MarkdownTextbox from '../markdown_textbox';
+import {pluginUrl} from 'src/browser_routing';
+import {postStatusUpdate} from 'src/client';
+
+const ID = `${pluginId}_${nameof(UpdateRunStatusModal)}`;
```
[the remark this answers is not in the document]
calebroseland: I was having second thoughts/doubts on this. Thanks for confirming! Will go back to a normal string.

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+function UpdateRunStatusModal({playbookRunId, playbookId, channelId, ...props}: Props) {
+    const {formatMessage} = useIntl();
+    const [message, setMessage] = useState<string | null>(null);
+    const playbook = usePlaybook(playbookId);
+    if (playbook && message == null) {
+        setMessage(playbook.reminder_message_template);
+    }
+    const currentUserId = useSelector(getCurrentUserId);
+    const channel = useSelector((state: GlobalState) => getChannel(state, channelId) || {display_name: 'Unknown Channel', id: channelId});
+    const team = useSelector((state: GlobalState) => getTeam(state, channel.team_id));
```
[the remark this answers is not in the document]
calebroseland: Mmmm, good point 👍

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+            cancelButtonText={'Cancel'}
+            handleCancel={() => true}
+            handleConfirm={onConfirm}
+            {...props}
+        >
+            <FormContainer>
+                <Description>
+                    {formatMessage({
+                        id: `${ID}_description`,
+                        defaultMessage: 'This update will be saved to the <OverviewLink>overview page</OverviewLink>{hasBroadcast, select, true { and broadcast to {broadcastChannel}} other {}}.',
```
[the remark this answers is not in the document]
calebroseland: Just using for the ICU conditional formatting! I'm not advocating we start a playbooks translations project! (Yet, anyway—I'd love to see it happen, but it's a lot of community work and maintenance.)

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export function makeModalDefinition(props: Props) {
+    return {
+        modalId: ID,
+        dialogType: UpdateRunStatusModal,
+        dialogProps: props,
+    };
+}
+
+function UpdateRunStatusModal({playbookRunId, playbookId, channelId, ...props}: Props) {
```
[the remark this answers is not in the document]
calebroseland: The typing is a bit loose at the moment, but it is required to route the props that `ModalController` provides so it can work properly with `global.WebappUtils.modals.openModal`.

webapp/src/components/rhs/rhs_post_update.tsx
```diff
@@ -92,6 +99,7 @@ const getTimestamp = (playbookRun: PlaybookRun, isNextUpdateScheduled: boolean)
 };
 
 const PastTimeSpec = [
+    {within: ['second', -45], display: 'just now'},
```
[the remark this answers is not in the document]
calebroseland: The default `now` only works with past-time, so it wasn't doing anything for the update reminder and just fell through to "Update due this minute" (which is perfectly accurate), only difference from default here is it'll say "Last update just now" instead of "Last update now"

webapp/src/components/widgets/generic_modal.tsx
```diff
@@ -90,41 +96,39 @@ export default class GenericModal extends React.PureComponent<Props, State> {
...
+                </TertiaryButton>
             );
         }
 
         return (
-            <Modal
+            <StyledModal
                 dialogClassName={classNames('a11y__modal GenericModal', this.props.className)}
-                show={this.props.show}
+                show={this.props.show ?? this.state.show}
```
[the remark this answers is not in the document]
calebroseland: This is required in order to be cross-compatible with existing usages (which `show` is managed externally), and `global.WebappUtils.modals.openModal` which expects internal state and function props to work.

webapp/src/components/widgets/generic_modal.tsx
```diff
@@ -136,19 +140,78 @@ export default class GenericModal extends React.PureComponent<Props, State> {
...
+        }
+        .modal-footer {
+            padding: 24px 0 0 0;
+        }
+        .close {
+            margin: 12px 12px 0 0;
+        }
+    }
+
+    &&&& {
```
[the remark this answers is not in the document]
calebroseland: 💯, these overrides were frustrating to deal with—will be great to comeback one day and refactor them.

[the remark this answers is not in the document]
calebroseland: Perhaps even in the polish stage of this feature, if it works out.

webapp/src/components/widgets/generic_modal.tsx
```diff
@@ -136,19 +140,78 @@ export default class GenericModal extends React.PureComponent<Props, State> {
...
 
 const FooterContainer = styled.div`
     display: flex;
     flex-direction: column;
-    align-items: center;
+    align-items: flex-end;
+`;
+
+const ModalHeading = styled.h1`
+    font-family: Metropolis;
```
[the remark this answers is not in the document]
calebroseland: Yep, probably leaked in through init figma styling.

webapp/src/components/markdown_textbox.tsx
calebroseland: renamed for clarity

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+import MarkdownTextbox from '../markdown_textbox';
+import {pluginUrl} from 'src/browser_routing';
+import {postStatusUpdate} from 'src/client';
+
+const ID = `${pluginId}_${nameof(UpdateRunStatusModal)}`;
+
+type Props = {
+    playbookRunId: PlaybookRun['id'];
+    playbookId: PlaybookRun['playbook_id'];
+    channelId: PlaybookRun['channel_id'];
```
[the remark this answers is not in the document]
calebroseland: I like that it points to where it is designed to come from

webapp/src/components/modals/update_run_status_modal.tsx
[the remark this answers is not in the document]
calebroseland: renamed for clarity

webapp/src/client.ts
```diff
@@ -66,6 +66,41 @@ export async function fetchPlaybookRun(id: string) {
     return data as PlaybookRun;
 }
 
+export async function postStatusUpdate(
+    playbookRunId: string,
+    payload: {
+        message: string,
+        reminder?: number},
```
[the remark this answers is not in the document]
calebroseland: Indeed

webapp/src/slash_command.ts
calebroseland: Done

webapp/src/actions.ts
```diff
@@ -89,9 +93,23 @@ export function startPlaybookRunById(teamId: string, playbookId: string) {
...
+    channelId: PlaybookRun['channel_id'],
+    defaultStatus?: PlaybookRunStatus,
+) {
     return async (dispatch: Dispatch<AnyAction>, getState: GetStateFunc) => {
-        await clientExecuteCommand(dispatch, getState, `/playbook update ${defaultStatus ?? ''}`, teamId);
+        const state = getState();
+        const experimentalFeaturesEnabled = selectExperimentalFeatures(state);
+        const hasPermission = canIPostUpdateForRun(state, channelId, teamId);
+
+        if (experimentalFeaturesEnabled) {
```
lieut-data, line 108: Is the plan to ship with this flag for v6.0?

webapp/src/actions.ts
```diff
@@ -89,9 +93,23 @@ export function startPlaybookRunById(teamId: string, playbookId: string) {
     };
 }
 
-export function updateStatus(teamId: string, defaultStatus?: string) {
+export function promptUpdateStatus(
+    teamId: string,
+    playbookRunId: PlaybookRun['id'],
```
lieut-data, line 98: nit: Assuming I parse this correctly, this is cool, but feels verbose. Is it valuable to scope this to the same type as the `id` property on `PlaybookRun`, or is it clearer just to write `string`?

webapp/src/components/modals/update_run_status_modal.tsx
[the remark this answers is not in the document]
lieut-data: Commented above, but this feels really verbose. If all these parameters are truly typed to the `PlaybookRun`, it also suggest that maybe we should just pass the `playbookRun` object directly?

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -0,0 +1,154 @@
...
+    hasPermission: boolean;
+};
+
+export function makeModalDefinition(props: Props) {
+    return {
+        modalId: ID,
+        dialogType: UpdateRunStatusModal,
+        dialogProps: props,
+    };
+}
```
lieut-data, line 44: nit: worth writing as an arrow function with implicit return?

webapp/src/components/widgets/generic_modal.tsx
```diff
@@ -6,13 +6,15 @@ import classNames from 'classnames';
...
+
 type Props = {
     className?: string;
     onHide: () => void;
     modalHeaderText: React.ReactNode;
-    show: boolean;
-    handleCancel?: () => void;
-    handleConfirm?: () => void;
+    show?: boolean;
+    handleCancel?: (() => void) | null;
```
lieut-data, line 16: For clarity, why is `| null` required if `handleCancel` is already optional?

lieut-data (approved):
Looks good -- just a handful of comments :)

webapp/src/actions.ts
calebroseland: No, will get removed as soon as https://mattermost.atlassian.net/browse/MM-37637 goes in

webapp/src/components/widgets/generic_modal.tsx
calebroseland: For usage over in run update modal
```tsx
handleConfirm={hasPermission ? onConfirm : null}
```
It allows passing an explicit `null` to disable the confirm button. It would otherwise necessitate doing a lot of round about work to not provide the prop at all.

webapp/src/actions.ts
calebroseland: It is verbose, but I think it clarifies where the value is designed to come from. If that's not valuable at this point in time I guess we could just use `string`.

calebroseland: In line with `PlaybookRun -> Run` verbiage, we could perhaps do something like 

```ts
export function promptUpdateStatus(
    id: Run['id'],
    teamId: string,
    playbookId: Run['playbook_id'],
    channelId: Run['channel_id'],
) {//...}
```

webapp/src/components/rhs/rhs_post_update.tsx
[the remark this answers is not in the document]
agarciamontoro: Oh, cool, thanks :)

webapp/src/components/widgets/generic_modal.tsx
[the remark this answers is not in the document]
agarciamontoro: I'm ok leaving it as is for now, but maybe we should diverge from `openModal`?

webapp/src/components/modals/update_run_status_modal.tsx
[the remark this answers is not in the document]
agarciamontoro: Yay, thanks!

webapp/src/components/modals/update_run_status_modal.tsx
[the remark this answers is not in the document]
agarciamontoro: But my question is: as we are already destructuring all fields from the `Props` type, wouldn't `props` (now `modalProps`) be always empty?

webapp/src/components/markdown_textbox.tsx
agarciamontoro: > I'm in a bit of a bind re: TS; I wanted to do something with ComponentProps<typeof Textbox>, but we don't have the infrastructure in place to import types from mattermost-webapp, so it's technically just any here.

This would be really nice.

Thanks for the renaming, it's more clear now!

agarciamontoro (approved):
Looks great, @calebroseland, thank you!

webapp/src/components/widgets/generic_modal.tsx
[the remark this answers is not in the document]
calebroseland: Possibly, though given the prompt can be requested from multiple places, I wanted to avoid the complexities of mounting manually wherever it might be invoked, but we could probably make a lighter in-house alternative.

webapp/src/components/modals/update_run_status_modal.tsx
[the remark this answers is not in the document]
calebroseland: Switched back to string, queued an item for discuission--there's multiple patterns for typing IDs.

webapp/src/components/widgets/generic_modal.tsx
lieut-data: Would it make sense to `: undefined` here instead?

webapp/src/components/modals/update_run_status_modal.tsx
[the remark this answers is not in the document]
calebroseland: Sorry for not addressing that… that would be true if the type signature was correct, I corrected it now so that it reads correctly.

webapp/src/components/widgets/generic_modal.tsx
calebroseland: Yes, but I wanted to avoid the ESlint comment for now. We can switch it after allowing undefined

PR #703, opened 2021-08-20
MM-37777 Unify runs lists

crspeller:
#### Summary
Unifies the two runs lists into one run list. 
Plenty of cleanup was done, but not all possible cleanup as this PR is already big enough. 

#### Review note:
Keep in mind that a lot of this code is copy pasted from the existing runs lists. There are certainly parts that could use some refactoring. 

#### Screenshots
All runs page:
image
Playbook specific:
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37777

#### Checklist
~~- [ ] Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/backstage/runs_list/row.tsx
```diff
@@ -47,7 +58,23 @@ const SmallStatusBadge = styled(StatusBadge)`
...
+interface Props {
+    playbookRun: PlaybookRun
+    fixedTeam?: boolean
+}
+
+const teamNameSelector = (teamId: string) => (state: GlobalState): string => getTeam(state, teamId).display_name;
+
+const Row = (props: Props) => {
+    // This is not optimal. One network request for every row.
+    const playbookName = usePlaybookName(props.fixedTeam ? '' : props.playbookRun.playbook_id);
```
crspeller, line 76: This is needed for the playbook name for each row. Not optimal, but it's probably OK for now since it's limited to the 15 requests for each page. Possibly we should pass down some extra info with the runs list form the server to fix this. 

Or make a graphql endpoint. That would be cool too.

calebroseland: > Possibly we should pass down some extra info with the runs list form the server to fix this.

My gut feel thinks might be worth it in this case, certainly not a blocker, but I think it would be good to do. cc @lieut-data

lieut-data: :+1: for adding `playbook_name` as computed metadata to the response payload. GraphQL would indeed be interesting... but complicated :p

cpoile: This could also be fixed by keeping the source of truth for playbooks in redux -- I think we discussed this awhile ago, but I can't find a ticket for it.

PR #704, opened 2021-08-20
MM-37776 Use query for filtering in run lists. 

crspeller:
#### Summary
Use query for filtering as well as fixing the status drop down so it syncs with query parameters. 

This allows flows like:
- Filter down to the runs you want.
- Click on one and view overview
- Click back button
- Back to the filtered list!

As well as sharing links to filtered lists. 

It works for both the main run list and the one for playbooks. 

Example query string:
```
http://localhost:9005/playbooks/runs?page=0&per_page=15&sort=last_status_update_at&direction=asc&statuses=InProgress&statuses=Finished&owner_user_id=mbbpisxh4igi7y3hyqupc8w4eh&team_id=wyqetf4o9jb73c8jx4ns31wchc
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37791

#### Checklist
~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


webapp/src/components/backstage/runs_list/status_filter.tsx
```diff
@@ -7,62 +7,54 @@ import {MultiSelect, SelectOption} from 'src/components/multi_select';
...
+    {value: '', display: 'All', selected: false, disabled: false},
+    {value: 'divider', display: '', selected: false, disabled: false},
+    {value: 'InProgress', display: 'In Progress', selected: false, disabled: false},
+    {value: 'Finished', display: 'Finished', selected: false, disabled: false},
 ];
 
-interface StatusOption {
-    value: string;
-    label: string;
+function statusesToOptions(statuses: string[]) {
```
crspeller, line 21: I really just forced this to behave the way I wanted. It's not pretty, but I think we are probably going to change this control soon anyway as it's kinda awkward for only the two statuses.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
```diff
@@ -213,7 +214,7 @@ const PlaybookRunBackstage = () => {
     }
 
     const closePlaybookRunDetails = () => {
-        navigateToPluginUrl('/runs');
+        history.goBack();
```
lieut-data, line 217: Does this make sense if I permalink to this page?

lieut-data (approved):
Wow, nice! That was easier than I expected -- one comment below, plus the eternal inquiry about e2e tests :p

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
crspeller: Really I don't have any clue what that back button means. This preserves the behavior of being able to go back to a filtered run list. 
I tried to see if there is a way to tell if you came from a direct link. There is not reliable way because of security. 
Really 0/5 on what happens here.

iomodo (approved):

webapp/src/hooks/general.ts
```diff
@@ -476,6 +491,11 @@ export function useRunsList(defaultFetchParams: FetchPlaybookRunsParams):
         };
     }, [fetchParams]);
 
+    // Update the query string when the fetchParams change
+    useEffect(() => {
+        history.replace({search: qs.stringify(fetchParams, {addQueryPrefix: false, indices: false})});
```
lieut-data, line 496: Trying this out, and I'm not sure `replace` is the right primitive to use here. It sort of makes sense when filtering, but when I browse to page 2 and then click back, it's very disorienting to be thrown somewhere else completely, only to click forward and end up back on page 2.

I think we should probably just stack the navigation changes normally instead. For better or worse, it's what Jira does, and seems less surprising.

lieut-data (changes requested):
Requesting changes re: history browsing, but going to effect them myself.

lieut-data:
Actually, this is proving trickier, since there's no apparent sync between history changes and the params. We'd need to listen to `location` changes as well for this to happen, but then it's awkward having two sources of truth. Maybe the query string should be authoritative?

I'll push the merge conflicts for now, but leaving the requested changes unresolved for now.

webapp/src/hooks/general.ts
crspeller: Totally preference here I think because I really disliked the back button controlling the paging and the filtering when I tried it out. It makes way more sense to me for that to not be controlled by history. This is an app, not a webpage. If this where android you wouldn't expect the back button to undo a filter.

lieut-data: Right, but the back button is present -- intentionally so, even, on desktop -- and I think the experience using it is worse with this feature than without. If we choose not to integrate with the browser's history navigation, the only benefit here is the ability to permalink to a particular view, and we should probably expose that in some other way instead (i.e. explicit "copy link to this page").

Let's raise with @itao and @abhijit-singh? Maybe we can spin up a cloud test server with this change installed for experimentation?

lieut-data:
@crspeller, can you sync with @itao on the desired user experience here / tradeoffs?

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
itao: To confirm, is it this back button?

Screen Shot 2021-10-05 at 9 30 28 AM

webapp/src/hooks/general.ts
itao: It'd be really helpful to click around as a user 🙏

crspeller:
@lieut-data Solved the awkwardness by having an infinite scroll list. Now there is no page / per_page in the query parameters.

webapp/src/components/backstage/runs_list/runs_list.tsx
```diff
@@ -29,6 +32,24 @@ const PlaybookRunList = styled.div`
...
 `;
 
+const Footer = styled.div`
+    margin: 10px 0 20px;
+    font-size: 14px;
+`;
+
+const Count = styled.div`
+    padding-top: 8px;
+	width: 100%;
```
lieut-data, line 42: nit: indentation here seems off (and in next component)?

lieut-data (approved):
:tada: -- is there a good way to e2e test this behaviour?

webapp/src/components/backstage/runs_list/runs_list.tsx
crspeller: It was. My vim has been doing some strange things with tabs as of late.

PR #705, opened 2021-08-21
Mm 37191 omit run deleted

Duaard:
#### Summary

Users can still Run deleted playbooks by going to previous playbook runs and clicking on the playbook. This PR aims to fix this by hiding away the Run button if the playbook is already deleted.

See [issue](https://mattermost.atlassian.net/browse/MM-37191) description for more details.

#### Ticket Link

Fixes: mattermost/mattermost-server/issues/18036
JIRA: https://mattermost.atlassian.net/browse/MM-37191

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/playbooks/playbook_backstage.tsx
+++ b/webapp/src/components/backstage/playbooks/playbook_backstage.tsx
@@ -128,20 +128,7 @@ const PlaybookBackstage = () => {
         subTitle = 'Everyone in this team can access this playbook';
     }
 
-    const RunButton = () => {
-        if (playbook?.delete_at !== 0) {
-            return null;
-        }
-        return (
-            <PrimaryButtonLarger onClick={runPlaybook}>
-                <RightMarginedIcon
-                    path={mdiClipboardPlayOutline}
-                    size={1.25}
-                />
-                {'Run'}
-            </PrimaryButtonLarger>
-        );
-    };
+    const enableRunPlaybook = playbook?.delete_at === 0;
 
     return (
         <OuterContainer>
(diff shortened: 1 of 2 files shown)


webapp/src/components/backstage/playbooks/playbook_backstage.tsx
```diff
@@ -128,6 +128,21 @@ const PlaybookBackstage = () => {
...
+            <PrimaryButtonLarger onClick={runPlaybook}>
+                <RightMarginedIcon
+                    path={mdiClipboardPlayOutline}
+                    size={1.25}
+                />
+                {'Run'}
+            </PrimaryButtonLarger>
+        );
+    };
+
```
Duaard, line 145: This could also be handled by an inline conditional rendering in the previous PrimaryButton location, but IMO this approach is more readable and cleaner.

Duaard:
I am not sure how to write Unit Tests for Front End components as I was not able to find current tests that I can use as an example. Appreciate it if you can point me in the right direction for this one. 😄

agarciamontoro:
Sorry for the late response on this, @Duaard, this PR fell off my radar!

> I am not sure how to write Unit Tests for Front End components as I was not able to find current tests that I can use as an example. Appreciate it if you can point me in the right direction for this one.

We have historically leaned towards end to end testing to validate the frontend changes. The few unit tests we have are in the files ending with `test.ts`, as in `webapp/src/license.test.ts`.

If you want to write an E2E test with Cypress, check the `test-e2e/` directory, you'll find plenty of examples. For running those, navigate to that directory, run `npm i` (you only need to do this once), and then `npm run cypress:open`. That will open Cypress, and you'll be able to run one of the multiple files we have.

webapp/src/components/backstage/playbooks/playbook_backstage.tsx
agarciamontoro: 100% agree with you, this is much easier to read.

webapp/src/components/backstage/playbooks/playbook_backstage.tsx
```diff
@@ -128,6 +128,21 @@ const PlaybookBackstage = () => {
         subTitle = 'Everyone in this team can access this playbook';
     }
 
+    const RunButton = () => {
```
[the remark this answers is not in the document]
Duaard: Hi @agarciamontoro modified my implementation on this basing on how other buttons were being disabled like the Add Comment button on empty text box. I think this one is much cleaner. :)

Duaard:
Hi @cpoile, just pinging in case you missed this. 😄

cpoile (approved):
Nice work @Duaard, thank you!

PR #715, opened 2021-08-27
Exclude err shadowing

agarciamontoro:
#### Summary

This is a proposal to exclude variable shadowing from the linter, but only with `err` variables. This error has not helped me a single time, only annoyed me. 

I added the exclude rule and a sample of what it lets us do (avoid `err2`, basically). I stole the idea from https://github.com/golangci/golangci-lint/issues/166#issuecomment-617517087.

What do you think?

#### Ticket Link
--

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):
Yes this is always causing me to do err2 err3 which is hideous.

cpoile:
Yes / no on this? I'm fine either way I guess. 🤷

cpoile (approved):

iomodo (approved):

lieut-data:
I'm willing to give it a go :)

calebroseland (approved):

agarciamontoro:
Merging this now, as it does not affect the code at all (I had to remove the sample I did when merging master here). The failing E2E test is not caused by this (it's failing everywhere).

PR #720, opened 2021-09-01
MM-38188 Adding missing primary keys

crspeller:
#### Summary
Adds missing primary keys to IR_PlaybookMember, IR_StatusPosts, IR_TimelineEvent, IR_ViewedChannel
Esentally does:
```sql
ALTER TABLE IR_PlaybookMember ADD PRIMARY KEY (MemberID, PlaybookID);
ALTER TABLE IR_StatusPosts ADD PRIMARY KEY (IncidentID, PostID);
ALTER TABLE IR_TimelineEvent ADD PRIMARY KEY (ID);
ALTER TABLE IR_ViewedChannel ADD PRIMARY KEY (ChannelID, UserID);
```
with more steps.

Added a test so we don't make this mistake again.

#### Ticket Link
http://mattermost.atlassian.net/browse/MM-38188

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/sqlstore/migrations.go
+++ b/server/sqlstore/migrations.go
@@ -1188,4 +1188,39 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.29.0"),
+		toVersion:   semver.MustParse("0.30.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DatabaseDriverMysql {
+				if err := addPrimaryKey(e, sqlStore, "IR_PlaybookMember", "(MemberID, PlaybookID)"); err != nil {
+					return err
+				}
+				if err := addPrimaryKey(e, sqlStore, "IR_StatusPosts", "(IncidentID, PostID)"); err != nil {
+					return err
+				}
+				if err := addPrimaryKey(e, sqlStore, "IR_TimelineEvent", "(ID)"); err != nil {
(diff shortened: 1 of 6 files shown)


lieut-data (approved):
thank you – – especially for the unit test!￼

agarciamontoro (commented):
Really cool!

I have a question: why are we modifying the original creation of the tables, and not simply adding the migration? I'm concerned that this somehow fake our tests on the migrations, as they are effectively exercising a different scenario than what the real databases have. For example, we're not testing the `ALTER TABLE` statements from the `addPrimaryKey`, if I understood the code correctly.

The new test is great, by the way :)

agarciamontoro:
@crspeller 

>  I mostly did it because I don't like leaving incorrect code around for people to copy paste mistakenly.

That's a good argument, but the table creation code is still incorrect, as we modify the tables in every migration. If we change the original code of the table creation, now we have a mix of correct and incorrect code, and I'd say that's worse than having an incorrect original code that is fixed through all the migrations.

In general, I understand the database schema as the application of every migration in the file, so it is always correct (as far as we know, of course, every new bug is fixed through a new migration).

I'm 3/5 about keeping the original code. Another solution would be to just squash all migrations and write an actual correct schema creation.

Thoughts?

crspeller:
@agarciamontoro Reverted the changes to the original table creations.

agarciamontoro (approved):
Cool, thanks for this!

PR #722, opened 2021-09-03
Tooling: i18n ICU MessageFormat syntax checking

calebroseland:
#### Summary

- Adds syntax checking for `<FormattedMessage/>`, `intl.formatMessage`, and so on. Catches syntax errors at buildtime instead of runtime. (\~‾⌣‾ )~ 
- Adds ESLint plugin, but no rules added yet.
- Adds `defaultMessage` -> AST precompilation (AOT instead of JIT)

Example error

CleanShot 2021-09-02 at 20 39 13@2x

#### Ticket Link
none

#### Checklist

~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


agarciamontoro (approved):
Oh, this looks super useful, thanks! :)

crspeller (approved):

PR #723, opened 2021-09-03
MM-38277: Fix broadcast channel description

calebroseland:
#### Summary
- Fixes broadcast channel description fragment in new Post update dialog to support new data type.
- Adds `<Tooltip/>` component/widget with built-in OverlayTrigger to make working with tooltips easier.

CleanShot 2021-09-02 at 20 32 18@2x

#### Ticket Link
- Fixes https://mattermost.atlassian.net/browse/MM-38277

#### Checklist

~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -73,29 +83,27 @@ const UpdateRunStatusModal = ({
...
                     ),
+                    ChannelsTooltip: (...chunks) => (
+                        <Tooltip
+                            id={`${ID}_broadcast_tooltip`}
+                            content={broadcastChannelNames}
+                        >
+                            <span tabIndex={0}>{chunks}</span>
+                        </Tooltip>
+                    ),
+                    hasBroadcast: Boolean(playbook?.broadcast_channel_ids?.length).toString(),
```
agarciamontoro, line 105: Why do we need to convert this to a string? (I don't actually know how the `formatMessage` works here)

webapp/src/components/widgets/tooltip.tsx
```diff
@@ -0,0 +1,48 @@
...
+const Tooltip = ({
+    id,
+    content,
+    children,
+    placement = 'top',
+    className = 'hidden-xs',
+    delay = OVERLAY_DELAY,
+    style,
+    ...props
+}: Props & Omit<ComponentProps<typeof OverlayTrigger>, 'overlay'>) => {
```
agarciamontoro, line 26: Nit: Does it make sense to define `type Props` as this, instead of defining `Props` and then adding the props of `OverlayTrigger`?

agarciamontoro (approved):
Looks good! I left a couple of nits/questions, but non-blocking at all.

Thanks for the Tooltip component, it makes total sense! :)

crspeller (approved):

agarciamontoro:
Merging this already, I want to use the Tooltip component in my branch :)

PR #724, opened 2021-09-05
MM-37637, MM-38257: Post update reminder timer and revamped FormattedDuration

calebroseland:
#### Summary

- MM-37637: Add Reminder timer to Post update dialog
  - default from last update (any: none, static, custom timer), else it will default to playbook-specified value, if applicable
    - playbook-supplied custom durations fully supported and sorted properly in static options list
  - can specify custom reminder timer using a mix of natural language or traditiona/ISO inputs: `in 2 days`, `23h9m`, `next monday at 8 pm`, `2021-12-01 14:30`, `12 may 2022 at 19:32`,  etc.
- New post update dialog made default, no longer experimental
  - E2E fixed and updated for new functionality
- Post update dialog _message_ now defaults to last undeleted status post, else the md template from the playbook.
- [MM-38257](https://mattermost.atlassian.net/browse/MM-38257): `<Duration/>` revamped into `<FormattedDuration/>`
  - Used in Reminder Timer 
  - Fix: add support for `years` (`2y 6d 3h 1m`). Could be further extended to support truncating, rounding to simplify.
  - Add support for `style='long'` which prints out full units (`2 years 6 days 3 hours 1 minute`) with manual formatting and pluralization. Once [Intl.UnitFormat](https://github.com/tc39/proposal-intl-unit-format) reaches maturity, Luxon may implement and could be utilized for i18n in the future. Else,
===
For _fixed-time_ i18n-enabled relative time (`in ...`, `... ago`), use` <Timestamp/> `or `Luxon.DateTime.toRelative`(see https://moment.github.io/luxon/api-docs/index.html#datetimetorelative)
- Added dependencies, see [size comparison](https://bundlephobia.com/scan-results?packages=chrono-node@2.3.1,luxon@2.0.2,moment@2.29.1,parse-duration@1.0.0)
  - `luxon` - natural successor to Moment (IMHO and it lives in the [moment org](https://github.com/moment/luxon/blob/master/docs/why.md)), it's immutable with a simpler API and built-in i18n via `Intl.` formatters.
  - `chrono-node` - parses fixed dates. i18n possible, but heavy implementation-wise
  - `parse-duration` - parses duration. i18n possible, simpler than chrono-node i18n

####  Commits
While this is a sizable PR, don't be too alarmed by the overall diff ... `+805, −17` is from `tests-e2e/package-lock`.

- Main feature work: https://github.com/mattermost/mattermost-plugin-playbooks/pull/724/files/1f6916483ef88f888e0175acf3eb3054b30a41f8
- E2E spec updates https://github.com/mattermost/mattermost-plugin-playbooks/pull/724/commits/4920fc98b96e69ef6a5bede33f12e807bd479bd6
- E2E add/fix cypress types https://github.com/mattermost/mattermost-plugin-playbooks/pull/724/commits/4920fc98b96e69ef6a5bede33f12e807bd479bd6
- After: merge updates/fixes
 

#### Screenshots

- Timer
image

- FormattedDuration
image

https://user-images.githubusercontent.com/11724372/132116963-c59784ab-e3f0-42d3-8a8a-d937159dfee7.mp4

- E2E add/fix cypress types (inline docs and intellisense)
CleanShot 2021-09-13 at 21 36 00@2x
CleanShot 2021-09-13 at 21 36 33@2x

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


webapp/package.json
```diff
@@ -80,11 +81,14 @@
   "dependencies": {
     "@mdi/js": "^5.9.55",
     "@mdi/react": "^1.5.0",
+    "chrono-node": "2.3.1",
     "core-js": "3.15.2",
     "debounce": "1.2.1",
+    "luxon": "2.0.2",
```
lieut-data, line 87: Separate from this PR, should we consider getting this added to mattermost-webapp and just mark it as an external dependency?

webapp/src/types/playbook_run.ts
```diff
@@ -21,6 +21,8 @@ export interface PlaybookRun {
     current_status: PlaybookRunStatus;
     last_status_update_at: number;
     reminder_post_id: string;
+
+    /** Previous reminder timer as nanoseconds */
```
lieut-data, line 25: Are we actually storing nanoseconds? Does that run into any JavaScript precision issues?

webapp/src/utils.ts
```diff
@@ -31,3 +31,12 @@ export const findLastUpdatedWithDefault = (playbookRun: PlaybookRun) => {
...
     return lastUpdated > 0 ? lastUpdated : playbookRun.create_at;
 };
+
+export const roundToNearest = (n: number, multiple: number) => {
+    if (n > 0) {
+        return Math.ceil(n / multiple) * multiple;
+    } else if (n < 0) {
+        return Math.floor(n / multiple) * multiple;
+    }
+    return n;
```
lieut-data, line 41: nit: would it be sufficient to just handle `0` in either of the other blocks?

webapp/src/components/datetime_input.tsx
```diff
@@ -0,0 +1,250 @@
...
+    .datetime-autocomplete__option--is-selected {
+        background-color: var(--center-channel-color-08);
+    }
+
+    .datetime-autocomplete__option--is-focused {
+        background-color: var(--center-channel-color-16);
+    }
+
+    .datetime-autocomplete__control {
+        -webkit-transition: all 0.15s ease;
```
lieut-data, line 41: Are we required to do the `-webkit` stuff? I had been imagining something in the pipeline would take care of this based on our configured target browsers (but maybe that's a pipedream :p).

webapp/src/components/datetime_input.tsx
```diff
@@ -0,0 +1,250 @@
...
+        const duration = durationFromQuery(query);
+
+        if (duration?.isValid) {
+            return DateTime.now().plus(duration);
+        }
+    }
+    return (date && DateTime.fromJSDate(date)) || null;
+};
+
+export function infer(query: string, mode: Mode.DateTimeValue): DateTime | null;
```
lieut-data, line 96: Are the three `export function infer` intended here?

webapp/src/components/formatted_duration.tsx
```diff
@@ -0,0 +1,73 @@
...
+    ago?: boolean;
+    style?: FormatStyle;
+}
+
+const label = (num: number, style: FormatStyle, narrow: string, singular: string, plural: string) => {
+    if (style === 'narrow') {
+        return narrow;
+    }
+
+    return num >= 2 ? plural : singular;
```
lieut-data, line 29: Not blocking, but I'll be interested to see how we transition this re: i18n, where I believe some languages have more complexity than just singular/plural.

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -51,12 +64,34 @@ const UpdateRunStatusModal = ({
     const {formatMessage} = useIntl();
     const [message, setMessage] = useState<string | null>(null);
     const playbook = usePlaybook(playbookId);
-    if (playbook && message == null) {
-        setMessage(playbook.reminder_message_template);
+    const run = useRun(playbookRunId);
+    const lastStatusPostMeta = run?.status_posts?.slice().reverse().find(({delete_at}) => !delete_at);
+    const lastStatusPost = usePost(lastStatusPostMeta?.id ?? '');
+    if (
```
lieut-data, line 70: This block feels brittle, but apart from moving to explicit state management (ala Redux), I don't have any suggestions :)

lieut-data (approved):
A handful of comments, but nothing strictly blocking -- awesome to see some e2e typing improvements too!

webapp/src/types/playbook_run.ts
calebroseland: It's a `time.Duration` in the struct. Realistically it's not an issue as we drop down to more common units before processing any official dates. We only use this as the last reminder interval—so unless the last reminder was ~290+ years prior, I think it should be fine. 😛

webapp/src/components/datetime_input.tsx
calebroseland: Yes, they provide the proper return types based on the `mode` provided.

webapp/src/components/formatted_duration.tsx
calebroseland: For this first revision, I didn't want to go too far past what was needed—this will indeed need to be replaced with something. I'm not directly aware of an API that does full-precision duration printing, and Intl.UnitFormat appears to be a decent ways out, so various formatted messags with a `{plural, =0, =1, =2, many {...}}` will probably be needed in the i18n epic.

webapp/src/components/modals/update_run_status_modal.tsx
calebroseland: I did not enjoy writing this one in particular 🙃. I think breaking this out to a custom hook at a minimum would be at least for now—we could then reinforce it more easily there.

calebroseland: Wrote `useDefaultMessage` below—should be much cleaner now.

webapp/src/utils.ts
calebroseland: simplified

webapp/package.json
calebroseland: Yes, that would be good to do! Opened a ticket for that, and an epic to contain smaller efforts. https://mattermost.atlassian.net/browse/MM-38572

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -142,6 +163,97 @@ const UpdateRunStatusModal = ({
...
+        // last status exist and should have a post-message
+        return lastStatusPost?.message;
+    }
+    if (run && !lastStatusPostMeta) {
+        // run loaded and was no last status post, but there might be a message template
+        return playbook?.reminder_message_template;
+    }
+
+    return null;
+};
```
cpoile, line 183: I thought we had this in a hook in the old hooks file? or was that deleted?

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -142,6 +163,97 @@ const UpdateRunStatusModal = ({
...
+const optionFromSeconds = (seconds: number) => {
+    const duration = Duration.fromObject({seconds});
+
+    return {
+        label: `in ${formatDuration(duration, 'long')}`,
+        value: duration,
+    };
+};
+
+export const useReminderTimer = (
```
cpoile, line 194: nit: could this be named better? If it returns the input and the reminder, maybe a different title would be more descriptive?  (I'm mentioning because I had to reread it a few times and then realized what it was doing, and the title may have thrown me off a bit)

webapp/src/hooks/crud.ts
```diff
@@ -110,3 +111,15 @@ export function usePlaybooksCrud(
...
+ * Read-only logic to fetch playbook run
+ */
+export function useRun(id: PlaybookRun['id']) {
+    const [run, setRun] = useState<PlaybookRun | undefined>();
+    useEffect(() => {
+        fetchPlaybookRun(id).then(setRun);
+    }, [id]);
+
+    return run;
+}
```
cpoile, line 125: shouldn't this be in the store already?

cpoile (changes requested):
Just minor questions, nothing serious. Very nice!

webapp/src/components/datetime_input.tsx
```diff
@@ -0,0 +1,246 @@
...
+        datetimeResults: DateTime[],
+        durationResults: Duration[],
+    ) => Option[] | null;
+} & Partial<ComponentProps<typeof StyledSelect>>;
+
+const DateTimeInput = ({
+    mode = Mode.DateTimeValue,
+    value,
+    defaultOptions,
+    makeOptions = (query, datetimes, durations) => {
```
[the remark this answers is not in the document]
calebroseland: Yeah, looks like its straightforward to do in this case.

webapp/src/components/modals/update_run_status_modal.tsx
calebroseland: `useReminderTimerSetting`? I agree something more specific would be more clear

webapp/src/components/modals/update_run_status_modal.tsx
calebroseland: I couldn't find one, but if one exists I would switch it. I skimmed some of this logic from server-side logic that composes the interactive dialog.

webapp/src/hooks/crud.ts
calebroseland: Indeed it is—I started down a line of fixing this, but got dragged into a rabbit trail of fixing types associated with `myPlaybookRunsByTeam`. I propose to leave this as-is for this PR and I'll open those changes as a follow-up PR.

cpoile: Cool, should we make a ticket to track it?

cpoile (approved):
Awesome, thanks @calebroseland !

PR #736, opened 2021-09-13
MM-38460 - Fix for: Delete entry confirmation dialog doesn't close

cpoile:
#### Summary
- The confirmation dialog would not close after confirming delete, so fixed that.
- That revealed a bigger problem: because we load the playbookRun in a parent component (it's not synced to the store), deleting the timeline event would not update the timeline. So we have to manually delete it at the level that the playbook is being loaded and stored. This means prop-drilling the delete function, which is not ideal. I think it's fine until we move to using the playbookRun in the store (which is a bigger change, out of scope for this bug fix PR).
- Did some refactoring cleanup.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38460

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data (approved):
Any chance of an e2e test to keep the modal working?

tests-e2e/cypress/integration/backstage/playbook_retro_timeline_spec.js
```diff
@@ -0,0 +1,114 @@
...
+            teamId: testTeam.id,
+            playbookId: testPublicPlaybook.id,
+            playbookRunName: runName,
+            ownerUserId: testUser.id,
+        }).then((run) => {
+            runId = run.id;
+        });
+    });
+
+    beforeEach(() => {
```
lieut-data, line 52: Should we collapse the two `beforeEach` blocks?

lieut-data (approved):

lieut-data:
Thanks for the e2e test, @cpoile :)

tests-e2e/cypress/integration/backstage/playbook_retro_timeline_spec.js
cpoile: 🤦 Thanks.

crspeller (approved):

PR #738, opened 2021-09-14
Fixing primary key test

crspeller:
#### Summary
Fixes the primary key test. 
The test was not incorrect. The addPrimaryKey migration's check for migration complete did not restrict itself to the selected DB. As a result it would not run if the migration had already been run on anouther DB. This fixes that error.


cpoile (approved):
that's awesome, learned stuff from this

calebroseland (approved):

PR #740, opened 2021-09-15
MM-38122 Update plugin ID

crspeller:
#### Summary
Updates the plugin id to `playbooks`.
Adds migration to kill off the old version of itself just like last time.

Server PR: https://github.com/mattermost/mattermost-server/pull/18392

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38122

#### Checklist
- [x] Telemetry updated
~~- [ ] Unit tests updated~~

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -1,5 +1,5 @@
 {
-    "id": "com.mattermost.plugin-incident-management",
+    "id": "playbooks",
     "name": "Playbooks",
     "description": "The playbooks plugin",
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-playbooks/",
--- a/plugin.json
+++ b/plugin.json
@@ -1,7 +1,7 @@
 {
     "id": "playbooks",
     "name": "Playbooks",
-    "description": "The playbooks plugin",
+    "description": "Mattermost Playbooks enable reliable and repeatable processes for your teams using checklists, automation, and retrospectives.",
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-playbooks/",
     "support_url": "https://github.com/mattermost/mattermost-plugin-playbooks/issues",
     "icon_path": "assets/plugin_icon.svg",
--- a/server/sqlstore/migrations.go
+++ b/server/sqlstore/migrations.go
@@ -1223,8 +1223,8 @@ var migrations = []Migration{
 		fromVersion: semver.MustParse("0.30.0"),
 		toVersion:   semver.MustParse("0.31.0"),
 		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
-			if _, err := e.Exec("UPDATE PluginKeyValueStore SET pluginid='playbooks' WHERE pluginid='com.mattermost.plugin-incident-management'"); err != nil {
-				return errors.Wrapf(err, "failed setting default value in column CategoryName of table IR_Playbook")
+			if _, err := e.Exec("UPDATE PluginKeyValueStore SET PluginId='playbooks' WHERE PluginId='com.mattermost.plugin-incident-management'"); err != nil {
+				return errors.Wrapf(err, "failed to migrate KV store plugin id")
 			}
 
 			return nil
--- a/server/sqlstore/migrations.go
+++ b/server/sqlstore/migrations.go
@@ -1219,4 +1219,15 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.30.0"),
+		toVersion:   semver.MustParse("0.31.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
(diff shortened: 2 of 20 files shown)


plugin.json
```diff
@@ -1,7 +1,7 @@
 {
```
lieut-data, line 1: While in here, should we bump `min_server_version` to `v6.0`?

tests-e2e/db-setup/mm534.sql
```diff
@@ -2759,7 +2759,7 @@ COPY public.outgoingwebhooks (id, token, createat, updateat, deleteat, creatorid
 --
 
 COPY public.pluginkeyvaluestore (pluginid, pkey, pvalue, expireat) FROM stdin;
```
lieut-data, line 2761: Are there any side effects from losing the data in the "old" key value store that we care about this time around?

lieut-data (approved):

plugin.json
crspeller: Let's do that separate. Plus I think community is still technically on 5.38 so we should probably wait for that update to happen.

tests-e2e/db-setup/mm534.sql
crspeller: No. We don't store anything in there anymore.

cpoile (approved):
noice

plugin.json
lieut-data: Actually, with community, I was thinking that hiding it would be a benefit given how it doesn't work well (i.e. no `X` to escape from the modal, etc.)

tests-e2e/db-setup/mm534.sql
lieut-data: What about the reminders implementation? Is that fully in the database now?

plugin.json
crspeller: Ok. Will make a PR then.

tests-e2e/db-setup/mm534.sql
crspeller: Hmm. You are right. My searching failed me because the implementation is hidden away in the plugin-api thing. 
We will loose current reminders for updates and retrospectives. 
Did we migrate the KV store last time? I don't think we did.

lieut-data: Indeed, I think we "didn't care" last time.

crspeller: How about we just do:
```sql
UPDATE PluginKeyValueStore SET pluginid='playbooks' WHERE pluginid='com.mattermost.plugin-incident-management'
```
We can do that now that we have DB access :)

lieut-data: Seems reasonable!

crspeller: Updated. Sorry for the rebase. The last commit is all that's changed.

server/sqlstore/migrations.go
```diff
@@ -1216,6 +1216,17 @@ var migrations = []Migration{
...
 
+			return nil
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.30.0"),
+		toVersion:   semver.MustParse("0.31.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if _, err := e.Exec("UPDATE PluginKeyValueStore SET pluginid='playbooks' WHERE pluginid='com.mattermost.plugin-incident-management'"); err != nil {
+				return errors.Wrapf(err, "failed setting default value in column CategoryName of table IR_Playbook")
```
lieut-data, line 1227: Error here is wrong?

server/sqlstore/migrations.go
```diff
@@ -1216,6 +1216,17 @@ var migrations = []Migration{
...
 			}
 
+			return nil
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.30.0"),
+		toVersion:   semver.MustParse("0.31.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if _, err := e.Exec("UPDATE PluginKeyValueStore SET pluginid='playbooks' WHERE pluginid='com.mattermost.plugin-incident-management'"); err != nil {
```
lieut-data, line 1226: The lowercase column name here might cause trouble for MySQL or Postgres, unless that's how's the server actually defined it? And given we're faking the schema for the test, I wonder if we'll run into issues?

server/sqlstore/migrations.go
crspeller: Fixed.

server/sqlstore/migrations.go
crspeller: That's what I get for copying and pasting. But it wouldn't have mattered, I don't think column names are case sensitive. It's table names where case matters, for whatever reason.

lieut-data (approved):

PR #751, opened 2021-09-21
[MM-37314] Sync sidebar category update

iomodo:
#### Summary
After some research I was able to reproduce the autocategorization bug locally, but only when I'm trying to autocategorize 2 channels at once. That is if I start two playbook runs. There we have the race condition on an HA environment. When there are 2 concurrent calls of `UserHasJoinedChannel`: 
```go 
GetSidebarCategories()
change_categories
UpdateSidebarCategories()
```
will not work, because what might happen is 

```go 
GetSidebarCategories() //for the first channel on node1
GetSidebarCategories() //for the second channel on node2
change_categories //first channel is updated on node1
change_categories //second channel is updated on node2
UpdateSidebarCategories()//first channel info is saved to DB from node1
UpdateSidebarCategories()//second channel info is saved to DB from node2
```
Hence the data will be overwritten and first channel will not be autocategorized.
Solution here is to have the distributed lock.

P.S. I was not able to reproduce the same for the single channel. Server code seems to be Ok as well.

#### Ticket Link

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	stripmd "github.com/writeas/go-strip-markdown"
 
+	"github.com/mattermost/mattermost-plugin-api/cluster"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/bot"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/config"
 	"github.com/mattermost/mattermost-server/v6/model"
@@ -39,6 +40,7 @@ type PlaybookRunServiceImpl struct {
 	scheduler     JobOnceScheduler
 	telemetry     PlaybookRunTelemetry
 	api           plugin.API
+	categoryMutex *cluster.Mutex
 }
 
 var allNonSpaceNonWordRegex = regexp.MustCompile(`[^\w\s]`)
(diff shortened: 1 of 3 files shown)


lieut-data (commented):
I guess this solves the case of when playbooks tries to manipulate the categories concurrently, but it doesn't help if the server is manipulating categories for other reasons, right?

I wonder if we should instead be fixing the app-layer APIs to effect more granular changes like this?

agarciamontoro (approved):
I'm ok merging this, but I agree with @lieut-data that we should still address the underlying issue with the API.

iomodo:
Yes I thought about that, but there is no easy solution here.
If we stick to the “state” model where we `get` and `update` the entire state of the sidebar categories we will need a distributed lock shared between webapp, server and plugins. There is no way around it.
If we change to the event-based model where we send only changes to the server. For example: `channel: X added to:categoryY` or `channel: X moved from: categoryY to: categoryZ` we will still need to verify the order somehow. Order of the changes should be the same on every node. We can do that with CRDTs.
I don’t see an easy way of solving this general problem. These solutions seem overkill. 

One ugly solution for this concrete problem we are having can be to wait until a new channel is categorized by the server. And only when we detect it in `some` category autocategorize it, hence making serial changes instead of concurrent ones.
Thoughts? @lieut-data @agarciamontoro

PR #754, opened 2021-09-21
MM-38744 - Fix for: Broadcast stops if one channel is read-only

cpoile:
#### Summary
- Don't stop broadcasting. Incidentally, this is how the broadcast start and finish messages are treated (which is why those seemed to keep working).
- Making a judgement call that it's not worth mocking unit tests to simulate this, and e2e tests would not be a valuable use of time given the amount of other bugs we need to finish and the edge-case-ness of this bug.
- I did reproduce on a cloud test server (using `--version release-6.0`) and the PR fixes it.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38744

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -693,7 +693,8 @@ func (s *PlaybookRunServiceImpl) broadcastStatusUpdate(post *model.Post, playboo
 		post.Id = "" // Reset the ID so we avoid cloning the whole object
 		post.ChannelId = channelID
 		if err := s.postMessageToThreadAndSaveRootID(playbookRunID, channelID, post); err != nil {
-			return errors.Wrap(err, "failed to post broadcast message")
+			s.pluginAPI.Log.Warn("failed to broadcast the status update to channel",
+				"channel_id", channelID, "error", err.Error())
 		}
 	}
 


agarciamontoro (approved):
Makes total sense, thanks!

server/app/playbook_run_service.go
```diff
@@ -693,7 +693,8 @@ func (s *PlaybookRunServiceImpl) broadcastStatusUpdate(post *model.Post, playboo
 		post.Id = "" // Reset the ID so we avoid cloning the whole object
 		post.ChannelId = channelID
 		if err := s.postMessageToThreadAndSaveRootID(playbookRunID, channelID, post); err != nil {
-			return errors.Wrap(err, "failed to post broadcast message")
+			s.pluginAPI.Log.Warn("failed to broadcast the status update to channel",
```
lieut-data, line 696: Given we always return `nil` here, should we remove the `err` return value and log on https://github.com/mattermost/mattermost-plugin-playbooks/blob/a3ef1d7ec459b90f01f74dce97b2b05622edccd9/server/app/playbook_run_service.go#L826-L828?

Another option might be to use something like https://github.com/hashicorp/go-multierror.

lieut-data (approved):

lieut-data:
@cpoile, heads up that the e2e failure here is broadcast related, but it doesn't seem likely to be related to your PR. Your call on when to merge :)

PR #755, opened 2021-09-21
MM-38533 Fixing themeing in markdown textboxes.

crspeller:
#### Summary
Simple CSS fix.

### DIFF
--- a/webapp/src/components/backstage/styles.tsx
+++ b/webapp/src/components/backstage/styles.tsx
@@ -76,7 +76,7 @@ export const StyledTextarea = styled.textarea`
 
 export const StyledMarkdownTextbox = styled(MarkdownTextbox)`
     .custom-textarea {
-		background-color: var(--center-channel-bg);
+        background-color: var(--center-channel-bg);
         border-radius: 4px;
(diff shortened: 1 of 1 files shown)


lieut-data (approved):
Looks good -- but whitespace is off?

crspeller:
@lieut-data Looks like a GitHub rendering issue.

webapp/src/components/backstage/styles.tsx
```diff
@@ -76,7 +76,7 @@ export const StyledTextarea = styled.textarea`
 
 export const StyledMarkdownTextbox = styled(MarkdownTextbox)`
     .custom-textarea {
-        background: white;
+		background-color: var(--center-channel-bg);
```
calebroseland, line 79: Looks like tabs
```suggestion
       background-color: var(--center-channel-bg);
```

calebroseland (approved):

webapp/src/components/backstage/styles.tsx
crspeller: Strange. How did that happen...

PR #759, opened 2021-09-22
MM-37699: improved finished run semantics

lieut-data:
#### Summary
Improved finish semantics:
* Add a Finished tag next to the run name. 
* Disable the Post update button. 
* Change the icon in the update box to a chequered flag 
* Change text to Run finished X minutes/hours/days ago. 
* Disable the checkbox for all the tasks (checked and unchecked)
* Remove/disable options to Add task, reorder tasks, add/edit assignee, edit task, delete task. 
* Hide the ability to “Run” any associated slash command
* Don't open the RHS by default for finished runs

CleanShot 2021-09-22 at 17 06 59@2x

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37699

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/webapp/src/components/checklist_item.tsx
+++ b/webapp/src/components/checklist_item.tsx
@@ -48,6 +48,7 @@ interface ChecklistItemDetailsProps {
     onRedirect?: () => void;
     draggableProvided: DraggableProvided;
     dragging: boolean;
+    disabled: boolean;
 }
 
 const RunningTimeout = 1000;
@@ -185,6 +186,10 @@ export const CheckboxContainer = styled.div`
 
     input[type="checkbox"]:checked::before {
         transform: scale(1) rotate(0deg);
+    }
+
+    input[type="checkbox"]:disabled {
+        opacity: 0.38;
     }
 
     label {
@@ -381,13 +386,15 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
                 data-testid='checkbox-item-container'
             >
                 <CheckboxContainer>
-                    {showMenu &&
+                    {showMenu && (!props.disabled || props.checklistItem.description !== '') &&
                     <HoverMenu>
-                        <HoverMenuButton
-                            title={'Drag me to reorder'}
-                            className={'icon icon-menu'}
-                            {...props.draggableProvided.dragHandleProps}
-                        />
+                        {!props.disabled &&
+                            <HoverMenuButton
+                                title={'Drag me to reorder'}
+                                className={'icon icon-menu'}
+                                {...props.draggableProvided.dragHandleProps}
+                            />
+                        }
                         {props.checklistItem.description !== '' &&
                         <StepDescription
                             text={props.checklistItem.description}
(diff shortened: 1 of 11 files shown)


webapp/src/components/rhs/rhs_post_update.tsx
```diff
@@ -34,21 +36,33 @@ const RHSPostUpdate = (props: Props) => {
...
 
     const timespec = (isDue || !isNextUpdateScheduled) ? PastTimeSpec : FutureTimeSpec;
 
+    let icon;
+    if (isFinished) {
+        icon = <Icon path={mdiFlagCheckered}/>;
+    } else if (isDue) {
+        icon = Exclamation;
+    } else {
+        icon = Clock;
```
agarciamontoro, line 56: The exclamation and clock icons are not visible now, we need to instantiate them. I don't know why Typescript is ok with the current code, by the way, shouldn't it complain?

```suggestion
        icon = <Exclamation/>;
    } else {
        icon = <Clock/>;
```

agarciamontoro (changes requested):
Looks great! There's just one comment we need to address.

Also, I wanted to bring up that the disabled button is almost unreadable
image

Should we make it a bit less transparent?

webapp/src/components/rhs/rhs_post_update.tsx
lieut-data: Ooh, good catch: didn't even notice that! I think TypeScript doesn't complain since I didn't type the variable, and rendering the function is "ok" but probably results in nothing? Fixed with typing:

CleanShot 2021-09-23 at 13 24 59@2x

lieut-data:
@agarciamontoro, can you share the theme on which you're seeing the really faded text?

agarciamontoro:
@lieut-data: Sure, it's the default, Denim.
image

lieut-data:
Thank you for that, @agarciamontoro! Missed that in testing -- looks like `TertiaryButton` needed some improved `:disable` love: fixed! The one failing CI test is unrelated to this PR, but should hopefully be fixed upstream soon.

agarciamontoro (approved):
Much better, thank you! :tada:

webapp/src/components/rhs/rhs_post_update.tsx
agarciamontoro: Cool!

webapp/src/components/checklist_item.tsx
```diff
@@ -381,58 +386,65 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
...
                             text={props.checklistItem.description}
                             channelNames={channelNamesMap}
                             team={team}
                         />
                         }
-                        <ProfileSelector
-                            selectedUserId={props.checklistItem.assignee_id}
-                            onlyPlaceholder={true}
-                            placeholder={
+                        {!props.disabled &&
```
cpoile, line 405: I haven't tested it, but if the entire <HoverMenu> is guarded with `!props.disabled`, do you need to put this guard here?

webapp/src/components/rhs/rhs_about_title.tsx
```diff
@@ -60,9 +72,14 @@ const RHSAboutTitle = (props: Props) => {
...
         return (
-            <RenderedTitle onClick={onRenderedTitleClick} >
-                {editedValue}
-            </RenderedTitle>
+            <TitleWrapper>
+                <RenderedTitle onClick={onRenderedTitleClick}>
+                    {editedValue}
+                </RenderedTitle>
+                {props.status === PlaybookRunStatus.Finished &&
+                    <StatusBadgeWrapper status={PlaybookRunStatus.Finished}/>
```
cpoile, line 80: 🙌

webapp/src/rhs_opener.ts
```diff
@@ -70,6 +72,11 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
             return;
         }
 
+        // Don't do anything if the playbook run is finished.
+        if (playbookRun && playbookRun.current_status === PlaybookRunStatus.Finished) {
+            return;
+        }
```
cpoile, line 78: I'm not sure if I like this -- I think I'd want the RHS to open: to remind me that this was an incident, so I know the playbooks product isn't broken, and so I can click on the `Run Details` link to quickly get to the overview/retro screens.

cpoile (approved):
Couple of comments, non-blocking. Thanks for fixing this @lieut-data!

webapp/src/rhs_opener.ts
lieut-data: Part of this comes from a direct ask from @coreyhulen, but also our general desire to begin a shift away from the tight coupling between channels and playbook runs. If a run could be standalone, or you could start multiple runs in a channel, do you think these semantics begin to make more sense?

webapp/src/components/checklist_item.tsx
lieut-data: There's two cases being handled: disabled and no description (hide the entire hover menu), and disabled but has a description (show the hover menu, but just the "(i)" icon). I suspect we can find better ways to surface the description in the first place, but I wanted to avoid the "empty" hover menu effect.

lieut-data:
(Merging, and we can continue discussing asynchronously?)

PR #760, opened 2021-09-23
MM-38775: Remove dependency on playbook in Update Status modal

agarciamontoro:
#### Summary

The bug (see the ticket) was caused by `playbook` being undefined: the request to fetch it failed because the user does not have permissions to open the playbook, so it was not set.

This affected the line below, in the call to `GenericModal`. The value of the `isConfirmDisabled` field was set to false because playbook was undefined.
```ts
isConfirmDisabled={!(hasPermission && message?.trim() && currentUserId && channelId && playbook?.team_id)}
```

This PR completely removes the dependency on the playbook in the Update Status modal: everything should be in the playbook run, as we cannot rely on the current information of a playbook, which has potentially been changed between the run creation and the time we fetch that information.

For doing that, I had to add the `ReminderTimerDefaultSeconds` field to the `PlaybookRun` model. Although `PreviousReminder` does have the same information when the run is created, it is changed as soon as an update is posted. Having this info lets us add it to the options list in the _Timer for next update_, as shown in the screenshot (`15 days` is not a hardcoded value in the list, but the one retrieved from the playbook)

image

Apart from that, I've removed every old mention to `broadcast_channel_id` (in singular). We still had a couple of them there.

I'm adding an E2E test while you guys review this :)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38775

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/api/api.yaml
+++ b/server/api/api.yaml
@@ -1443,10 +1443,13 @@ paths:
                     type: string
                     description: User ID of the playbook member.
                     example: ilh6s1j4yefbdhxhtlzt179i6m
-                broadcast_channel_id:
-                  description: The ID of the channel where all the status updates will be broadcasted. The team of the broadcast channel must be the same as the playbook's team.
-                  type: string
-                  example: 2zh7rpashwfwapwaqyslmhwbax
+                broadcast_channel_ids:
+                  description: The IDs of the channels where all the status updates will be broadcasted. The team of the broadcast channel must be the same as the playbook's team.
+                  type: array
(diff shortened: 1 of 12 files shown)


lieut-data (approved):
Nice!!

iomodo (approved):

agarciamontoro:
The failing E2E test is expected. Merging now :)

PR #761, opened 2021-09-23
[MM-37314] Fix the categorization bug

iomodo:
#### Summary
An ugly fix for the categorization bug.
Comment in the code is pretty self-explanatory.
When user starts a run, channel is created, and user added to the channel. At that moment server sends the `user_added` socket event to the webapp. Webapp gets the event and tries to categorize the channel for the user by adding the channel to the default `CHANNELS` category. 
Concurrently we are trying to auto-categorize the channel into our category. Depending on what happens first channel might get stuck in `CHANNELS` category.

Unfortunately we can not reliably detect when webapp finished updating the state, that's why we are waiting for 5 seconds(A very magic number) to let webapp do its job.

#### Ticket Link

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -1693,25 +1693,28 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
 	}
 
 	if playbookRun.CategoryName != "" {
-		err = s.createOrUpdatePlaybookRunSidebarCategory(userID, channelID, channel.TeamId, playbookRun.CategoryName)
-		if err != nil {
-			s.logger.Errorf("failed to categorize channel; error: %s", err.Error())
-		}
+		// Update sidebar category in the go-routine not to block the UserHasJoinedChannel hook
+		go func() {
+			// Wait for 5 seconds(a magic number) for the webapp to get the `user_added` event,
+			// finish channel categorization and update it's state in redux.
+			// Currently there is no way to detect when webapp finishes the job.
+			// After that we can update the categories safely.
+			// Technically if user starts multiple runs simultaneously we will still get the race condition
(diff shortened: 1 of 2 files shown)


agarciamontoro (approved):
I'm ok merging this if we want it for 1.20, but we should make sure to apply a proper fix as soon as we can. Can we maybe create a follow-up ticket for this and continue the investigation?

server/app/playbook_run_service.go
```diff
@@ -1703,6 +1703,15 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
...
 // it does not already exist and adds the channel within the sidebar category
 func (s *PlaybookRunServiceImpl) createOrUpdatePlaybookRunSidebarCategory(userID, channelID, teamID, categoryName string) error {
+	// Wait for 5 seconds(a magic number) for the webapp to get the `user_added` event,
+	// finish channel categorization and update it's state in redux.
+	// Currently there is no way to detect when webapp finishes the job.
+	// After that we can update the categories safely.
+	// Technically if user starts multiple runs simultaneously we will still get the race condition
+	// on category update. Since that's not realistic at the moment we are not adding the
+	// distributed lock here.
+	time.Sleep(5 * time.Second)
```
lieut-data, line 1713: Let's not actually block the completion of the hook here, though, and run instead spawn a goroutine that blocks for the required workaround time.

lieut-data (changes requested):
Nice investigation :D

One comment below.

lieut-data (approved):

lieut-data:
@iomodo, I /think/ the failures here are due to my requested changes re: making the categorization happen async: I guess the unit tests need to be updated?

lieut-data (approved):

PR #764, opened 2021-09-24
MM-38734 Fix unknown channels being removed when somone without premissions saves playbook.

crspeller:
#### Summary
Unknown channels ie channels you don't have permissions to where being removed incorrectly on save. This fixes that. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38734

### DIFF
--- a/server/api/playbooks.go
+++ b/server/api/playbooks.go
@@ -262,18 +262,6 @@ func doPlaybookModificationChecks(playbook *app.Playbook, userID string, pluginA
 		playbook.DefaultOwnerEnabled = false
 	}
 
-	filteredBroadcastChannelIDs := []string{}
-	for _, channelID := range playbook.BroadcastChannelIDs {
-		if !pluginAPI.User.HasPermissionToChannel(userID, channelID, model.PermissionCreatePost) {
-			pluginAPI.Log.Warn("broadcast channel is not valid, removing channel from list", "channel_id", channelID)
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


server/api/playbooks.go
```diff
@@ -262,18 +262,6 @@ func doPlaybookModificationChecks(playbook *app.Playbook, userID string, pluginA
 		playbook.DefaultOwnerEnabled = false
 	}
 
-	filteredBroadcastChannelIDs := []string{}
```
crspeller, line 265: This is actually implemented correctly in the permissions check: https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/app/permissions.go#L342 so the fix was just to delete the bad implementation. 

This area really needs a refactor. Especially after the app changes because the names are confusing. I will take care of it when I come in to do the changes for advanced permissions.

server/api/playbooks_test.go
```diff
@@ -1041,7 +1041,7 @@ func TestPlaybooks(t *testing.T) {
 		requireErrorWithStatusCode(t, err, http.StatusForbidden)
 	})
 
-	t.Run("update playbook but no premissions in broadcast channel, but no edit", func(t *testing.T) {
+	t.Run("update playbook but no premissions in broadcast channel, but it already exists", func(t *testing.T) {
```
crspeller, line 1044: When the implementation was messed up the test was changed too.

agarciamontoro (approved):
Perfect :tada: More code always means more bugs!

lieut-data (approved):

PR #765, opened 2021-09-24
MM-38251: workaround "Run" from Playbooks

lieut-data:
#### Summary
Navigating to a playbook and clicking "Run" sometimes doesn't work, even if Channels has already loaded, because something is forcing the websocket to disconnect and reconnect on the transition to Channels, losing the interactive dialog event. Workaround this with a timeout on dispatching the event in question.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-38251

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/playbooks/playbook_backstage.tsx
+++ b/webapp/src/components/backstage/playbooks/playbook_backstage.tsx
@@ -123,7 +123,7 @@ const PlaybookBackstage = () => {
 
         if (playbook?.id) {
             telemetryEventForPlaybook(playbook.id, 'playbook_dashboard_run_clicked');
-            dispatch(startPlaybookRunById(team.id, playbook.id));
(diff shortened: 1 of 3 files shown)


webapp/src/components/backstage/playbooks/playbook_backstage.tsx
```diff
@@ -123,7 +123,7 @@ const PlaybookBackstage = () => {
 
         if (playbook?.id) {
             telemetryEventForPlaybook(playbook.id, 'playbook_dashboard_run_clicked');
-            dispatch(startPlaybookRunById(team.id, playbook.id));
+            dispatch(startPlaybookRunById(team.id, playbook.id, 3000));
```
crspeller, line 126: Three seconds? We might as well remove the button. 🤦

crspeller (approved):

calebroseland (approved):

PR #766, opened 2021-09-24
MM-38731 - Fix for: Old status update dialog partially exposed in web/desktop

cpoile:
#### Summary
- Trigger the new style dialog using a custom post - it works for webapp and automatically falls back to the interactive dialog for mobile

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38731

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -836,16 +836,16 @@ func (h *PlaybookRunHandler) reminderButtonDismiss(w http.ResponseWriter, r *htt
 	ReturnJSON(w, nil, http.StatusOK)
 }
 
-func (h *PlaybookRunHandler) removeReminderPost(w http.ResponseWriter, userID, playbookID, channelID string) {
-	storedPlaybookID, err := h.playbookRunService.GetPlaybookRunIDForChannel(channelID)
+func (h *PlaybookRunHandler) removeReminderPost(w http.ResponseWriter, userID, playbookRunID, channelID string) {
+	storedPlaybookRunID, err := h.playbookRunService.GetPlaybookRunIDForChannel(channelID)
 	if err != nil {
 		h.log.Errorf("reminderButtonDismiss: no playbook run for requestData's channelID: %s", channelID)
 		h.HandleErrorWithCode(w, http.StatusBadRequest, "no playbook run for requestData's channelID", err)
 		return
 	}
-	if storedPlaybookID != playbookID {
+	if storedPlaybookRunID != playbookRunID {
 		h.HandleErrorWithCode(w, http.StatusBadRequest, "error removing reminder",
-			fmt.Errorf("storedPlaybookRunID: %s for channelID: %s did not match playbookID in url path: %s", storedPlaybookID, channelID, playbookID))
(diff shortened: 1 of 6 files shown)


server/api/playbook_runs.go
```diff
@@ -801,23 +802,54 @@ func (h *PlaybookRunHandler) reminderButtonUpdate(w http.ResponseWriter, r *http
...
 		return
 	}
 
-	playbookRunID, err := h.playbookRunService.GetPlaybookRunIDForChannel(requestData.ChannelId)
+	h.removeReminderPost(w, userID, playbookRunID, requestData.ChannelId)
+
+	ReturnJSON(w, nil, http.StatusOK)
+}
+
+func (h *PlaybookRunHandler) removeReminderPost(w http.ResponseWriter, userID, playbookID, channelID string) {
```
agarciamontoro, line 839: I think this is a typo, right?
```suggestion
func (h *PlaybookRunHandler) removeReminderPost(w http.ResponseWriter, userID, playbookRunID, channelID string) {
```

webapp/src/components/update_request_post.tsx
```diff
@@ -0,0 +1,91 @@
...
+    post: Post;
+}
+
+export const UpdateRequestPost = (props: Props) => {
+    const dispatch = useDispatch();
+    const channel = useSelector<GlobalState, Channel>((state) => getChannel(state, props.post.channel_id));
+    const team = useSelector<GlobalState, Team>((state) => getTeam(state, channel.team_id));
+    const currentRun = useSelector(currentPlaybookRun);
+    const targetUsername = props.post.props.targetUsername ?? '';
+    const playbookId = currentRun?.id;
```
agarciamontoro, line 32: Same typo here:
```suggestion
    const playbookRunId = currentRun?.id;
```

agarciamontoro (changes requested):
Nice and clean! Just a couple of naming typos that we should fix to not confuse our future selves.

server/api/playbook_runs.go
cpoile: Ahhh, it strikes again! Thanks for catching that.

webapp/src/components/update_request_post.tsx
cpoile: 👍

cpoile:
@abhijit-singh Does this look good for replacing the interactive dialog (when the user is using the webapp -- ie, not mobile):

image

agarciamontoro (approved):
:tada:

crspeller (approved):

abhijit-singh:
Thanks @cpoile! A couple of suggested changes:

* Can we wrap the buttons in a box similar to the one we have for status updates?
image

* I think the buttons in the post can always have the standard blue color instead of red, as in the screenshot above.

abhijit-singh (changes requested):
Thanks @cpoile! A couple of suggested changes:

* Can we wrap the buttons in a container similar to what we have for the status update post?
image

* I think the buttons in the post can always be of the standard blue color (instead of red), as in the screenshot above.

cpoile:
@abhijit-singh Look good?
image

abhijit-singh (approved):

PR #768, opened 2021-09-27
MM-38779: test with globalheader

lieut-data:
#### Summary
Add a test that exercises our e2e suite with the global header enabled. Fix a recent regression in the e2e tests around the bot rename. Clean up some CircleCI stuff while we're in there.

The one failing test should be fixed with https://github.com/mattermost/mattermost-server/pull/18464.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38779

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -229,24 +229,6 @@ jobs:
           command: |
             ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -b "$(< ./dist/release-notes.md)" -c ${CIRCLE_SHA1} -n ${CIRCLE_TAG} -delete ${CIRCLE_TAG} dist/*.tar.gz
 
-  test-MySQL56-Postgres96:
-    docker:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -229,24 +229,6 @@ jobs:
           command: |
             ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -b "$(< ./dist/release-notes.md)" -c ${CIRCLE_SHA1} -n ${CIRCLE_TAG} -delete ${CIRCLE_TAG} dist/*.tar.gz
 
-  test-MySQL56-Postgres96:
```
cpoile, line 232: ah, yah, that should go. :)

cpoile (approved):

.circleci/config.yml
lieut-data: Upon reflection, I suppose we actually still need the `MySQL56` part of this, don't we? I'll pull this back but without Postgres96.

PR #772, opened 2021-09-27
MM-38608 - Include list of Runs in Progress

cpoile:
#### Summary
- Does what it says on the tin

- Design
image

- Actual (I made some slight changes to the english, but there will be a polish pass later)
image

#### Progress: 
- POC ✅ 
- Include list of ongoing runs (<--- we're here)
- Include list of overdue updates
- Send digest once a day when user logs on
- Checklist directly in the digest (custom post)
- Design polish (if needed)
- Telemetry (total number of in-notification link clicks, unique users who click)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38608

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/sqlstore/playbook_run.go
+++ b/server/sqlstore/playbook_run.go
@@ -921,13 +921,13 @@ func (s *playbookRunStore) GetAssignedTasks(userID string) ([]app.AssignedRun, e
 		ChecklistsJSON json.RawMessage
 	}
 
-	query := s.store.builder.Select("i.ID AS PlaybookRunID", "i.Name AS PlaybookRunName",
-		"t.Name AS TeamName", "c.Name AS ChannelName", "c.DisplayName AS ChannelDisplayName",
+	query := s.store.builder.Select("i.ID AS PlaybookRunID", "t.Name AS TeamName",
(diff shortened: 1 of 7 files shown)


server/sqlstore/playbook_run.go
```diff
@@ -969,6 +969,33 @@ func (s *playbookRunStore) GetAssignedTasks(userID string) ([]app.AssignedRun, e
 	return ret, nil
 }
 
+// GetParticipatingRuns returns the list of active runs with userID as a participant
+func (s *playbookRunStore) GetParticipatingRuns(userID string) ([]app.RunLink, error) {
```
iomodo, line 973: Quite a heavy query for the DB.
What do you think about having the `Participants` table? It will allow us to decouple from the channels as well.
Not a part of this PR

iomodo (approved):
Great!

server/sqlstore/playbook_run.go
cpoile: True. However, it's simpler and more constrained than the query used in the `GetPlaybookRuns` call (when a participantID is included). And we call that one more frequently. But a `Participants` table would be useful, I agree. Let's keep that in mind for the future?

PR #774, opened 2021-09-29
MM-38935: Webapp, i18n mvp tooling

calebroseland:
#### Summary
- Adds minimum tooling for message extraction to JSON (`webapp/i18n/en.json`)
  (i18n-check will come in a later PR)
- Adds registration to provide messages to upstream `<IntlProvider/>`
- Adds a few sanity eslint rules to prevent `id`s and enforce `defaultMessage`s

An important note about `id`s: they are automatically generated via a checksum of the content of a descriptor's `defaultMessage`—this prevents stale translations and ensures duplicate strings will not be created (`"Help"` will always be `"Help"`—if a particular instance of `"Help"` is changed, it will automatically kick out a new `id` yet leave other unaffected instances as-is.)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38935

#### Checklist

~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~

### DIFF
--- a/webapp/i18n/en.json
+++ b/webapp/i18n/en.json
@@ -1,7 +1,7 @@
 {
-  "playbooks./HtNUp": "Select or specify a {mode, select, DurationValue {time span (\"4 hours\", \"7 days\"...)} DateTimeValue {time (\"in 4 hours\", \"May 1\", \"Tomorrow at 1 PM\"...)} other {time or time span}}",
-  "playbooks.AF9wda": "This update will be saved to the <OverviewLink>overview page</OverviewLink>{hasBroadcast, select, true { and broadcast to <ChannelsTooltip>{broadcastChannelCount, plural, =1 {one channel} other {{broadcastChannelCount, number} channels}}</ChannelsTooltip>} other {}}.",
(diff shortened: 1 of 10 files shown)


webapp/i18n/en.json
```diff
@@ -1 +1,7 @@
-{}
+{
+  "/HtNUp": "Select or specify a {mode, select, DurationValue {time span (\"4 hours\", \"7 days\"...)} DateTimeValue {time (\"in 4 hours\", \"May 1\", \"Tomorrow at 1 PM\"...)} other {time or time span}}",
```
lieut-data, line 2: I love that we don't have to craft new ids for each translated string. But I'm wondering if it's possible to avoid using special characters in the generated names? I could see it being harder than necessary to search for `/HtNUp` (kinda like how Zoom's passwords are often copy/pasted wrong).

lieut-data (approved):
Sweet!

webapp/i18n/en.json
calebroseland: 100% agree with the search ability re: special chars, but as-is, there aren't any IDs to search for—you'll have to search by defaultMessage.

crspeller (approved):
This is how I always imagined translations should work. Thanks for making dreams a reality :)

PR #775, opened 2021-09-30
MM-38668 - Send digest once a day when user logs on

cpoile:
#### Summary
- Refactoring of where the code lives (it belongs in `playbookRunService` because it's used by multiple dependents)
- Totally cribbing how Github plugin knows when to send daily notifications (`onClick` after an hour of inactivity -> server side checks if you've been sent today's message). On the face it seems like this is bad, but after thinking about it it's not that much work on the client (compared to what react does on any single rerender!). Basically a couple of math ops on every click. Only after inactivity for an hour do you get one request to the server, so no big deal.
- The one difficult design decision was making a UserInfo app struct, service, and sqlstore. There were a couple options here:
  - add the struct to the playbookRun file/service/store, but that felt like it completely violated separation of concerns.
  - use the KVStore. Initially I leaned towards this. But then I remembered Jesse's argument that we should keep out of the KVStore, so we know where all our data is.
  - that left it's own service/sqlstore. This seemed the technically "correct" option, even though it's a lot more typing. But I bit the bullet and now it's done. And now we can add fields whenever we need to store user-level information (which we will almost definitely need to in the future).

Slight redesign, now looks like:
image

#### Progress: 
- POC ✅ 
- Include list of ongoing runs ✅ 
- Include list of overdue updates ✅ 
- Send digest once a day when user logs on (<--- we're here)
- Telemetry (total number of in-notification link clicks, unique users who click)
- Design polish (if needed)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38668

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


server/app/user_info.go
```diff
@@ -0,0 +1,25 @@
...
+// See LICENSE.txt for license information.
+
+package app
+
+type UserInfo struct {
+	ID       string `json:"id"`
+	LastDMAt int64  `json:"last_dm_at"`
+}
+
+type UserInfoService interface {
```
crspeller, line 11: Do we need a service here? Maybe a raw store makes more sense since there really won't be any business logic here.

server/sqlstore/migrations.go
```diff
@@ -1254,4 +1254,32 @@ var migrations = []Migration{
...
+	{
+		fromVersion: semver.MustParse("0.32.0"),
+		toVersion:   semver.MustParse("0.33.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DatabaseDriverMysql {
+				if _, err := e.Exec(`
+					CREATE TABLE IF NOT EXISTS IR_UserInfo
+					(
+						ID       VARCHAR(26) PRIMARY KEY,
+						LastDMAt BIGINT
```
crspeller, line 1266: Maybe we should be more specific. Like `LastDailySummaryDMAt`

server/api/bot.go
```diff
@@ -151,3 +162,44 @@ func (h *BotHandler) promptForFeedback(w http.ResponseWriter, r *http.Request) {
 
 	w.WriteHeader(http.StatusOK)
 }
+
+// connect handles the GET /bot/connect endpoint (a notification sent when the client wakes up or reconnects)
+func (h *BotHandler) connect(w http.ResponseWriter, r *http.Request) {
```
crspeller, line 167: There doesn't seem to be an option to turn this off. Should we have a user setting for that? Maybe I don't want these.

crspeller (changes requested):

server/sqlstore/migrations.go
cpoile: Yep, makes sense. Done.

server/api/bot.go
cpoile: This is by design. Muting the playbooks bot = turning it off. @stephenvanhemmen for visibility.

crspeller: What? How do you mute a bot?

cpoile: I meant mute the DM channel

crspeller: Oh. I guess that works. You just have to make sure it never mentions the user otherwise it will break the mute. 
I suspect people aren't going to think about using the mute functionality in this way. Also, when we implement the following runs feature people may want to receive those messages but not these ones. So it seems like we will need a setting then?

cpoile: I'm 0/5 on which solution. I'm fine with a setting, that's what I suggested originally. But @tao and @stephenvanhemmen wanted to go with the mute. Up to you @tao and @stephenvanhemmen.
Personally, I love github messaging me every day, I rely on it all the time, so I've never thought of muting it.

stephenvanhemmen: The current thinking around preventing notifications from being turned off is to always maintain a place where these cross-run notifications can be referenced for each user. Muting the bot may help the user if they do not wish to receive these notifications while still having a reference location for their responsibilities over all runs

crspeller: I feel like I am going to mute notifications to suppress this, then never turn it back on for the follow feature.

crspeller (approved):

iomodo (approved):
Looks great!

PR #776, opened 2021-09-30
MM-38872 - Telemetry for the todo message

cpoile:
#### Summary
- Will record interactions with each of the todo list sections. We can total the number of interactions, group based on unique userId, and by section (to see which section is most useful).
- If found, the search param is removed from the url (leaves other search params untouched)
- I added it to the RHSRunDetails component so that in the future we can use this params method to scroll to and/or highlight specific checklist items (e.g., click on the checklist item due, and we'll take you right to that item -- or as close as possible to that)
- cc @stephenvanhemmen 

#### Progress: 
- POC ✅ 
- Include list of ongoing runs ✅ 
- Include list of overdue updates ✅ 
- Send digest once a day when user logs on ✅ 
- Telemetry (<--- we're here)
- Design polish

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38872

#### Checklist

- [x] Telemetry updated
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/rhs/rhs_run_details.tsx
+++ b/webapp/src/components/rhs/rhs_run_details.tsx
@@ -14,9 +14,23 @@ import {
 import {currentPlaybookRun} from 'src/selectors';
 import RHSAbout from 'src/components/rhs/rhs_about';
 import RHSChecklists from 'src/components/rhs/rhs_checklists';
+import {browserHistory} from 'src/webapp_globals';
+import {telemetryEventForPlaybookRun} from 'src/client';
 
 const RHSRunDetails = () => {
     const playbookRun = useSelector(currentPlaybookRun);
+    const url = new URL(window.location.href);
+    const searchParams = new URLSearchParams(url.searchParams);
+
+    if (searchParams.has('telem') && playbookRun) {
+        const action = searchParams.get('telem');
+        if (action) {
+            telemetryEventForPlaybookRun(playbookRun.id, action);
+        }
+        searchParams.delete('telem');
+        url.search = searchParams.toString();
+        browserHistory.push({pathname: url.pathname, search: url.search});
+    }
 
     if (!playbookRun) {
         return null;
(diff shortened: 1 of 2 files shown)


agarciamontoro (approved):
Looks good!

A nit: is it possible to remove the telemetry params when the user is copying the link? If they copy it and pass it to someone else, we'll track that as a telemetry event, although they did not click on it. This is probably an edge case, so I'm ok merging without that.

webapp/src/components/rhs/rhs_run_details.tsx
```diff
@@ -14,9 +14,23 @@ import {
...
+    const searchParams = new URLSearchParams(url.searchParams);
+
+    if (searchParams.has('telem') && playbookRun) {
+        const action = searchParams.get('telem');
+        if (action) {
+            telemetryEventForPlaybookRun(playbookRun.id, action);
+        }
+        searchParams.delete('telem');
+        url.search = searchParams.toString();
+        browserHistory.push({pathname: url.pathname, search: url.search});
```
lieut-data, line 32: Should we use `replace` here instead of `push`? I imagine clicking back right now results in this running a second time, making it hard to actually go back to previous locations.

lieut-data (changes requested):
Thanks, @cpoile! I think my feedback is the same as Alejandro's, with just the observation that I think this breaks clicking back.

webapp/src/components/rhs/rhs_run_details.tsx
cpoile: It didn't make it run a second time, it would put the query string back into the location url--you could press back again to really go back. But definitely `replace` works better here. One-click to go back. Thanks!!

cpoile:
@agarciamontoro that's a good point. On the one hand, we don't want to get accidental telemetry if someone posts a copied link. On the other hand, this is probably a very rare edge case.  But just in case, do you know of a way to fix it? These are markdown urls -- is there a way to edit the link if copied, or have the click be different from the copied link? I can't think of a way to do that without javascript.

agarciamontoro:
@cpoile, I don't know of any way, sorry. And I agree that it's a rare case. I'm ok merging this without it.

lieut-data (approved):

PR #779, opened 2021-10-01
[MM-38151] MM-38843, MM-37172: i18n, convert playbook overview

calebroseland:
#### Summary
- Wrap strings in `formatMessage` calls
- Use i18n-enabled time formatting
  Note: formatting true-to-value time ranges may now shift days displayed in stats. (see [MM-39004](https://mattermost.atlassian.net/browse/MM-39004))

CleanShot 2021-10-01 at 14 08 25@2x
[Green: converted] [Red: not converted]
#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-37172
- https://mattermost.atlassian.net/browse/MM-38843

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


tests-e2e/cypress/integration/backstage/playbook_overview_spec.js
```diff
@@ -72,7 +72,7 @@ describe('playbook overview', () => {
             });
 
             // # Verify permissions text
-            cy.findByTestId('playbookPermissionsDescription').contains(`Everyone in ${testTeam.name} can access this playbook`);
+            cy.findByTestId('playbookPermissionsDescription').contains(`Everyone in ${testTeam.display_name} can access this playbook`);
```
calebroseland, line 75: Show display name instead of url slug

webapp/src/components/widgets/pill.tsx
```diff
@@ -1,13 +1,13 @@
...
 
-import React from 'react';
+import React, {ReactNode} from 'react';
 import styled from 'styled-components';
 
 import ClearIcon from 'src/components/assets/icons/clear_icon';
 
 interface Props {
-    text: string;
+    text: ReactNode;
```
calebroseland, line 10: Fix dev-tools warning

webapp/src/components/backstage/playbooks/stats_view.tsx
```diff
@@ -110,61 +129,52 @@ const StatsView = (props: Props) => {
...
                     />
                 </GraphBox>
             </BottomRow>
             <BottomRow>
                 <GraphBox>
                     <BarGraph
-                        title={'ACTIVE RUNS per day over the last 14 days'}
-                        labels={props.stats.active_runs_per_day_labels}
+                        title={formatMessage({defaultMessage: 'ACTIVE RUNS per day over the last 14 days'})}
+                        labels={props.stats.active_runs_per_day_times.map(([start, _]) => DateTime.fromMillis(start).toLocaleString(DateTimeFormats.DATE_MED_NO_YEAR))}
```
cpoile, line 163: I'm worried that because the `start` time is the exact start of the day in millis, this might cause problems with timezones and daylight savings, etc. Is there a way to test it, or maybe add a bit of time to avoid the potential edge cases?  (same for the other start times)

cpoile (approved):
Nice! Just a question about edge cases, otherwise looks great.

webapp/src/components/backstage/playbooks/stats_view.tsx
calebroseland: I agree, but I think from a formatting perspective it shouldn't matter too much. 

CleanShot 2021-10-04 at 09 46 36@2x


I would propose we tackle this in [MM-39004](https://mattermost.atlassian.net/browse/MM-39004) by adding `start`, `end` to `api/stats`/`sqlstore/stats` and create the filter times client-side via something like `DateTime.now().startOf('week')` to base the timezone in

webapp/src/components/backstage/playbooks/playbook_backstage.tsx
```diff
@@ -58,10 +60,11 @@ const RunListContainer = styled.div`
 
 const PlaybookBackstage = () => {
     const dispatch = useDispatch();
+    const {formatMessage} = useIntl();
     const match = useRouteMatch<MatchParams>();
     const location = useLocation();
     const [playbook, setPlaybook] = useState<PlaybookWithChecklist | null>(null);
-    const [filterPill, setFilterPill] = useState<JSX.Element | null>(null);
+    const [filterPill, setFilterPill] = useState<ReactNode | null>(null);
```
crspeller, line 67:
```suggestion
    const [filterPill, setFilterPill] = useState<ReactNode>(null);
```

webapp/src/components/backstage/runs_list/runs_list.tsx
```diff
@@ -19,7 +19,7 @@ interface Props {
     totalCount: number
     fetchParams: FetchPlaybookRunsParams
     setFetchParams: React.Dispatch<React.SetStateAction<FetchPlaybookRunsParams>>
-    filterPill: React.ReactNode
+    filterPill: React.ReactNode | null
```
crspeller, line 22: Unnecessary since ReactNode already includes null.
```suggestion
    filterPill: React.ReactNode
```

crspeller (approved):
Just a couple nits.

webapp/src/components/backstage/runs_list/runs_list.tsx
calebroseland: Technically yes, and I would much prefer that, but seems react-devtools-backend isn't quite smart enough to infer that. It converts:
- `{filterPill: React.ReactNode}` to be `{filterPill: propTypes.node.isRequired}`
- `{filterPill: React.ReactNode | null}` to be `{filterPill: propTypes.oneOfType([propTypes.node, propTypes.oneOf([null])])}`

So this is required to suppress the devtools warning.

crspeller: Ah I think this functionality might be the source of the error: https://github.com/milesj/babel-plugin-typescript-to-proptypes#strict-boolean
It might look just a bit nicer if you made it optional instead of having an `| null` if it accepts that. 

Any idea what the prop types conversion gets us? I know there is a runtime check but does it actually catch anything beyond what Typescript would?

calebroseland: Ahh, that makes sense. Thanks for the link. 

I don't personally find the auto propTypes very helpful. I'd much rather see a linting or TS error than a runtime error, and it would appear the propTypes conversion isn't even very accurate in cases like this. Given that they're generated from TS, I would think the TS errors would actually be more thorough.

And turning off propTypes generation might speed up development builds.

calebroseland: Re: `{filterPil?: React.ReactNode}` vs `{filterPill: React.ReactNode | null}` I don't really care much, but am 1/5 towards the latter, because the state lives in the parent.

I'd propose we leave as-is for now, and re-address when/if we take on any `babel-plugin-typescript-to-proptypes` changes.

PR #781, opened 2021-10-01
[GH-17954] migrate profile_button to styled component

lex111:
#### Summary

This will migrate webapp/src/components/profile/profile_button.scss to styled-components

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/17954

#### Checklist

~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


webapp/src/components/profile/profile_button.tsx
```diff
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React from 'react';
+import styled from 'styled-components';
 
 import './profile_button.scss';
```
crspeller, line 7: Looks like you just forgot to remove the import. That's why the build is failing.

crspeller (changes requested):
Thanks for the PR!

webapp/src/components/profile/profile_button.tsx
lex111: Oh, that's right, I forgot all about that.

crspeller (approved):

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @lex111

PR #784, opened 2021-10-02
[GH-18377] style react select pills to look like rounded pills

joshalling:
#### Summary

This updates the styles of the react-select pills in the keyword input to make them look like the other rounded pills. See image below for the update.

Screen Shot 2021-10-01 at 9 43 31 PM

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/18377
#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


webapp/src/components/backstage/automation/input_keywords.tsx
```diff
@@ -120,24 +120,32 @@ const selectStyles: StylesConfig = {
...
+        borderRadius: '10px',
+        paddingLeft: '8px',
+        overflow: 'hidden',
     }),
     multiValueLabel: (provided) => ({
         ...provided,
-        backgroundColor: 'var(--center-channel-color-08)',
-        color: 'var(--center-channel-color)',
+        padding: 0,
+        paddingLeft: 0,
```
joshalling, line 131: For some reason just setting `padding: 0` would leave `6px` of padding on the left so I had to set both properties. I'm not familiar with `react-select` so I wasn't sure what else to do here.

joshalling:
@agarciamontoro It's no problem at all. We should be good now. 👍

PR #787, opened 2021-10-04
MM-37999

Phrynobatrachus:
Hello y'all. :wave: 

Kept it pretty basic. I also attempted to add a third test wherein a private playbook is created and the test validates that it is _not_ shown to a non-member, but I was having some trouble with API 403s and left it out for now. I used `playbook_list_permissions_spec.js` for guidance, but I noticed that test actually fails similarly when run. If that DOES seem like a good test to have, let me know and I dig a bit more around that.

#### Summary
-  adds a new spec for Playbooks Home

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37999

#### Checklist
~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/frontstage/rhs/home_spec.js
+++ b/tests-e2e/cypress/integration/frontstage/rhs/home_spec.js
@@ -1,3 +1,11 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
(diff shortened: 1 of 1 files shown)


lieut-data:
Thanks, @Phrynobatrachus! I've asked @saturninoabril and @crspeller to take a look :)

crspeller (approved):

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @Phrynobatrachus

PR #789, opened 2021-10-05
[MM-38151] Convert: Playbooks -> Run-Overview

shadowshot-x:
#### Summary

- Added formatMessage function to all possible hard-coded strings in src/components/backstage/playbook_runs/playbook_run_backstage  and neighbouring files.

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/18514

#### Checklist

- ~[ ] Detected Hard-coded strings~
- ~[ ] Added the formatMessage function as per the hard-coded strings~

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
@@ -23,6 +23,8 @@ import {
 import {useEnsureProfiles} from 'src/hooks';
 import {navigateToUrl} from 'src/browser_routing';
 
+import {useIntl} from 'react-intl';
+
 const StyledContent = styled(Content)`
     padding: 8px 20px 24px 24px;
 `;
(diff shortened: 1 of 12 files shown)


webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
```diff
@@ -52,27 +53,29 @@ const Participants = (
...
         .filter((id) => id !== playbookRun.owner_user_id && id !== playbookRun.reporter_user_id);
 
+    const {formatMessage} = useIntl();
     const team = useSelector<GlobalState, Team>((state) => getTeam(state, playbookRun.team_id));
 
     return (
         <TabPageContainer>
-            <Title>{`Participants (${participant_ids.length})`}</Title>
+            {/* Should this be of type Plural? */}
+            <Title>{formatMessage({defaultMessage: 'Participants ({participants})'}, {participants: participant_ids.length})}</Title>
```
calebroseland, line 62:
```suggestion
            <Title>{formatMessage({defaultMessage: 'Participants ({participants})'}, {participants: participant_ids.length})}</Title>
```
No, it doesn't need to be a `{plural}`—as this is just a title, it's perfectly fine as you have it. 👍

calebroseland (changes requested):
Thanks for the PR, @shadowshot-x! This looks really good, just a couple parts left in the updates section (in `webapp/src/components/rhs/post_card.tsx`), I think the timestamp and a string or two.

Also, `en.json` got some new strings in master, so you'll probably have to merge from master and re-extract to resolve that conflict.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
shadowshot-x: Right, Caleb, I will leave it as is. Thank you!

webapp/src/components/rhs/post_card.tsx
```diff
@@ -94,17 +95,18 @@ interface Props {
...
     const [authorProfileUrl, authorUserName] = useAuthorInfo(props.post?.user_id || '');
 
     if (!props.post) {
         return (
             <NoRecentUpdates>
-                {'No recent updates. '}<a onClick={() => dispatch(promptUpdateStatus(props.team.id, props.playbookRunId, props.playbookId, props.channelId))}>{'Click here'}</a>{' to update status.'}
+                {formatMessage({defaultMessage: 'No recent updates. '})}<a onClick={() => dispatch(promptUpdateStatus(props.team.id, props.playbookRunId, props.playbookId, props.channelId))}>{formatMessage({defaultMessage: 'Click here'})}</a>{formatMessage({defaultMessage: ' to update status.'})}
             </NoRecentUpdates>
         );
     }
```
calebroseland, line 107: Sorry I didn't catch this last review, but after taking a look at this if block, it seems like dead code. (`<PostContent/>` has an `if (!post) {return null}` check.)

If you wouldn't mind, could you remove it entirely and remove the `null` from `Props['post']`? Technically out of scope, but it doesn't hurt to remove dead code when it comes to light 🙂.

calebroseland (changes requested):

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @shadowshot-x

PR #791, opened 2021-10-05
MM-38929: Add a navbar under the playbook name to switch between Preview and Usage

agarciamontoro:
#### Summary
This is the first PR in the Playbook Preview epic. It lays the groundwork for the actual development, but it does not do much on itself:

- Add a navigation bar to switch between the (yet to be implemented) Preview page and the Usage page. This is behind the Experimental Features flag.
- Refactor the `PlaybookBackstage` component: it is now named `PlaybookUsage`, and it only contains the logic to display the usage page. The rest of the code, used to render the bar with the playbook name, the left arrow to get back to the playbooks list and the Edit and Run buttons have been added to the new Playbook component.
- Add a `Playbook component`, which is in charge of:
  - Rendering the bar with the playbook name and the Edit and Run buttons (the Edit button will be moved to the Preview page soon, as per the designs).
  - Rendering the navigation bar to switch between the Preview and Usage tabs
  - Handling the logic to show one of those two tabs (with the new `/playbooks/playbooks/:playbookId/{preview, usage}` routes), redirecting to `/playbooks/playbooks/:playbookId/usage` when the plain `/playbooks/playbooks/:playbookId` is visited. This default will change soon (the design is to show `/preview` as the default, but as it is empty as of today, we will have `/usage` as the default while I finish the actual implementation of that view).
  
image

#### Plan
1. Add a navigation bar with Preview and Usage tabs <-- we're here
2. Implement the layout of the Preview page
3. Add the sticky navigation widget
4. Remove the Experimental Feature flag and make /preview the default view
5. Add telemetry

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38929

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbooks/playbook_usage.tsx
+++ b/webapp/src/components/backstage/playbooks/playbook_usage.tsx
@@ -31,7 +31,7 @@ interface Props {
     playbook: PlaybookWithChecklist;
 }
 
-const PlaybookBackstage = (props: Props) => {
+const PlaybookUsage = (props: Props) => {
     const [filterPill, setFilterPill] = useState<JSX.Element | null>(null);
     const [stats, setStats] = useState(EmptyPlaybookStats);
     const [playbookRuns, totalCount, fetchParams, setFetchParams] = useRunsList(defaultPlaybookFetchParams);
@@ -108,4 +108,4 @@ const BottomInnerContainer = styled.div`
     }
 `;
 
-export default PlaybookBackstage;
+export default PlaybookUsage;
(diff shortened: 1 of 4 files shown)


webapp/src/components/backstage/playbooks/playbook_usage.tsx
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
agarciamontoro, line 1: All the code in this file comes from the removed code in `playbook_backstage.tsx` (now called simply `playbook.tsx`)

PR #806, opened 2021-10-12
[MM-39116] Moved null check for post to happen before reference to channel_id

devinbinnie:
#### Summary
There was a null reference happening when opening the post menu if the message was a system message. We were trying to reference the `channel_id` before we did the null check, so I've moved the null check up.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39116


amyblais:
Thanks!

PR #807, opened 2021-10-12
MM-39158: avoid playbook edit console errors

lieut-data:
#### Summary
Wrap `dispatch` in `useEffect` to effectively avoid a `setState` call on an unmounted component. Also fix an unrelated warning from React re: `data-testId` instead of `data-testid`. Finally, don't `dispatch` at all until we have an actual team.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-39158

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/runs_list/search_input.tsx
+++ b/webapp/src/components/backstage/runs_list/search_input.tsx
@@ -59,7 +59,7 @@ export default function SearchInput(props: Props) {
 
     return (
         <RunListSearch
-            data-testId={props.testId}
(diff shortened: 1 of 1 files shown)


lieut-data:
The e2e failure here is unrelated to my changes -- raised this with DevOps at https://community-daily.mattermost.com/core/pl/r5ychapoe3ysbqsqwyf95iayxo.

PR #813, opened 2021-10-14
MM-39351 - Fix e2e tests by updating the db test data

cpoile:
#### Summary
- And they removed some users, because. So had to edit the tests. It's green again!!
- I'm not totally sure if we still need https://mattermost.atlassian.net/browse/MM-39351 -- we do if it's possible for us to get into that state in a real server.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-39351

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


lieut-data:
(Still fixing the SQL :p)

PR #815, opened 2021-10-15
Fix model changes around server upgrade.

crspeller:
#### Summary
- Fix ability to upgrade server dependency. 
- Update server dependency.


crspeller:
@agarciamontoro The one Jesse pointed out was actually the only one with any danger. But when I was looking at the other ones I noticed they where eating the error if it wasn't nil. So I added it. (That error handler doesn't barf when getting a nil since its optional in that case)

crspeller:
Unfortunately it looks like the updated version of our testing library breaks most of our tests because of a mistake we made when writing them. Going to be quite the project to fix. Reverted the upgrade for now.

PR #829, opened 2021-10-22
MM-39126 Use natural lang for default update time

pascalhein:
#### Summary
When specifying the default update timer, we currently
have to use a modal with hour/minute/day for custom
duration. Instead, let's re-use the natural language
input from the status update modal. This supports
entering durations such as "2 minutes", "6 days",
"3 hours 5 seconds".

Replaces a lot of custom-written code with a call to
DateTimeInput.

image

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/18566
JIRA: https://mattermost.atlassian.net/browse/MM-39126

#### Checklist
- [ ] ~Telemetry updated~
- [ ] ~Unit tests updated~


pascalhein:
Hi @abhijit-singh and @calebroseland, thanks for your feedback!

I've implemented your suggestions: removing 'Never' and the X-to-clear button. However, that means that once an option has been clicked or typed, there is no way to reset it to the initial empty state. Is that intended?

calebroseland (approved):
@pascalhein I think due to the `never` option being removed, and reminder timers being required now—the value should always be _something_. (I believe that is intended.)

abhijit-singh (approved):
Yep, that's intended. Thank you @pascalhein!

lieut-data:
@calebroseland, I wasn't sure of the status of this PR -- heads up I just looped you back in for a review to make sure this gets merged or reworked as needed :)

PR #831, opened 2021-10-25
[MM-39119] Auto complete categories in categorize playbook run

dunak-debug:
#### Summary
When configuring the optional sidebar category into which to categorize a new playbook run channel, it uses react-select `Creatable` component to expose user existing categories and also allow creating non existing category as before.

Before
image

After
image

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/18562
https://mattermost.atlassian.net/browse/MM-39119

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/backstage/category_selector.tsx
```diff
@@ -0,0 +1,65 @@
...
+    isClearable?: boolean;
+    selectComponents?: SelectComponentsConfig<ChannelCategory>;
+    isDisabled: boolean;
+    captureMenuScroll: boolean;
+    shouldRenderValue: boolean;
+    placeholder?: string;
+}
+
+const getCategoriesForTeam = makeGetCategoriesForTeam();
+const getMyCategories = (state: GlobalState) => getCategoriesForTeam(state, state.entities.teams.currentTeamId);
```
iomodo, line 24: Could we avoid using redux and use `GetChannelSidebarCategories` from the plugin API instead?

webapp/src/components/backstage/automation/categorize_playbook_run.tsx
```diff
@@ -20,5 +31,90 @@ export const CategorizePlaybookRun = (props: Props) => (
...
+            </StyledScrollbars>
+        </MenuListWrapper>
+    );
+};
+
+const ClearIndicator = ({clearValue}: {clearValue: () => void}) => (
+    <div onClick={clearValue}>
+        <ClearIcon/>
+    </div>
+);
```
iomodo, line 120: Enter missing here

webapp/src/components/backstage/automation/categorize_playbook_run.tsx
```diff
@@ -20,5 +31,90 @@ export const CategorizePlaybookRun = (props: Props) => (
...
+const ThumbVertical = styled.div`
+    background-color: rgba(var(--center-channel-color-rgb), 0.24);
+    border-radius: 2px;
+    width: 4px;
+    min-height: 45px;
+    margin-left: -2px;
+    margin-top: 6px;
+`;
+
+const MenuList = (props: MenuListComponentProps<ChannelCategory>) => {
```
iomodo, line 102: It would be ideal to reuse this component

webapp/src/components/backstage/automation/categorize_playbook_run.tsx
```diff
@@ -20,5 +31,90 @@ export const CategorizePlaybookRun = (props: Props) => (
...
+                autoHeight={true}
+                renderThumbVertical={({style, ...thumbProps}) => <ThumbVertical {...thumbProps}/>}
+            >
+                {props.children}
+            </StyledScrollbars>
+        </MenuListWrapper>
+    );
+};
+
+const ClearIndicator = ({clearValue}: {clearValue: () => void}) => (
```
iomodo, line 116: Would be ideal to reuse this as well

iomodo (changes requested):
Looks great @dunak-debug !  Thanks for the contribution.
Couple of suggestions below

webapp/src/components/backstage/category_selector.tsx
iomodo: @dunak-debug My bad. Missed the `fetchMyCategories` on playbook edit. Getting from the redux will actually be Ok here.

iomodo: @agarciamontoro thanks for pointing out ^_^

dunak-debug: Actually yes, we discussed something at mattermost and it would be OK to use redux here.

dunak-debug:
Thank you for review. I'll get home and look onto requested changes

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @dunak-debug

agarciamontoro:
Hi, @dunak-debug, is there anything we can do to help you with this?

dunak-debug:
Hi, sorry I had corona and I couldnt really get into work on computer 😏 I feel bad for this big delay on this PR and as soon as I arrive home I will push changes and we can resolve any other reviews. Is that ok? 🙂

agarciamontoro:
Oh please, don't worry at all! Take all the time you need and get well soon, your health is way more important than any PR. Ping us when you're ready, there's no rush at all. Stay safe :)

webapp/src/components/backstage/automation/categorize_playbook_run.tsx
dunak-debug: Do you want to specify explicitly menu header title or inherit from placeholder?

dunak-debug:
I did some changes. Please @agarciamontoro, take a look and let me know

webapp/src/components/backstage/category_selector.tsx
```diff
@@ -0,0 +1,65 @@
...
+            props.onCategorySelected('');
+        } else {
+            props.onCategorySelected(option.value);
+        }
+    };
+
+    const components = props.selectComponents || defaultComponents;
+
+    return (
+        <StyledCreatable
```
agarciamontoro, line 46: Could we use `menuPlacement={'top'}`? It is not completely visible when it extends below:

image

webapp/src/components/backstage/category_selector.tsx
```diff
@@ -0,0 +1,65 @@
...
+            className={props.className}
+            id={props.id}
+            controlShouldRenderValue={props.shouldRenderValue}
+            options={options}
+            onChange={onChange}
+            defaultMenuIsOpen={false}
+            openMenuOnClick={true}
+            isClearable={props.isClearable}
+            value={{value: props.categoryName, label: props.categoryName}}
+            placeholder={props.placeholder || 'Add channel to category'}
```
agarciamontoro, line 56: Since not that long ago, we can internationalize text in Playbooks, so this is a good opportunity to use it.

You simply just add `const {formatMessage} = useIntl();` at the beginning of the component (the import for this is `import {useIntl} from 'react-intl';`) and change this line to
```suggestion
            placeholder={props.placeholder || formatMessage({defaultMessage: 'Add channel to category'})}
```

agarciamontoro (changes requested):
Thank you, @dunak-debug! I left just a couple of minor comments, otherwise looks great :)

dunak-debug:
Thank you, useful comments ;) Didn't know about intl and menuPlacement.. Updating it

webapp/src/components/backstage/category_selector.tsx
dunak-debug: Also I had to remove `placeholder` from upper component `categorize_playbook_run` so it use `formatMessage` instead of un-translated `props.placeholder`

iomodo (approved):
LGTM! Thanks @dunak-debug !

iomodo:
Could you merge with master @dunak-debug ? Also please run `make i18n-extract` and commit `en.json` file changes <3

dunak-debug:
Also I wanted to ask where could I push tests for this PR.  It's likely a part of my school assignment :D to do these tests... So if I can push it inside this PR or make a new one with the same ticket number?

iomodo:
@dunak-debug you can push Tests here.
Squash will happen automatically on merge

dunak-debug:
> @dunak-debug you can push Tests here. Squash will happen automatically on merge

So if I am right these specific tests should go inside https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/tests-e2e/cypress/integration/playbooks/edit_spec.js

agarciamontoro:
> So if I am right these specific tests should go inside https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/tests-e2e/cypress/integration/playbooks/edit_spec.js

@dunak-debug, that would a be a great place, yes! Thank you :)

dunak-debug:
I am experiencing an issue when I run tests on `is disabled in a new playbook`

```
cy.request() failed on:

http://localhost:8065/plugins/playbooks/api/v0/playbooks

The response we received from your web server was:

  > 403: Forbidden
```

iomodo:
@dunak-debug sometimes tests are a bit flaky. Just fix the lint error, tests should be ok.

dunak-debug:
> @dunak-debug sometimes tests are a bit flaky. Just fix the lint error, tests should be ok.

I did but I get still this 403 error.

```
 POST /plugins/playbooks/api/v0/playbooks
                        Status: 403 - Forbidden
                        Response body: {
                          "error": "Not authorized"
                        }
                        ```

**Comment by dunak-debug** [comment]
I saw warning.. `Not authorized; internal details: restricting playbook editing to specific users is not available with your current subscription: not covered by current server license` This may be the issue

**Comment by dunak-debug** [comment]
Not sure why CI wants from me to extract i18n. Maybe somebody forgot to do?

**Comment by agarciamontoro** [comment]
> I saw warning.. Not authorized; internal details: restricting playbook editing to specific users is not available with your current subscription: not covered by current server license This may be the issue

Do you have the testing and developer modes enabled? That should fix it

> Not sure why CI wants from me to extract i18n. Maybe somebody forgot to do?

Probaly the merge added some conflicting change to `en.json`. Run `make i18n-extract` and commit whatever comes out of that, it will fix the errors.

**Comment by dunak-debug** [comment]
> Do you have the testing and developer modes enabled? That should fix it

I have, still 403. `✘  POST /plugins/playbooks/api/v0/playbooks` 

I am really confused 😕

**Comment by agarciamontoro** [comment]
Oh, I think why this might be. Can you try opening cypress like this?

```
CYPRESS_developerMode=true npm run cypress:open
```

Also, first double-check that the testing and developer modes are indeed enabled, running the Cypress tests without the env var might have overwritten that at some point.

**Comment by dunak-debug** [comment]
Cypress developerMode fixed 🎊

**Comment by agarciamontoro** [comment]
> Cypress developerMode fixed :tada: 

Yay! Sorry, should have warned you about that, but it's a relatively new thing and I forgot!

**Comment by dunak-debug** [comment]
Hope I covered all possible cases 👍 If you can think of any I forgot let me know :)

**Comment by agarciamontoro** [comment]
I'll get to this tomorrow morning, @dunak-debug, sorry for the late reply! Jesse also added Cass to help review the E2E tests :tada:

File: tests-e2e/cypress/integration/playbooks/edit_spec.js [hunk 7]
```diff
@@ -1195,5 +1195,240 @@ describe('playbooks > edit', () => {
...
         });
+
+        describe('when a new member joins the channel', () => {
+            describe('add the channel to sidebary category', () => {
+                it('is disabled in a new playbook', () => {
+                    // # Visit the selected playbook
+                    cy.visit(`/playbooks/playbooks/${testPlaybook.id}/edit`);
+
+                    // # Switch to Actions tab
+                    cy.get('#root').findByText('Actions').click();
```
Phrynobatrachus, line 1206:
```suggestion
                    cy.findByText('Actions').click();
```
I don't think the `get('#root')` is necessary on any of these. This `visit -> switch to 'Actions'` sequence also happens at the start of all the `it` blocks, so it could probably go in a `beforeEach` for the parent `describe` block.

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -1195,5 +1195,240 @@ describe('playbooks > edit', () => {
...
+
+                    // * Verify that the toggle is unchecked
+                    cy.get('#user-joins-channel-categorize label input').should(
```
Phrynobatrachus, line 1209: Going forward we'd like to avoid using this method of selection in favor of [accessible queries] (https://developers.mattermost.com/contribute/webapp/e2e/which-query-to-use/) when possible. In this particular case I'm thinking of an `aria-label` on the `input`, which would enable the use of `cy.findByLabelText` instead (see the last section of the [RTL example](https://testing-library.com/docs/queries/bylabeltext/)) and decouples the test a bit more from the DOM semantics. You don't need to do that in this PR unless you'd like, just something to keep in mind.

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -1195,5 +1195,240 @@ describe('playbooks > edit', () => {
...
+
+                        // # Click on the toggle to enable the setting
+                        cy.get('label input').click({force: true});
+
+                        // * Verify that the toggle is unchecked
+                        cy.get('label input').should('be.checked');
+                    });
+                });
+
+                it('does not let select a category when disabled', () => {
```
Phrynobatrachus, line 1233:
```suggestion
                it('prevents category selection when disabled', () => {
```

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -1195,5 +1195,240 @@ describe('playbooks > edit', () => {
...
+                        // * Verify that the control still shows the selected category
+                        cy.get('#playbook-automation-categorize-playbook-run .channel-selector__control').contains(
+                            'Favorites'
+                        );
+                    });
+                });
+
+                it('shows new category name when category was created', () => {
+                    // # Visit the selected playbook
+                    cy.visit('/playbooks/playbooks/' + testPlaybook.id + '/edit');
```
Phrynobatrachus, line 1390: note that here you're using string concatenation, in the other `visit` calls you're using string interpolation. similar to the review comment above, I suggest you pull the `visit -> switch to 'Actions'` sequence out into a `beforeEach` or helper function.

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -1195,5 +1195,240 @@ describe('playbooks > edit', () => {
...
+                    .type('Custom category' + '{enter}', {delay: 200});
+
+                    // # Save the playbook
+                    cy.findByTestId('save_playbook').click();
+
+                    // # Visit the selected playbook
+                    cy.visit('/playbooks/playbooks/' + testPlaybook.id + '/edit');
+
+                    // # Switch to Actions tab
+                    cy.get('#root').findByText('Actions').click();
```
Phrynobatrachus, line 1418: same note as above

Phrynobatrachus (changes requested):

tests-e2e/cypress/integration/playbooks/edit_spec.js
dunak-debug: I will keep this in mind for future, but honestly I don't have that much time from now so somebody else could take care of this specific reusable toggle component to introduce aria-label

agarciamontoro (approved):
This is fantastic, @dunak-debug, thank you very much for all your hard work! Also, the coverage on the E2E tests is great, thanks!

Phrynobatrachus (approved):
Thanks for the changes, LGTM!

dunak-debug:
Thank you 👍 I am happy

PR #843, opened 2021-10-28
[MM-39570] Fix number of webhooks check

iomodo:
#### Summary
This PR Fix number of webhooks check

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-39570

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/automation/patterned_text_area.tsx
+++ b/webapp/src/components/backstage/automation/patterned_text_area.tsx
@@ -33,7 +33,7 @@ export const PatternedTextArea = (props: Props) => {
             return;
         }
 
-        if (props.maxRows && urls.split(props.delimiter || '\n').filter((v) => v.trim().length).length > props.maxRows) {
+        if (props.maxRows && urls.split(props.delimiter || '\n').filter((v) => v.trim().length > 0).length > props.maxRows) {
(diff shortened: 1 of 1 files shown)


webapp/src/components/backstage/automation/patterned_text_area.tsx
```diff
@@ -33,7 +33,7 @@ export const PatternedTextArea = (props: Props) => {
             return;
         }
 
-        if (props.maxRows && urls.length > props.maxRows) {
+        if (props.maxRows && urls.split(props.delimiter || '\n').filter((v) => v.trim().length).length > props.maxRows) {
```
agarciamontoro, line 36: Super nit: is it possible to return a literal boolean in the filter function? Something like this:
```suggestion
        if (props.maxRows && urls.split(props.delimiter || '\n').filter((v) => v.trim().length > 0).length > props.maxRows) {
```

I sometimes find difficult to read when these implicit conversions to boolean happen.

agarciamontoro (approved):
Makes sense, thank you! I left a non-blocking comment below, but please feel free to ignore if you don't agree!

PR #844, opened 2021-10-28
MM-39285 - Support snoozing status update reminders

cpoile:
#### Summary
- Snooze reminders
- @abhijit-singh I am also adding the last reminder that was set, (that's the `in less than 1 minute` option in the list there. Let me know if you don't want that.  (the `35 minute` option is the default timer for that playbook)

Looks like:
image
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-39285

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


server/api/playbook_runs.go
```diff
@@ -818,6 +819,55 @@ func (h *PlaybookRunHandler) reminderDelete(w http.ResponseWriter, r *http.Reque
...
+	if err != nil {
+		err = errors.Wrapf(err, "reminderReset: no playbook run for payload's channelID: %s", payload.ChannelID)
+		h.HandleErrorWithCode(w, http.StatusBadRequest, "no playbook run for payloads's channelID", err)
+		return
+	}
+	if storedPlaybookRunID != playbookRunID {
+		h.HandleErrorWithCode(w, http.StatusBadRequest, "error resetting reminder",
+			fmt.Errorf("storedPlaybookRunID: %s for channelID: %s did not match playbookID in url path: %s", storedPlaybookRunID, payload.ChannelID, playbookRunID))
+		return
+	}
```
agarciamontoro, line 851: This just checks that the provided channel ID in the payload corresponds to the run ID provided in the path, right? Probably I'm missing something, but can't we avoid sending the channel ID and simply using the run ID in the path?

server/app/playbook_run.go
```diff
@@ -475,9 +475,9 @@ type PlaybookRunService interface {
 	// NukeDB removes all playbook run related data.
 	NukeDB() error
 
-	// SetReminder sets a reminder. After timeInMinutes in the future, the owner will be
-	// reminded to update the playbook run's status.
-	SetReminder(playbookRunID string, timeInMinutes time.Duration) error
+	// SetReminder sets a reminder. After fromNow (time.Now().Add(fromNow)) in the future,
```
agarciamontoro, line 478: Super nit:
```suggestion
	// SetReminder sets a reminder. After `time.Now().Add(fromNow)` in the future,
```

I first read this as `fromNow` being a function and `time.Now().Add(fromNow)` being an argument. Probably because it's late and I'm tired, but I guess it doesn't hurt to mention.

server/app/reminder.go
```diff
@@ -182,3 +168,57 @@ func (s *PlaybookRunServiceImpl) ResetReminderTimer(playbookRunID string) error
...
+	// Remove reminder post (if any)
+	if playbookRunToModify.ReminderPostID != "" {
+		if err = s.removePost(playbookRunToModify.ReminderPostID); err != nil {
+			return err
+		}
+		playbookRunToModify.ReminderPostID = ""
+	}
+
+	playbookRunToModify.PreviousReminder = newReminder
+	playbookRunToModify.LastStatusUpdateAt = model.GetMillis()
```
agarciamontoro, line 193: I understand that this is for making the countdown timer work with the snoozed timer, but we're technically not making a status update, so maybe we (probably me, because it's my fault, hehe) should refactor that logic of `PreviousReminder + LastStatusUpdateAt`, introducing an actual `NextStatusUpdateAt` or similar.
Something for a future PR, though, no changes requested here, but I feel like this current weirdness could introduce some bugs in the future if we're not super careful.

server/app/playbook_run_service_test.go
```diff
@@ -540,11 +540,11 @@ func TestUpdateStatusWebhookFailure(t *testing.T) {
 	store.EXPECT().CreateTimelineEvent(gomock.AssignableToTypeOf(&app.TimelineEvent{}))
 	store.EXPECT().UpdatePlaybookRun(gomock.AssignableToTypeOf(&app.PlaybookRun{})).Return(nil)
 	store.EXPECT().UpdateStatus(gomock.AssignableToTypeOf(&app.SQLStatusPost{})).Return(nil)
-	store.EXPECT().GetPlaybookRun(gomock.Any()).Return(playbookRun, nil).Times(4)
+	store.EXPECT().GetPlaybookRun(gomock.Any()).Return(playbookRun, nil).Times(5)
```
agarciamontoro, line 543: Cannot wait for @crspeller's new tests and forget about these :joy:

agarciamontoro (approved):
Awesome! I left a couple of comments: something I don't understand, a super nit and two literal comments that do not request any change.

So approving already, thank you :)

webapp/src/components/update_request_post.tsx
```diff
@@ -37,6 +44,33 @@ export const UpdateRequestPost = (props: Props) => {
         return null;
     }
 
+    const options = [
+        makeOption('in 60 minutes', Mode.DurationValue),
+        makeOption('in 24 hours', Mode.DurationValue),
+        makeOption('in 7 days', Mode.DurationValue),
```
calebroseland, line 50: @abhijit-singh / @cpoile would it be worth adding 15 &/ 30 minute options? When @lieut-data mentioned this feature one time, I remember him pitching more short-term snooze options (perhaps: 15 mins, 60 mins, 24 hrs).

calebroseland (commented):
This is great, @cpoile! 

Some UX observations:

If `Click to open threads` user setting is on, when trying to open the select, RHS's create_comment steals focus and it closes right away. (This may be unfixable here or an effect of that feature, as it has it's own [check function](https://github.com/mattermost/mattermost-webapp/pull/8342/files#diff-efa0d8b3bc74afa1fa4a36331dfa53b1549a04392979e2a2d83fdd914ffc2f4a) for elements to allow/ignore clicks from.)

CleanShot 2021-11-02 at 17 30 00


Also, much more of a minor nit, but it also disappears without css hover state on the post. (It'd be nice to fix, but non-blocker for me; I'm not sure it's fixable without a mattermost-webapp PR.)
CleanShot 2021-11-02 at 17 33 32

server/app/playbook_run.go
cpoile: Hah, I see it, fixed. :)

server/app/reminder.go
cpoile: Yep, it is a little weird, agreed. Future PR for sure.

webapp/src/components/update_request_post.tsx
```diff
@@ -60,6 +94,18 @@ export const UpdateRequestPost = (props: Props) => {
...
+                    classNamePrefix='channel-selector'
+                    filterOption={null}
+                    isMulti={false}
+                    menuPlacement={'top'}
+                    components={{IndicatorSeparator: () => null}}
+                    placeholder={formatMessage({defaultMessage: 'Snooze'})}
+                    options={options}
+                    onChange={snoozeFor}
+                    styles={customStyles}
+                />
```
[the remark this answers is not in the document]
cpoile: Good eye, thanks!

webapp/src/components/update_request_post.tsx
```diff
@@ -37,6 +44,33 @@ export const UpdateRequestPost = (props: Props) => {
...
+    }
+    options.sort((a, b) => ms(a.value) - ms(b.value));
+
+    const snoozeFor = (option: Option) => {
+        resetReminder(currentRun.id, channel.id, ms(option.value) / 1000);
+    };
+
+    const customStyles: ComponentProps<typeof Select>['styles'] = {
+        menu: (provided) => ({...provided, zIndex: 7}),
+    };
```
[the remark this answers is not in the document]
cpoile: Not exactly sure what you mean by "static" here -- can you elaborate?

webapp/src/components/update_request_post.tsx
cpoile: I'll ping in the channel just in case, thanks!

cpoile:
@calebroseland Thank you for pointing out the menu overlapping issue! It's fixed by using a portal to `document.body`. Pretty clean (compared to the other few things I tried to make work... yeesh.)

calebroseland (approved):

PR #850, opened 2021-11-01
Add useRelative with playbooks

clovis1122:
#### Summary

Changes the way we display dates for playbooks by making them relative.

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


webapp/src/components/rhs/post_card.tsx
```diff
@@ -124,6 +124,7 @@ const PostCard = (props: Props) => {
                             timestamp: (
                                 <Timestamp
                                     value={props.post.create_at}
+                                    useRelative
```
crspeller, line 127: Linter wants this to be: 
```suggestion
                                    useRelative={true}
```

crspeller (changes requested):
Thanks for the PR! Just one lint fix.

webapp/src/components/rhs/post_card.tsx
clovis1122: Thanks! Done!

crspeller (approved):
Thanks!

PR #854, opened 2021-11-01
MM-39165: Make "Go to channel" more prominent

cpoile:
#### Summary
- Style the "Go to channel" button as primary
- Render it to the right of all other buttons
- The icon should always be the one in the ticket

@abhijit-singh The icons are the same size as before, but I was wondering if you would prefer them a bit smaller?
image

#### Ticket Link
- 

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_runs/shared.tsx
```diff
@@ -70,13 +70,23 @@ export const SecondaryButtonLargerRight = styled(SecondaryButtonLarger)`
     margin-left: auto;
 `;
 
+export const ExpandRight = styled.div`
+    margin-left: auto;
+`;
```
agarciamontoro, line 75: Neat!

agarciamontoro (approved):
Thanks!

iomodo (approved):
Looks Great!

abhijit-singh (approved):
I agree @cpoile. They can be 16px instead of the current 18px. Everything else looks great!

webapp/src/components/backstage/playbook_runs/shared.tsx
```diff
@@ -70,16 +70,32 @@ export const SecondaryButtonLargerRight = styled(SecondaryButtonLarger)`
...
     position: unset;
     height: unset;
     white-space: nowrap;
 `;
+
+export const Icon16 = styled.div`
+    > i {
+      font-size: 16px;
+    }
+`;
```
cpoile, line 101: @calebroseland I hate doing this, but directly styling the `i` element doesn't work (the `font-size` stays `18px`). Do you know why?

calebroseland: Selector specificity, I think. It's not that pretty, but if you need to arbitrarily increase it, there's:

```tsx
export const Icon16 = styled.i`
    && {
        font-size: 16px;
    }
`;
```

cpoile: Thank you, @calebroseland! That was what I was looking for.

PR #858, opened 2021-11-02
MM-39145: Add description to playbook

agarciamontoro:
#### Summary

This PR addresses the missing description in playbooks, adding a way to edit it and showing it in the preview page.

Before this PR, the `IR_Playbook` table already had a `Description` column, but it was actually being used as the run summary template: the one we used in the `Templates` section and whose value we retrieved when creating a new run to pre-populate the run summary in the RHS. To solve this, I've added a new `RunSummaryTemplate` column, and copied the data from the `Description` column into it, setting the `Description` column to the empty string. I've adapted the places in the server and the client where we were using the description to now use the run summary template, so everything should be a bit more clear now.

Now that we have an actual description of the playbook (the repurposed `Description` column), we can show it in the playbook preview, as well as edit it in the playbook edit page, for which we're using a modal that updates both the title and the optional description. As we had a long while ago, the description is previewed in the edit page right below the title.

https://user-images.githubusercontent.com/3924815/139922904-a9d8e015-a4d8-4145-92aa-099553c48056.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39145

#### Checklist

- [ ] ~~Telemetry updated~~
- [X] Unit tests updated


server/api/playbook_runs.go
```diff
@@ -377,7 +377,7 @@ func (h *PlaybookRunHandler) createPlaybookRun(playbookRun app.PlaybookRun, user
 		playbookRun.Checklists = pb.Checklists
 		public = pb.CreatePublicPlaybookRun
 
-		playbookRun.Description = pb.Description
+		playbookRun.Description = pb.RunSummaryTemplate
```
lieut-data, line 380: Would it make sense to rename /this/ `Description` to `Summary` as well? Otherwise, the competing names here seem unfortunate.

server/sqlstore/migrations.go
```diff
@@ -1462,6 +1462,32 @@ var migrations = []Migration{
...
+	},
+	{
+		fromVersion: semver.MustParse("0.38.0"),
+		toVersion:   semver.MustParse("0.39.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DatabaseDriverMysql {
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "RunSummaryTemplate", "TEXT"); err != nil {
+					return errors.Wrapf(err, "failed adding column RunSummaryTemplate to table IR_Playbook")
+				}
+				if _, err := e.Exec("UPDATE IR_Playbook SET RunSummaryTemplate = '' WHERE RunSummaryTemplate IS NULL"); err != nil {
```
lieut-data, line 1477: Why do we need to do this? Can we `COALESCE` at the query instead, gaining the protection against `NULL` values forever?

server/sqlstore/migrations.go
```diff
@@ -1462,6 +1462,32 @@ var migrations = []Migration{
...
+					return errors.Wrapf(err, "failed updating default value of column RunSummaryTemplate from table IR_Playbook")
+				}
+			} else {
+				if err := addColumnToPGTable(e, "IR_Playbook", "RunSummaryTemplate", "TEXT DEFAULT ''"); err != nil {
+					return errors.Wrapf(err, "failed adding column RunSummaryTemplate to table IR_Playbook")
+				}
+			}
+
+			// Copy the values from the Description column, historically used for the run summary template, into the new RunSummaryTemplate column
+			if _, err := e.Exec("UPDATE IR_Playbook SET RunSummaryTemplate = Description, Description = '' WHERE Description <> ''"); err != nil {
```
lieut-data, line 1487: nit: This swapping of variables seems like the kind of thing MySQL would get wrong. Is there a way we can unit test this transition? Or maybe just break up into separate queries? (I might be paranoid...)

webapp/src/components/backstage/playbooks/playbook_preview_description.tsx
```diff
@@ -0,0 +1,54 @@
...
+    id: string;
+    playbook: PlaybookWithChecklist;
+}
+
+const PlaybookPreviewDescription = (props: Props) => {
+    const {formatMessage} = useIntl();
+    const markdownOptions = useDefaultMarkdownOptionsByTeamId(props.playbook.team_id);
+    const renderMarkdown = (msg: string) => messageHtmlToComponent(formatText(msg, markdownOptions), true, {});
+
+    if (props.playbook.description === '') {
```
lieut-data, line 25: Should we handle pure whitespace descriptions?

webapp/src/components/backstage/playbooks/playbook_preview_navbar.tsx
```diff
@@ -34,7 +35,7 @@ const headersOffset = 140;
 const PlaybookPreviewNavbar = ({playbookId, runsInProgress}: Props) => {
     const {formatMessage} = useIntl();
     const match = useRouteMatch();
-    const [activeId, setActiveId] = useState(SectionID.Checklists);
+    const [activeId, setActiveId] = useState(SectionID.Description);
```
lieut-data, line 38: I'm having trouble parsing this -- why do we initialize to the description here, and why did we initialize to checklists before?

lieut-data (approved):
Looks good, but a handful of comments.

iomodo (approved):
Looks great Alejandro!

server/api/playbook_runs.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/858/commits/e38bf811da5bf77785374590ae6a28477c5e6c70 and https://github.com/mattermost/mattermost-plugin-playbooks/pull/858/commits/1f5534d4d279aae297466449e1c57876a6c91cb0!

server/sqlstore/migrations.go
agarciamontoro: I kept this and added the `COALESCE`. Just in case. See https://github.com/mattermost/mattermost-plugin-playbooks/pull/858/commits/0be4c2a86162b0cbd30d499d53e6d7c5476a8593

server/sqlstore/migrations.go
agarciamontoro: Looks like it worked :) Tested in https://github.com/mattermost/mattermost-plugin-playbooks/pull/858/commits/352d5f0f670c2e065c297b84644569f4834808fc

webapp/src/components/backstage/playbooks/playbook_preview_description.tsx
agarciamontoro: Nice catch. Fixed in https://github.com/mattermost/mattermost-plugin-playbooks/pull/858/commits/485807ff15b6565edbafba2ccab677c835753ef0

webapp/src/components/backstage/playbooks/playbook_preview_navbar.tsx
agarciamontoro: We need to store which section is highlighted. So we need to initialize the state with *some* data. If we initialize this to something else (null or undefined), we would need to call `updateActiveSection` when the component mounts. We avoid that initializing the state to the first item in the list (which before was checklists, now is description).

Unless I'm misunderstanding your parsing troubles. Were you thinking of something else?

agarciamontoro:
Ugh, I found a bug, I'll merge after I fix it.

webapp/src/components/backstage/playbooks/playbook_preview_navbar.tsx
lieut-data: Oh! I completely misparsed this -- somehow, I thought we actually stored the checklists before and the description, and missed that this was just a section id for those categories. Disregard :D

agarciamontoro: But... I just realized that this breaks as soon as the Description is not there! I'll use the first rendered section (using the new logic in https://github.com/mattermost/mattermost-plugin-playbooks/pull/858/commits/dbe0bd3791f5c668a722e93cb557e17b21dce536) as the initial state.

agarciamontoro: Nevermind, I actually ended up calling  `updateActiveSection` on component mount. My first idea was a bit too complex.

agarciamontoro:
Sorry for the double request, @lieut-data, but I wanted to ask for a second review on the last two commits, which I came up with after your first review.

lieut-data:
Looks good to me, @agarciamontoro!

lieut-data (approved):

PR #859, opened 2021-11-02
Reorganize E2E Tests

lieut-data:
#### Summary
Organize into the following directory structure, rendering the term `backstage` obsolete in this context and treating `channels` as an integration and not the `frontstage`.
  - cypress/integration (all specs live here)
    - playbooks/
    - runs/
    - channels/
      - rhs/
      - slash_command/
    - (other global tests)

The top level `describe` of each spec now contains the path (e.g.  `channels > rhs > spec_name`). There's probably still room for improvement here, both in the organization and naming, but this gets us past the hump as it were.

I'm proposing we do something similar for the `webapp` component. Open for discussion -- hoping this makes the code easier to navigate.

#### Ticket Link
N/A

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated?

### DIFF
--- a/tests-e2e/cypress/integration/channels/actions_spec.js
+++ b/tests-e2e/cypress/integration/channels/actions_spec.js
@@ -6,7 +6,7 @@
 // - [*] indicates an assertion (e.g. * Check the title)
 // ***************************************************************
 
-describe('playbook run automation', () => {
+describe('channels > actions', () => {
     let testTeam;
     let testUser;
     let testPublicChannel;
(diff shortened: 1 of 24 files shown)


cpoile (approved):
I was confused about the `runs/` directory being top level -- shouldn't it be under the `playbooks` directory?

lieut-data:
> I was confused about the `runs/` directory being top level -- shouldn't it be under the `playbooks` directory?

I was trying to avoid the `/playbooks/playbooks` problem, and instead assume the global directory was "our product", one of whose features is a `channels` integration. (I guess I was also implicitly shortcutting `playbook_run` as `run` everywhere as previously proposed.)

iomodo (approved):
Wonderful! Shouldn't we be doing same for the code in `webapp/componenets`?

crspeller (approved):
Yay organization!

lieut-data:
> Wonderful! Shouldn't we be doing same for the code in webapp/componenets?

Yes, that would be ideal. Definitely something to move towards after we get this merged and it continues to "feel good."

PR #860, opened 2021-11-02
MM-39501: Close existing RHS when running playbook

cpoile:
#### Summary
- When running a playbook, force the run's RHS open when landing on that new run's channel (even if there was a thread or any other kind of mattermost RHS open)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-39501

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/rhs_opener.ts
```diff
@@ -58,13 +59,26 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
...
 
-        // Don't do anything unless we're in a playbook run channel.
-        if (!currentChannelIsPlaybookRun) {
+        // Should we force open the RHS?
+        const url = new URL(window.location.href);
+        const searchParams = new URLSearchParams(url.searchParams);
+        if (searchParams.has('forceRHSOpen')) {
+            searchParams.delete('forceRHSOpen');
+            url.search = searchParams.toString();
+            browserHistory.replace({pathname: url.pathname, search: url.search});
```
cpoile, line 73: We could have also used redux, but this allows us to force the RHS open from markdown links too, so I settled on this.

webapp/src/types/actions.ts
```diff
@@ -30,6 +30,8 @@ export const SET_RHS_ABOUT_COLLAPSED_STATE = pluginId + '_set_rhs_about_collapse
 export const SET_CHECKLIST_COLLAPSED_STATE = pluginId + '_set_checklist_collapsed_state';
 export const SET_ALL_CHECKLISTS_COLLAPSED_STATE = pluginId + '_set_all_checklists_collapsed_state';
 export const SET_CHECKLIST_ITEMS_FILTER = pluginId + '_set_checklist_items_filter';
+export const MM_UPDATE_RHS_STATE = 'UPDATE_RHS_STATE';
+export const MM_SELECT_POST = 'SELECT_POST';
```
cpoile, line 34: These are copies of the Webapp's constants -- I couldn't import them because it drew in all the other files the constants file depends on and it broke webpack (probably fixable, but not worth the effort).

webapp/src/actions.ts
```diff
@@ -174,6 +171,23 @@ export function toggleRHS() {
...
+                state: null,
+            },
+            {
+                type: MM_SELECT_POST,
+                postId: '',
+                channelId: '',
+                timestamp: 0,
+            },
+        ]));
+    };
```
cpoile, line 188: Also copied from Webapp.

webapp/src/websocket_events.ts
```diff
@@ -97,9 +97,14 @@ export function handleWebsocketPlaybookRunCreated(getState: GetStateFunc, dispat
 
         const currentTeam = getCurrentTeam(getState());
 
+        // Pause to let the server catch up (sometimes we navigate so quickly, the server says
+        // the channel does not exist)
+        await new Promise((r) => setTimeout(r, 500));
```
cpoile, line 102: I'm not totally happy with this, but I can't think of any other way to fix the issue the e2e tests were having -- they were navigating to the newly created channel and the server was returning a "channel not found" error screen. I doubt this would happen in a real situation, but at least this short delay is not really noticeable (it completes just after the interactive dialog closes, so I can't tell any difference from the original without the delay).

webapp/src/browser_routing.ts
```diff
@@ -6,7 +6,12 @@ import React from 'react';
 // @ts-ignore
 const WebappUtils = window.WebappUtils;
 
-export const navigateToUrl = (urlPath: string) => {
+type PathLike = {
+    pathname: string;
+    search: string;
+}
+
+export const navigateToUrl = (urlPath: string | PathLike) => {
```
agarciamontoro, line 14: Neat!

webapp/src/websocket_events.ts
agarciamontoro: If this does not happen in a real situation, would it be worth it simply fixing it in the e2e tests? I'm concerned that in the future this websocket event is received in scenarios different than the one with the dialog, so this could be not needed.

I don't know, I think we've not received any report of this happening in real life, right? Or did we?

agarciamontoro (approved):
Looks good, I like the solution! The only mild concern I may have is what you pointed out you don't like (we can discuss in the comment I left there), but I'm ok merging it, so approving to unblock.

cpoile:
@agarciamontoro For some reason I can't respond to your last comment directly. Anyway, I can't think of a way to force the delay in the e2e tests, since the navigation is triggered by the websocket event itself. Alternatively we could delay the websocket event, to give the server some time to "catch up". But I'm totally open to suggestions if you can think of anything?

webapp/src/websocket_events.ts
```diff
@@ -97,9 +97,14 @@ export function handleWebsocketPlaybookRunCreated(getState: GetStateFunc, dispat
...
+
         // Navigate to the newly created channel
-        const url = `/${currentTeam.name}/channels/${playbookRun.channel_id}`;
-        navigateToUrl(url);
+        const pathname = `/${currentTeam.name}/channels/${payload.channel_name}`;
```
lieut-data, line 105: Can you elaborate on the need to switch from navigating using the `channel_id` to using the `channel_name` instead? Does it have to do with the redirect from `id` to `name` losing the desired search parameters?

webapp/src/websocket_events.ts
lieut-data: If we need this workaround, should we do it (plugin) server-side instead, delaying the websocket event until the server "catches up"?

tests-e2e/cypress/integration/frontstage/rhs/open_spec.js
```diff
@@ -284,6 +284,36 @@ describe('playbook run rhs', () => {
...
+            // # Wait until the channel loads enough to show the post textbox.
+            cy.get('#post-create').should('exist');
+
+            // # Open the flagged posts RHS
+            cy.get('body').then(($body) => {
+                if ($body.find('#channelHeaderFlagButton').length > 0) {
+                    cy.get('#channelHeaderFlagButton').click({force: true});
+                } else {
+                    cy.findByRole('button', {name: 'Select to toggle a list of saved posts.'}).click({force: true});
+                }
```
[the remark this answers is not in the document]
lieut-data: Yeah, we can probably rip this out anytime -- I had added support for this while we were testing against both configurations, but a non-global-header is no longer in scope.

lieut-data (approved):
A handful of comments, but nothing strictly blocking -- one less sharp edge to the product! :)

webapp/src/websocket_events.ts
cpoile: Yep, server side works for me. There are two places we send the event, but I only delayed the event from the dialog (there's no need to delay the runs enpoint, and doing so would add unneeded complexity).

tests-e2e/cypress/integration/frontstage/rhs/open_spec.js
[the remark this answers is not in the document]
cpoile: Okay, I ripped it out (and will for the other PR too). Thanks!

webapp/src/websocket_events.ts
cpoile: Yes, that's exactly why. We could file a ticket to fix that behaviour, if we think it should be?

lieut-data: 0/5, but thank you for clarifying :)

lieut-data (approved):

webapp/src/websocket_events.ts
agarciamontoro: I like that solution :)

PR #862, opened 2021-11-03
[MM-38586] Skip instead of delete playbook run tasks

shadowshot-x:
#### Summary
Added the Skip and Restore Functionality to Playbook tasks. Eliminated the Delete feature. 

1. Users can now Skip the task. This will result in the Task being crossed off from the list. Earlier the user could delete the task which led to the vanishing of the task from the checklist items.

2. The Skipped tasks can be restored by the user. This will lead the tasks to be restored and a normal task title will be displayed.

3. When a task is skipped, the status of the task is changed to "skipped". Also, a LastSkipped field or "deleted_at" is added which contains the current timestamp.
4. When Restoring the task, only the status is updated to "open". The LastSkipped field is not changed.
5. Whenever Skipping or Restoring the Task, the Playbooks Bot will notify the user.

Screenshot from 2021-11-06 21-52-31
Screenshot from 2021-11-03 12-20-06
Screenshot from 2021-11-06 21-55-36
Screenshot from 2021-11-03 12-19-46

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/18596

#### Checklist

- ~[ ]~ Telemetry updated
- ~[ ]~ Store Updated to contain deleted_at
- ~[ ]~ New Routes and Methods added in Server
- ~[ ]~ Update the Progress bar based on deletion
- ~[ ]~ Add End Test Cases


lieut-data:
Wow, this looks like a pretty comprehensive change! Thank you, @shadowshot-x!! I've invited @cpoile and @crspeller to help review :)

shadowshot-x:
Thank you @lieut-data. Happy to help. :D

server/app/playbook.go
```diff
@@ -142,8 +142,8 @@ type ChecklistItem struct {
 	// Title is the content of the checklist item.
 	Title string `json:"title"`
 
-	// State is the state of the checklist item: "closed" if it's checked, the empty string
-	// otherwise.
+	// State is the state of the checklist item: "closed" if it's checked, "skipped" if it has
+	// been skipped, the empty string otherwise.
```
cpoile, line 146: That's a good choice!

server/app/playbook_run.go
```diff
@@ -458,6 +458,11 @@ type PlaybookRunService interface {
 	// RemoveChecklistItem removes an item from the specified checklist
 	RemoveChecklistItem(playbookRunID, userID string, checklistNumber int, itemNumber int) error
 
+	// SkipChecklistItem removes an item from the specified checklist
+	SkipChecklistItem(playbookRunID, userID string, checklistNumber int, itemNumber int) error
+
+	RestoreChecklistItem(playbookRunID, userID string, checklistNumber int, itemNumber int) error
```
cpoile, line 464: Could we add a code comment here (although it's pretty self-explanatory, we should probably be consistent).

server/app/playbook.go
```diff
@@ -174,6 +174,10 @@ type ChecklistItem struct {
 
 	// Description is a string with the markdown content of the long description of the item.
 	Description string `json:"description"`
+
+	// LastSkipped is the timestamp, in milliseconds since epoch, of the last time the item
+	// was skipped. 0 if it was never skipped.
+	LastSkipped int64 `json:"deleted_at"`
```
cpoile, line 180: I was reading the ticket description, and it mentioned setting a `delete_at` field to be consistent with how we model channels -- was that specification changed to this `LastSkipped` field?

webapp/src/components/checklist_item.tsx
```diff
@@ -462,18 +465,22 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
                                     }}
                                 />
                                 <HoverMenuButton
-                                    title={formatMessage({defaultMessage: 'Delete'})}
-                                    className={'icon-trash-can-outline icon-16 btn-icon'}
+                                    title={(props.checklistItem.state === ChecklistItemState.Skip) ? 'Restore' : 'Skip'}
+                                    className={(props.checklistItem.state === ChecklistItemState.Skip) ? 'icon-refresh icon-16 btn-icon' : 'icon-close-circle-outline icon-16 btn-icon'}
```
cpoile, line 469: I think you'll have to use `formatMessage`s for `Restore` and `Skip`, for internationalization.

webapp/src/components/checklist_item.tsx
```diff
@@ -486,7 +493,7 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
                             <div
                                 onClick={((e) => handleFormattedTextClick(e, relativeTeamUrl))}
                             >
-                                {messageHtmlToComponent(formatText(title, markdownOptions), true, {})}
+                                {(props.checklistItem.state === ChecklistItemState.Skip) ? <StrikeThrough>{messageHtmlToComponent(formatText(props.checklistItem.title, markdownOptions), true, {})}</StrikeThrough> : messageHtmlToComponent(formatText(props.checklistItem.title, markdownOptions))}
```
cpoile, line 496: This is getting a little long, maybe we should refactor it into a const?

webapp/src/components/checklist_item.tsx
```diff
@@ -532,14 +539,34 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
...
             <ConfirmModal
-                show={showDeleteConfirm}
-                title={formatMessage({defaultMessage: 'Delete task'})}
-                message={formatMessage({defaultMessage: 'Are you sure you want to delete this task? This will be removed from this run but will not affect the playbook.'})}
-                confirmButtonText={formatMessage({defaultMessage: 'Delete'})}
-                onConfirm={() =>
-                    clientRemoveChecklistItem(props.playbookRunId, props.checklistNum, props.itemNum)
+                show={showSkipConfirm}
+                title={'Skip task'}
+                message={'Are you sure you want to skip this task? This will be crossed from this run but will not affect the playbook.'}
```
cpoile, line 544: We should probably use formatMessages for all these strings.

cpoile (changes requested):
This is awesome work @shadowshot-x , really nicely done, thank you! A few requests for changes (mostly to bring things a bit more in line with our current style/patterns). But otherwise this is almost ready to go. Great work!

server/app/playbook.go
shadowshot-x: Hi @cpoile, The ticket does mention `delete_at` field. However, I used LastSkipped because we are actually removing the Deletion functionality as a whole for the tasks. Do you want me to change this to `DeleteAt` instead of `LastSkipped`?

server/app/playbook_run.go
shadowshot-x: Right, I might have missed it.

server/app/playbook.go
shadowshot-x: Thank You

webapp/src/components/checklist_item.tsx
shadowshot-x: Done!

webapp/src/components/checklist_item.tsx
shadowshot-x: Refactored it.

webapp/src/components/checklist_item.tsx
shadowshot-x: Done! Thank You for reminding me.

server/app/playbook.go
crspeller: We should at least be consistent here with the json representation.

crspeller (approved):
Great! Thanks for the contribution! Just the one comment on the outstanding discussion.

server/app/playbook.go
shadowshot-x: Sure, I will change this to `LastSkipped int64` `json:"delete_at"`.

Phrynobatrachus:
hi @shadowshot-x! I [opened a PR](https://github.com/shadowshot-x/mattermost-plugin-playbooks/pull/1) to your branch to add a touch more detail to the Cypress test, but otherwise it's LGTM.

Phrynobatrachus (approved):

PR #866, opened 2021-11-03
New go e2e style tests

crspeller:
#### Summary
These are the new E2E/server style golang tests I have been talking about extracted from my permissions branch. 

The fun is in `server/main_test.go` with the other `_test` files mirroring the tests in the api directory they are designed to replace.  This is just a start so that others can get the ball rolling. 

Mentioning everyone as an FYI feel free to review: @cpoile @lieut-data @agarciamontoro @calebroseland @iomodo


crspeller:
@lieut-data or @cpoile Could one or both of you check to make sure this works for you locally? Plugins stuff has some platform differences.

lieut-data:
It almost passed, though I had a weird failure -- maybe specific to me? Seems like MySQL died under the weight?
```
=== Failed
=== FAIL: server/sqlstore TestPlaybookRunStore_UpdateTimelineEvent (7.66s)
Created temporary postgres database dbbtz8ttqn678ejftnrrnzs3pf7o
    store.go:126: Unexpected call to *mock_bot.MockLogger.Errorf([Failed to rollback transaction; err: %v invalid connection]) at /Users/jesse/Documents/mattermost/src/plugins/playbooks/server/bot/mocks/mock_logger.go:60 because: there are no expected calls of the method "Errorf" for that receiver
[mysql] 2021/11/04 11:48:25 packets.go:37: unexpected EOF
[mysql] 2021/11/04 11:48:25 packets.go:37: unexpected EOF
[mysql] 2021/11/04 11:48:25 packets.go:37: unexpected EOF
Dropped temporary database dbbtz8ttqn678ejftnrrnzs3pf7o
Created temporary mysql database dbtdof7waanibbxx1njex48nf53e
[mysql] 2021/11/04 11:48:25 packets.go:37: unexpected EOF
    --- PASS: TestPlaybookRunStore_UpdateTimelineEvent/Save_4_and_delete_2_timeline_events (0.03s)
panic: failed to drop temporary database dbtdof7waanibbxx1njex48nf53e: failed to execute `DROP DATABASE dbtdof7waanibbxx1njex48nf53e` against mysql database as root: driver: bad connection [recovered]
        panic: failed to drop temporary database dbtdof7waanibbxx1njex48nf53e: failed to execute `DROP DATABASE dbtdof7waanibbxx1njex48nf53e` against mysql database as root: driver: bad connection
```

client/playbooks.go
```diff
@@ -68,6 +68,11 @@ func (s *PlaybooksService) List(ctx context.Context, teamId string, page, perPag
 
 // Create a playbook.
 func (s *PlaybooksService) Create(ctx context.Context, opts PlaybookCreateOptions) (*Playbook, error) {
+	// For ease of use set the default if not specificed so it doesn't just error
+	if opts.ReminderTimerDefaultSeconds == 0 {
```
lieut-data, line 72: `0` used to mean no reminder, but now I think that's disallowed, right @cpoile?

server/main_test.go
```diff
@@ -0,0 +1,321 @@
...
+	BasicPublicChannel      *model.Channel
+	BasicPublicChannelPost  *model.Post
+	BasicPrivateChannel     *model.Channel
+	BasicPrivateChannelPost *model.Post
+	BasicPlaybook           *client.Playbook
+	AdminUser               *model.User
+	RegularUser             *model.User
+}
+
+func RunTest(t *testing.T, test func(t *testing.T, e *TestEnvironment)) {
```
lieut-data, line 84: I'm not a big fan of the callback, mostly because of the need to indent every test. Is this mostly so we can automate running against MySQL and Postgres?

Could we instead just have some external configuration of which database to use that, and just test against the configured database? It means it's likely we forget to test against one database locally, but it feels like the tradeoffs would be in our favour regardless. And this would be amazing in CircleCI to clearly call out which set of tests were failing vs. having to mesh the permutations inline.e

server/main_test.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/mattermost/mattermost-server/v6/store/storetest"
+	"github.com/mattermost/mattermost-server/v6/utils"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMain(m *testing.M) {
+	// Run the plugin under test if the server is trying to run us as a plugin.
+	value := os.Getenv("MATTERMOST_PLUGIN")
+	if value == "Securely message teams, anywhere." {
```
lieut-data, line 30: This is a string we exchange in plugin startup, right? Is it something we can reference from `model`?

lieut-data (approved):

client/playbooks.go
crspeller: That's why I had to add that here. When I updated to the latest playbooks right before submitting everything started failing complaining about `ReminderTimerDefaultSeconds` being zero.

server/main_test.go
crspeller: I do want to add ability to specify it externally.
I did add this for the purpose of automating multiple runs with different DBs. I think I like the external solution better anyway as long as we can live with `go test` just testing one DB. Which now that I think about it, is fine. 

Going to fix this.

crspeller: @lieut-data Removed the wrapper. For now the tests default to postgres and you can run them in mysql by doing `env TEST_DATABASE_DRIVERNAME=mysql go test` 
Future PRs to unify these with the other DB tests and then fix the CI.

lieut-data: Brilliant! Thank you :)

server/main_test.go
lieut-data: Yeah, not worried about change, mostly about "what in the world is this here for?" :D

client/playbooks.go
cpoile: Yep, and yep.

server/api_playbooks_test.go
```diff
@@ -0,0 +1,84 @@
...
+		})
+		assert.Nil(t, err)
+	})
+
+	t.Run("create playbook, unlicensed with one pre-existing playbook in the team, should fail via licencing", func(t *testing.T) {
+		result, err := e.PlaybooksClient.Playbooks.Create(context.Background(), client.PlaybookCreateOptions{
+			Title:  "test2",
+			TeamID: e.BasicTeam.Id,
+		})
+		requireErrorWithStatusCode(t, err, http.StatusForbidden)
```
cpoile, line 40: How do we know this failed via licencing? :)

server/api_playbooks_test.go
```diff
@@ -0,0 +1,84 @@
...
+		assert.Nil(t, result)
+	})
+
+	e.SetE10Licence()
```
cpoile, line 44: If we turn on randomize test order... Would it increase setup time too much to do server setup in each test to make them isolated?

server/main_test.go
cpoile: I was wondering "what in the world is this" too. :D

server/main_test.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func getEnvWithDefault(name, defaultValue string) string {
+	if value := os.Getenv(name); value != "" {
+		return value
+	}
+	return defaultValue
+}
+
+func Setup(t *testing.T) *TestEnvironment {
```
cpoile, line 91: This whole thing is really cool. Nice work.

cpoile (approved):
I like it, thank you! Just minor questions.

go.mod
```diff
@@ -5,47 +5,22 @@ go 1.14
...
-	google.golang.org/genproto v0.0.0-20211005153810-c76a74d43a8e // indirect
-	google.golang.org/grpc v1.41.0 // indirect
-	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
-	gopkg.in/ini.v1 v1.63.2 // indirect
+	gotest.tools v2.2.0+incompatible
 )
 
 replace github.com/mattermost/mattermost-plugin-playbooks/client => ./client
+
+replace github.com/golang/mock v1.6.0 => github.com/golang/mock v1.4.4
```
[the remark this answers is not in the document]
crspeller: Otherwise the server version update auto updates this and breaks all the tests we are about to delete.

server/api_playbooks_test.go
crspeller: This passes with `-shuffle=on` since we specify the dependencies explicitly.

server/api_playbooks_test.go
crspeller: This passes with `-shuffle=on` since we specify the dependencies explicitly.

server/api_playbooks_test.go
crspeller: We don't. We could change the status code for a licensing error. Separate PR.

PR #870, opened 2021-11-08
[MM-39857] Add auto-follow playbook runs

iomodo:
#### Summary
This PR adds support of auto-following a playbook run

https://user-images.githubusercontent.com/2340127/140747258-4ed90396-fcc8-4a36-899b-9711a4b73429.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-39857

#### Checklist

- ~~[ ] Telemetry updated~~
- [X] Unit tests updated


abhijit-singh (commented):
Thanks @iomodo! A few minor tweaks:
* Can we add the tooltips for when the user hovers on the checkbox-button like in the design? They'll help people understand what the feature is. 
* Can we reduce the left and right spacing within the box to 20px?
image

* Can we change the colors of the text and the button to the following ([figma link](https://www.figma.com/file/Lep365ubDUjUcgV0757zFF/MM-33880-Run-Overview?node-id=744%3A277092)): 
  * Unchecked state - outline & checkbox color: 24% opacity `center channel text`
  * Unchecked state - text color: 56% opacity `center channel text` 
  * Checked state - outline, checkbox & text color: `button background`
image

server/sqlstore/migrations.go
```diff
@@ -1455,11 +1455,56 @@ var migrations = []Migration{
...
+			return nil
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.38.0"),
+		toVersion:   semver.MustParse("0.39.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DatabaseDriverMysql {
+				if _, err := e.Exec(`
+					CREATE TABLE IF NOT EXISTS IR_Playbook_Participants (
```
lieut-data, line 1474: Let's avoid overloading the term participants here, since this concept is currently only defined for a run. Perhaps we simplify to just "PlaybookAutoFollowers" for now?

lieut-data (changes requested):
Thanks, @iomodo! I'm inclined to think we should call the underlying concept, "AutoFollow", as outlined below in a few different comments. Otherwise, looks good!

server/api/playbooks.go
```diff
@@ -54,6 +54,11 @@ func NewPlaybookHandler(router *mux.Router, playbookService app.PlaybookService,
 	playbookRouter.HandleFunc("", handler.deletePlaybook).Methods(http.MethodDelete)
 	playbookRouter.HandleFunc("/restore", handler.restorePlaybook).Methods(http.MethodPost)
 
+	followersRouter := playbookRouter.PathPrefix("/followers").Subrouter()
+	followersRouter.HandleFunc("", handler.follow).Methods(http.MethodPut)
+	followersRouter.HandleFunc("", handler.unfollow).Methods(http.MethodDelete)
+	followersRouter.HandleFunc("/check", handler.isFollower).Methods(http.MethodGet)
```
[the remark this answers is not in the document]
iomodo: `GET` on `/follow` endpoint should return all followers(we will/might need this in the future). Another way to implement this is to introduce filtering on  `GET /follow` endpoint, but I decided to avoid the complexity for now.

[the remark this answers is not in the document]
lieut-data: I could see `/follow` be implicitly for the active user, and `/followers` for the full list. Or maybe `/followers/<user_id>` for the `GET`/`PUT`/`DELETE` with the `/followers`/ for the full list?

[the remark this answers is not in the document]
cpoile: I do like combining `PUT` and `DELETE` to `/followers`, agree with avoiding the complexity of adding a flag for the check.

server/api/playbooks.go
```diff
@@ -531,3 +536,56 @@ func (h *PlaybookHandler) validateCategoryName(categoryName string) error {
...
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	if err := app.PlaybookAccess(userID, playbookID, h.playbookService, h.pluginAPI); err != nil {
+		h.HandleErrorWithCode(w, http.StatusForbidden, "User doesn't have permissions to playbook.", err)
+		return
+	}
+
+	var isFollower bool
+	var err error
+	if isFollower, err = h.playbookService.IsFollower(playbookID, userID); err != nil {
```
cpoile, line 585: I'm starting to hate this pre-declaring vars -- it's `incidental complexity` noise in the worst way. Can we turn off shadowing linting yet? @lieut-data

cpoile (approved):
Nice!

server/api/playbooks.go
lieut-data: This isn't a linting issue, is it? It's that `isFollower` is needed after the conditional? Personally, I'd just lean towards writing all these blocks as:
```go
err := CheckSomething()
if err != nil {
    // ...
}

result, err := DoSomethingElse()
if err != nil {
    // ...
}
```

server/sqlstore/migrations.go
```diff
@@ -1455,11 +1455,56 @@ var migrations = []Migration{
...
+						PlaybookID VARCHAR(26) NULL REFERENCES IR_Playbook(ID),
+						UserID VARCHAR(26) NOT NULL,
+						IsFollower BOOLEAN NOT NULL,
+						INDEX IR_Playbook_Participants_UserID (UserID),
+						INDEX IR_Playbook_Participants_PlaybookID (PlaybookID)
```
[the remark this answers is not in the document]
iomodo: There might be a lot of followers so I think it's more scalable approach.

iomodo:
Take a look @abhijit-singh 
https://user-images.githubusercontent.com/2340127/141168476-dc512c1f-f33f-4342-aadf-61fad3b6eb77.mp4

abhijit-singh (approved):
Thanks @iomodo! Looks great! A couple of tiny nits:
* The font size for `Auto-follow runs` seems to be bit smaller than the font size for the `Run` button. Can you ensure it is the same as that of `Run`?
* In the tooltip when it is unchecked, there is a comma(",") missing after the word `checked`.
Might be good to have @justinegeffen review this.
image

server/api/playbooks.go
```diff
@@ -55,9 +55,11 @@ func NewPlaybookHandler(router *mux.Router, playbookService app.PlaybookService,
 	playbookRouter.HandleFunc("/restore", handler.restorePlaybook).Methods(http.MethodPost)
 
 	followersRouter := playbookRouter.PathPrefix("/followers").Subrouter()
-	followersRouter.HandleFunc("", handler.follow).Methods(http.MethodPut)
-	followersRouter.HandleFunc("", handler.unfollow).Methods(http.MethodDelete)
-	followersRouter.HandleFunc("/check", handler.isFollower).Methods(http.MethodGet)
+	followersRouter.HandleFunc("", handler.autoFollow).Methods(http.MethodPut)
```
lieut-data, line 58: nit: since these operate on the calling user, I'd be inclined to group them all under `/follow` until such time as we expose actions on the set of `/followers`. Thoughts?

server/api/playbooks.go
```diff
@@ -55,9 +55,11 @@ func NewPlaybookHandler(router *mux.Router, playbookService app.PlaybookService,
...
+	followersRouter.HandleFunc("", handler.autoUnfollow).Methods(http.MethodDelete)
+
+	followerRouter := playbookRouter.PathPrefix("/follow").Subrouter()
```
lieut-data, line 61: Looking at this now, I'm wondering if we push through to `/autoFollower` and `/autoFollowers` for completeness?

server/app/playbook.go
```diff
@@ -235,17 +235,17 @@ type PlaybookService interface {
 	// Restores a deleted playbook
 	Restore(playbook Playbook, userID string) error
 
-	// Follow method lets user follow a all runs of a specific playbook
-	Follow(playbookID, userID string) error
+	// AutoFollow method lets user follow a all runs of a specific playbook
```
lieut-data, line 238: Should we push through to terminology too, like:
```suggestion
	// AutoFollow method lets a user automatically follow a all runs of a specific playbook
```

server/sqlstore/migrations.go
```diff
@@ -1468,69 +1468,69 @@ var migrations = []Migration{
...
+		fromVersion: semver.MustParse("0.39.0"),
+		toVersion:   semver.MustParse("0.40.0"),
 		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
 			if e.DriverName() == model.DatabaseDriverMysql {
 				if _, err := e.Exec(`
-					CREATE TABLE IF NOT EXISTS IR_Playbook_Participants (
+					CREATE TABLE IF NOT EXISTS PlaybookAutoFollowers (
 						PlaybookID VARCHAR(26) NULL REFERENCES IR_Playbook(ID),
 						UserID VARCHAR(26) NOT NULL,
 						IsFollower BOOLEAN NOT NULL,
```
lieut-data, line 1502: Should we drop `IsFollower` now that this is just a table of auto followers?

server/sqlstore/migrations.go
```diff
@@ -1468,69 +1468,69 @@ var migrations = []Migration{
...
-						INDEX IR_Playbook_Participants_UserID (UserID),
-						INDEX IR_Playbook_Participants_PlaybookID (PlaybookID)
+						INDEX PlaybookAutoFollowers_UserID (UserID),
+						INDEX PlaybookAutoFollowers_PlaybookID (PlaybookID)
```
lieut-data, line 1504: Can we rely on the primary key to effectively give us this set? (I believe we can get partial results out of an index when the conditions are specified first.)

lieut-data (approved):
Thanks, Shota! A few more thoughts -- mostly wanting to get the schema "right", and we can always iterate on terminology going forward.

webapp/i18n/en.json
```diff
@@ -37,6 +37,7 @@
   "6n0XDG": "Are you sure you want to remove the checklist? All tasks will be removed.",
   "6uhSSw": "Select a channel",
   "8hDbW6": "Send an outgoing webhook",
+  "8pXxyF": "You'll automatically be added as a follower for all new runs of this playbook.",
```
justinegeffen, line 40:
```suggestion
  "8pXxyF": "You're automatically added as a follower to all new runs of this playbook.",
```

justinegeffen (commented):
Two slight rewording suggestions and I also picked up two unrelated nits that would be awesome to resolve. :) 

As an aside, do we have a definition of "follower"? I wonder if I could tighten up the language based on what the role entails.

server/api/playbooks.go
itao: I suggest `isFollowing` as opposed to `isFollower` to minimize confusion as we anticipate introducing the concept of specific positions for participants. In general, I'd err towards referring to "following" as an action that someone has taken as opposed to a role that they're fulfilling.

itao (changes requested):
I propose we intentionally always refer to `following` as an action rather than `follower` role to minimize user confusion. We ourselves already mix up the purpose of `owner`, `participant`, and `followers` as it implies someone can be a combination of those. And it'll only get worse when we introduce customizable roles/positions per playbook run. I can be convinced otherwise but want to raise it for consideration cc @abhijit-singh @justinegeffen 

Sample places where this would need to be addressed: `... automatically added as a follower ...` tooltip would say something like `... receive notification for every run and status updates ... ` Which would make the consequence more clear to user.

I'm 3/5 that `Following` should be consistently referred to as an action/state as opposed to role/position, and 1/5 that developer-facing terms are consistent with user-facing mental model.

justinegeffen:
> I'm 3/5 that `Following` should be consistently referred to as an action/state as opposed to role/position, and 1/5 that developer-facing terms are consistent with user-facing mental model.

That's a good point, @itao, and I've revised my revisions. I didn't think of it as an implied role and agree that it will make things very tangled. Let me know what you think of the new wording.

webapp/i18n/en.json
```diff
@@ -212,6 +213,7 @@
   "edxtzC": "Create playbook",
   "egvJrY": "Assignee Changed",
   "eiPBw7": "Retrospective reminder interval",
+  "fRh92E": "If checked you'll automatically be added as a follower for all new runs of this playbook.",
```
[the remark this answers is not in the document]
abhijit-singh: I agree with not using `follower` anywhere but notifications will not be limited to status updates and also cover other run events.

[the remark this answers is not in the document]
justinegeffen: Gotcha, I'll rework these.

server/api/playbooks.go
```diff
@@ -542,13 +541,19 @@ func (h *PlaybookHandler) validateCategoryName(categoryName string) error {
 func (h *PlaybookHandler) autoFollow(w http.ResponseWriter, r *http.Request) {
 	playbookID := mux.Vars(r)["id"]
 	userID := r.Header.Get("Mattermost-User-ID")
+	followID := mux.Vars(r)["followID"]
```
lieut-data, line 544: nit: `followID` makes me think this is a resource for something other than a user. Maybe we just set this to `userID`, while still checking that it's the same as the header value?

server/api/playbooks.go
```diff
@@ -542,13 +541,19 @@ func (h *PlaybookHandler) validateCategoryName(categoryName string) error {
...
 
-	if err := app.PlaybookAccess(userID, playbookID, h.playbookService, h.pluginAPI); err != nil {
+	if followID != userID {
```
lieut-data, line 546: Can we allow system admins to do this, though?

server/sqlstore/playbook.go
```diff
@@ -665,68 +665,56 @@ INSERT INTO IR_PlaybookMember(PlaybookID, MemberID)
 }
 
 func (p *playbookStore) AutoFollow(playbookID, userID string) error {
-	return p.autoFollowHelper(playbookID, userID, true)
+	if _, err := p.store.execBuilder(p.store.db, sq.
+		Insert("IR_PlaybookAutoFollow").
+		Columns("PlaybookID", "UserID").
+		Values(playbookID, userID)); err != nil {
```
lieut-data, line 671: Should we still handle the `ON DUPLICATE KEY` case so that this doesn't error out if the user is already following?

server/sqlstore/playbook.go
```diff
@@ -665,68 +665,56 @@ INSERT INTO IR_PlaybookMember(PlaybookID, MemberID)
...
+	var returnedPlaybookID string
+	err := p.store.getBuilder(p.store.db, &returnedPlaybookID, query)
 	if err == sql.ErrNoRows {
 		return false, nil
 	} else if err != nil {
 		return false, errors.Wrapf(err, "failed to get follower status for playbook '%s'", playbookID)
 	}
 
-	return isFollower, nil
+	return returnedPlaybookID == playbookID, nil
```
lieut-data, line 717: It's not obvious why we nee to check `returnedPlaybookID` against `playbookID`, given that's the condition already in the `WHERE` clause. Could we tweak the SQL query to just effectively do `SELECT TRUE FROM ...` and verify that instead?

webapp/src/client.ts
```diff
@@ -490,20 +490,20 @@ export const unfollowPlaybookRun = async (playbookRunId: string) => {
     });
 };
 
-export const followPlaybook = async (playbookId: string) => {
-    await doFetchWithoutResponse(`${apiUrl}/playbooks/${playbookId}/followers`, {
+export const followPlaybook = async (playbookId: string, userId: string) => {
```
lieut-data, line 493: should we name these `autoFollowPlaybook`, etc.?

lieut-data (approved):
A handful of comments, though not strictly blocking. Thanks, Shota!

lieut-data (approved):

justinegeffen (approved):

itao (approved):
Assuming copy has been changed to refer to `following` as an action rather than `follower` as a role, then LGTM!

PR #872, opened 2021-11-09
MM-39938: react-select theming, support portaling

calebroseland:
#### Summary
- Refactor and globalize `react-select` theming overrides to support portaling.
- Mount/unmount global style in index-init., as we don't have an always-active, central mounting point.
- Replaces classNamePrefix `channel-selector` with `playbooks-rselect` as the shared select is used for multiple things.

CleanShot 2021-11-09 at 09 18 32@2x

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-39938

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated


lieut-data (approved):
Super neat!

PR #874, opened 2021-11-09
MM-39939: Hide enabled actions that have no actual content

agarciamontoro:
#### Summary
The original ticket focused on the action that automatically selects a default owner when the run starts: when said action was enabled but no actual user was added (which is actually possible, and hopefully something we'll fix when the edit screen is revamped), then its preview rendered in a weird state.

But this happens with most actions: we need to check both that they are enabled and that they have actual content (the owner is selected, there is at least one webhook, there is at least one user invited...).

This PR addresses all these cases, also changing the negative boolean conditions into positive ones, which are way easier to read.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39939

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbooks/playbook_preview_actions.tsx
+++ b/webapp/src/components/backstage/playbooks/playbook_preview_actions.tsx
@@ -29,26 +29,34 @@ const PlaybookPreviewActions = (props: Props) => {
     // hiding them if they don't have any visible subentries.
     // If a new CardSubEntry is added or the conditions are changed, these booleans need to be updated.
 
-    const emptyPromptEntry =
-        !props.playbook.signal_any_keywords_enabled;
+    const showPromptCardEntry = props.playbook.signal_any_keywords_enabled && props.playbook.signal_any_keywords.length !== 0;
 
-    const emptyRunStartEntry =
-        !props.playbook.invite_users_enabled &&
-        !props.playbook.default_owner_enabled &&
-        !props.playbook.broadcast_enabled &&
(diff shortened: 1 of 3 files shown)


webapp/src/components/backstage/playbooks/playbook_preview_actions.tsx
```diff
@@ -130,35 +138,35 @@ const PlaybookPreviewActions = (props: Props) => {
                         title={formatMessage({
                             defaultMessage: 'Update run summary',
                         })}
-                        enabled={props.playbook.webhook_on_status_update_enabled}
+                        enabled={runSummaryEnabled}
```
agarciamontoro, line 141: The condition for this one was completely wrong, so I also fixed it.

webapp/src/components/backstage/playbooks/playbook_preview_checklists.tsx
```diff
@@ -22,6 +22,10 @@ const PlaybookPreviewChecklists = (props: Props) => {
     const initialArray = Array(props.playbook.checklists.length).fill(false);
     const [checklistsCollapsed, setChecklistsCollapsed] = useState(initialArray);
 
+    if (props.playbook.checklists.length === 0) {
+        return null;
+    }
+
```
agarciamontoro, line 28: I did not know this, but we technically allow empty checklists! This covers that weird edge case.

iomodo (approved):
Looks good!

PR #876, opened 2021-11-10
Build just one binary on a developer environment

agarciamontoro:
#### Summary
After [this conversation](https://community-daily.mattermost.com/core/pl/bf4kq4i77fnr5qzqrpwqz3befa) in Mattermost, I decided to set up this simple proof of concept to build only one binary when developing locally.

Right now I'm using the environment variable `MM_SERVICESETTINGS_ENABLEDEVELOPER`, but we could also use any ad-hoc variable. This PR is just intended to continue the conversation so it does not die in a poor abandoned thread.

Let me know what you think!

#### Ticket Link
--

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ GO_TEST_FLAGS ?= -race
 GO_BUILD_FLAGS ?=
 MM_UTILITIES_DIR ?= ../mattermost-utilities
 DLV_DEBUG_PORT := 2346
+DEFAULT_GOOS := $(shell go env GOOS)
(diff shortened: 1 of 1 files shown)


agarciamontoro:
Also, I just remembered that you said you recalled something similar to this in the incident-management Makefile, @cpoile. Did it resemble this? Do we want something different?

lieut-data (approved):
This is great! We can iterate on the exact way of triggering it, but it seems like a great start :)

cpoile (approved):
Yep, this is even a bit simpler than what we had earlier. I like it!

PR #879, opened 2021-11-11
MM-38593: Improve semantics for archived playbooks

cpoile:
#### Summary
- Replacing delete playbook semantics with archive playbook

#### Notes
- /Not/ changing the db, so we're using `DeleteAt` to signify that a playbook is archived. That seems fine to me because it resembles what we do on the server.
- Also /not/ changing then telemetry, so there is still a delete telemetry event. I'm keeping that because I don't want to cause one more data translation task (and having to remember somewhere
- I'm reusing sing the Badge component, but Badge used to be for runs only. Now it's being used for Runs and Playbooks. I chose to use a union instead of adding 'archived' to the run statuses (which they currently can't be). But since we don't have a concept of playbook "state" I didn't create a playbook state type.

Looks like (compare to the ticket):

image

image

image

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38593

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


server/api/playbook_runs_test.go
```diff
@@ -761,6 +761,56 @@ func TestPlaybookRuns(t *testing.T) {
 		require.Nil(t, resultPlaybookRun)
 	})
 
+	t.Run("create invalid playbook run - playbook is archived", func(t *testing.T) {
```
crspeller, line 764: Have you tried using the new tests?

webapp/src/components/backstage/playbook_runs/status_badge.tsx
```diff
@@ -9,7 +9,7 @@ import {FormattedMessage} from 'react-intl';
 import {PlaybookRunStatus} from 'src/types/playbook_run';
 
 interface BadgeProps {
-    status: PlaybookRunStatus;
+    status: PlaybookRunStatus | 'Archived';
```
crspeller, line 12: For ease of use maybe defining a type for this? Call it something like "BadgeType".

crspeller (approved):

webapp/src/components/backstage/playbook_runs/status_badge.tsx
agarciamontoro: Agree, and maybe move this file out of the playbook_runs/ directory?

webapp/src/components/backstage/playbooks/playbook.tsx
```diff
@@ -120,9 +121,16 @@ const Playbook = () => {
                             <SubTitle>{subTitle}</SubTitle>
                         </HorizontalBlock>
                     </VerticalBlock>
+                    {
+                        archived &&
+                        <StatusBadge
+                            data-testid='archived-badge'
+                            status='Archived'
+                        />
```
agarciamontoro, line 129: Doesn't the linter enforce the curly braces here? I thought that was needed

```suggestion
                        <StatusBadge
                            data-testid={'archived-badge'}
                            status={'Archived'}
                        />
```

tests-e2e/cypress/integration/channels/rhs_spec.js
```diff
@@ -180,6 +183,9 @@ describe('channels > rhs', () => {
                 cy.apiFinishRun(playbookRun.id);
             });
 
+            // # Wait because this test is flaky if we move too quickly
+            cy.wait(TIMEOUTS.FIVE_SEC);
```
agarciamontoro, line 187: Wow, do we need five seconds?

tests-e2e/cypress/integration/playbooks/overview_spec.js
```diff
@@ -122,4 +125,48 @@ describe('playbooks > overview', () => {
             cy.findByText('Start run').should('exist');
         });
     });
+
+    describe.only('archiving', () => {
```
agarciamontoro, line 129: Should we remove the only?

tests-e2e/cypress/integration/playbooks/overview_spec.js
```diff
@@ -122,4 +125,48 @@ describe('playbooks > overview', () => {
...
+            // # Programmatically archive it
+            cy.apiArchivePlaybook(testPlaybook.id);
+
+            // # Visit the selected playbook
+            cy.visit(`/playbooks/playbooks/${testPlaybook.id}`);
+
+            // * Verify we're on the right playbook
+            cy.get('[class^="Title-"]').contains(playbookTitle);
+
+            // * Verify we an see the archived badge
```
agarciamontoro, line 153: Typo
```suggestion
            // * Verify we can see the archived badge
```

tests-e2e/cypress/integration/playbooks/overview_spec.js
```diff
@@ -122,4 +125,48 @@ describe('playbooks > overview', () => {
...
+            cy.findByTestId('edit-playbook').should('be.disabled');
+
+            // # Attempt to edit the playbook
+            cy.apiGetPlaybook(testPlaybook.id).then((playbook) => {
+                // # New title
+                playbook.title = 'new Title!!!';
+
+                // * Verify update fails
+                cy.apiUpdatePlaybook(playbook, 400);
+            });
```
agarciamontoro, line 169: I really like this, but maybe this should be coded as one of the new server E2E tests?

agarciamontoro (approved):
Nice! I left a couple of nits and minor suggestions, nothing blocking :) Thank you :tada:

server/api/playbook_runs_test.go
cpoile: Yep, done. Had to fix a pretty bad bug in the client, which shows that using the client in tests like this is useful.

webapp/src/components/backstage/playbook_runs/status_badge.tsx
cpoile: Okay, sounds good. We have to convert types now, by using `BadgeType[playbookRun.current_status]` for example, but that's easy.

tests-e2e/cypress/integration/channels/rhs_spec.js
cpoile: I'm not sure how much of a delay would make the test consistently pass, but this is safe (and far less time than we're wasting by having to rerun the test)

tests-e2e/cypress/integration/playbooks/overview_spec.js
cpoile: 🤦

tests-e2e/cypress/integration/playbooks/overview_spec.js
cpoile: yep, done. But I kept this, cause why not. :)

crspeller (approved):

PR #882, opened 2021-11-12
[MM-39849] Add followers to my runs filter

iomodo:
#### Summary
This ticket has changed a bit. As discussed in the meeting we decided to just show run followers too(with the participants) when we toggle `my runs only` checkbox. 
Basically `my runs` = `runs I'm following` + `runs I'm participating in`

No UX changes

#### Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-39849

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


stephenvanhemmen:
Does the Runs page default to having `My runs only` selected? If not, the runs page should only show the user's participating/following runs by default. This is to test the assumption that users use the Runs page to see their specific runs more often than other peoples' runs.

iomodo:
> Does the Runs page default to having `My runs only` selected? If not, the runs page should only show the user's participating/following runs by default. This is to test the assumption that users use the Runs page to see their specific runs more often than other peoples' runs.

It does not. I'll make the change. Thanks @stephenvanhemmen !

iomodo:
`My runs only` will be checked by default in the runs page as well as the usage page

Screen Shot 2021-11-12 at 18 30 51

server/sqlstore/playbook_run.go
```diff
@@ -232,6 +232,24 @@ func (s *playbookRunStore) GetPlaybookRuns(requesterInfo app.RequesterInfo, opti
...
+			FROM IR_Run_Participants as rp
+			WHERE rp.IncidentID = i.ID
+			AND rp.UserID = ?
+			AND rp.IsFollower = ?)`
+		participantFilter := `EXISTS(SELECT 1
+			FROM ChannelMembers AS cm
+			WHERE cm.ChannelId = i.ChannelID
+			AND cm.UserId = ?)`
+		query := fmt.Sprintf("(%s OR %s)", followerFilter, participantFilter)
+		myRunsClause := sq.Expr(query, userIDFilter, true, userIDFilter)
```
agarciamontoro, line 247: Nit: for the sake of simplicity, can we hardcode the `true` in the followerFilter`'s `rp.IsFollower` condition? That way we can remove one parameter to `myRunsClause`.

agarciamontoro (approved):
Looks good :tada: Just a minor, non-blocking, nit. Thanks!

stephenvanhemmen:
> `My runs only` will be checked by default in the runs page as well as the usage page
> 
> Screen Shot 2021-11-12 at 18 30 51

Just one more thing, could we keep this option unchecked by default on the Usage page but checked by default on the Runs list? Users seem to have different uses for these two lists, even though they both display runs. Sorry about that, I should have specified. Thanks!

iomodo:
> Just one more thing, could we keep this option unchecked by default on the Usage page but checked by default on the Runs list? Users seem to have different uses for these two lists, even though they both display runs. Sorry about that, I should have specified. Thanks!

No worries, done

stephenvanhemmen (approved):

iomodo:
Adjusted input field size
Screen Shot 2021-11-17 at 18 31 16

crspeller (approved):

PR #883, opened 2021-11-12
[MM-39529] Adding an item in the run's RHS dropdown menu to edit a run summary

JulienTant:
#### Summary
This PR is modifying the hamburger menu of a run in the RHS to add an explicit "Edit run summary" item.

image

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/18972

#### Checklist
~~- [ ] Telemetry updated~~
- [x] Unit tests updated

#### Dev note
1) My ReactJS skills are not perfect, so I'm not quite sure that the way I implemented the communication between components is the "state of the art" way to do it. I'm happy to change that to whatever is best!
2) I had to use a bit of trickery to have the hamburger menu open in the e2e tests because it appears only when the mouse goes over the container. I tried using `trigger('mouseover')` on the container but it did not work and [cypress has no implementation of `hover` yet](https://docs.cypress.io/api/commands/hover#Workarounds) was not enough. To remedy that during the test I force the buttons the show.

### DIFF
--- a/webapp/src/components/rhs/rhs_about_buttons.tsx
+++ b/webapp/src/components/rhs/rhs_about_buttons.tsx
@@ -6,13 +6,13 @@ import {FormattedMessage, useIntl} from 'react-intl';
 import styled from 'styled-components';
 
 import Icon from '@mdi/react';
-import {mdiClipboardPlayOutline, mdiNotebookOutline} from '@mdi/js';
+import {mdiClipboardPlayOutline, mdiNotebookOutline, mdiPencil} from '@mdi/js';
 
 import {PlaybookRun} from 'src/types/playbook_run';
 
 import {navigateToPluginUrl} from 'src/browser_routing';
 import {HoverMenuButton} from 'src/components/rhs/rhs_shared';
-import DotMenu, {DotMenuButton, DropdownMenuItem} from 'src/components/dot_menu';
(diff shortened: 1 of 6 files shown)


webapp/src/components/rhs/rhs_about_buttons.tsx
```diff
@@ -44,11 +47,27 @@ const RHSAboutButtons = (props: Props) => {
                     }
                 }}
             />
-            <DotMenu
+            <ClosableDotMenu
```
crspeller, line 50: You don't actually need this workaround. I fixed this issue on a branch of mine. 
If you add this to dot_menu.jsx:122 (In the `<DropDownMenu>` component)  It will automatically close when an item is clicked. 
```jsx
  onClick={(e) => {
      e.stopPropagation();
      setOpen(false);
  }}
```

crspeller (changes requested):
Thanks for the contribution! Just one comment to simplify.

webapp/src/components/rhs/rhs_about_buttons.tsx
JulienTant: So much simpler and elegant, thanks @crspeller!

crspeller (approved):
Looks good :)

iomodo (approved):
Thanks for the contribution @JulienTant 
Looks great

PR #892, opened 2021-11-18
MM-38954: Scroll the RHS to top when the run is finished

agarciamontoro:
#### Summary

Scroll to top when the run is finished. To actually finish the run, the user needs to click on an interactive dialog, to which we don't have direct access. Thus, what this PR does is listening to any change of `run.current_status`, and fires the scrolling event when it changes to finished.

https://user-images.githubusercontent.com/3924815/142421094-957bc309-c69a-4257-a3f8-45659dd53518.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38954

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


iomodo (approved):
👍

calebroseland (approved):

PR #893, opened 2021-11-18
Cypress default config & README updates

Phrynobatrachus:
#### Summary
update `on_prem` `MaxFileSize`  default config and add a README note about `CYPRESS_developerMode`. I'll keep thinking about whether there's a better way to account for config stuff but it's probably worth documenting while we're using it.

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ To quickly test Mattermost Playbooks, use the following test commands to create 
 
   * An example command looks like: `/playbook test bulk-data 10 3 342 2`
 
-## Running E2E Tests
+## Running E2E tests
 
 When running E2E tests, local `mattermost-server` configuration may be unexpectedly modified if values are different than a corresponding key default in either `on_prem_default_config.json` or `cloud_default_default_config.json`. This can be avoided by setting `CYPRESS_developerMode=true` when calling Cypress scripts. For example: `CYPRESS_developerMode=true npm run cypress:open`.
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ To quickly test Mattermost Playbooks, use the following test commands to create 
 
   * An example command looks like: `/playbook test bulk-data 10 3 342 2`
 
+## Running E2E Tests
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,6 +38,10 @@ To quickly test Mattermost Playbooks, use the following test commands to create
 
   * An example command looks like: `/playbook test bulk-data 10 3 342 2`
 
+## Running E2E Tests
+
+When running E2E tests, local `mattermost-server` configuration may be unexpectedly modified if values are different than a corresponding key default in either `on_prem_default_config.json` or `cloud_default_default_config.json`. This can be avoided by setting `CYPRESS_developerMode=true` when calling Cypress scripts. For example: `CYPRESS_developerMode=true npm run cypress:open`.
```
justinegeffen, line 43:
```suggestion
When running E2E tests, the local `mattermost-server` configuration may be unexpectedly modified if the values for `on_prem_default_config.json` or `cloud_default_default_config.json` are different to the corresponding default key . This can be avoided by setting `CYPRESS_developerMode=true` when calling Cypress scripts. For example: `CYPRESS_developerMode=true npm run cypress:open`.
```

README.md
```diff
@@ -38,6 +38,10 @@ To quickly test Mattermost Playbooks, use the following test commands to create
 
   * An example command looks like: `/playbook test bulk-data 10 3 342 2`
 
+## Running E2E Tests
```
justinegeffen, line 41:
```suggestion
## Running E2E tests
```

justinegeffen (commented):
Just two comments - LMK your thoughts on the rewording one. :)

README.md
Phrynobatrachus: I think this phrasing shouldn't have the "default" in it. My thinking is that the aforementioned JSON files are the test defaults, not the local server config, but this implies the latter.

how about:

When running E2E tests, the local `mattermost-server` configuration may be unexpectedly modified if either `on_prem_default_config.json` or `cloud_default_config.json` (depending on the server edition) has conflicting values for the the same keys. This can be avoided by setting `CYPRESS_developerMode=true` when calling Cypress scripts. For example: `CYPRESS_developerMode=true npm run cypress:open`.

justinegeffen: Thank you for that insight - great suggestion! I've applied it. :)

justinegeffen (approved):
This looks great, thank you!

lieut-data (approved):
Thanks, @Phrynobatrachus!

PR #894, opened 2021-11-22
Add playbook filter to runs

iomodo:
#### Summary
Adds support of the playbook filter in the run list

https://user-images.githubusercontent.com/2340127/142857373-5fdaaa4d-9b62-4366-b50a-4be88b181b4d.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-38581

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


webapp/src/components/backstage/runs_list/filters.tsx
```diff
@@ -120,6 +139,11 @@ const Filters = ({fetchParams, setFetchParams, fixedTeam}: Props) => {
         });
     }
 
+    async function fetchPlaybooks() {
+        const playbooks = await clientFetchPlaybooks(fetchParams.team_id || '', {team_id: fetchParams.team_id || '', sort: 'title'});
+        return playbooks ? playbooks.items : {} as Playbook[];
```
agarciamontoro, line 144: Nit: can we do this instead? It feels weird to convert an empty object to a `Playbook[]`
```suggestion
        return playbooks ? playbooks.items : [] as Playbook[];
```

webapp/src/components/backstage/runs_list/playbook_selector.tsx
```diff
@@ -0,0 +1,281 @@
...
+    showOnRight?: boolean;
+    className?: string;
+}
+
+export default function PlaybookSelector(props: Props) {
+    const [isOpen, setOpen] = useState(false);
+    const toggleOpen = () => {
+        if (!isOpen) {
+            fetchPlaybooks();
+        }
```
agarciamontoro, line 44: Do we need to fetch the playbooks every time we toggle the menu?

webapp/src/components/backstage/runs_list/playbook_selector.tsx
```diff
@@ -0,0 +1,281 @@
...
+            return;
+        }
+
+        const playbook = playbookOptions.find((option: Option) => option.playbookId === props.selectedPlaybookId);
+        if (playbook) {
+            setSelected(playbook);
+        } else {
+            setSelected(null);
+        }
+    }, [playbookOptions, props.selectedPlaybookId]);
```
agarciamontoro, line 93: If I'm not mistaken, this will probably run in every render, because `playbookOptions` will be a new array every time. I'm not sure this specific case would be a performance problem, but I thought it's worth to point out.

webapp/src/components/backstage/runs_list/playbook_selector.tsx
```diff
@@ -0,0 +1,281 @@
...
+        </Dropdown>
+    );
+}
+
+// styles for the select component
+const selectStyles: StylesConfig<Option, boolean> = {
+    control: (provided) => ({...provided, minWidth: 240, margin: 8}),
+    menu: () => ({boxShadow: 'none'}),
+    option: (provided, state) => {
+        const hoverColor = 'rgba(20, 93, 191, 0.08)';
```
agarciamontoro, line 206: This is not themed. Should it be?

webapp/src/components/backstage/runs_list/playbook_selector.tsx
```diff
@@ -0,0 +1,281 @@
...
+    customControl?: (props: ControlProps<Option, boolean>) => React.ReactElement;
+    controlledOpenToggle?: boolean;
+    getPlaybooks: () => Promise<Playbook[]>;
+    onSelectedChange?: (playbookId?: string) => void;
+    customControlProps?: any;
+    showOnRight?: boolean;
+    className?: string;
+}
+
+export default function PlaybookSelector(props: Props) {
```
agarciamontoro, line 39: The code in this file is very similar to `ProfileSelector` and `TeamSelector`, right? Should we generalize all these components (here or in a future PR)?

agarciamontoro (changes requested):
Looks great, thank you! I left a handful of comments, but mostly questions and nits. The only blocking one is the theming issue.

webapp/src/components/backstage/runs_list/playbook_selector.tsx
iomodo: Created a ticket https://mattermost.atlassian.net/browse/MM-40326

agarciamontoro (approved):
Great, thank you! :tada:

abhijit-singh (approved):
This is great @iomodo! A couple of small things:
1. The owner and team filter also expand based on the value selected. Can we ensure the layout doesn't break if all three have values with lots of characters selected? It probably already works perfectly, but just wanted to make sure. 
2. The cursor changes to the hand while hovering on the Owner & Team filters. Can we have the same for the Playbook filter as well?
image
image

webapp/src/components/backstage/runs_list/playbook_selector.tsx
cpoile: I don't think so -- it's a state object, so the object identity should only change when the state changes (or unmounts-remounts).

cpoile (approved):
Nice, thanks @iomodo !!

webapp/src/components/backstage/runs_list/playbook_selector.tsx
agarciamontoro: I thought that using arrays there had the problem of new objects. @cpoile, do you have a link to some docs that explain this? If not, I'll test it and see what happens.

PR #895, opened 2021-11-22
[MM-39140] Add copy URL to run and playbook

iomodo:
#### Summary
This PR adds support of the run and playbook url copy

https://user-images.githubusercontent.com/2340127/142882834-b238d007-6975-4d2a-9f5d-6e8caee2b6aa.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-39140

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
```diff
@@ -6,9 +6,12 @@ import {useSelector} from 'react-redux';
 import styled from 'styled-components';
 import {Redirect, Route, useRouteMatch, NavLink, Switch} from 'react-router-dom';
 import {useIntl} from 'react-intl';
+import {Tooltip, OverlayTrigger} from 'react-bootstrap';
 
 import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
 
+import {copyToClipboard, getSiteURL} from '../../../../utils';
```
cpoile, line 13: super nit: this would probably be easier to read as an absolute path

webapp/src/utils.ts
```diff
@@ -38,3 +38,72 @@ export const nearest = (
...
+        siteURL = siteURL.substring(0, siteURL.length - 1);
+    }
+
+    return siteURL;
+}
+
+// copied from webapp
+export function getSiteURL(): string {
+    return getSiteURLFromWindowObject(window);
+}
```
cpoile, line 109: I think we should have the `basePath` in the `client.ts`, set when the plugin gets initialized--could we use that instead of new code?

cpoile (changes requested):
This is awesome @iomodo, thanks!

Also, do you think we should make an e2e test for it? (Personally, I'm trying to add e2e tests for anything that isn't highly interactive)

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
```diff
@@ -268,6 +285,36 @@ const PlaybookRunBackstage = () => {
         followButton = (<Button onClick={onUnfollow}>{formatMessage({defaultMessage: 'Unfollow'})}</Button>);
     }
 
+    const copyRunLink = () => {
+        const siteUrl = getSiteURL() + '/playbooks/runs/' + playbookRun.id;
+        copyToClipboard(siteUrl);
```
calebroseland, line 290: nit: Perhaps `runUrl` or direct param?

webapp/src/utils.ts
calebroseland: +1

calebroseland (approved):
Nice! Aside from @cpoile's comments and my own minor nit, this looks good to me!

webapp/src/utils.ts
calebroseland: +1

cpoile (approved):
Awesome Shota!

abhijit-singh (changes requested):
Thanks @iomodo. A couple of requests:
1. Is it possible to make the status tag and the link icon align/be in the same row as the `run name` as in the attached mockup? It looks odd right now if the run name takes less space than the playbook name.
image

(The same problem is there for the playbook page as well, but that'll probably go away with @crspeller's changes for the permissions feature as we're removing the bottom row of text from the playbook page header.)
2. Add 8px space between the status tag and the link icon box.
image

3. Extremely minor nit, but can we make the hover state background colour and icon colour grey instead of the current blue?
image
image

iomodo:
Is it what you had in mind @abhijit-singh ?

https://user-images.githubusercontent.com/2340127/143444856-dadf386b-872f-445f-af5d-dc4ee2a2ab83.mp4

abhijit-singh:
@iomodo The placement and spacing is perfect 👌!

For the hover state, it should have a grey background while hovering. The exact colors on-hover would be:
- Background color: 8% opacity of `center-channel-text`
- Icon color: 72% opacity of `center-channel-text`

Also, the icon container that has the background color should have a border radius of 4px.

iomodo:
@abhijit-singh check it :)


https://user-images.githubusercontent.com/2340127/143470211-0ab2b2ad-a249-452d-aa68-49973ad06db8.mp4

abhijit-singh (approved):
Perfect! Thank you @iomodo!

iomodo:
Thanks for your patience @abhijit-singh <3

PR #898, opened 2021-11-22
Strip "Time: "

lieut-data:
#### Summary
Strip `Time: `  from the timeline to avoid this:
image

#### Ticket Link
None.

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


calebroseland (commented):
re: https://github.com/mattermost/mattermost-server/issues/19019#issuecomment-974534821, I pushed a few commits so long as we're here. Feel free to rewind/`push -f` either/both.

It: 
- Switches to luxon for simpler forced `h23` & UTC
- Moves `time` to the space above each time entry to give the impression of illustrating the timespan between events in the timeline

CleanShot 2021-11-22 at 11 48 41@2x

CleanShot 2021-11-22 at 11 50 13@2x

lieut-data:
@calebroseland, yours is so much better than my tiny change. Feel free to take the driver's seat here :D

calebroseland:
A bit of a rabbit-trail refactor 😬, but I think it looks more readable. A11y also improved. The time-between is added, @abhijit-singh or @stephenvanhemmen what do you think—is that helpful or distracting?

CleanShot 2021-11-22 at 17 42 11@2x

stephenvanhemmen:
> 

I like how this has both the time-between and the total time elapsed for each entry in a way that isn't distracting or noisy. The time-between is likely a useful feature for many use-cases, especially for rapid incident collaboration. I like this, and I'll keep an eye out for confusion around this change if it gets added.

lieut-data:
I really like the simpler left hand side of the timeline. I'm somewhat confused by the more detailed estimate to the right of the date. Standing alone, it's not obvious what that is measuring, plus it adds a lot of visual noise. Would we consider rendering just the data in the left hand side, and show the more detailed amount on hover?

cpoile:
@calebroseland Do you want me to hold off on the review until you update?

calebroseland:
> @calebroseland Do you want me to hold off on the review until you update?

Yes, I’ll re-request once I simplify it 👍

Thanks!

calebroseland:
Clock-tooltip with local time and `{duration} after run started`

image

cpoile (approved):
This is great!

lieut-data (commented):
Love it! I feel badly for getting some PR credit for all your hard work, @calebroseland :p

PR #901, opened 2021-11-22
MM-38736: Prompt before discarding unsaved retrospective

cpoile:
#### Summary
- Added the prompt warning about discarding unsaved retrospective.

- Found a new bug while testing this:

https://user-images.githubusercontent.com/1490756/142929709-18a4a147-641c-4f9e-9946-97a916eb753a.mp4

- Need to lift the report's state into the playbook state (which will get overwritten when the playbook is loaded, but that's fine because it'll be the correct current state).

Completed with bug fix and prompt:

https://user-images.githubusercontent.com/1490756/142929762-d1970e01-58ac-4eb3-898c-59a80bfae14a.mp4

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-38736

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


iomodo (approved):
Looks good

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/retrospective/report.tsx
```diff
@@ -120,20 +124,24 @@ const Report = (props: ReportProps) => {
...
-                        text={report}
+                        text={props.playbookRun.retrospective}
                         team={team}
                     />
                 </PostTextContainer>
             }
+            <RouteLeavingGuard
+                navigate={(path) => WebappUtils.browserHistory.push(path)}
+                shouldBlockNavigation={() => editing}
+            />
```
calebroseland, line 144: Aside mention: I'm not precisely sure how electron would handle it, but just a shout out for future consideration of also using [onbeforeunload](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload)—it's great for handling refreshes, tab/window closures.

calebroseland (approved):
🎉

PR #902, opened 2021-11-23
Action channel name template

lieut-data:
#### Summary
Allow a playbook to specify a template for the channel name to be created, noting that the channel creation itself is disabled since it cannot currently be turned off:

CleanShot 2021-11-22 at 22 50 33@2x

When one is configured:

CleanShot 2021-11-22 at 22 51 06@2x

it appears in the preview:

CleanShot 2021-11-22 at 22 51 32@2x

Indeed, this section now appears at all times even when no template is specified:

CleanShot 2021-11-22 at 22 52 21@2x

The start run dialog automatically fills in when choosing to run this playbook:

CleanShot 2021-11-22 at 22 52 56@2x

#### Ticket Link
None?

#### Checklist
- [x] Telemetry updated
- [x] Unit tests updated


lieut-data:
@abhijit-singh, I've been sitting on this one for a while as you might recall. Decided to polish it up -- would love your feedback!

server/api/urls.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/pkg/errors"
+
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
+)
+
+const defaultBaseAPIURL = "plugins/playbooks/api/v0"
+
+func getAPIBaseURL(pluginAPI *pluginapi.Client) (string, error) {
```
cpoile, line 16: This is a nice QoL improvement, thank you!

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -162,6 +167,45 @@ describe('playooks > edit', () => {
...
+                    // # Visit the selected playbook
+                    cy.visit(`/playbooks/playbooks/${testPlaybook.id}/edit`);
+
+                    // # Switch to Actions tab
+                    cy.get('#root').findByText('Actions').click();
+
+                    // * Verify that the toggle is unchecked
+                    cy.get('#create-channel label input').should(
+                        'be.checked'
+                    );
```
cpoile, line 181: nit: just wondering why these are on separate lines? (automatic wrapping in the editor?)

webapp/src/components/backstage/playbook_edit.tsx
```diff
@@ -744,6 +752,8 @@ const PlaybookEdit = (props: Props) => {
                                     onToggleCategorizePlaybookRun={handleToggleCategorizePlaybookRun}
                                     categoryName={playbook.category_name}
                                     categoryNameChange={handleCategoryNameChange}
+                                    channelNameTemplate={playbook.channel_name_template}
+                                    onChannelNameTemplateChange={handleChannelNameTemplateChange}
```
cpoile, line 756: This whole component feels like it's due for a rewrite. Not an actionable comment, just an observation for later. :)

cpoile (approved):
Nice!

tests-e2e/cypress/integration/playbooks/edit_spec.js
lieut-data: Purely copy-pasta in this case -- probably unnecessary, but maybe something to pass over holistically as we keep driving improvements here.

abhijit-singh (approved):
Looks awesome @lieut-data! 

I was wondering if it makes sense to add the public/private configuration here as a part of this effort?
image
Can be done as a separate feature as well but that'll soon be the only thing in the `Permissions` tab as the new permission scheme rolls in.

lieut-data:
Ooh, that's such an awesome extension @abhijit-singh! I'm going to pull that out as a separate effort, but maybe I can recruit @crspeller to help polish that off :p

lieut-data:
(Either way, I'll update the copy as per your screenshot.)

server/sqlstore/playbook.go
```diff
@@ -109,6 +109,7 @@ func NewPlaybookStore(pluginAPI PluginAPIClient, log bot.Logger, sqlStore *SQLSt
 			"NumStages",
 			"NumSteps",
 			`(
+				1 + -- Channel creation is hard-coded
```
agarciamontoro, line 112: Good catch!

webapp/src/components/backstage/automation/settings.tsx
```diff
@@ -80,6 +82,20 @@ export const AutomationSettings = (props: Props) => {
...
                     {formatMessage({defaultMessage: 'When a run starts'})}
                 </SectionTitle>
+                <Setting id={'create-channel'}>
+                    <PatternedInput
+                        enabled={true}
+                        disableToggle={true}
+                        onToggle={() => null}
+                        input={props.channelNameTemplate}
+                        onChange={props.onChannelNameTemplateChange}
+                        pattern={'[\\s\\S]{2,}'} // at least two characters
```
agarciamontoro, line 92: This allows whitespace-only strings, which are not allowed in channel names. We can either come up with a convoluted regex or refactor `PatternedInput` a bit so that `pattern` also accepts a `(string) => boolean` function, and then we can simply pass here `(s) => s.trim().length > 1`. Thoughts?

webapp/src/components/backstage/automation/toggle.tsx
```diff
@@ -37,11 +43,10 @@ const RoundSwitch = styled.span`
     width: 40px;
     height: 24px;
     border-radius: 14px;
-    background: rgba(var(--center-channel-color-rgb), 0.24);
-
+    background: rgba(var(--center-channel-color-rgb), ${(props) => (props.disabled ? '0.08' : '0.24')});
```
agarciamontoro, line 46: Super nit: we seem to converge on using destructured arguments in styled-components when it's this short:
```suggestion
    background: rgba(var(--center-channel-color-rgb), ${({disabled}) => (disabled ? '0.08' : '0.24')});
```

webapp/src/components/backstage/automation/toggle.tsx
```diff
@@ -56,13 +61,14 @@ const RoundSwitch = styled.span`
...
     }
 
-    input:checked + & {
-        background-color: var(--button-bg);
+    input:checked + && {
+        background-color: ${(props) => (props.disabled ? 'var(--button-bg-30)' : 'var(--button-bg)')}
     }
 
-    input:checked + &:before {
+    input:checked + &&::before {
```
agarciamontoro, line 68: Why do we need the double ampersand here?

webapp/src/components/backstage/automation/toggle.tsx
```diff
@@ -56,13 +61,14 @@ const RoundSwitch = styled.span`
         transition: .4s;
     }
 
-    input:checked + & {
-        background-color: var(--button-bg);
+    input:checked + && {
+        background-color: ${(props) => (props.disabled ? 'var(--button-bg-30)' : 'var(--button-bg)')}
```
agarciamontoro, line 65: Super nit (bis)
```suggestion
        background-color: ${({disabled}) => (disabled ? 'var(--button-bg-30)' : 'var(--button-bg)')}
```

webapp/src/components/backstage/playbook_edit.tsx
agarciamontoro: Absolutely, yes, this is way too big and complex.

webapp/src/components/backstage/playbooks/playbook_preview_actions.tsx
```diff
@@ -32,13 +32,15 @@ const PlaybookPreviewActions = (props: Props) => {
...
 
+    const createChannelEnabled = true;
     const inviteUsersEnabled = props.playbook.invite_users_enabled && props.playbook.invited_user_ids.length !== 0;
     const defaultOwnerEnabled = props.playbook.default_owner_enabled && props.playbook.default_owner_id !== '';
     const broadcastEnabled = props.playbook.broadcast_enabled && props.playbook.broadcast_channel_ids.length !== 0;
     const runSummaryEnabled = props.playbook.run_summary_template !== '';
     const webhookOnCreationEnabled = props.playbook.webhook_on_creation_enabled && props.playbook.webhook_on_creation_urls.length !== 0;
 
     const showRunStartCardEntry =
+        createChannelEnabled ||
```
agarciamontoro, line 43: Another good catch :)

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -162,6 +167,45 @@ describe('playooks > edit', () => {
...
         describe('when a playbook run starts', () => {
+            describe('create channel setting', () => {
+                it('is enabled in a new playbook', () => {
+                    // # Visit the selected playbook
+                    cy.visit(`/playbooks/playbooks/${testPlaybook.id}/edit`);
+
+                    // # Switch to Actions tab
+                    cy.get('#root').findByText('Actions').click();
+
+                    // * Verify that the toggle is unchecked
```
agarciamontoro, line 178: Typo
```suggestion
                    // * Verify that the toggle is checked
```

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -162,6 +167,45 @@ describe('playooks > edit', () => {
...
+
+                it('can not be disabled', () => {
+                    // # Visit the selected playbook
+                    cy.visit(`/playbooks/playbooks/${testPlaybook.id}/edit`);
+
+                    // # Switch to Actions tab
+                    cy.get('#root').findByText('Actions').click();
+
+                    cy.get('#create-channel').within(() => {
+                        // * Verify that the toggle checked
```
agarciamontoro, line 197: Typo 
```suggestion
                        // * Verify that the toggle is checked
```

tests-e2e/cypress/integration/channels/actions_spec.js
```diff
@@ -17,6 +18,10 @@ describe('channels > actions', () => {
             testTeam = team;
             testUser = user;
 
+            cy.apiCreateCustomAdmin().then(({sysadmin}) => {
+                testSysadmin = sysadmin;
+            });
```
agarciamontoro, line 23: Just curious: why the test sysadmin everywhere? Is that simply not to depend on existing users?

agarciamontoro (changes requested):
Really cool! :tada:

My comments are mostly observations and nits, the only blocking one is the one about the pattern.

webapp/src/components/backstage/automation/toggle.tsx
lieut-data: Hmm, I wrote this part of the PR quite a while back, but I do recall wrestling with specificity issues. I can take a closer look if we think it's an issue?

tests-e2e/cypress/integration/channels/actions_spec.js
lieut-data: In this case, I think the specific issue was that I didn't have the `id` of the default sysadmin, while creating a custom one guaranteed that. I need the id to make that sysadmin a member of the playbook in order to be able to return it when it's created.

webapp/src/components/backstage/automation/settings.tsx
lieut-data: Ooh, good catch. Given we just pass the `pattern` directly to the `input` and rely on CSS thereafter, I'm inclined to stick with the Regular expression for now. I think `[\\S][\\s\\S]*[\\S]` is better?

agarciamontoro: Yeah, that works, it looks we trim the channel name in any case. Thanks! :)

webapp/src/components/backstage/automation/toggle.tsx
agarciamontoro: Oh, if it's a specificity issue then it makes sense. Styled components are great, but still CSS :sob:

tests-e2e/cypress/integration/channels/actions_spec.js
agarciamontoro: Sounds good, thanks for the context!

agarciamontoro (approved):
Thanks!

PR #904, opened 2021-11-23
Replace all CSS variables with opacity with rgba

agarciamontoro:
#### Summary

This PR replaces CSS variables like `var(--center-channel-color-16)` to something like `rgba(var(--center-channel-color-rgb),0.16)`. I've never liked that we use both styles, as it can be confusing for new (and old, actually) contributors. After [discussing which one is preferred](https://community-daily.mattermost.com/core/channels/webapp/wip46mny1tde7cud8nsha5kikr), the most flexible `rgba` option won. So I came up with this simple sed command:

```sh
sed -ri 's/var\((--[a-z-]+)([0-9]{2})\)/rgba(var(\1rgb),0.\2)/g'
```

This is super low priority, but it was so easy to do it that I could not stop myself from opening the PR. I did a quick, but broad, smoke testing to see that nothing was broken. I did not see any errors, but if you want to spend five minutes taking a look, I'd appreciate it. If there is a theming problem going forward, we can always blame this PR.

#### Ticket Link
--

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


cpoile (approved):
Cool, I was thinking of doing the same actually. :)
Looks good to me after smoke testing for a few minutes.

One super nit -- do you think we should add a space after the `,` to keep it consistent with what we're already doing?

Approving so that I can blame all future theming troubles on @agarciamontoro !

calebroseland (approved):
LGTM!


> One super nit -- do you think we should add a space after the , to keep it consistent with what we're already doing?

+1

agarciamontoro:
Force-pushed to fix the conflicts and get the space after the comma :)

PR #814, opened 2021-10-14
[MM-38575] Import luxon as external dependency from mattermost-webapp, bump min version

mathiasvr:
#### Summary

In playbooks, define/import luxon as external dependency provided from mattermost-webapp.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/18588

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Gated by experimental feature flag~~
- ~~[ ] Unit tests updated~~

PR #905, opened 2021-11-23
temporarily drop min version to 6.0

lieut-data:
#### Summary
Community has the requisite code (Luxon), but didn't actually get bumped to 6.1 like we expected. Let's drop the min version for a few days until it catches up to restore Playbooks to community.

#### Ticket Link
None.

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -5,7 +5,7 @@
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-playbooks/",
     "support_url": "https://github.com/mattermost/mattermost-plugin-playbooks/issues",
     "icon_path": "assets/plugin_icon.svg",
-    "min_server_version": "6.1.0",
+    "min_server_version": "6.0.0",
     "server": {
         "executables": {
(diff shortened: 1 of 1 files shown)


cpoile (approved):

cpoile:
@lieut-data It's possible this will cause calls to luxon to error, since this was updated to 6.1.0 because we moved to depending on it externally: #814

lieut-data:
Thanks, @cpoile! The tag currently on commit -- `cloud-2021-10-27-1`, as old as it is -- does indeed have Luxon. It's just that the database version that tracks "6.1" isn't present in that branch (or the one that should be live today). It gets fixed tomorrow, but the Luxon has actually been there for quite a while.

I'm guessing we need a "better" way to track version dependencies here given how fluidly things are changing.

cpoile:
@lieut-data Thanks!

PR #907, opened 2021-11-24
MM-37549: Avoid repetitive notifications on completing tasks

cpoile:
#### Summary
- Removed in-channel notifications. DMing now according to this ruleset:
  - No in-channel notifications for any task activity or owner assignee activity
  - DM new run owners that they are the owner of the run, but only if they are not the run creator
  - DM new task assignees that they are the assignee on a task, unless they were the person who initiated the assignee change
- Removed the `state_modified_post_id` and `assignee_modified_post_id` from checklist items, since we're not making posts to refer to anymore. 
- Modified the logic in the timeline entries so that the cursor won't turn into a pointer when hovering over an item that doesn't have a post_id link.

#### Request for UX/Editing
Here are the texts, please feel free to make suggestions and I'll make the changes:
image
image
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-37549

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/checklist_item.tsx
```diff
@@ -513,7 +513,7 @@ export const ChecklistItemDetails = (props: ChecklistItemDetailsProps): React.Re
                             <div
                                 onClick={((e) => handleFormattedTextClick(e, relativeTeamUrl))}
                             >
-                                {(props.checklistItem.state === ChecklistItemState.Skip) ? <StrikeThrough>{labelText}</StrikeThrough> : labelText}
+                                {(props.checklistItem.state === ChecklistItemState.Skip) ? <StrikeThrough className={'skipped'}>{labelText}</StrikeThrough> : labelText}
```
agarciamontoro, line 516: What's this className for?

server/app/playbook_run_service.go
```diff
@@ -1314,53 +1313,61 @@ func (s *PlaybookRunServiceImpl) SetAssignee(playbookRunID, userID, assigneeID s
...
 	itemToCheck.AssigneeID = assigneeID
 	itemToCheck.AssigneeModified = model.GetMillis()
-	itemToCheck.AssigneeModifiedPostID = post.Id
 	playbookRunToModify.Checklists[checklistNumber].Items[itemNumber] = itemToCheck
 
 	if err = s.store.UpdatePlaybookRun(playbookRunToModify); err != nil {
 		return errors.Wrapf(err, "failed to update playbook run; it is now in an inconsistent state")
 	}
 
+	// Do we send a DM to the new assignee?
```
agarciamontoro, line 1344: Really like these comments :)

agarciamontoro (approved):
Cool! Just a question below :)

iomodo (approved):
Looks great @cpoile !

webapp/src/components/checklist_item.tsx
cpoile: At the moment it's for testing -- it's difficult/error-prone to test for css attr on elements, so Cypress recommends attaching a class instead. Open to other options too!

agarciamontoro: Oh, ok, I see the problem now. I made a super quick google search and from the strategies people use (basically ideas from https://github.com/cypress-io/cypress/issues/1212), I think I like better using something like `data-cy` (in the same vein we use `data-testid`). Having an actual class in the element, the first question someone reading the code will ask themselves is "where is this class defined and what does it do?", but then they won't find it at all. But if there's a `data-cy` attribute, I think it's more clear that this is something cypress-specific.

That said, this is something I have not put a lot of thought into, so I'll defer to whatever you think is better!

abhijit-singh:
This is great @cpoile! A few thoughts:
1. For task assignment DMs - It feels more natural to say "`Person X` assigned `you` the task `task title`", similar to what Jira does, instead of saying "`X` changed the assignee of `task title` from `Person A` to `Person B`". However, I also think it is important to mention who was assigned that task previously. Maybe we can add something "Previously assigned to `Person A`" at the end. Will defer to @justinegeffen for a better recommendation here. 
2. Is it possible and would it make sense to add a delay and club some of these messages if they were done in quick succession? Eg. "`Person X` assigned 3 tasks to `you` for the run `Everything is on fire, send nurses!`" and then list down the tasks in a bullet list. 
3. I can imagine these getting lost in the `playbooks` bot DM as we add things. Could adding #hashtags to help filter the messages help with that? Eg. #taskassigned 

Also, it'll probably be a good idea for @stephenvanhemmen or @itao to take a look at this PR and also review some of the suggestions above. Thanks!

stephenvanhemmen:
> This is great @cpoile! A few thoughts:
> 
> 1. For task assignment DMs - It feels more natural to say "`Person X` assigned `you` the task `task title`", similar to what Jira does, instead of saying "`X` changed the assignee of `task title` from `Person A` to `Person B`". However, I also think it is important to mention who was assigned that task previously. Maybe we can add something "Previously assigned to `Person A`" at the end. Will defer to @justinegeffen for a better recommendation here.
> 2. Is it possible and would it make sense to add a delay and club some of these messages if they were done in quick succession? Eg. "`Person X` assigned 3 tasks to `you` for the run `Everything is on fire, send nurses!`" and then list down the tasks in a bullet list.
> 3. I can imagine these getting lost in the `playbooks` bot DM as we add things. Could adding #hashtags to filter the messages help with that? Eg. #taskassigned
> 
> Also, it'll probably be a good idea for @stephenvanhemmen or @itao to take a look at this PR and also review some of the suggestions above. Thanks!

These are amazing suggestions. One of the biggest risks with moving in-channel notifications to DM is noise, even without sending the task complete notifications. 

+1 for adding a delay to group these messages, although it might be nice to open up the copy for different assigners, for example, "`You` were assigned 3 tasks for the run `run name`" with the normal messages in a bullet list underneath, which would allow for different assigners. @justinegeffen any copy feedback would be great :)

Being able to manage messages by hashtag would be nice if the other types of DM notifications have them as well, #digest, #statusupdate, #retrospective, for example

justinegeffen:
Thanks, @abhijit-singh and @stephenvanhemmen.

I took a look through this PR and couldn't find the copy to edit inline so I'd suggest the following:

- "``Member`` assigned you the task ``task title`` (previously assigned to ``member``)". I really like your suggestion, @abhijit-singh and I don't think there's anything that needs to be changed. :)
- @stephenvanhemmen, do you think "You've been assigned three tasks for the run ``run name``" would work? Our house style is that numbers from 0-9 are spelled out and from 10 upwards are digits. Not sure if this will create any design issues with spacing though so I'm flexible on this.

LMK if that works and if so, I'm happy to make the changes in the PR. :)

webapp/src/components/checklist_item.tsx
cpoile: Awesome suggestion, doing it with a data-cy now. Thanks!

cpoile:
@justinegeffen @abhijit-singh @stephenvanhemmen Here's the updated message:
image

The idea of batching messages is nice, but it would take a bit of effort to do that. You would need to buffer them, for a specific receiver, then clear the buffer after a certain time period -- I'd propose to leave that for later (if ever). If we really want it, we should make a ticket and prioritize it.

stephenvanhemmen:
> @justinegeffen @abhijit-singh @stephenvanhemmen Here's the updated message: image
> 
> The idea of batching messages is nice, but it would take a bit of effort to do that. You would need to buffer them, for a specific receiver, then clear the buffer after a certain time period -- I'd propose to leave that for later (if ever). If we really want it, we should make a ticket and prioritize it.
> 
> @stephenvanhemmen I am adding a telemetry event for clicking on the run link: `dm_assignedtask_clicked` -- I'm not sure where you're keeping the list of them, but you should add that one and an explanation of what it means? Thanks!

I support leaving batched messages for now. To keep it lean, we can gather feedback on this iteration and see if it is important to customers. 

Thanks for adding this event, I hadn't thought of that. I'll categorize it with the other DM notification events. I currently don't have a combined list of telemetry events for the new features, so I'll make a list of them and share it with our team.

stephenvanhemmen:
@cpoile Can we adjust the messaging to put the hashtag at the end of the message instead of the beginning? Abhijit and I felt that having them at the beginning distracts user from reading out each message, working against the objective of this change, although it's important to have them somewhere in the message for ease of filtering

cpoile:
@stephenvanhemmen Of course, no problem. How does this look?
image

stephenvanhemmen:
> @stephenvanhemmen Of course, no problem. How does this look? image

Looks great, thank you!

PR #912, opened 2021-11-25
Improved run permissions specs

lieut-data:
#### Summary
Rewrite our existing run permissions spec to more clearly breakdown the current behaviour. Notably, while system admins only see runs from teams to which they belong, they can (currently) directly navigate to any playbook overview. Or rather, they should be able to, but there was a JavaScript exception preventing same. Now the list and overview permissions are explicitly tested:

CleanShot 2021-11-25 at 17 16 40@2x

This is effectively a pre-PR to my work to introduce "sudo admins", which let admins behave like normal users all the time until they explicitly ask for their complete admin privileges.

#### Ticket Link
None.

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


tests-e2e/cypress/integration/runs/permissions_spec.js
```diff
@@ -0,0 +1,505 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
```
lieut-data, line 1: Note that I deleted the file above and effectively rewrote it as this one. (It was in the wrong folder anyway: it was testing runs, not playbooks.)

tests-e2e/cypress/integration/runs/permissions_spec.js
```diff
@@ -0,0 +1,505 @@
...
+            });
+
+            it('to admins in the team', () => {
+                assertRunIsVisible(run, sysadminInTeam);
+            });
+
+            // XXX: The following asserts that while sysadmins don't see runs from other teams in
+            // the list, they still have access to view the overview directly. Once we support
+            // sudo-admins, we should change this behaviour to be consistent with normal users.
+            it('to admins not in the team (overview only)', () => {
```
lieut-data, line 183: Note that I'm intentionally breaking out `it` blocks instead of trying to generate these from some kind of key => expectedResult pattern. Here, I was able to deviate from the above and tweak the logic without having to complicate things.

tests-e2e/cypress/support/plugin/api_commands.js
```diff
@@ -3,6 +3,8 @@
 
 const playbookRunsEndpoint = '/plugins/playbooks/api/v0/runs';
 
+const StatusOK = 200;
```
lieut-data, line 6: The warning about `200` as a magic number was annoying me...

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/overview/participants.tsx
```diff
@@ -64,12 +64,12 @@ const Participants = (
                 <Participant
                     userId={playbookRun.owner_user_id}
                     isOwner={true}
-                    teamName={team.name}
+                    teamName={team?.name}
```
lieut-data, line 67: This doesn't really "solve" the problem of not have access to the team, but at least lets the page load for the admin. We can look at improving this in the future.

lieut-data:
@itao, if you could just read through the spec output and confirm that it's all what you expected, we should be good :)

tests-e2e/cypress/integration/runs/permissions_spec.js
```diff
@@ -0,0 +1,505 @@
...
+                cy.apiAddUserToTeam(testTeam.id, createdUser.id);
+            });
+
+            // # Create a public playbook and corresponding run with a public channel in
+            // team 1. This is to ensure the list isn't empty for users who can't access the
+            // run under test.
+            cy.apiCreatePlaybook({
+                teamId: testTeam.id,
+                title: 'Playbook (Team 1)',
+                memberIDs: [],
```
crspeller, line 70: Could you add `public: true` or `public: false` on these? That way these should just work with the permissions stuff coming in.

crspeller (approved):
Nice!

itao (approved):
Yup looks good to me!

cpoile (approved):

cpoile:
Very cool, I like how you split the tests into easy-to-understand verification statements.

Phrynobatrachus (approved):
runs passing and definitely is easier to follow. :+1: 

for the upcoming `public` Playbook field, perhaps we can set that based on `memberIDs` in the API custom command?

lieut-data:
> for the upcoming public Playbook field, perhaps we can set that based on memberIDs in the API custom command?

I suspect we'll definitely need to start introducing some kind of "member" of a playbook, even if it's public, i.e. to show things on the LHS like we do for channels. It will indeed help once we can decouple permission to do something on that from purely the memberlist.

PR #915, opened 2021-11-30
[MM-32517] Disallow archived channels

iomodo:
#### Summary

This PR does couple of three things:
1. Removes archived channels from the suggested list in boradcast
2. If arhived channels are already selected they are shown as "Unknown Channel"
3. Refactors broadcast to channels and broadcast to users who follow, without changeing the behaviour

https://user-images.githubusercontent.com/2340127/144041580-8f6a0dff-7c2f-440c-b250-c93b4f0a0305.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-32517

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


agarciamontoro (changes requested):
Wow, this was an old ticket. Thank you for tackling this! :tada:

Just a question about consistency between the broadcasting functions (and a nit you can forget about if you want to).

server/app/playbook_run_service.go
```diff
@@ -371,20 +354,12 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
...
-		for _, broadcastChannelID := range pb.BroadcastChannelIDs {
-			if err = s.broadcastPlaybookRunCreation(broadcastChannelID, message, playbookRun); err != nil {
-				s.pluginAPI.Log.Warn("failed to broadcast the playbook run creation to channel", "ChannelID", playbookRun.BroadcastChannelIDs, "error", err)
-
-				if _, err = s.poster.PostMessage(channel.Id, "Failed to announce the creation of this playbook run in the configured channel."); err != nil {
-					return nil, errors.Wrapf(err, "failed to post to channel")
-				}
-			}
+		if err := s.broadcastPlaybookRunCreationToChannels(pb.BroadcastChannelIDs, message, playbookRun, channel.Id); err != nil {
+			return nil, errors.Wrap(err, "failed to broadcast to channels")
```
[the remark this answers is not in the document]
iomodo: Agree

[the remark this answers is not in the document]
iomodo: Actually, it throws error when it can't post failure message on the configured channel. So maybe if message posting is failing playbook creation should fail as well?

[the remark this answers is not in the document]
agarciamontoro: @iomodo: But we don't fail with the other broadcasting messages, right? I guess that if this does fail, there's a bigger problem going on :thinking:

agarciamontoro (approved):
Thanks, @iomodo! I answered in the thread above about failing on the broadcast errors, but 0/5 on the solution, so approving now :tada:

webapp/src/components/backstage/channel_selector.tsx
```diff
@@ -24,16 +24,26 @@ export interface Props {
 }
 
 const getMyPublicAndPrivateChannels = (state: GlobalState) => getMyChannels(state).filter((channel) =>
-    channel.type !== General.DM_CHANNEL && channel.type !== General.GM_CHANNEL,
+    channel.type !== General.DM_CHANNEL && channel.type !== General.GM_CHANNEL &&
+    channel.type !== General.ARCHIVED_CHANNEL && channel.delete_at === 0,
```
crspeller, line 28: Archived channel is not a channel type. I think that constant is for something else. The delete at here is what counts.

webapp/src/components/backstage/channel_selector.tsx
```diff
@@ -24,16 +24,26 @@ export interface Props {
...
 );
 
 type GetChannelType = (channelID: string) => Channel
 
-const ChannelSelector = (props: Props & { className?: string }) => {
+const getChannel = (state: GlobalState, channelID: string): Channel => {
+    const channel = getChannelFromRedux(state, channelID);
+
+    if (channel && channel.type !== General.ARCHIVED_CHANNEL && channel.delete_at === 0) {
```
crspeller, line 36: Same with the channel type here.

webapp/src/components/backstage/channel_selector.tsx
```diff
@@ -24,16 +24,26 @@ export interface Props {
...
+    }
+    return {display_name: 'Unknown Channel', id: channelID} as Channel;
+};
+
+const ChannelSelector = (props: Props & {className?: string}) => {
     const {formatMessage} = useIntl();
     const selectableChannels = useSelector(getMyPublicAndPrivateChannels);
 
-    const getChannelFromID = useSelector<GlobalState, GetChannelType>((state) => (channelID) => getChannel(state, channelID) || {display_name: 'Unknown Channel', id: channelID});
+    const getChannelFromID = useSelector<GlobalState, GetChannelType>((state) => (channelID) => getChannel(state, channelID));
```
crspeller, line 46: Since we have our own selector here now can we just combine this inline function with the above?

server/api/playbook_runs.go
```diff
@@ -416,7 +416,15 @@ func (h *PlaybookRunHandler) createPlaybookRun(playbookRun app.PlaybookRun, user
 		}
 
 		if pb.BroadcastEnabled {
-			playbookRun.BroadcastChannelIDs = pb.BroadcastChannelIDs
+			channelIDs := []string{}
```
crspeller, line 419: This seems like a weird place to check for this since we have to check again anyways and this isn't even user visible.  Maybe remove?

server/app/broadcast.go
```diff
@@ -0,0 +1,98 @@
+package app
+
+import (
+	"fmt"
+
+	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/pkg/errors"
+)
+
+// broadcasting to channels
```
crspeller, line 10: All of these broadcast functions are quite similar. I feel like these should be unified and more generic. With inputs varying per location of use? They also all just use broadcast but go to different places like channels or DMs.

crspeller (changes requested):

server/api/playbook_runs.go
iomodo: This is for old playbooks which already have archived channels as a broadcast channel. When we create their run, we filter out the archived channels. Old runs do not change behaviour on the other hand.

webapp/src/components/backstage/channel_selector.tsx
iomodo: Not sure I understood what you mean. `selectableChannels` is the selector for my channels. `getChannelFromID` gets channels from `props.channelIds`, how could we combine these?

server/app/broadcast.go
iomodo: Totally agree, that's why I put them in one file to be more visible. Unification will change behaviour though. In this PR messages and when you get what message is not changed. I'll create new PR for that(as well as issue Alejandro mentioned), probably it could be tied to the whole notification settings effort.

server/api/playbook_runs.go
crspeller: Right, but in that case (or in the case that the channel becomes archived) we do a check when we are about to send the reminder. Shouldn't that cover it?

server/app/broadcast.go
crspeller: This file is so confusing that I would suggest fixing it in this pr.

webapp/src/components/backstage/channel_selector.tsx
crspeller: Took a closer look at this, it's stranger then I thought. It doesn't return state, it returns a function that takes a value then calls a selector on the "stolen" state. 
This is bad because this component will re-render every time redux changes. 

The easiest way to fix this would be to just select all the channels then reference the selected array to create the `values` below.

crspeller (changes requested):

iomodo:
Adding @calebroseland on a review since @crspeller is OOO

webapp/src/components/backstage/channel_selector.tsx
```diff
@@ -24,17 +24,35 @@ export interface Props {
...
+    channels.forEach((channel: Channel) => channelsMap.set(channel.id, channel));
+
+    const result: Channel[] = [];
+    channelIDs.forEach((id: string) => {
+        let filteredChannel: Channel;
+        const channel = channelsMap.get(id);
+        if (channel && channel.delete_at === 0) {
+            filteredChannel = channel;
+        } else {
+            filteredChannel = {display_name: 'Unknown Channel', id} as Channel;
```
calebroseland, line 45:
```suggestion
            filteredChannel = {display_name: formatMessage({defaultMessage: 'Unknown Channel'}), id} as Channel;
```

calebroseland (approved):
LGTM! Just one translatable string if possible

webapp/src/components/backstage/channel_selector.tsx
calebroseland: Else 
```suggestion
            filteredChannel = {display_name: '', id} as Channel;
```

With ... down in ChannelSelector
```tsx
                {channel.display_name || formatMessage({defaultMessage: 'Unknown Channel'})}
```

PR #918, opened 2021-12-01
Fix broken select in RHS

iomodo:
#### Summary
This PR fixes broken status update reminder select in RHS

https://user-images.githubusercontent.com/2340127/144250795-8355b767-c7f9-4a69-9a49-f65871b7dd13.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-40170

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/webapp/src/components/update_request_post.tsx
+++ b/webapp/src/components/update_request_post.tsx
@@ -124,7 +124,7 @@ export const UpdateRequestPost = (props: Props) => {
                         }),
                         menuPortal: (base: CSSProperties) => ({
                             ...base,
-                            zIndex: 9999,
+                            zIndex: 22,
                         }),
                     }}
                 />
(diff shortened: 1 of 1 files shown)


webapp/src/components/update_request_post.tsx
```diff
@@ -120,6 +120,11 @@ export const UpdateRequestPost = (props: Props) => {
                         control: (base: CSSProperties) => ({
                             ...base,
                             height: '40px',
+                            minWidth: '100px',
+                        }),
+                        menuPortal: (base: CSSProperties) => ({
+                            ...base,
+                            zIndex: 9999,
```
cpoile, line 127: I'm really hesitant to put such a high zIndex -- there might be modals and other UIs that need to be on top. Earlier with the sticky headers, I was able to get by with a zIndex of something like 3 I think. Can you try something lower, just in case?

cpoile (approved):
Awesome, thanks for fixing this @iomodo !

abhijit-singh (approved):
This is awesome @iomodo!

Just one tiny nit, can we make sure the space between the two buttons as well as the buttons and the snooze input field is equal? Thanks!
image

lieut-data:
Also going to merge this one for v1.22 without the second dev review.

PR #920, opened 2021-12-03
MM-38588: Manage checklists

agarciamontoro:
#### Summary

This PR adds the needed endpoints and UI components to create, delete and rename checklists. The video below shows the addition:

https://user-images.githubusercontent.com/3924815/144587186-87aaa3d4-854f-44d6-bdfb-eb0012dddfb6.mp4

Although there are more than 1000 lines added, do not despair! A big chunk corresponds to tests (and to telemetry) and the actual code is pretty straightforward. You can review by commit if you want to.

About the tests: I added E2E tests both server-side (thanks, @crspeller, I really enjoyed writing those!) and with Cypress. @Phrynobatrachus, I'd love if you can take a look at those.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38588

#### Checklist

- [X] Telemetry updated
- [X] Unit tests updated


tests-e2e/cypress/integration/channels/rhs/checklist_spec.js
```diff
@@ -223,5 +235,71 @@ describe('channels > rhs > checklist', () => {
...
+            cy.get('#rhsContainer').within(() => {
+                cy.findByText('Checklists').trigger('mouseover').within(() => {
+                    cy.findByTitle('Add checklist').click();
+                });
+            });
+
+            // # Type a title and click on the Add button
+            const title = 'Checklist - ' + Date.now();
+            cy.findByLabelText('Checklist name').type(title);
+            cy.findByRole('button', {name: 'Add'}).click();
```
Phrynobatrachus, line 250: oh def going to use this selector more

Phrynobatrachus (approved):
JS e2e looks good!

client/playbook_runs.go
```diff
@@ -151,3 +151,36 @@ func (s *PlaybookRunService) Finish(ctx context.Context, playbookRunID string) e
 
 	return nil
 }
+
+func (s *PlaybookRunService) CreateChecklist(ctx context.Context, playbookRunID string, checklist Checklist) error {
+	createURL := fmt.Sprintf("runs/%s/checklists", playbookRunID)
+	req, err := s.client.newRequest(http.MethodPut, createURL, checklist)
```
cpoile, line 157: should probably be `Post` since you're creating new, and never replacing, right?

server/app/playbook_run_service.go
```diff
@@ -1442,6 +1442,70 @@ func (s *PlaybookRunServiceImpl) RunChecklistItemSlashCommand(playbookRunID, use
...
+
+	if !s.hasPermissionToModifyPlaybookRun(playbookRunToModify, userID) {
+		return errors.New("user does not have permission to modify playbook run")
+	}
+
+	if !s.hasPermissionToModifyPlaybookRun(playbookRunToModify, userID) {
+		return errors.New("user does not have permission to modify playbook run")
+	}
+
+	playbookRunToModify.Checklists = append([]Checklist{checklist}, playbookRunToModify.Checklists...)
```
cpoile, line 1460: This confused me for a second... Is there a reason we're prepending?

server/api_runs_test.go
```diff
@@ -102,3 +104,228 @@ func TestRuns(t *testing.T) {
...
+		run := createNewRunWithNoChecklists(t)
+
+		// Try to create a new checklist with no title
+		err := e.PlaybooksClient.PlaybookRuns.CreateChecklist(context.Background(), run.ID, client.Checklist{
+			Title: "",
+			Items: []client.ChecklistItem{},
+		})
+		require.Error(t, err)
+
+		// Make sure that the checklist was not added
```
cpoile, line 186: I like these comments -- like a trimmed down version of the e2e tests -- you can scan them quickly to see what's happening, without having to read the code. We should adopt this as the standard for the new tests.

cpoile (approved):
Nice work @agarciamontoro !

client/playbook_runs.go
agarciamontoro: That was my first instinct, but then I saw that the endpoint to add a checklist item is also a PUT. The rationale being (I guess), that we're modifying the run, not the checklist itself.
That said, if your first instinct was also to have a POST here, I'll change it, I think it makes more sense.
Do you think we should change the addChecklistItem endpoint to POST, too?

server/app/playbook_run_service.go
agarciamontoro: This is by design. If we were to add the new checklist at the end, the user may not see it, as they would need to scroll to get into view. We could solve that by, well, automatically scrolling, but I think this is a bit cleaner.

client/playbook_runs.go
iomodo: Easy way to determine `put` vs `post` is whether second call should change resource or not. In this case second call adds another checklist, so I'd stick to `POST`. Same for the `addChecklistItem`

iomodo (changes requested):
Looks wonderful @agarciamontoro, just one comment regarding `put` vs `post`

client/playbook_runs.go
agarciamontoro: Done: https://github.com/mattermost/mattermost-plugin-playbooks/pull/920/commits/768cf3716dd31cc8045f8a6ff6a542ca87ac57e6

And I created a ticket for addChecklistItem: https://mattermost.atlassian.net/browse/MM-40490

server/api_runs_test.go
agarciamontoro: Indeed, I copied the idea from the Cypress tests! I even thought of getting the `#` and `*` convention, but seemed like too much structure, actually. We can try to follow this as a best practice, yup :)

iomodo (approved):

PR #921, opened 2021-12-03
Reduce Cypress waits

Phrynobatrachus:
Every `cy.visit` and `cy.reload` has some additional waiting applied to it, but it would be nice to reduce that if it doesn't increase test flake significantly. Several local runs seemed ok with these changes, this is to test CI. If that has problems, I'll change this to be included only with `CYPRESS_developerMode` so at least test development will be faster.

### DIFF
--- a/tests-e2e/cypress/support/extended_commands.js
+++ b/tests-e2e/cypress/support/extended_commands.js
@@ -3,12 +3,12 @@
 
 import * as TIMEOUTS from '../fixtures/timeouts';
 
-Cypress.Commands.overwrite('reload', (originalFn, forceReload, options, duration = TIMEOUTS.THREE_SEC) => {
+Cypress.Commands.overwrite('reload', (originalFn, forceReload, options, duration = TIMEOUTS.ONE_SEC) => {
     originalFn(forceReload, options);
     cy.wait(duration);
 });
 
-Cypress.Commands.overwrite('visit', (originalFn, url, options, duration = TIMEOUTS.THREE_SEC) => {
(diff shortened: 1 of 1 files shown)


tests-e2e/cypress/support/extended_commands.js
```diff
@@ -3,12 +3,12 @@
...
 import * as TIMEOUTS from '../fixtures/timeouts';
 
-Cypress.Commands.overwrite('reload', (originalFn, forceReload, options, duration = TIMEOUTS.THREE_SEC) => {
+Cypress.Commands.overwrite('reload', (originalFn, forceReload, options, duration = TIMEOUTS.ONE_SEC) => {
     originalFn(forceReload, options);
     cy.wait(duration);
 });
 
-Cypress.Commands.overwrite('visit', (originalFn, url, options, duration = TIMEOUTS.THREE_SEC) => {
+Cypress.Commands.overwrite('visit', (originalFn, url, options, duration = TIMEOUTS.ONE_SEC) => {
```
lieut-data, line 11: What happens if we drop this to zero? (i.e. remove the `overwrite` altogether)

lieut-data (approved):
This makes a difference! Before:

CleanShot 2021-12-03 at 16 06 47@2x

after:

CleanShot 2021-12-03 at 16 07 43@2x

tests-e2e/cypress/support/extended_commands.js
Phrynobatrachus: Slight time increase but I figure the failures are really the relevant thing, and all passed on both. I'm cool with no wait and keep an eye on rate.

lieut-data:
Woot! 
CleanShot 2021-12-03 at 17 02 20@2x

PR #922, opened 2021-12-06
MM-40305: Preliminary work to simplify RHSChecklists component

agarciamontoro:
#### Summary

This PR just moves code around to make the old `RHSChecklists` component a bit more manageable:
- `RHSChecklists` is now `RHSChecklistList` (just so we don't mistake the names `RHSChecklist` and `RHSChecklists`)
- The map over all checklists now return a new `RHSChecklist` component.
- The map over all items now return a new `RHSChecklistItem` component.

The goal is to simplify the work needed for MM-40305. Behaviour was not changed (at least not intentionally), so if tests pass I'm pretty confident everything should work. I've manually tested that the reordering of checklist items do still work.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40305

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/rhs/rhs_checklist.tsx
```diff
@@ -0,0 +1,140 @@
...
+                >
+                    {formatMessage({defaultMessage: '+ Add task'})}
+                </AddTaskLink>
+            </EmptyChecklistContainer>
+        );
+    }
+
+    if (!visibleTasks(props.checklist, checklistItemsFilter, myUser.id)) {
+        return null;
+    }
```
iomodo, line 79: Why do we return `null` here? No `Add task` needed?

webapp/src/components/rhs/rhs_checklist_item.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    checklistNum={props.checklistIndex}
+                    itemNum={props.itemIndex}
+                    channelId={props.playbookRun.channel_id}
+                    playbookRunId={props.playbookRun.id}
+                    onChange={(newState: ChecklistItemState) => {
+                        setChecklistItemState(props.playbookRun.id, props.checklistIndex, props.itemIndex, newState);
+                    }}
+                    onRedirect={() => {
+                        if (isMobile()) {
+                            dispatch(toggleRHS());
```
iomodo, line 43: Not sure I get this, could you elaborate why are we doing this?

iomodo (approved):
Looks good @agarciamontoro, just couple questions below for me to better understand this. Thanks!

webapp/src/components/rhs/rhs_checklist.tsx
agarciamontoro: You're absolutely right. Fixed in https://github.com/mattermost/mattermost-plugin-playbooks/pull/922/commits/5ecb791e09129a8fbe231d539f9068594d9ab1b3.

webapp/src/components/rhs/rhs_checklist_item.tsx
agarciamontoro: I simply copy-pasted the component, but I took a look now and it seems like the prop was not being used at all in `ChecklistItemDetails`, so I just got rid of it (https://github.com/mattermost/mattermost-plugin-playbooks/pull/922/commits/eb83e4ad7870d9d4d766245f2f4868e9fbca42e9). Thanks for catching that!

crspeller (approved):
Nice!

PR #926, opened 2021-12-06
Cypress: Reduce numTestsKeptInMemory

Phrynobatrachus:
This lowers the numbers of test debug snapshots available in the Cypress runner from the default `50` to `5` in order to avoid OOM crashes when running specs with a large number of tests (`edit_spec.js` for example).


agarciamontoro (approved):
:tada: Thanks!

PR #929, opened 2021-12-07
MM-40490: Endpoint to add a checklist item should be a POST, not PUT

cpoile:
#### Summary
- Changed endpoint to PUT
- client method didn't exist so added that (and tests)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40490

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/client/playbook_runs.go
+++ b/client/playbook_runs.go
@@ -184,3 +184,15 @@ func (s *PlaybookRunService) RenameChecklist(ctx context.Context, playbookRunID 
 	_, err = s.client.do(ctx, req, nil)
 	return err
 }
+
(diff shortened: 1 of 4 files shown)


iomodo (approved):

agarciamontoro (changes requested):
That was fast!

We also need to change the `clientAddChecklistItem` to do a POST. And maybe add a super simple E2E test that covers that use case? Because if this PR passed CI is because we're not exercising that feature at all!

cpoile:
@agarciamontoro Ahh, there's a reason I didn't notice that was there -- we never use it, we only use the slash command route. So I removed it. There, problem solved. :) If we need to add it at some point in the future, we can add the e2e test then. :)

agarciamontoro (approved):
That's a good solution :sunglasses:

PR #930, opened 2021-12-07
MM-40511: Improve retro editing/saving flow; Fix: Guard against publishing retro accidentally

cpoile:
#### Summary
- Clicking Publish will give you a confirmation modal now
- I took the initiative to change the editing/saving flow to be more like the RHS 
  - Hovering over the retro text area will change the cursor to a text icon. 
  - Clicking will open the editor. 
  - Typing will be debounced for 2 seconds and then persisted to the backend.
  - The way this is done with hooks is pretty sweet -- the textarea is the only component rerendering while typing, and persisting will not change the cursor location.
  - Clicking outside the text area will flush the debounce, persisting the text, the user sees the correct retro text -- even if the user does something like clicking a link before waiting for the debounce to send.
  - @crspeller I wonder if this is how we should be doing all inputs in the app, to fix typing delay issues. It's performant.
- e2e tests

#### UX/PM review

https://user-images.githubusercontent.com/1490756/145110772-b958691c-8caa-421a-bbac-8b8d9d33ee0f.mp4

#### Editing review
- @justinegeffen is the text ok?
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40511

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


tests-e2e/cypress/integration/runs/retrospective_spec.js
```diff
@@ -49,16 +49,26 @@ describe('runs > retrospective', () => {
...
-        cy.findByText('Save').click();
+
+        // # Save it by clicking outside the text area
+        cy.findByText('Report').click();
+
+        // # Publish
         cy.findByText('Publish').click({force: true});
 
+        // * Verify we're showing the publish retro confirmation modal
+        cy.get('#confirmModalLabel').contains('Publish retrospective');
```
agarciamontoro, line 67: Here we may have the possibility to use the dialog role:
```suggestion
        cy.findByRole('dialog').contains('Publish retrospective');
```
I have not tested it, though.

tests-e2e/cypress/integration/runs/retrospective_spec.js
```diff
@@ -49,16 +49,26 @@ describe('runs > retrospective', () => {
...
+
+        // # Publish
+        cy.get('#confirmModalButton').click({force: true});
```
agarciamontoro, line 70: And similarly here, we could probably look for the Publish button (maybe `within` the dialog, if needed)
```suggestion
        cy.findByRole('button', {name: 'Publish'}).click({force: true});
```

agarciamontoro (approved):
Much cleaner and much more intuitive :tada:

I left a handful of non-blocking nits below.

justinegeffen:
@cpoile, text looks good to me thank you! :)

stephenvanhemmen (approved):
The new editing/saving flow looks fantastic, thanks for everything that went into this! 🎉

justinegeffen (approved):

webapp/i18n/en.json
```diff
@@ -192,6 +193,7 @@
   "WAHCT2": "Notify System Admin",
   "WIxhrv": "Run name must have at least two characters",
   "WTQpnI": "Take action now using playbooks",
+  "WbsomC": "Publish retrospective",
   "X/koAN": "Invalid entry: the maximum number of webhooks allowed is 64",
   "X2K92H": "Checklist name",
   "X3DLGJ": "Everyone in this workspace can create playbooks.",
```
justinegeffen, line 199:
```suggestion
  "X3DLGJ": "Everyone in this workspace can create playbooks",
```

justinegeffen: Nit, but this is the only one that ends in punctuation other than the question.

crspeller (approved):
Nice! I would like to see this in other edit fields as well. Generic textbox component incoming? Starting with @agarciamontoro's nit?

webapp/i18n/en.json
cpoile: Hey @justinegeffen! This is from the backstage settings page -- can you look at that as a whole? (there are other sentences with punctuation on the page)

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/retrospective/report.tsx
```diff
@@ -104,34 +60,41 @@ const Report = (props: ReportProps) => {
...
             </Header>
-            {editing &&
-                <ReportTextarea
-                    autoFocus={true}
-                    value={props.playbookRun.retrospective}
-                    onChange={(e) => {
-                        props.setRetrospective(e.target.value);
+            {
+                editing &&
+                <ReportTextArea
```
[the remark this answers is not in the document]
cpoile: Yep, that's a good idea, done!

tests-e2e/cypress/integration/runs/retrospective_spec.js
```diff
@@ -49,16 +49,26 @@ describe('runs > retrospective', () => {
         cy.visit(`/playbooks/runs/${runId}/retrospective`);
 
         // # Start editing
-        cy.findByText('Edit').click();
+        cy.findByTestId('retro-report-text').click();
```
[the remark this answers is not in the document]
cpoile: The problem is that at that point it's not a text box, it's a `PostTextContainer-styledcomponentshash` and a post. I'm open to any other way, just can't think of one at the moment.

tests-e2e/cypress/integration/runs/retrospective_spec.js
cpoile: Yah, there's only one role in that modal (it's `document`, probably not specific enough) -- the webapp gives us this id, so we can just use it. I don't know what the benefit of using something less specific would be...?

tests-e2e/cypress/integration/runs/retrospective_spec.js
cpoile: Similar here --not sure why we should bother if the id is already there? Personally, I think it's a poor use of our effort to refactor the tests in this way? (2/5)

cpoile:
@crspeller turns out, we don't need to make the textArea controlled, if we're lifting the persist to the parent. Sweet.

tests-e2e/cypress/integration/runs/retrospective_spec.js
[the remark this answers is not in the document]
agarciamontoro: Oooh, that makes sense. I guess it's ok here, then.

tests-e2e/cypress/integration/runs/retrospective_spec.js
agarciamontoro: I agree that refactoring existing tests that already work just for the sake of making them assess accessibility levels is not worth it. But when we write new tests, or are already modifying ones, I think it may be a good practice to use these queries so that we don't only test the feature, but the accessibility itself: if it's hard to make a query, maybe there's room for improvement for making the site more accessible.

That said, in this specific example we may not need it, so 0/5 on changing this line of code. Having this conversation was already valuable :nerd_face:

tests-e2e/cypress/integration/runs/retrospective_spec.js
agarciamontoro: From what I've seen writing the tests, thinking this way has forced me to change the DOM when needed to make it more accessible and semantic. For example, in this case it would be good that the modal had the `dialog` role. But it's not something we need to do now in this PR.

PR #936, opened 2021-12-09
MM-40155: Remove Dismiss Prompt

cpoile:
#### Summary
- Removed the dismiss server code and the UI elements
- A more reasonable max-width for the snooze input
- Migrating old runs whose reminder was dismissed. Leaving runs that have an overdue reminder unchanged. (don't want to reset a reminder that recently fired, for ex.)

image

#### UX question
- @abhijit-singh I'm making the `Post update` button not wrap. Let me know if you want it back to original

New:
image

Original:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40155

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


server/sqlstore/migrations.go
```diff
@@ -1584,4 +1584,49 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.43.0"),
+		toVersion:   semver.MustParse("0.44.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
```
crspeller, line 1590: Should we really do this migration? I am unsure of this decision to not allow dismiss. But resurrecting old reminders seems like it will do way more harm then good. Why not just have this decision effect future reminders?

crspeller (changes requested):
Not sure I get why we are removing this functionality. Maybe we should keep it on as an option? In ether case the migration seems extreme if I am understanding it correctly.

server/sqlstore/migrations.go
iomodo: I agree with @crspeller here.

abhijit-singh (commented):
Thanks, @cpoile, looks good! You're right, the button text should not wrap. 

0/5 on the migration but I'm leaning towards not resurrecting any old reminders. 

A couple of minor visual points - 
1. In the RHS state, can we make the dropdown open below the input box instead of above it?
2. Can we make the width of the dropdown popup greater than that of the input box?

Current:
Screenshot 2021-12-15 at 2 20 14 AM

Suggested:
image

server/sqlstore/migrations.go
cpoile: @crspeller we did do almost this same migration earlier, resurrecting old reminders, so this isn't new. But I understand your point. @itao or @lieut-data, can you weigh in on this?

lieut-data: Indeed, the intended behaviour of the previous change was to prohibit not having reminders at all so that runs don't just get forgotten. This just "completes" that effort.

@cpoile, one thought though: does this take into account the possibility that status updates might be disabled for a run from @gbochora's changes? We should make sure we don't set reminders for a run which can't then post a status update :)

iomodo (approved):

server/sqlstore/migrations.go
crspeller: @lieut-data Do we understand that this is on the `IR_Incident` table? It resurrects old incidents and not just updates some default in the playbook? 

If I understand correctly, practical behavior of this migration is one week afterwards a flurry of updates will hit users that dismissed updates.

lieut-data: Indeed, the goal here is to find all incidents without an active "reminder" and force them to have one, one week out from today. Hopefully, we caught most of them last time, and this change should only touch the select few that happened to squeak through due to the ability to "ignore" the reminder.

crspeller: Alright. If that's what we want to do.

PR #941, opened 2021-12-13
MM-35035: Cypress parallelism

Phrynobatrachus:
#### Summary
This enables parallel test runs in CircleCI. Based on some test runs it cuts the Cypress jobs to ~9 minutes, which means the whole workflow will also be ~9 min since those are the longest running jobs.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35035

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,10 +140,8 @@ commands:
           name: Run Cypress Tests
           no_output_timeout: 30m
           command: |
-            export FAILURE_MESSAGE="At least one test has failed."
-            export RESULTS_OUTPUT="results-output.txt"
             cd tests-e2e
-            circleci tests glob "cypress/integration/**/*.js" | circleci tests split > ./tests-to-run
+            circleci tests glob "cypress/integration/**/*.js" | circleci tests split --split-by=timings > ./tests-to-run
             npm run test -- --spec "$(cat ./tests-to-run | tr '\n' ',')"
 
       - *save_cypress_cache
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -140,9 +140,10 @@ commands:
           name: Run Cypress Tests
           no_output_timeout: 30m
           command: |
-            export FAILURE_MESSAGE="At least one test has failed."
-            export RESULTS_OUTPUT="results-output.txt"
-            cd tests-e2e && npm run test |& tee $RESULTS_OUTPUT; if grep "$FAILURE_MESSAGE" "$RESULTS_OUTPUT"; then exit 1; fi
+            cd tests-e2e
+            circleci tests glob "cypress/integration/**/*.js" | circleci tests split --split-by=timings > ./tests-to-run
+            npm run test -- --spec "$(cat ./tests-to-run | tr '\n' ',')"
```
lieut-data, line 145: Any idea what the `results-output.txt` used to do, @cpoile?

lieut-data (approved):
Crazy that 6 lines of code will save us hours and hours of waiting....

.circleci/config.yml
Phrynobatrachus: Oh yeah meant to note this removal, I couldn't find anywhere that output file was used and the job ends with `exit 1` on failure even without it. Definitely can put it back if there's something I missed with it though.

cpoile: ...Nope. :)

cpoile (approved):
Great # lines to performance improvement ratio, @Phrynobatrachus!

PR #943, opened 2021-12-14
MM-38150 Advanced Permissions support for playbooks.

crspeller:
#### Summary
Advanced permissions support for Playbooks. 

- Adding members and roles to playbooks.
- Adding ability in the system console to designate which roles have which permissions.

Permissions:
- PermissionPublicPlaybookCreate
- PermissionPublicPlaybookManageProperties
- PermissionPublicPlaybookManageMembers
- PermissionPublicPlaybookView
- PermissionPublicPlaybookMakePrivate
- PermissionPrivatePlaybookCreate
- PermissionPrivatePlaybookManageProperties
- PermissionPrivatePlaybookManageMembers
- PermissionPrivatePlaybookView
- PermissionPrivatePlaybookMakePublic
- PermissionRunCreate
- PermissionRunManageProperties
- PermissionRunManageMembers
- PermissionRunView

- Adding the full fleged concept of public and private playbooks.
- New playbook creation modal with support for templates and public/private selection.
- New permissions service to handle all permissions related queries.
- Lots of tests being migrated to the new system.


crspeller:
The cypress tests pass for me locally. Still need to figure out why they don't work on CI.

crspeller:
Tests are fixed now. Cloud e2e not expected to pass until the server changes reach that branch. (not sure when that will be)

server/app/permissions_service.go
```diff
@@ -0,0 +1,445 @@
...
+
+// IsGuest returns true if the userID is a system guest
+func IsGuest(userID string, pluginAPI *pluginapi.Client) (bool, error) {
+	user, err := pluginAPI.User.Get(userID)
+	if err != nil {
+		return false, errors.Wrapf(err, "Unable to get user to determine permissions, user id `%s`", userID)
+	}
+
+	return user.IsGuest(), nil
+}
```
agarciamontoro, line 430: Nit: should this return just a bool like `IsSystemAdmin` does?

server/command/command.go
```diff
@@ -222,15 +232,10 @@ func (r *Runner) actionRun(args []string) {
 		postID = args[1]
 	}
 
-	if !app.CanViewTeam(r.args.UserId, r.args.TeamId, r.pluginAPI) {
-		r.postCommandResponse("Must be a member of the team to run playbooks.")
-		return
-	}
-
```
agarciamontoro, line 229: I guess we got rid away of these because this is centralized in the `playbookRunService`?

server/sqlstore/migrations.go
```diff
@@ -1584,4 +1584,35 @@ var migrations = []Migration{
...
+					return errors.Wrapf(err, "failed adding column Roles to table IR_Playbook")
+				}
+				if err := addColumnToMySQLTable(e, "IR_Playbook", "Public", "BOOLEAN DEFAULT FALSE"); err != nil {
+					return errors.Wrapf(err, "failed adding column Roles to table IR_Playbook")
+				}
+				if _, err := e.Exec("UPDATE IR_PlaybookMember SET Roles = 'playbook_member playbook_admin' WHERE Roles IS NULL"); err != nil {
+					return errors.Wrapf(err, "failed setting default value in column Roles of table IR_Playbook")
+				}
+			} else {
+				if err := addColumnToPGTable(e, "IR_PlaybookMember", "Roles", "TEXT DEFAULT 'playbook_member playbook_admin'"); err != nil {
```
agarciamontoro, line 1602: This is a bit different than in MySQL: should we default to null, and then only add `playbook_member playbook_admin` for existing playbooks?
If we keep this and we make a mistake in the future, the default will be that everyone is an admin.

server/api/bot.go
```diff
@@ -70,9 +70,17 @@ func (h *BotHandler) notifyAdmins(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func CanStartTrialLicense(userID string, pluginAPI *pluginapi.Client) error {
+	if !pluginAPI.User.HasPermissionTo(userID, model.PermissionManageLicenseInformation) {
+		return errors.Wrap(app.ErrNoPermissions, "no permission to manage license information")
+	}
+
+	return nil
+}
+
```
agarciamontoro, line 80: Nit: Shouldn't this live in the permissions file?

server/api/bot_test.go
```diff
@@ -1,134 +0,0 @@
-package api
```
agarciamontoro, line 1: Why did we get rid of these tests?

server/api/error_handler.go
```diff
@@ -23,3 +23,15 @@ func (h *ErrorHandler) HandleError(w http.ResponseWriter, internalErr error) {
 func (h *ErrorHandler) HandleErrorWithCode(w http.ResponseWriter, code int, publicErrorMsg string, internalErr error) {
 	HandleErrorWithCode(h.log, w, code, publicErrorMsg, internalErr)
 }
+
+// PermissionsCheck handles the output of a permisions check
+// Automaticly does the proper error handling.
+// Retruns true if the check passed and false on failure. Correct use is: if !h.PermissionsCheck(w, check) { return }
```
agarciamontoro, line 29: Some typos
```suggestion
// Automatically does the proper error handling.
// Returns true if the check passed and false on failure. Correct use is: if !h.PermissionsCheck(w, check) { return }
```

server/api/playbook_runs.go
```diff
@@ -123,12 +133,7 @@ func (h *PlaybookRunHandler) checkEditPermissions(next http.Handler) http.Handle
 			return
 		}
 
-		if err := app.EditPlaybookRun(userID, playbookRun.ChannelID, h.pluginAPI); err != nil {
-			if errors.Is(err, app.ErrNoPermissions) {
-				h.HandleErrorWithCode(w, http.StatusForbidden, "Not authorized", err)
-				return
-			}
-			h.HandleError(w, err)
+		if !h.PermissionsCheck(w, h.permissions.RunManageProperties(userID, playbookRun.ID)) {
```
agarciamontoro, line 136: We should maybe add a RunManagePropertiesByRun (probably better named) so that we don't need to retrieve the run again, given that we already have it here.

agarciamontoro (commented):
Submitting the first pass, I'll continue later. Some initial thoughts:
- I'm not sure what the differences between `Roles` and `SchemeRoles` are.
- I'd love for the permissions to have an interface that let us do something like `permissions.Playbook.Create()`, instead of `permissions.PlaybookCreate`. But that's just a nice-to-have.

webapp/src/components/backstage/select_users_below.tsx
```diff
@@ -23,53 +44,150 @@ const UserList = styled.div`
...
+    if (roles.includes(PlaybookRole.Admin)) {
+        return <FormattedMessage defaultMessage='Playbook Admin'/>;
+    }
+
+    return <FormattedMessage defaultMessage='Playbook Member'/>;
+}
+
 const SelectUsersBelow = (props: SelectUsersBelowProps) => {
+    const permissionToEditMembers = useHasPlaybookPermission(PlaybookPermissionGeneral.ManageMembers, props.playbook);
+    const teamnameNameDisplaySetting = useSelector(getTeammateNameDisplaySetting) || '';
```
agarciamontoro, line 78: Typo here:
```suggestion
    const teammateNameDisplaySetting = useSelector(getTeammateNameDisplaySetting) || '';
```

webapp/src/components/backstage/select_users_below.tsx
```diff
@@ -23,53 +44,150 @@ const UserList = styled.div`
...
+    onRemoveUser: (userid: string) => void;
+    onMakeAdmin: (userid: string) => void;
+    onMakeMember: (userid: string) => void;
+}
+
+const MemberButton = styled.div`
+    display: inline-flex;
+    border-radius: 4px;
+    fill: rgb(var(--link-color-rgb));
+    color: rgb(var(--link-color-rgb));
```
agarciamontoro, line 130: We can probably simplify this:
```suggestion
    fill: var(--link-color);
    color: var(--link-color);
```

webapp/src/components/backstage/public_private_selector.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+const StackedText = styled.div`
+	flex-grow: 1;
+	text-align: left;
+`;
+
+const GiantIcon = styled.i<{active?: boolean}>`
+	font-size: 36px;
+	line-height: 42px;
+	color: ${(props) => (props.active ? 'rgb(var(--button-bg-rgb))' : 'rgba(var(--center-channel-color-rgb), 0.56)')};
```
agarciamontoro, line 45: We can simplify this too
```suggestion
	color: ${(props) => (props.active ? 'var(--button-bg)' : 'rgba(var(--center-channel-color-rgb), 0.56)')};
```

webapp/src/components/backstage/public_private_selector.tsx
```diff
@@ -0,0 +1,125 @@
...
+`;
+
+const CheckIcon = styled.i`
+	font-size: 24px;
+	line-height: 24px;
+	color: rgb(var(--button-bg-rgb));
```
agarciamontoro, line 51: And this:
```suggestion
	color: var(--button-bg);
```

webapp/src/components/backstage/public_private_selector.tsx
```diff
@@ -0,0 +1,125 @@
...
+	font-size: 12px;
+	line-height: 16px;
+	color: rgba(var(--center-channel-color-rgb), 0.56);
+`;
+
+const PublicPrivateSelector = (props: Props) => {
+    const {formatMessage} = useIntl();
+    const handleButtonClick = (pub: boolean) => {
+        props.setPlaybookPublic(pub);
+    };
```
agarciamontoro, line 70: We can probably define this as 
```suggestion
    const handleButtonClick = props.setPlaybookPublic;
```

tests-e2e/cypress/support/index.js
```diff
@@ -197,9 +197,9 @@ function sysadminSetup(user) {
 
     // # Reset roles
     cy.apiGetClientLicense().then(({isLicensed, isCloudLicensed}) => {
-        if (isLicensed) {
-            cy.apiResetRoles();
-        }
+        //if (isLicensed) {
+        //cy.apiResetRoles();
+        //}
```
agarciamontoro, line 202: Should we remove if not needed?

tests-e2e/cypress/integration/runs/permissions_spec.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
agarciamontoro, line 1: Only you could catch this error :joy:

agarciamontoro (commented):
Second pass, all new code is now reviewed (although I mostly skimmed through the webapp code).
The only missing part are the sqlstore tests and the api tests. I'll get to them tomorrow morning.

agarciamontoro:
@crspeller: I talked to @cpoile and @lieut-data and we all agreed that it seems too risky to merge this right before a two-week (three for me!) break. The main reasons being:
- We'll definitely miss stuff during the review.
- The main goal of merging this now is to get some soaking time before 6.3, but if there's nobody around, it will be difficult to find bugs, and more difficult to fix them.
- Also, it turns out that 6.3 is an ESR, so if something's broken, we'll have to patch it for the next six months.

So, although I know we wanted this in before the holidays, what do you think about deferring it until after we're back in January? @lieut-data offered to talk to the PMs so the decision to defer to 6.4 is well understood by everyone.

server/main_test.go
```diff
@@ -335,7 +434,10 @@ func (e *TestEnvironment) SetE20Licence() {
 func (e *TestEnvironment) CreateBasic() {
 	e.CreateClients()
 	e.CreateBasicServer()
-	e.CreateBasicPlaybooks()
+	e.SetE20Licence()
+	e.CreateBasicPlaybook()
+	e.CreateBasicRun()
+	e.CreateAdditionalPlaybooks()
```
agarciamontoro, line 440: I'm slightly concerned that this `CreateBasic` function will grow indefinitely, taking a lot of time every time we call it.

server/api_playbooks_test.go
```diff
@@ -6,36 +6,41 @@ import (
...
+	"github.com/stretchr/testify/require"
 )
 
 func TestPlaybooks(t *testing.T) {
 	e := Setup(t)
 	e.CreateClients()
 	e.CreateBasicServer()
 
-	t.Run("unlicenced servers can't create a playbook with members", func(t *testing.T) {
+	t.Run("unlicenced servers can't create a private playbooks", func(t *testing.T) {
```
agarciamontoro, line 20: Typo
```suggestion
	t.Run("unlicenced servers can't create a private playbook", func(t *testing.T) {
```

server/api_playbooks_test.go
```diff
@@ -183,3 +190,539 @@ func TestPlaybooks(t *testing.T) {
 		assert.Equal(t, 1, playbookResults.TotalCount)
 	})
 }
+
+func TestPlaybooksRetrival(t *testing.T) {
```
agarciamontoro, line 194: Typo
```suggestion
func TestPlaybooksRetrieval(t *testing.T) {
```

server/api_playbooks_test.go
```diff
@@ -183,3 +190,539 @@ func TestPlaybooks(t *testing.T) {
...
+
+	t.Run("update playbook without chaning existing broadcast channel", func(t *testing.T) {
+		e.BasicPlaybook.BroadcastChannelIDs = []string{e.BasicPrivateChannel.Id}
+		err := e.PlaybooksAdminClient.Playbooks.Update(context.Background(), *e.BasicPlaybook)
+		require.NoError(t, err)
+
+		e.BasicPlaybook.Description = "unrelated update"
+		err = e.PlaybooksClient.Playbooks.Update(context.Background(), *e.BasicPlaybook)
+		require.NoError(t, err)
+	})
```
agarciamontoro, line 235: These tests just check that the request did not error, but we should probably retrieve the playbooks again and check that the description has indeed changed.

server/api_playbooks_test.go
```diff
@@ -183,3 +190,539 @@ func TestPlaybooks(t *testing.T) {
...
+
+func TestPlaybooksPermissions(t *testing.T) {
+	e := Setup(t)
+	e.CreateBasic()
+
+	t.Run("test no permissions to create", func(t *testing.T) {
+		defaultRolePermissions := e.Permissions.SaveDefaultRolePermissions()
+		defer func() {
+			e.Permissions.RestoreDefaultRolePermissions(defaultRolePermissions)
+		}()
```
agarciamontoro, line 627: We could probably use `t.Cleanup` instead of `defer`

server/api_playbooks_test.go
```diff
@@ -183,3 +190,539 @@ func TestPlaybooks(t *testing.T) {
...
+		_, err := e.PlaybooksClient2.Playbooks.Get(context.Background(), e.BasicPrivatePlaybook.ID)
+		requireErrorWithStatusCode(t, err, http.StatusForbidden)
+	})
+
+	t.Run("list playbooks", func(t *testing.T) {
+		t.Run("user in private", func(t *testing.T) {
+			results, err := e.PlaybooksClient.Playbooks.List(context.Background(), e.BasicTeam.Id, 0, 100, client.PlaybookListOptions{})
+			require.NoError(t, err)
+
+			expectedIDs := getPlaybookIDsList([]client.Playbook{*e.BasicPlaybook, *e.BasicPrivatePlaybook})
```
agarciamontoro, line 659: Should we create the playbooks in the test? I'm concerned we'll have to change these every time we change the basic playbooks (which may not be super frequently, but still)

server/api_playbooks_test.go
```diff
@@ -183,3 +190,539 @@ func TestPlaybooks(t *testing.T) {
...
+		t.Run("user not in private", func(t *testing.T) {
+			err := e.PlaybooksClient2.Playbooks.Update(context.Background(), *e.BasicPrivatePlaybook)
+			requireErrorWithStatusCode(t, err, http.StatusForbidden)
+		})
+
+		t.Run("public with no permissions", func(t *testing.T) {
+			defaultRolePermissions := e.Permissions.SaveDefaultRolePermissions()
+			defer func() {
+				e.Permissions.RestoreDefaultRolePermissions(defaultRolePermissions)
+			}()
```
agarciamontoro, line 710: Same comment here, we can use `t.Cleanup`

server/api_runs_test.go
```diff
@@ -87,22 +88,216 @@ func TestRunCreation(t *testing.T) {
...
+
+	t.Run("create valid run without playbook", func(t *testing.T) {
+		run, err := e.PlaybooksClient.PlaybookRuns.Create(context.Background(), client.PlaybookRunCreateOptions{
+			Name:        "No playbook",
+			OwnerUserID: e.RegularUser.Id,
+			TeamID:      e.BasicTeam.Id,
+		})
+		assert.NoError(t, err)
+		assert.NotNil(t, run)
+	})
```
agarciamontoro, line 183: Oh, we can create runs without playbooks?

server/api_runs_test.go
```diff
@@ -87,22 +88,216 @@ func TestRunCreation(t *testing.T) {
...
+			Name:        "Basic create",
+			OwnerUserID: e.RegularUser.Id,
+			TeamID:      e.BasicTeam.Id,
+			PlaybookID:  e.ArchivedPlaybook.ID,
+		})
+		assert.Error(t, err)
+	})
+}
+
+func TestRunRetreval(t *testing.T) {
```
agarciamontoro, line 224: Typo
```suggestion
func TestRunRetrieval(t *testing.T) {
```

agarciamontoro (changes requested):
Ok, I think I reviewed *everything* :tada: This is an amazing piece of work, thank you for tackling this, it's incredible!

I left some more comments, and I also have an additional general question:
- What are exacly the `Default{Playbook,Run}{Admin,Member}Role`a fields?

server/api/playbooks.go
```diff
@@ -363,32 +324,14 @@ func (h *PlaybookHandler) getPlaybooks(w http.ResponseWriter, r *http.Request) {
...
-	if isGuest, errg := app.IsGuest(userID, h.pluginAPI); errg != nil {
-		h.HandleError(w, errg)
-		return
-	} else if isGuest {
-		h.HandleErrorWithCode(w, http.StatusForbidden, "Not authorized", errors.Errorf(
-			"userID %s does not have permission to get playbooks on teamID %s because they are a guest",
-			userID,
-			teamID,
-		))
+	if teamID != "" && !h.PermissionsCheck(w, h.permissions.PlaybookList(userID, teamID)) {
```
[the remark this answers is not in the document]
crspeller: Yes but it handles it with failure which it should. We actually want to succeed here since a blank team ID is valid and permissions are handled differently.

tests-e2e/cypress/integration/runs/permissions_spec.js
crspeller: Only my automatic linting on save could catch this error. 😆

server/app/permissions_service.go
```diff
@@ -0,0 +1,445 @@
...
+		if member.UserID == userID {
+			return member.SchemeRoles
+		}
+	}
+
+	// Public playbooks
+	if playbook.Public {
+		if playbook.DefaultPlaybookMemberRole == "" {
+			return []string{playbook.DefaultPlaybookMemberRole}
+		}
```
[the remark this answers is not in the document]
crspeller: No "" means use the default role.

go.mod
```diff
@@ -3,24 +3,26 @@ module github.com/mattermost/mattermost-plugin-playbooks
...
 	github.com/rudderlabs/analytics-go v3.3.1+incompatible
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.7.0
 	github.com/writeas/go-strip-markdown v2.0.1+incompatible
-	gotest.tools v2.2.0+incompatible
 )
 
 replace github.com/mattermost/mattermost-plugin-playbooks/client => ./client
 
+replace github.com/HdrHistogram/hdrhistogram-go => github.com/codahale/hdrhistogram v1.1.2
```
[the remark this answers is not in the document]
crspeller: The build fails without it.

server/app/permissions_service.go
```diff
@@ -0,0 +1,445 @@
...
+		return []string{PlaybookRoleMember}
+	}
+
+	return []string{}
+}
+
+func (p *PermissionsService) getRunRole(userID string, run *PlaybookRun) []string {
+	// For now, everyone with access to the channel is a run admin
+	if p.pluginAPI.User.HasPermissionToChannel(userID, run.ChannelID, model.PermissionReadChannel) {
+		return []string{"run_member", "run_admin"}
```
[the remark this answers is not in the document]
crspeller: Done.

server/app/permissions_service.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	if p.hasPermissionsToPlaybook(userID, playbook, model.PermissionPrivatePlaybookView) {
+		return nil
+	}
+
+	return noAccessErr
+}
+
+func (p *PermissionsService) PlaybookMakePrivate() {}
```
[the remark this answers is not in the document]
crspeller: Good catch. This wasn't being checked properly.

server/api/bot.go
crspeller: It's only used in this file and probably unlikely to be used elsewhere.

server/api/playbook_runs.go
crspeller: https://mattermost.atlassian.net/browse/MM-40793

server/api/playbooks.go
```diff
@@ -333,17 +295,16 @@ func (h *PlaybookHandler) restorePlaybook(w http.ResponseWriter, r *http.Request
...
-		return
-	}
-
 	playbookToRestore, err := h.playbookService.Get(playbookID)
 	if err != nil {
 		h.HandleError(w, err)
 		return
 	}
 
+	if !h.PermissionsCheck(w, h.permissions.DeletePlaybook(userID, playbookToRestore)) {
```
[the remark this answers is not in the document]
crspeller: Ya that makes sense I think for it to be symmetric with the ability to archive/delete. We could fracture it into to permissions if we wanted to.

server/api/stats.go
```diff
@@ -89,9 +90,8 @@ func (h *StatsHandler) playbookStats(w http.ResponseWriter, r *http.Request) {
 	activeParticipantsPerDay, activeParticipantsPerDayTimes := h.statsStore.ActiveParticipantsPerDayLastXDays(14, filters)
 
 	ReturnJSON(w, &PlaybookStats{
-		RunsInProgress:                h.statsStore.TotalInProgressPlaybookRuns(filters),
-		ParticipantsActive:            h.statsStore.TotalActiveParticipants(filters),
-		RunsFinishedPrev30Days:        runsFinishedLast30Days,
+		RunsInProgress:     h.statsStore.TotalInProgressPlaybookRuns(filters),
+		ParticipantsActive: h.statsStore.TotalActiveParticipants(filters), RunsFinishedPrev30Days: runsFinishedLast30Days,
```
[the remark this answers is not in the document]
crspeller: Somehow gofmt was ok with that.

server/sqlstore/playbook.go
```diff
@@ -98,71 +98,84 @@ func applyPlaybookFilterOptionsSort(builder sq.SelectBuilder, options app.Playbo
...
 		).
-		From("IR_Playbook")
-
-	memberIDsSelect := sqlStore.builder.
-		Select("PlaybookID", "MemberID").
-		From("IR_PlaybookMember").
-		OrderBy("MemberID ASC") // Entirely for consistancy for the tests
+		From("IR_Playbook p").
+		LeftJoin("Teams t ON t.Id = p.TeamID").
+		LeftJoin("Schemes s ON t.SchemeId = s.Id")
```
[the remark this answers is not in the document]
crspeller: We will probably always need to check permissions.

webapp/src/components/backstage/archive_playbook_modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {FormattedMessage, useIntl} from 'react-intl';
+
+import {archivePlaybook} from 'src/client';
+import {Banner} from 'src/components/backstage/styles';
+import {Playbook} from 'src/types/playbook';
+import ConfirmModal from '../widgets/confirmation_modal';
+
+const ArchiveBannerTimeout = 5000;
+
+const useConfirmPlaybookArchiveModal = (): [React.ReactNode, (pb: Playbook) => void] => {
```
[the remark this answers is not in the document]
crspeller: Its more of a component than a hook and it's very specific not a general purpose tool.

webapp/src/types/permissions.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export type Permission = RunPermission | PlaybookPermission
+
+export enum RunPermission {
+    ManageProperties = 'run_manage_properties',
+    ManageMembers = 'run_manage_members',
+    View = 'run_view',
+}
+
+export enum PlaybookPermissionGeneral {
```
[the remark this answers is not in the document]
crspeller: It's entirely for simplification. The easier we make it the less permissions issues we have hopefully :)

webapp/src/types/playbook.ts
```diff
@@ -187,7 +197,7 @@ export function isPlaybook(arg: any): arg is PlaybookWithChecklist {
         typeof arg.create_public_playbook_run === 'boolean' &&
         typeof arg.delete_at === 'number' &&
         arg.checklists && Array.isArray(arg.checklists) && arg.checklists.every(isChecklist) &&
-        arg.member_ids && Array.isArray(arg.member_ids) && arg.checklists.every((id: any) => typeof id === 'string') &&
+        arg.members && Array.isArray(arg.members) && arg.checklists.every((id: any) => typeof id === 'string') &&
```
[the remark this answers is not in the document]
crspeller: It's not used. Deleted.

webapp/src/components/widgets/generic_modal.tsx
```diff
@@ -177,6 +178,8 @@ const StyledModal = styled(Modal)`
         }
     }
 
+	z-index: 1040;
```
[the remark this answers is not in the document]
crspeller: One could say magical. It is the number needed to make it so modals can appear above other modals in sequence.

tests-e2e/cypress/support/index.js
crspeller: We could, but I kept it around to show differences between our version and the webapp version.

server/main_test.go
crspeller: Me too. It's a trade off between ease of use and speed. If we want to optimize it should be easy enough to decompose later.

server/api_playbooks_test.go
crspeller: Probably. I was just mindlessly copying the originals which also never tested that. I guess they had the excuse that everything was faked so there really was no point. 
https://mattermost.atlassian.net/browse/MM-40794

server/api_playbooks_test.go
crspeller: `t.Cleanup` is only necessary if you are worried about state not getting cleaned up on a test panicking. In these cases I don't think we care.

server/api_playbooks_test.go
crspeller: The solution here is to not change the basic playbooks. If you need a specialized one then create it for the test.

server/api_runs_test.go
crspeller: Apparently. There was a test for it that I copied.

webapp/src/types/playbook.ts
```diff
@@ -177,39 +187,6 @@ export const ChecklistItemsFilterDefault: ChecklistItemsFilter = {
     others: true,
 };
 
-// eslint-disable-next-line
-export function isPlaybook(arg: any): arg is PlaybookWithChecklist {
```
cpoile, line 181: yah, I guess we should give up on this.

cpoile (approved):
👍

build/custom.mk
```diff
@@ -4,7 +4,6 @@ GO_BUILD_FLAGS += -ldflags "-X main.rudderDataplaneURL=$(MM_RUDDER_DATAPLANE_URL
 ## Generate mocks.
 mocks:
 ifneq ($(HAS_SERVER),)
-	go install github.com/golang/mock/mockgen
```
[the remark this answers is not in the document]
crspeller: I think it was breaking because of a version mismatch.

server/api/playbook_runs.go
```diff
@@ -334,7 +342,7 @@ func (h *PlaybookRunHandler) addToTimelineDialog(w http.ResponseWriter, r *http.
 		return
 	}
 
-	if err := app.EditPlaybookRun(userID, playbookRun.ChannelID, h.pluginAPI); err != nil {
+	if !h.PermissionsCheck(w, h.permissions.RunManageProperties(userID, playbookRun.ID)) {
```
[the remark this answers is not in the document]
crspeller: We did change the frontend to say Manage Configurations. We could change it to that? Really 0/5.

lieut-data:
@crspeller, I've dismissed @agarciamontoro's review to allow us to choose to merge this. Can you help get this deployed to community today so we can use it for bug bash tomorrow?

PR #933, opened 2021-12-08
MM-40304: Move items between checklists

agarciamontoro:
#### Summary

This PR allows for items to be moved between checklists in the RHS. The gist of it is:
- In the server, I refactored the old `/reorder` endpoint to be a more generic `/move`, to which we send the checklist number and item number we want to move, as well as the checklist number and item number we want to move it to.
- In the webapp, every checklist was wrapped inside a `DragDropContext`, so the checklists did not know about each other. I changed that, so now there is a single `DragDropContext` for all the checklists, and I tweaked the `onDragEnd` function to recompute the new state and call the new `/move` endpoint. Also, I converted the component returned when there are no visible items in the checklist to be a `Droppable`, so we can move items there too.

https://user-images.githubusercontent.com/3924815/145248569-59691d37-f8c6-4573-8c4e-4e742aa7cf44.mp4

I did not add E2E Cypress tests for this, but I did test it extensively with the server integration tests, which I'm *loving*.

Awaiting #922.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40304

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Gated by experimental feature flag~~
- [X] Unit tests updated

PR #944, opened 2021-12-15
MM-40305: Reorder checklists in RHS

agarciamontoro:
#### Summary

This PR adds the ability to reorder checklists in the RHS:
- It adds an endpoint to move a checklist (the old `/move` now moves checklists, whereas the one to move items is now called `/move-item`)
- It refactors the `DragDropContext` structure from react-beautiful-dnd:
  - it first contains a `Droppable`, where all checklists can be dropped,
  - every checklist is now a `Draggable`, with a new icon that shows on hover that acts as the handle,
  - each checklist is still a `Droppable`, with every item being a `Draggable`.
- To keep the collapsed state of every checklist in sync, I had to add a new action to the store so that we can overwrite the whole state of `Record<number, boolean>` that keeps track of it: if we move the i-th checklist to the j-th position, we need to update the state of every k-th checklist, with `i <= k <= j` (assuming `i < j`). Otherwise, the checklists would change their collapsed state depending on the position they are in. Maybe it's time to add proper IDs to checklists.

https://user-images.githubusercontent.com/3924815/146217995-a6a09321-7896-41e5-b171-02e2010493b2.mp4

This PR is based off #933, and so it should be merged after that one is merged.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40305

#### Checklist

- [X] Telemetry updated
- [X] Unit tests updated

### DIFF
--- a/server/api_runs_test.go
+++ b/server/api_runs_test.go
@@ -402,7 +402,7 @@ func TestChecklistManagement(t *testing.T) {
 
 	type ExpectedError struct{ StatusCode int }
 
-	tests := []struct {
+	moveItemTests := []struct {
 		Title              string
 		Checklists         [][]string
 		SourceChecklistIdx int
@@ -526,7 +526,7 @@ func TestChecklistManagement(t *testing.T) {
 		},
 	}
 
-	for _, test := range tests {
+	for _, test := range moveItemTests {
 		t.Run(test.Title, func(t *testing.T) {
 			// Create a new empty run
 			run := createNewRunWithNoChecklists(t)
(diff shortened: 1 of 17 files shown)


server/api_runs_test.go
```diff
@@ -573,4 +573,139 @@ func TestChecklistManagement(t *testing.T) {
 			}
 		})
 	}
+
+	moveChecklistTests := []struct {
```
cpoile, line 577: nice tests (again!)

server/app/playbook_run_service.go
```diff
@@ -1641,6 +1641,38 @@ func (s *PlaybookRunServiceImpl) EditChecklistItem(playbookRunID, userID string,
...
+
+	if destChecklistIdx < 0 || destChecklistIdx >= len(playbookRunToModify.Checklists) {
+		return errors.New("invalid destChecklist")
+	}
+
+	// Get checklist to move
+	checklistMoved := playbookRunToModify.Checklists[sourceChecklistIdx]
+
+	// Delete checklist to move
+	playbookRunToModify.Checklists = append(playbookRunToModify.Checklists[:sourceChecklistIdx], playbookRunToModify.Checklists[sourceChecklistIdx+1:]...)
```
cpoile, line 1659: This time, because `Checklist` contains `[]ChecklistItem` it's possible that we'll have a memory leak, right? (see: https://github.com/golang/go/wiki/SliceTricks#delete).  Because in the future `checklistMoved` could be deleted from `destChecklist`, but it might still be held by `sourceChecklist`.

cpoile (changes requested):
This is awesome! Nice work again @agarciamontoro! Only one thing (I think) should be fixed, otherwise great!

server/app/playbook_run_service.go
agarciamontoro: Great catch! Please take a thorough look to https://github.com/mattermost/mattermost-plugin-playbooks/pull/944/commits/64ca0f4f575dff82820b545acc755ab67ccfa9a7: I explain in the commit what I did, and I *think* it's correct, but I'd love if you can double-check I'm right.

server/app/playbook_run_service.go
```diff
@@ -1656,10 +1656,10 @@ func (s *PlaybookRunServiceImpl) MoveChecklist(playbookRunID, userID string, sou
 	checklistMoved := playbookRunToModify.Checklists[sourceChecklistIdx]
 
 	// Delete checklist to move
-	playbookRunToModify.Checklists = append(playbookRunToModify.Checklists[:sourceChecklistIdx], playbookRunToModify.Checklists[sourceChecklistIdx+1:]...)
+	copy(playbookRunToModify.Checklists[sourceChecklistIdx:], playbookRunToModify.Checklists[sourceChecklistIdx+1:])
+	playbookRunToModify.Checklists[len(playbookRunToModify.Checklists)-1] = Checklist{}
```
cpoile, line 1660: You'll also want to shorten the checklist, right?
```suggestion
	playbookRunToModify.Checklists[len(playbookRunToModify.Checklists)-1] = Checklist{}
	playbookRunToModify.Checklists = playbookRunToModify.Checklists[:len(playbookRunToModify.Checklists)-1]
```

cpoile (changes requested):

server/app/playbook_run_service.go
agarciamontoro: @cpoile: Immediately after the deletion we insert the new element, so instead of shortening the checklist and then appending a new item:

```go
// Delete
copy(c[i:], c[i+1:])
c[len(c)-1] = Checklist{}
c = c[:len(c)-1]

// Insert
c = append(c, Checklist{})
copy(c[j+1:], c[j:])
c[j] = checklistMoved
```

I simply keep the same length and copy the elements to their position:

```go
// Delete
copy(c[i:], c[i+1:])
c[len(c)-1] = Checklist{}

// Insert
copy(c[j+1:], c[j:])
c[j] = checklistMoved
```

Does that make sense?

cpoile: ah, I was afraid I overlooked something, and that was it. Thanks!

cpoile (approved):
Nice work!

calebroseland (approved):
About as good as DND can get!

PR #947, opened 2021-12-17
Playbooks v2

lieut-data:
#### Summary
There's no "rush" to merge this -- but let's plan to bump this before cutting v2 for inclusion in the ESR/v6.3 in January.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-40365

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (changes requested):
Why?

lieut-data:
> Why?

It's a good question. Mostly this was to "finish" v2 as we pivot to focussing on v3 this year. We could do that "in spirit" instead of actually bumping the version and introducing all the Go module pain though. We could also just bump the plugin version I think and not the code.

Thoughts?

lieut-data:
We've decided this change isn't needed, nor high impact. We've "hit" v2 internally as a team, but will leave the in-code version untouched at this time.

PR #948, opened 2021-12-17
MM-40268: Restructure edit screen tabs

cpoile:
#### Summary
- This turned out to be bigger than expected, but it's broken into easy to digest commits
- See the desired figma designs: https://www.figma.com/file/7cTztN73jmHGrlDdpu7qCG/Playbook-preview?node-id=127%3A32894
- There are a few things in the designs not implemented
  - Add as a follower (broken into a separate ticket: 
  - When a new member joins -> send an outgoing webhook (ticket: 
  - When a run is finished -> export channel (ticket: 
- We probably need to add a proper retrospective tab upgrade notice/flow (ticket: 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40268

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


server/api/playbook_runs.go
```diff
@@ -402,7 +402,9 @@ func (h *PlaybookRunHandler) createPlaybookRun(playbookRun app.PlaybookRun, user
 		playbookRun.Checklists = pb.Checklists
 		public = pb.CreatePublicPlaybookRun
 
-		playbookRun.Summary = pb.RunSummaryTemplate
+		if pb.RunSummaryTemplateEnabled {
```
lieut-data, line 405: Is the rationale for a new "bit" to allow the user to disable the playbook while preserving the actual content of the template?

lieut-data (approved):
I took a quick scan -- all seems to be in order. I didn't exhaustively review the frontend changes, but given the obviously working demo + warranty guarantee (;P), I'm content. Only one comment re: the schema.

server/api/playbook_runs.go
cpoile: @lieut-data That's what I assumed, since I didn't want the UX of: spending time editing a template, then turning the toggle off, then changing your mind and turning it back on, only to find it's been wiped out. So we're saving the template, and using the bit to decide on `createPlaybookRun` whether or not to use the saved template (if any).

webapp/src/components/backstage/playbook_edit/automation/channel_access.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        <RadioInput
+                            type='radio'
+                            checked={playbook.create_public_playbook_run}
+                        />
+                        <Icon
+                            active={playbook.create_public_playbook_run}
+                            className={'icon-globe'}
+                        />
+                        <BigText>{'Public'}</BigText>
+                    </Button>
```
calebroseland, line 62: These two should probably be wrapped in labels instead of buttons, they're naturally clickable and don't interfere with tab focus indexing.

I18n, also

```suggestion
                    <ButtonLabel>
                        <RadioInput
                            type='radio'
                            checked={playbook.create_public_playbook_run}
                            onChange={() => handlePublicChange(true)}
                        />
                        <Icon
                            active={playbook.create_public_playbook_run}
                            className={'icon-globe'}
                        />
                        <BigText>{formatMessage({defaultMessage: 'Public'})}</BigText>
                    </ButtonLabel>
```

calebroseland (changes requested):
This is great! Much nicer. Just one comment on the radio buttons.

webapp/src/components/backstage/playbook_edit/automation/channel_access.tsx
```diff
@@ -0,0 +1,127 @@
...
+                />
+                <div>{formatMessage({defaultMessage: 'Create a channel'})}</div>
+            </AutomationTitle>
+            <HorizontalSplit>
+                <VerticalSplit>
+                    <ButtonLabel onClick={() => handlePublicChange(true)}>
+                        <RadioInput
+                            type='radio'
+                            checked={playbook.create_public_playbook_run}
+                        />
```
calebroseland, line 56:
```suggestion
                    <ButtonLabel>
                        <RadioInput
                            type='radio'
                            checked={playbook.create_public_playbook_run}
                            onChange={() => handlePublicChange(true)}
                        />
```

webapp/src/components/backstage/playbook_edit/automation/channel_access.tsx
```diff
@@ -0,0 +1,127 @@
...
+                            className={'icon-globe'}
+                        />
+                        <BigText>{formatMessage({defaultMessage: 'Public'})}</BigText>
+                    </ButtonLabel>
+                    <HorizontalSpacer size={8}/>
+                    <ButtonLabel onClick={() => handlePublicChange(false)}>
+                        <RadioInput
+                            type='radio'
+                            checked={!playbook.create_public_playbook_run}
+                        />
```
calebroseland, line 68:
```suggestion
                    <ButtonLabel>
                        <RadioInput
                            type='radio'
                            checked={!playbook.create_public_playbook_run}
                            onChange={() => handlePublicChange(false)}
                        />
```

calebroseland (changes requested):

webapp/src/components/backstage/playbook_edit/automation/channel_access.tsx
cpoile: @calebroseland Hmm, I'm not sure why -- is that better?

calebroseland: It'll respond to non-click input events aimed at the radio inputs (e.g. arrow keys once keyboard-focused)

calebroseland:
```
Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.
```

cpoile: That warning is a little weird -- I guess it can't tell it's a controlled component for some reason?

calebroseland: Even though it's wrapped, it probably still applies the validation in the same way.

calebroseland (approved):

PR #949, opened 2021-12-17
MM-40656: Z-index issue between sticky headers and user profile dropdown

cpoile:
#### Summary
- A recent change made the checklist headers change their z-index based on how many checklists there are -- so I'm changing the profile dropdown to have a z-index of 50, under the assumption that we won't ever have that many checklists -- unless you think we should go even higher?

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40656

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data (approved):
Works for me!

calebroseland (approved):
😃
Something something [Occam's Razor](https://en.wikipedia.org/wiki/Occam%27s_razor)...

PR #950, opened 2021-12-21
MM-40748: Fix cannot assign tasks in RHS

calebroseland:
#### Summary
Fix cannot assign tasks in RHS

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40748

#### Checklist

Item | Y / N / NA [^1]
---|---
Telemetry updated? | NA
Gated by experimental feature flag? | NA
Unit/E2E tests updated? | NA


iomodo (approved):
Great! Could you fix the e2e test as well?

Phrynobatrachus (approved):
Duped the issue locally and verified the fix. I looked into the failed test but I haven't had any failures re-running locally, and it passed on a CI re-run as well. I'm thinking there might be a more general inconsistency with viewports in CI, not necessarily something in the test itself, since the `copy playbook link` test that was failing a bunch the other week was for a similar reason.

PR #951, opened 2021-12-21
MM-40731: Split Cypress by timings 

Phrynobatrachus:
#### Summary
Cypress runs are now parallelized after #941, but I realized they were not actually being split by timings despite passing `--split-by=timings`. I believe this is because the output of `mocha-junit-reporter` doesn't quite line up with CircleCI's expectations for the recorded XML results. Here's an example result of the current reporter output:

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="Mocha Tests" time="8.6360" tests="2" failures="0">
  <testsuite name="Root Suite" timestamp="2021-12-21T16:12:42" tests="0" file="cypress/integration/navigation_spec.js" time="0.0000" failures="0">
  </testsuite>
  <testsuite name="navigation" timestamp="2021-12-21T16:12:43" tests="2" time="8.6360" failures="0">
    <testcase name="navigation switches to playbooks list view via header button" time="6.9570" classname="switches to playbooks list view via header button">
    </testcase>
    <testcase name="navigation switches to playbook runs list view via header button" time="1.6790" classname="switches to playbook runs list view via header button">
    </testcase>
  </testsuite>
</testsuites>
```
There's 2 issues with this.

- CircleCI looks for a `file` attribute on the `<testcase>` elements, which is not present. There's not really much documentation around how splitting/timing/storing results actually works, I'm gathering this from threads such as [1] and messing around. 

- The `<testcase>` elements are individual tests from the same spec file, but splitting happens at the file level.

My attempted solution for this is to simply cobble together some XML manually, as opposed to trying to patch `mocha-junit-reporter` or fix up its output post-hoc. My version looks like this instead:

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="Cypress tests">
  <testsuite>
    <testcase name="cypress/integration/channels/actions_spec.js" file="cypress/integration/channels/actions_spec.js" time="34.202"/>
  </testsuite>
  <testsuite>
    <testcase name="cypress/integration/channels/rhs/header_spec.js" file="cypress/integration/channels/rhs/header_spec.js" time="10.537"/>
  </testsuite>
  <testsuite>
    <testcase name="cypress/integration/channels/run_dialog_spec.js" file="cypress/integration/channels/run_dialog_spec.js" time="18.487"/>
  </testsuite>
  <testsuite>
    <testcase name="cypress/integration/navigation_spec.js" file="cypress/integration/navigation_spec.js" time="5.481"/>
  </testsuite>
  <testsuite>
    <testcase name="cypress/integration/playbooks/creation_button_spec.js" file="cypress/integration/playbooks/creation_button_spec.js" time="9.578"/>
  </testsuite>
  <testsuite>
    <testcase name="cypress/integration/playbooks/edit_spec.js" file="cypress/integration/playbooks/edit_spec.js" time="91.958"/>
  </testsuite>
</testsuites>
```

This appears to work. Compare this run [2] that has parallelism but not these changes, to this run [3] that has both. [2]'s Cypress job step mentions:

Screenshot from 2021-12-21 12-18-33

And the timings for the parallel runners are a bit unbalanced:

Screenshot from 2021-12-21 12-18-13

Whereas for [3], the Cypress job detects timings:

Screenshot from 2021-12-21 12-22-09

And the runner times are all quite close:

Screenshot from 2021-12-21 12-22-21

[1] https://discuss.circleci.com/t/split-by-timings-on-elixir-project-fails-to-find-stored-timings-possibly-missing-file-names-in-junit-xml/34813
[2] https://app.circleci.com/pipelines/github/mattermost/mattermost-plugin-playbooks/5094/workflows/6c9c31e9-add3-4f6e-8d91-6eebf5ad50d1/
[3] https://app.circleci.com/pipelines/github/mattermost/mattermost-plugin-playbooks/5107/workflows/37cad505-26c3-4ee2-ae84-048bc84ee97e

#### Open Questions / Follow-Up
- If a Cypress job fails, it seems like a subsequent run will not have timing data and will fallback to splitting by filename.  My original hope was that results would be automagically averaged over time by Circle, but I'm not sure that's the case. Is there a way to rely on a previous good result instead?

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40731


Phrynobatrachus:
There's definitely more tuning I can follow up on here, but I think it should be good to go in as is.

.circleci/config.yml
```diff
@@ -143,12 +143,16 @@ commands:
             cd tests-e2e
             circleci tests glob "cypress/integration/**/*.js" | circleci tests split --split-by=timings > ./tests-to-run
             npm run test -- --spec "$(cat ./tests-to-run | tr '\n' ',')"
-
+            echo "<?xml version=\"1.0\" encoding=\"UTF-8\"?><testsuites name=\"Cypress tests\">" > ./cypress/results/junit.xml
```
lieut-data, line 146: Heh, this feels like an epic solution :D

lieut-data (approved):
Nice hack :)

PR #952, opened 2021-12-22
[MM-39899] Make restore idempotent

iomodo:
#### Summary
This PR makes `/restore` endpoint idempotent

#### Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-39899

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


Phrynobatrachus:
Maybe a silly question but, why does removing that error make `Restore` idempotent? Given that:

```
if playbook.ID == "" {
		return errors.New("can't restore a playbook without an ID")
}
```
also returns a similar error.

iomodo:
> Maybe a silly question but: why does removing that error make `Restore` idempotent, given that:
> 
> ```
> if playbook.ID == "" {
>   return errors.New("can't restore a playbook without an ID")
> }
> ```
> 
> still can return a similar error?

PUT method is idempotent, POST is not. Meaning with PUT if you send the same request multiple times, that should be equivalent to the single request. With POST if you send the request N times, you will end up having N resources(or the changed state).

In this context when you restore already restored playbooks it should not error.

Phrynobatrachus (approved):

calebroseland (approved):

PR #954, opened 2021-12-29
[MM-38732] Disallow publishing retro repeatedly

iomodo:
#### Summary
This PR will disallow publishing retros repeatedly 

https://user-images.githubusercontent.com/2340127/147648692-a9ee8559-488b-4c08-8bf3-4a9e1d675303.mp4

#### Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-38732

#### Checklist

- ~~[ ] Telemetry updated~~
- [X] Unit tests updated


webapp/src/components/backstage/playbook_runs/playbook_run_backstage/retrospective/report.tsx
```diff
@@ -22,26 +24,44 @@ interface ReportProps {
...
+                </DisabledPrimaryButtonSmaller>
+                <FormattedDuration
+                    from={publishedAtThisSession || props.playbookRun.retrospective_published_at}
+                    ago={true}
+                />
+                <span>{'Published'}</span>
+                <i className={'icon icon-check-all'}/>
             </>
         );
-    } else if (props.playbookRun.retrospective_published_at && !props.playbookRun.retrospective_was_canceled) {
```
Phrynobatrachus, line 43: this was the only usage of `playbookRun.retrospective_was_canceled` I could find. is that needed for a distinct UI state, or can we get rid of the field entirely from the Go struct and corresponding TS interface?

calebroseland (approved):
LGTM! I like the new UX 🙂

I hope it's okay, but I pushed a change to add i18n to the published check. Timestamp is a bit easier to translate around, so I applied the same time spec Alejandro had applied in `rhs_post_update`. Of course feel free to revert or make any further changes necessary!

CleanShot 2022-01-05 at 00 27 08@2x
CleanShot 2022-01-05 at 00 24 09@2x

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/retrospective/report.tsx
iomodo: Good catch, btu we might still need this. We are still tracking the retrospective cancelation in the DB. For example when user canceled the retro we might want to display that here somewhere. Like `retrospective was canceled you can't edit the text`. It's not part of this PR though.

PR #961, opened 2022-01-05
MM-40837: Client side telemetry

cpoile:
#### Summary
- This moves the client-side telemetry to a single point: the RHSOpener. Why here? Because putting it anywhere else would conflict with the changes the RHSOpener makes to the same url. 
  - Open to renaming `makeRHSOpener` to something like `makeRHSOpenerAndTelemetry` -- but I 1/5 felt that was a bit too long, and a little confusing.
- To track a client-side event in a link, add the two params to the link: `telem=telemetry_action` and `id=playbookid`, which will be removed immediately on arriving at a channel which is involved in a playbook run.
  -  Q for @iomodo -- will this work with the way you're using telemetry?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40837

#### Checklist
- [x] Telemetry updated
- [ ] ~~Unit tests updated~~


webapp/src/rhs_opener.ts
```diff
@@ -46,6 +46,19 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
             store.dispatch(receivedTeamPlaybookRuns(fetched.items));
         }
 
+        // Record and remove telemetry
```
iomodo, line 49: Shouldn't this be global instead of only in `rhs`?
I though we would use `history.listen` to detect history changes.

server/app/playbook_run_service.go
```diff
@@ -2613,8 +2613,8 @@ func buildAssignedTaskMessageAndTotal(runs []AssignedRun, locale string) string
 	msg += T("app.user.digest.tasks.num_outstanding", total) + "\n\n"
 
 	for _, run := range runs {
-		msg += fmt.Sprintf("[%s](/%s/channels/%s?telem=todo_assignedtask_clicked&forceRHSOpen)\n",
-			run.ChannelDisplayName, run.TeamName, run.ChannelName)
+		msg += fmt.Sprintf("[%s](/%s/channels/%s?telem=todo_assignedtask_clicked&id=%s&forceRHSOpen)\n",
```
iomodo, line 2616: Maybe we could use `telem_action` and `telem_run_id`?
Since it will be global `id` might collide with other parameters.

iomodo (changes requested):

webapp/src/rhs_opener.ts
cpoile: That's what I was saying in the PR summary -- I'm not happy with how this works now. Let me rethink and try putting it all together in the listener.

webapp/src/rhs_opener.ts
```diff
@@ -46,6 +46,19 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
...
+        const url = new URL(window.location.href);
+        const searchParams = new URLSearchParams(url.searchParams);
+
+        if (searchParams.has('telem') && searchParams.has('id')) {
```
crspeller, line 53: I wonder if we should have one parameter and then encode all the values in one. That way it's a bit easier to add additional values for specific telemetry points?

webapp/src/rhs_opener.ts
cpoile: @iomodo I don't think there's a way to do this cleanly only in the `history.listen` -- we have to keep the url-rewriting logic in one place, for at least 2 reasons:
1. simply because it keeps the it encapsulated in one place
2. but also because if another section of the code tries to read and rewrite the url, they will rewrite the url multiple times (the telemetry will be called twice, for example, because the history.listen is called once for the first url change, then again for when the `rhsOpener` rewrites the url).

So, I think it's cleanest to keep everything in the `rhsOpener` -- you can't do the rhsOpen logic in the `history.listen` function (I've tried), because it needs to be evaluated  to the rest of the rhsOpener logic, and you can't move all of the `rhsOpener` logic into the `history.listen` function (I've tried) -- it's super finicky code, and I don't want to mess with it for this.

So, I propose we move the telemetry into the `rhsOpener`. I'm pretty sure it will detect telemetry even in the backstage, since the plugin code is still receiving state changes from the store. But if it doesn't for some reason, we can revisit then. Sound good?

webapp/src/rhs_opener.ts
cpoile: @crspeller Not exactly sure what you mean here, actually. Something like: `telem_action_plus_id=someaction+playbookid`? If that's the suggestion, I think I would prefer keeping them separate--let the search params do what they're made to do. 

As for adding additional values for telemetry in the future: I think the way we're doing telemetry is limiting us -- if we want to add additional data, we would have to add a new endpoint that would take a second/third/etc parameter. The way we're doing it now, we keep the single action endpoint and you have to use a new action: `telem_action` turns into `telem_action_and_extra_data`. Then on the data side we have to split `telem_action_and_extra_data` into `telem_action` and `extra_data`.

I suppose we could allow for multiple telem actions, since that's possible with search params. Maybe wait until we need that? It will be easy enough to add.

server/app/playbook_run_service.go
cpoile: Yep, that's a good idea, for url cleanliness. Will do. 👍

webapp/src/rhs_opener.ts
crspeller: I meant something like `telem=<encoded json blob>` so we are just basically passing a JS object though. 
Probably overthinking this a bit. Either way would work.

crspeller (approved):

webapp/src/rhs_opener.ts
iomodo: Sounds good, I'll test the telemetry detection in the backstage. I can address that in the separate PR if it's not working for some reason.

iomodo (approved):

PR #964, opened 2022-01-06
[MM-38732] Fix retro publish

iomodo:
#### Summary
This PR fixes retrospective publish

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-38732

#### Checklist

- ~~[ ] Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/runs/retrospective_spec.js
+++ b/tests-e2e/cypress/integration/runs/retrospective_spec.js
@@ -56,6 +56,17 @@ describe('runs > retrospective', () => {
 
     it('prevents repeated publishing', () => {
         editAndPublishRetro(runId);
+
+        // * Verify that can't publish
+        cy.findByText('Publish').should('not.be.enabled');
+
+        // # Navigate to the overview tab
+        cy.findByText('Overview').click();
+
+        // # Navigate directly to the retro tab
+        cy.findByText('Retrospective').click();
+
+        // * Verify that can't publish
         cy.findByText('Publish').should('not.be.enabled');
     });
 });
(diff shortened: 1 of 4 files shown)


Phrynobatrachus (approved):
nice, my b for not catching this on the first time through. :zap:

agarciamontoro (commented):
The fix seems fine, but I have two questions (which are independent of this fix, as far as I'm concerned, but worth bringing up):

1. Should we remove the hover state when the button is disabled?

https://user-images.githubusercontent.com/3924815/149809991-04ccefee-b292-4bad-8689-762d07b6c6a4.mp4

2. Should we add a question mark at the end of the title in the modal, so it reads "Are you sure you want to publish?"?

image

cpoile (approved):
Looks good, and I second @agarciamontoro 's two suggestions.

abhijit-singh (commented):
I agree with @agarciamontoro suggestions. 

Also, I just wanted to confirm my understanding of the logic we're going forward with. Please let me know if any of the below points are inaccurate:
 – If a user clicks on `skip retrospective` on any of the reminders in the channel, they can still go to the retrospective tab in the run page, edit the report, and publish it. 
 – On clicking the `publish` button, we show a confirmation modal, informing the user that they won't be able to edit it after publishing. 
 – Once published, the report becomes non-editable and we don't allow editing or publishing again.  
 
 
Lastly, I tried testing this but I'm not able to create any playbooks on this branch. It'll probably be because of me doing something wrong during setup though so don't take this too seriously. Here's a screenshot of what I see -
Screenshot 2022-01-19 at 6 08 45 PM

iomodo:
> – If a user clicks on skip retrospective on any of the reminders in the channel, they can still go to the retrospective tab in the run page, edit the report, and publish it.

@abhijit-singh This part was not correct. I fixed it now. You will be able to publish retrospective even if you canceled it previously.
Also added the question mark on the modal.
Regarding the button hover there's another ticket tracking that.

agarciamontoro (approved):
Cool! :tada:

abhijit-singh (approved):
Awesome, thanks @iomodo!

PR #965, opened 2022-01-06
Fix autoselection of team in playbook creation modal.

crspeller:
#### Summary
Fix autoselection of teams in playbook creation modal so there is always a team selected. 

#### Ticket Link
Part of: https://mattermost.atlassian.net/browse/MM-40862

### DIFF
--- a/webapp/src/components/create_playbook_modal.tsx
+++ b/webapp/src/components/create_playbook_modal.tsx
@@ -65,7 +65,7 @@ const PlaybookCreateModal = ({startingName, startingTeamId, startingTemplate, st
     const {formatMessage} = useIntl();
     const teams = useSelector(selectTeamsIHavePermissionToMakePlaybooksOn);
     const [name, setName] = useState(startingName);
-    const [teamId, setTeamId] = useState(teams.length === 1 ? teams[0].id : startingTeamId);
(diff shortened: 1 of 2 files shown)


webapp/src/components/create_playbook_modal.tsx
```diff
@@ -65,7 +65,7 @@ const PlaybookCreateModal = ({startingName, startingTeamId, startingTemplate, st
     const {formatMessage} = useIntl();
     const teams = useSelector(selectTeamsIHavePermissionToMakePlaybooksOn);
     const [name, setName] = useState(startingName);
-    const [teamId, setTeamId] = useState(teams.length === 1 ? teams[0].id : startingTeamId);
+    const [teamId, setTeamId] = useState<string>(teams.length === 1 ? teams[0].id : (startingTeamId || teams[0].id));
```
lieut-data, line 68: Do we need to guard against `teams.length === 0` for some unexpected reason?

lieut-data (approved):

webapp/src/components/create_playbook_modal.tsx
crspeller: It's not possible to get to this point without a team.

cpoile (approved):

PR #966, opened 2022-01-06
MM-40845 Fix the menu button opening too far right.

crspeller:
#### Summary
Fix the dot menus floating too far to the right.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40845

### DIFF
--- a/webapp/src/components/dot_menu.tsx
+++ b/webapp/src/components/dot_menu.tsx
@@ -57,7 +57,6 @@ export const DropdownMenu = styled.div<DropdownMenuProps>`
     list-style: none;
 
     padding: 10px 0;
-	margin-left: 50px;
     font-family: Open Sans;
     font-style: normal;
     font-weight: normal;


webapp/src/components/dot_menu.tsx
```diff
@@ -57,7 +57,6 @@ export const DropdownMenu = styled.div<DropdownMenuProps>`
     list-style: none;
 
     padding: 10px 0;
-	margin-left: 50px;
```
lieut-data, line 60: That happened to actually be my first fix too... but I was worried it was added to fix something else?

lieut-data (approved):

webapp/src/components/dot_menu.tsx
crspeller: I thought so too, but I checked the other cases and it seemed fine without it.

cpoile: I looked at the blame, and it says you added it yesterday? Do you remember why?

cpoile (approved):

webapp/src/components/dot_menu.tsx
crspeller: It came in with the advanced permissions stuff. I don't remember why.

calebroseland (approved):

PR #967, opened 2022-01-07
Mm 40660 enable retros and status updates for test data

gbochora:
#### Summary
This PR will update test data generation logic(`/playbooks test create-playbooks` and `/playbooks test bulk-data`).

**New Logic**:
* `/playbooks test create-playbooks n` command randomly selects unique n dummy playbooks and creates them.

* `/playbooks test bulk-data` command: 
    1. Gets playbooks list for a specific user from DB
    2. If there are no playbooks in DB, it will create dummy playbooks.
    3. Creates each run using a random playbook.

**Before**: In `command.generateTestData` and `command.actionTestGeneratePlaybooks` methods playbooks where cteated without specifying `pb.Members`. Because of that in `command.generateTestData` method, `playbookService.GetPlaybooksForTeam` call result always was empty list. 

**After**: Added playbook creator user to `Members` and `playbookService.GetPlaybooksForTeam` works properly. 

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-40660

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


server/command/command.go
```diff
@@ -1184,11 +1203,18 @@ func (r *Runner) actionTestGeneratePlaybooks(params []string) {
 	}
 
 	playbookIds := make([]string, 0, numPlaybooks)
+	randPlaybooksIndexes := getRandomUniqueNumbers(0, len(dummyListPlaybooks), numPlaybooks, time.Now().UTC().UnixNano())
 	for i := 0; i < numPlaybooks; i++ {
-		dummyPlaybook := dummyListPlaybooks[i]
+		dummyPlaybook := dummyListPlaybooks[randPlaybooksIndexes[i]]
```
crspeller, line 1208: You can simplify this by using https://pkg.go.dev/math/rand#Shuffle instead of the custom function.

crspeller (changes requested):
Wondering what the purpose of the extra changes not specified in the ticket are? Not sure if there was extra context I don't have.

gbochora:
Extra context from the discussion with @lieut-data  is that we agreed it would be better if some of the runs(roughly 25%) had disabled retros and status updates for diversity. 

Regarding other changes, I assumed generating unique random playbooks or choosing a random playbook for each run would give us more diverse test data.

crspeller:
>  that we agreed it would be better if some of the runs(roughly 25%) had disabled retros and status updates for diversity.

@gbochora Looks like this PR doesn't do that? Maybe I am missing it, but it looks like they all now have it enabled.

gbochora:
> @gbochora Looks like this PR doesn't do that? Maybe I am missing it, but it looks like they all now have it enabled.
In the dummy playbooks list, one item(`Blank Playbook`) out of five has disabled retros and status updates. So when a run is created, there is a 20% chance to select it. Because exactly 25% was not required, I assumed it was an acceptable approach for test data generation.

agarciamontoro (approved):
Oof, this code was a bit messy and complex, thanks for tackling these improvements! Approving now, as the code is great, but I do have one question about the PR description:

> After: Added playbook creator user to Members and playbookService.GetPlaybooksForTeam works properly.

I don't fully understand this. The call to `GetPlaybooksForTeam` in `generateTestData` grabs the playbooks the user has already defined. How is that behavior modified by adding the playbook creator user to the dummy playbooks? I suspect I misunderstood something here.

gbochora:
>I don't fully understand this. The call to GetPlaybooksForTeam in generateTestData grabs the playbooks the user has already defined. How is that behavior modified by adding the playbook creator user to the dummy playbooks? I suspect I misunderstood something here.

Generally, if we create a playbook from the web interface, 'api.playbook.createPlaybook' will be called. If members are not specified, it sets the creator user as sole admin and then calls `app.playbookService.Create` method.
In the case of the `command.generateTestData` method, we directly call `app.playbookService.Create` to create a playbook, which doesn't check/set members.

agarciamontoro:
> In the case of the command.generateTestData method, we directly call app.playbookService.Create to create a playbook, which doesn't check/set members.

That's exactly what I had missed. Thanks for the thorough explanation, @gbochora :)

PR #968, opened 2022-01-07
Client v0.7.0

cpoile:
#### Summary
- also updated godoc: https://pkg.go.dev/github.com/mattermost/mattermost-plugin-playbooks/client@v0.7.0

#### Ticket Link
- none

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Gated by experimental feature flag~~
- [ ] ~~Unit tests updated~~

PR #969, opened 2022-01-07
Server dependency updates

crspeller:
#### Summary
Update dependencies!


go.mod
```diff
@@ -1,28 +1,31 @@
...
+	github.com/Masterminds/squirrel v1.5.2
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/golang/mock v1.6.0
 	github.com/gorilla/mux v1.8.0
 	github.com/jmoiron/sqlx v1.3.4
 	github.com/lib/pq v1.10.4
 	github.com/mattermost/mattermost-plugin-api v0.0.22-0.20211207232216-3faec618d311
-	github.com/mattermost/mattermost-plugin-playbooks/client v0.6.0
+	github.com/mattermost/mattermost-plugin-playbooks/client v0.0.0-00010101000000-000000000000
```
cpoile, line 24: this is updated in https://github.com/mattermost/mattermost-plugin-playbooks/pull/968 :P

go.mod
```diff
@@ -1,28 +1,31 @@
...
-go 1.14
+go 1.16
+
+replace github.com/mattermost/mattermost-plugin-playbooks/client => ./client
+
+replace github.com/HdrHistogram/hdrhistogram-go => github.com/codahale/hdrhistogram v1.1.2
+
+replace github.com/golang/mock => github.com/golang/mock v1.4.4
+
+replace github.com/mattermost/mattermost-server/v6 => github.com/mattermost/mattermost-server/v6 v6.0.0-20220107143254-93a9c6c2931f
```
cpoile, line 11: hmm, why do we need this separate from below?

cpoile (changes requested):

go.mod
crspeller: Why? The replace line make this line meaningless. Might as well be the 000000 thing right?

go.mod
crspeller: Makes updates easier by locking the server version since we need this.

lieut-data: Is there an opportunity to document this somehow? At first glance, it feels like a "broken" usage of the replace directive for something that's supposed to be in the `require` block. Is it an oddity related to how we version the server?

go.mod
```diff
@@ -1,28 +1,31 @@
 module github.com/mattermost/mattermost-plugin-playbooks
 
-go 1.14
+go 1.16
+
+replace github.com/mattermost/mattermost-plugin-playbooks/client => ./client
```
lieut-data, line 5: nit: Why before the `require` instead of after? (Seems slightly weird to start reading replace directives when I haven't even encountered what they replace.)

lieut-data (approved):

go.mod
crspeller: It's what `go mod tidy` did. 🤷

go.mod
crspeller: It's an oddity because there is no released server version. If you try to install anything else it will redo this and grab the most recent server version (which would actually be a downgrade). This just helpfully pins the version so we don't have to worry about it downgrading randomly.

go.mod
crspeller: I actually like it because the require stuff is mostly meaningless. The replace stuff is more important.

go.mod
cpoile: imo: because you can comment the replace directive out and run the tests to make sure the actual published module is correct

go.mod
cpoile: I'm still not sure I understand -- why would the hashed version dependency (that we've always used) be downgraded?

go.mod
crspeller: @cpoile But the replace directive is part of the module? There isn't some other version of the module as far as I am aware?

go.mod
crspeller: If you do a `go get` and add another module without the replace directive it's going to do it's minimal version selection and likely try to downgrade mattermost-server. With the replace we force it to stay there until we explicitly change it.

go.mod
cpoile: I'm not sure what you mean by "the replace directive is part of the module", sorry. There is version v0.7.0 which we just published... Unless you're talking about something else?

go.mod
cpoile: ok, if it works it works, but we haven't had a problem with it before, have we?

cpoile (approved):

go.mod
crspeller: Synced offline and figured out what each other where saying. This line is totally ignored by go mod when the replace is active. But you can test mismatched versions when you comment out a replace and fill in a version here.

PR #970, opened 2022-01-07
Client dependency updates

crspeller:
#### Summary
Update dependencies!


webapp/package.json
```diff
@@ -21,94 +21,95 @@
...
+    "@types/react-beautiful-dnd": "13.1.2",
     "@types/react-bootstrap": "1.0.1",
-    "@types/react-custom-scrollbars": "4.0.8",
-    "@types/react-dom": "16.9.8",
-    "@types/react-redux": "7.1.18",
-    "@types/react-router-dom": "5.1.8",
-    "@types/react-select": "^3.1.2",
-    "@types/react-test-renderer": "16.9.3",
+    "@types/react-custom-scrollbars": "4.0.10",
+    "@types/react-dom": "17.0.11",
```
cpoile, line 44: Only one q: why is this being updated but the dep isn't?

cpoile (approved):

webapp/package.json
crspeller: Because I didn't notice and it somehow doesn't break. Will make them match though so nobody steps on a mine.

webapp/package.json
```diff
@@ -21,94 +21,95 @@
...
+    "@types/debounce": "1.2.1",
     "@types/history": "4.7.8",
-    "@types/jest": "26.0.24",
-    "@types/luxon": "2.0.2",
+    "@types/jest": "27.4.0",
+    "@types/luxon": "2.0.8",
     "@types/qs": "6.9.7",
-    "@types/react": "16.9.35",
-    "@types/react-beautiful-dnd": "13.1.1",
+    "@types/react": "17.0.38",
```
calebroseland: Because we're still on `react@16.x.x`, does this need to stay  on the `16.x` track?

calebroseland (approved):

webapp/package.json
```diff
@@ -21,94 +21,95 @@
...
-    "react-redux": "7.2.4",
+    "react-intl": "5.24.1",
+    "react-redux": "7.2.6",
     "react-select": "3.2.0",
-    "redux": "4.1.0",
-    "styled-components": "5.3.0",
-    "typescript": "4.3.5"
+    "redux": "4.1.2",
+    "styled-components": "5.3.3",
+    "typescript": "4.3.4"
```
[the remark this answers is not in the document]
crspeller: 198 Most of them being unfixable from redux/webapp. Hence not upgrading until the webapp does.

PR #973, opened 2022-01-10
e2e: fix config plumbing for EnableOnboardingFlow

Phrynobatrachus:
`getDefaultConfig()` was trying to fetch `config.enableOnboardingFlow` from `tests-e2e/cypress.json` but that key is actually present under `config.env.enableOnboardingFlow`. This should prevent the view from appearing and failing tests both in CI (which @cpoile already fixed, so should be double good now) and locally.


lieut-data (approved):
"Deep magic." Is this broken in the webapp version of this code as well?

Phrynobatrachus:
Nope. I changed this to match what `webapp` does. Still room for improvement with the support package hopefully.

PR #974, opened 2022-01-12
MM-40944: Lift playbook limits

agarciamontoro:
#### Summary
This PR removes the restriction on the number of playbooks that applied to Team Edition (TE) and Starter Edition (old E0) servers. Before this, these servers could create only one playbook per team. After this, all servers can create unlimited playbooks :tada: And the key icons in the backstage are gone:

| Before 	| After 	|
|---	|---	|
| image 	| image 	|

I expected this to be a 20-30 lines removed PR, but we had a lot of code to handle this particular case, as we can create playbooks in the backstage, both with templates and with the button, and in the RHS. We also had some action/reducer/selector boilerplate to track the number of playbooks in each team, as well as hooks for checking it. All of that is now gone.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40944

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Gated by experimental feature flag
- [X] Unit tests updated


agarciamontoro:
Whoops, @iomodo is on PTO. Sorry about that.

cpoile (approved):
💙

PR #975, opened 2022-01-12
Allow creating runs in an existing channel via API

lieut-data:
#### Summary
Fundamentally, we don't require a /new/ channel for a run, but currently tie the two steps together. Let's start decoupling that by allowing -- at an API level -- the ability to specify an existing channel into which to create a playbook run.

Note that I'm using Cypress E2E tests to exercise this API vs. unit tests.

#### Ticket Link
None, but the backstory here is a desire to apply a playbook retroactively to an existing channel on community.

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -154,6 +154,7 @@ func (h *PlaybookRunHandler) createPlaybookRunFromPost(w http.ResponseWriter, r 
 		app.PlaybookRun{
 			OwnerUserID: playbookRunCreateOptions.OwnerUserID,
 			TeamID:      playbookRunCreateOptions.TeamID,
+			ChannelID:   playbookRunCreateOptions.ChannelID,
 			Name:        playbookRunCreateOptions.Name,
 			Summary:     playbookRunCreateOptions.Description,
 			PostID:      playbookRunCreateOptions.PostID,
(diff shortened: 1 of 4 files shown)


server/api/playbook_runs.go
```diff
@@ -365,23 +366,34 @@ func (h *PlaybookRunHandler) createPlaybookRun(playbookRun app.PlaybookRun, user
...
+		}
+
+		if playbookRun.TeamID == "" {
+			playbookRun.TeamID = channel.TeamId
+		} else if channel.TeamId != playbookRun.TeamID {
+			return nil, errors.Wrap(app.ErrMalformedPlaybookRun, "playbook run channel not in given team")
+		}
+	} else if playbookRun.TeamID == "" {
 		return nil, errors.Wrap(app.ErrMalformedPlaybookRun, "missing team id of playbook run")
 	}
```
gbochora, line 390: It seems this code is unreachable. If the program reaches `if playbookRun.TeamID == ""` line, that means: 
- `playbookRun.TeamID` and `playbookRun.ChannelID` both are not empty
- `playbookRun.ChannelID` is empty

So `playbookRun.TeamID` must not be empty.

server/app/playbook_run_service.go
```diff
@@ -219,26 +219,46 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
...
+		} else if existingPlaybookRunID != "" {
+			return nil, errors.Wrapf(ErrMalformedPlaybookRun, "playbook run %s already exists for channel %s", existingPlaybookRunID, playbookRun.ChannelID)
+		}
+
+		channel, err = s.pluginAPI.Channel.Get(playbookRun.ChannelID)
+		if err != nil {
+			return nil, err
+		}
+
+		playbookRun.Name = channel.Name
```
gbochora, line 258: When creating a playbook run in an existing channel, do we want its name always to be `channel.Name`? Even when `playbookRun.Name` parameter is passed from the caller?

gbochora (approved):
Just a few non-blocking comments. I liked a lot idea of using Cypress to test API : ))

server/api/playbook_runs.go
```diff
@@ -454,14 +466,30 @@ func (h *PlaybookRunHandler) createPlaybookRun(playbookRun app.PlaybookRun, user
...
+		permissionMessage := "You are not able to create a private channel"
+		if public {
+			permission = model.PermissionCreatePublicChannel
+			permissionMessage = "You are not able to create a public channel"
+		}
+		if !h.pluginAPI.User.HasPermissionToTeam(userID, playbookRun.TeamID, permission) {
+			return nil, errors.Wrap(app.ErrNoPermissions, permissionMessage)
+		}
+	} else {
+		permission := model.PermissionManagePublicChannelProperties
```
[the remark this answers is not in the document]
lieut-data: Yeah, I was thinking channel admins so as to avoid spamming other folks channels.

Phrynobatrachus (approved):
tried to clean up the tests a little bit more (in linked PR), otherwise LGTM

server/app/playbook_run_service.go
lieut-data: Whoops, forgot to reply to this, but yes: I think we want to force the name to match the channel vs. rename the channel. The two are supposed to always be in sync.

PR #976, opened 2022-01-13
test revisions

Phrynobatrachus:
changes to #975 to shave off some extra boilerplate. 

- `apiRunPlaybook` now accepts a status code and `failOnStatus` so it can handle both fail/success instead of just `201`

### DIFF
--- a/tests-e2e/cypress/support/plugin/api_commands.js
+++ b/tests-e2e/cypress/support/plugin/api_commands.js
@@ -4,6 +4,7 @@
 const playbookRunsEndpoint = '/plugins/playbooks/api/v0/runs';
 
 const StatusOK = 200;
+const StatusCreated = 201;
 
 /**
  * Get all playbook runs directly via API
@@ -75,8 +76,13 @@ Cypress.Commands.add('apiRunPlaybook', (
         playbookId,
         playbookRunName,
         ownerUserId,
+        channelId,
         description = ''
-    }) => {
+    },
+    {
+        expectedStatusCode = StatusCreated,
+        failOnStatusCode = true,
+    } = {}) => {
     return cy.request({
         headers: {'X-Requested-With': 'XMLHttpRequest'},
         url: playbookRunsEndpoint,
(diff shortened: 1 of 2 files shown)


tests-e2e/cypress/support/plugin/api_commands.js
```diff
@@ -75,8 +76,13 @@ Cypress.Commands.add('apiRunPlaybook', (
         playbookId,
         playbookRunName,
         ownerUserId,
+        channelId,
         description = ''
```
lieut-data, line 80: Should we remove the default in this case so that the API is "pure" (vs. having implied semantics that we may not test as thoroughly?)

tests-e2e/cypress/support/plugin/api_commands.js
```diff
@@ -86,10 +92,12 @@ Cypress.Commands.add('apiRunPlaybook', (
             owner_user_id: ownerUserId,
             team_id: teamId,
             playbook_id: playbookId,
+            channel_id: channelId,
             description,
         },
+        failOnStatusCode,
```
lieut-data, line 98: Would we consider inferring `failOnStatusCode` from `expectedStatusCode` (unless it's set explicitly)?

lieut-data (approved):
Sweet, thank you! I had imagined our `api*` functions were "higher level", but indeed they look like a much better way to abstract things.

tests-e2e/cypress/support/plugin/api_commands.js
Phrynobatrachus: for `description`? I'll check without it.

tests-e2e/cypress/support/plugin/api_commands.js
Phrynobatrachus: the `failOnStatusCode = true` default reduces to the same behavior doesn't it?

`{expectedStatusCode: 400}` will fail
`{expectedStatusCode: 400, failOnStatusCode: false}` won't

Phrynobatrachus: oh do you mean like if `expectedStatusCode` is passed assume `failOnStatusCode: false`? since we probably want to assert on the result even if it's a failure. let me think about how to write that.

PR #977, opened 2022-01-14
Stop notifying others' overdue runs

lieut-data:
#### Summary
Following up on [this thread](https://community-daily.mattermost.com/core/pl/gza8jybbrtn5ifeqq9r5bbmwhr), let's stop notifying participants about runs with overdue status updates of which they are not the owner. We're not finding this to be an actionable event at this time, and want to "stop the spam" while we revisit the overall approach.

#### Ticket Link
N/A

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -2676,8 +2676,8 @@ func buildRunsOverdueMessage(runs []RunLink, locale string) string {
 	msg += T("app.user.digest.overdue_status_updates.num_overdue", total) + "\n"
 
 	for _, run := range runs {
-		msg += fmt.Sprintf("- [%s](/%s/channels/%s?telem_action=todo_overduestatus_clicked&telem_run_id=%s&forceRHSOpen)",
-			run.ChannelDisplayName, run.TeamName, run.ChannelName, run.PlaybookRunID) + "\n"
+		msg += fmt.Sprintf("- [%s](/%s/channels/%s?telem_action=todo_overduestatus_clicked&telem_run_id=%s&forceRHSOpen)\n",
+			run.ChannelDisplayName, run.TeamName, run.ChannelName, run.PlaybookRunID)
 	}
 
 	return msg
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -2676,12 +2676,8 @@ func buildRunsOverdueMessage(runs []RunLink, locale string) string {
 	msg += T("app.user.digest.overdue_status_updates.num_overdue", total) + "\n"
 
 	for _, run := range runs {
-		values := map[string]interface{}{
-			"Username": run.OwnerUserName,
-		}
-		appended := " " + T("app.user.digest.overdue_status_updates.md_link_item_appended", values)
 		msg += fmt.Sprintf("- [%s](/%s/channels/%s?telem_action=todo_overduestatus_clicked&telem_run_id=%s&forceRHSOpen)",
-			run.ChannelDisplayName, run.TeamName, run.ChannelName, run.PlaybookRunID) + appended + "\n"
+			run.ChannelDisplayName, run.TeamName, run.ChannelName, run.PlaybookRunID) + "\n"
 	}
 
 	return msg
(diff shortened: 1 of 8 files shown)


tests-e2e/cypress/support/api/user.js
```diff
@@ -246,6 +246,7 @@ Cypress.Commands.add('apiCreateUser', ({
 
         if (hideOnboarding) {
             cy.apiSaveOnboardingPreference(createdUser.id, 'hide', 'true');
+            cy.apiSaveOnboardingPreference(createdUser.id, 'skip', 'true');
```
lieut-data, line 249: @Phrynobatrachus, I recall your investigating something to this effect. I kept having this problem locally, and found that the webapp now sends `skip` instead of `hide` as the preference name. 🤷🏼

server/app/playbook_run_service.go
```diff
@@ -2676,12 +2676,8 @@ func buildRunsOverdueMessage(runs []RunLink, locale string) string {
 	msg += T("app.user.digest.overdue_status_updates.num_overdue", total) + "\n"
 
 	for _, run := range runs {
-		values := map[string]interface{}{
-			"Username": run.OwnerUserName,
-		}
-		appended := " " + T("app.user.digest.overdue_status_updates.md_link_item_appended", values)
 		msg += fmt.Sprintf("- [%s](/%s/channels/%s?telem_action=todo_overduestatus_clicked&telem_run_id=%s&forceRHSOpen)",
```
cpoile, line 2679:
```suggestion
		msg += fmt.Sprintf("- [%s](/%s/channels/%s?telem_action=todo_overduestatus_clicked&telem_run_id=%s&forceRHSOpen)\n",
```

server/app/playbook_run_service.go
```diff
@@ -2676,12 +2676,8 @@ func buildRunsOverdueMessage(runs []RunLink, locale string) string {
...
-			run.ChannelDisplayName, run.TeamName, run.ChannelName, run.PlaybookRunID) + appended + "\n"
+			run.ChannelDisplayName, run.TeamName, run.ChannelName, run.PlaybookRunID) + "\n"
```
cpoile, line 2680:
```suggestion
			run.ChannelDisplayName, run.TeamName, run.ChannelName, run.PlaybookRunID)
```

server/sqlstore/playbook_run.go
```diff
@@ -1047,26 +1047,18 @@ func (s *playbookRunStore) GetParticipatingRuns(userID string) ([]app.RunLink, e
...
-		Suffix(")")
-
 	query := s.store.builder.
 		Select("i.ID AS PlaybookRunID", "t.Name AS TeamName",
 			"c.Name AS ChannelName", "c.DisplayName AS ChannelDisplayName", "u.UserName AS OwnerUserName").
 		From("IR_Incident AS i").
 		Join("Teams AS t ON (i.TeamID = t.Id)").
 		Join("Channels AS c ON (i.ChannelId = c.Id)").
-		LeftJoin("Users AS u ON i.CommanderUserID = u.Id").
+		Join("Users AS u ON i.CommanderUserID = u.Id").
```
cpoile, line 1058: I think we might be able to remove this join, and remove the `u.UserName AS OwnerUserName` above, and then remove the `OwnerUserName` field from `RunLink`

cpoile (approved):
Awesome!

crspeller (approved):
Yay! No more annoying playbooks bot messages!

lieut-data:
@Phrynobatrachus, I'm going to merge this one to get the changes live on `community` for tomorrow, but I'm definitely open to any E2E feedback you might have and can circle back to address retroactively :)

tests-e2e/cypress/support/api/user.js
Phrynobatrachus: Do you know why the server configuration wouldn't already prevent a new user from seeing it (even without the explicit preference being set for them here)? Doesn't quite make sense to me.

Phrynobatrachus (commented):
:+1: on the tests other than the onboarding question

tests-e2e/cypress/support/api/user.js
lieut-data: I'm wondering if there's something that prevents the e2e tests from clobbering a developer's config environment? I know we had an issue in the past where e2e would stomp all over the presumably custom config, but I'm not 100% how we resolved that.

Phrynobatrachus: there is `CYPRESS_developerMode: true`. here's what happens without that env var:

server default: `EnableOnboardingFlow: true` 
on_prem_default: `EnabledOnboardingFlow: false` -> overwrites server config since different value for same key

I'm not sure that's the issue here.

PR #978, opened 2022-01-14
MM-40840 Server side of export functionality.

crspeller:
#### Summary
Server side of Export functionality. 

Decided to go with this approach for ease of use and to allow for checks to verify that export is being considered when fields are being added.

For an explanation of use see the comment above the Playbooks struct. (Improved wording of that comment welcomed!) 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40840

### DIFF
--- a/server/app/export.go
+++ b/server/app/export.go
@@ -1,0 +1,61 @@
+package app
+
+import (
+	"encoding/json"
+	"reflect"
+)
+
+const CurrentPlaybookExportVersion = 1
+
+func getFieldsForExport(in interface{}) map[string]interface{} {
+	out := map[string]interface{}{}
+
+	inType := reflect.TypeOf(in)
+	inValue := reflect.ValueOf(in)
+	for i := 0; i < inType.NumField(); i++ {
+		field := inType.Field(i)
+		tag := field.Tag.Get("export")
+		fieldValue := inValue.Field(i)
+		if tag != "" && tag != "-" && !fieldValue.IsZero() {
+			out[tag] = fieldValue.Interface()
+		}
+	}
+
+	return out
+}
+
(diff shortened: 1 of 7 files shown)


lieut-data:
Swapping myself out for @iomodo.

webapp/package-lock.json
```diff
@@ -10,7 +10,7 @@
         "chrono-node": "2.3.5",
         "core-js": "3.20.2",
         "debounce": "1.2.1",
-        "luxon": "2.3.0",
+        "luxon": "2.0.2",
```
iomodo, line 13: why do we have this change?

server/app/export.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return exported
+}
+
+// GeneratePlaybookExport returns a playbook in export format.
+// Fields marked with the stuct tag "export" are included using the given string.
+func GeneratePlaybookExport(playbook Playbook) ([]byte, error) {
+	export := getFieldsForExport(playbook)
+	export["version"] = CurrentPlaybookExportVersion
+	export["checklists"] = generateChecklistExport(playbook.Checklists)
```
iomodo, line 53: why not do the recursive implementation, it would be resilient to all the struct changes in the future.

iomodo (approved):
Looks great @crspeller 
Just couple of questions below

webapp/package-lock.json
crspeller: That's the package-lock. I am guessing when that other PR went in it wasn't updated so this PR is doing it.

server/app/export.go
crspeller: We could possibly detect slice type and do that. But I think I would rather force some thought here. If we create some super complicated structure we can change it.

calebroseland (approved):

PR #991, opened 2022-01-22
Mm 40931 playbook edit update playbook metrics backend

gbochora:
#### Summary
This PR adds backend functionality for storing / updating / retrieving Playbook metrics configurations.

**Change Log:**
1) Add two tables([SQL tables spec](https://docs.google.com/document/d/11wzPKfsHesAzEOX_7ljEe7893xywv-239P0J-VJWoqU/edit?usp=sharing)):
-IR_MetricConfig - to store playbooks key metrics configurations
-IR_Metric - to store runs metrics values

2) Introduced PlaybookMetric struct and Add field to playbook struct

3) Add functions to `sqlstore.playbook` for storing / updating / retrieving metrics configs.
4) Add tests for `sqlstore.playbook`. Cleaned metrics IDs in the playbook object to compare the original playbook with one from the database.

5) Add metrics configs validation on API layer:
-Validate if numbers of metrics are not more than 4
-Validate that metrics titles are unique

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40931
Epic Link: https://mattermost.atlassian.net/browse/MM-40926

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated


server/sqlstore/playbook.go
```diff
@@ -688,6 +709,61 @@ func (p *playbookStore) replacePlaybookMembers(q queryExecer, playbook app.Playb
...
+		).
+		From("IR_MetricConfig mc").
+		Where(sq.Eq{"DeleteAt": 0}).
+		OrderBy("mc.Sort ASC")
+}
+
+// replacePlaybookMetrics replaces the metric configs of a playbook
+func (p *playbookStore) replacePlaybookMetrics(q queryExecer, playbook app.Playbook) error {
+	// Mark as deleted all metrics for this playbook
+	updateBuilder := sq.Update("IR_MetricConfig").
```
crspeller, line 730: Nit: Little confusing because we restore them later if they are the same. Maybe add that to the comment?

crspeller (changes requested):

server/sqlstore/playbook.go
```diff
@@ -688,6 +709,61 @@ func (p *playbookStore) replacePlaybookMembers(q queryExecer, playbook app.Playb
...
+	return p.store.builder.
+		Select(
+			"mc.ID",
+			"mc.PlaybookID",
+			"mc.Title",
+			"mc.Description",
+			"mc.Type",
+			"mc.Target",
+		).
+		From("IR_MetricConfig mc").
```
cpoile, line 722: just fyi, if there's no join (and no chance for ambiguity) you don't need to add the `mc.` prefix to the columns here -- I realize it mirrors what we (blame says @crspeller ?) did for the `makeMembersSelect` but I'm not sure why we did that there either. :)
In fact, for both, we can add those selects right to the `playbookStore` struct, like we do for the `playbookSelect` statement, on line 164. That keeps them together so we remember to add new columns to it, like we do with `playbookSelect`

cpoile (changes requested):
Looks great, nice first big PR @gbochora! 
Some minor comments, one question about `sort` vs `order` but otherwise look great.

server/sqlstore/playbook.go
gbochora: Sure, Good idea.

server/sqlstore/migrations.go
```diff
@@ -1708,6 +1708,89 @@ var migrations = []Migration{
...
+			if e.DriverName() == model.DatabaseDriverMysql {
+				if _, err := e.Exec(`
+					CREATE TABLE IF NOT EXISTS IR_MetricConfig (
+					   ID VARCHAR(26) PRIMARY KEY,
+					   PlaybookID VARCHAR(26) NOT NULL REFERENCES IR_Playbook(ID),
+					   Title VARCHAR(512) NOT NULL,
+					   Description VARCHAR(4096) NOT NULL,
+					   Type VARCHAR(32) NOT NULL,
+					   Target BIGINT NOT NULL,
+					   Sort TINYINT NOT NULL DEFAULT 0,
```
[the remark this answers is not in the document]
gbochora: I agree Order is much better, but I avoided using it because it's a reserved keyword. 
If we are ok with using reserved keywords, then I'll change. It just needs to be enclosed in backticks/double quotes.

server/app/playbook.go
```diff
@@ -14,65 +14,83 @@ import (
...
 	SchemeRoles []string `json:"scheme_roles"`
 }
 
+type PlaybookMetricConfig struct {
+	ID          string `json:"id"`
+	PlaybookID  string `json:"playbook_id"`
+	Title       string `json:"title"`
+	Description string `json:"description"`
+	Type        string `json:"type"`
+	Target      int64  `json:"Target"`
```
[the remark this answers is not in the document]
gbochora: Store layer returns metrics list sorted by order value. So I thought there was no need for an additional `order` parameter, Index of each element in the list defines an order. 
Do you think we may still need order value?

server/sqlstore/playbook.go
gbochora: I'll remove a prefix to the column. I didn't notice that there is no need for that.
Excellent point keeping selects together.

server/app/playbook.go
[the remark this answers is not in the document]
cpoile: Nope, you're right -- let's use the array ordering, I'll change the frontend code for that.

server/sqlstore/migrations.go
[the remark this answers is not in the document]
cpoile: Hmm, I just really don't want to use `Sort` -- it's too confusing. Howabout `Ordering`?

cpoile (approved):
One last suggestion (for the export), otherwise ready to merge. Thanks @gbochora!

server/app/playbook.go
cpoile, line 91: This caused a bug in the frontend:

```suggestion
	Target      int64  `json:"target"`
```

gbochora: Oh, I missed that : )) Fixed.

server/app/playbook.go
```diff
@@ -14,53 +14,53 @@ import (
...
+	CategorizeChannelEnabled             bool                   `json:"categorize_channel_enabled" export:"categorize_channel_enabled"`
+	CategoryName                         string                 `json:"category_name" export:"category_name"`
+	RunSummaryTemplateEnabled            bool                   `json:"run_summary_template_enabled" export:"run_summary_template_enabled"`
+	RunSummaryTemplate                   string                 `json:"run_summary_template" export:"run_summary_template"`
+	ChannelNameTemplate                  string                 `json:"channel_name_template" export:"channel_name_template"`
+	DefaultPlaybookAdminRole             string                 `json:"default_playbook_admin_role" export:"-"`
+	DefaultPlaybookMemberRole            string                 `json:"default_playbook_member_role" export:"-"`
+	DefaultRunAdminRole                  string                 `json:"default_run_admin_role" export:"-"`
+	DefaultRunMemberRole                 string                 `json:"default_run_member_role" export:"-"`
+	Metrics                              []PlaybookMetricConfig `json:"metrics" export:"-"`
```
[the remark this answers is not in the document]
gbochora: Fixed, Thanks.

server/app/playbook.go
```diff
@@ -14,65 +14,83 @@ import (
...
+	CategorizeChannelEnabled             bool                   `json:"categorize_channel_enabled" export:"categorize_channel_enabled"`
+	CategoryName                         string                 `json:"category_name" export:"category_name"`
+	RunSummaryTemplateEnabled            bool                   `json:"run_summary_template_enabled" export:"run_summary_template_enabled"`
+	RunSummaryTemplate                   string                 `json:"run_summary_template" export:"run_summary_template"`
+	ChannelNameTemplate                  string                 `json:"channel_name_template" export:"channel_name_template"`
+	DefaultPlaybookAdminRole             string                 `json:"default_playbook_admin_role" export:"-"`
+	DefaultPlaybookMemberRole            string                 `json:"default_playbook_member_role" export:"-"`
+	DefaultRunAdminRole                  string                 `json:"default_run_admin_role" export:"-"`
+	DefaultRunMemberRole                 string                 `json:"default_run_member_role" export:"-"`
+	Metrics                              []PlaybookMetricConfig `json:"metrics" export:"metrics"`
```
crspeller, line 63: One more step for the export here since this is a struct. You need to add the `export` struct tag to the `PlaybookMetricConfig` struct and also add it to the GeneratePlaybookExport function here: https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/app/export.go#L53

Sorry this is so complicated. I thought that since there was only one exported struct in all this time they would be rare. Seems like I was wrong 😆

crspeller (changes requested):
Just one thing for export.

crspeller (approved):

PR #993, opened 2022-01-24
Export and Import UI

crspeller:
#### Summary
UI for import and export of playbooks

Export Link (downloads on click):
image
image

Import button (instant select file if you only have one team):
image

Import team selection:
image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40840
https://mattermost.atlassian.net/browse/MM-40842


crspeller:
@Phrynobatrachus These jest test failures don't seem related to what I edited here. Have you seen these before? Could you take a look?

Phrynobatrachus:
I haven't seen this before. `ChecklistItemState` is an enum so it should always be "initialized" right? Some quick searching suggests that it might be a circular dependency issue, I'll keep looking.

webapp/src/types/playbook.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {getAPIUrl} from 'src/client';
```
Phrynobatrachus, line 4: I think this is the reason for the Jest failures. 

`src/client` imports `src/types/playbook`

crspeller: Thanks! Weird that it only complains in the tests.

Phrynobatrachus: Yeah it is a bit odd, there's apparently a decent amount more (23 according to [this Webpack plugin](https://github.com/aackerman/circular-dependency-plugin#basic-usage) I tested with real quick) that don't seem to cause issues.

webapp/src/components/backstage/playbook_list.tsx
```diff
@@ -123,6 +151,41 @@ const PlaybookList = () => {
...
+                        { teams.length <= 1 &&
+                        <ImportButton
+                            onClick={() => {
+                                if (fileInputRef && fileInputRef.current) {
+                                    fileInputRef.current.click();
+                                }
+                            }}
+                            spin={importUploading}
+                        />
+                        }
```
iomodo, line 188:
```js
teams.length > 1 ? <TeamSelector...   :  <ImportButton..
```
might be easier to digest.

webapp/src/client.ts
```diff
@@ -667,3 +676,9 @@ export const doFetchWithoutResponse = async (url: string, options = {}) => {
         url,
     });
 };
+
+export const playbookExportProps = (playbook: Playbook) => {
+    const href = `${getAPIUrl()}/playbooks/${playbook.id}/export`;
```
iomodo, line 681: shouldn't we be consistent using `getAPIUrl()` vs `apiUrl`? I'd remove `getAPIUrl()` altogether for consistency

iomodo (approved):
Looks great! Just minor nits you can ignore

webapp/src/client.ts
crspeller: Yup. This happened when I fixed the curricular dependency.

webapp/src/components/backstage/playbook_list.tsx
crspeller: Personally I think the ternary is more confusing.

webapp/src/components/backstage/playbook_list.tsx
```diff
@@ -125,6 +153,41 @@ const PlaybookList = () => {
                             onSearch={setSearchTerm}
                             placeholder={formatMessage({defaultMessage: 'Search for a playbook'})}
                         />
+                        <HorizontalSpacer size={12}/>
+                        <input
+                            type='file'
```
calebroseland, line 158:
```suggestion
                            type='file'
                            accept='*.json,application/JSON'
```

calebroseland (changes requested):
Nice to see the whole flow! Works well, though I have a request or two:
- `accept` attr on the input to make file format more obvious
- after successful import: either refresh playbooks list (mirroring the `duplicate` flow), or preferably—navigate to the imported playbook to see/confirm details (same as what happens after `create -> save`)


Some sort of success message would also be nice, but along with `duplicate`, this would perhaps be better as a follow-up ticket.

webapp/src/client.ts
```diff
@@ -231,6 +231,11 @@ export async function archivePlaybook(playbookId: Playbook['id']) {
     return data;
 }
 
+export async function importFile(file: any, teamId: string) {
+    const data = await doPost(`${apiUrl}/playbooks/import?team_id=${teamId}`, file);
+    return data;
+}
```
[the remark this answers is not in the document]
crspeller: Maybe, but I don't want to break the current pattern in that file. Maybe we can do that as a holistic change?

crspeller:
@calebroseland Nice suggestions! I did the redirect to the view page of the new playbook. Feels nicer!

calebroseland (approved):
Thanks, @crspeller!

abhijit-singh (approved):
Awesome stuff @crspeller!

A couple of tiny non-blocking nits:
1. Might be valuable to add `playbook` to the name of the exported file? Maybe something like `customer_onboarding_playbook.json` instead of adding `export`. While looking for the file later I'm personally much more likely to search by the word `playbook` than `export` for the playbook I downloaded. Also might help build some brand recognition if that file is shared.
2. Can the team selector dropdown be positioned just below `import` the button instead of overlapping on it?
image
Thanks!

PR #998, opened 2022-01-25
MM-40446: Adopt a mixed license

agarciamontoro:
#### Summary

Adopt a mixed license model, where all the code is licensed under Apache 2.0, except for the `server/enterprise` directory, which is still licensed under the Source Available license. That new directory contains a single file, `license.go`, that defines the checks for the enterprise-only features.

A couple of technical considerations:
- Instead of defining the `LicenseChecker` interface in the new `enterprise` package, I defined it in the `app` package, which acts as the client in this case.
- In order not to create an import loop, I don't accept an `app.Playbook` in the `PlaybookAllowed` function, only the bits of information we need to decide if it's allowed or not. For now, those bits are the one an only `is this playbook public?`. 

And a legal consideration:
- The license changes are based on [this document](https://developers.mattermost.com/integrate/plugins/source-available-license/#how-do-i-apply-the-license-to-a-mixed-license-plugin) and inspired on how [the Jira plugin](https://github.com/mattermost/mattermost-plugin-jira) implemented the mixed license model.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40446

#### Checklist

- [ ] ~~Telemetry updated~~
- [X] Unit tests updated


agarciamontoro:
@jamiehurewitz, this is the PR I mentioned in the Legal channel in Mattermost. I'd love to have your opinion on this :)

jamiehurewitz (approved):

jamiehurewitz:
Looks good!

server/enterprise/license.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// PlaybookAllowed returns true if the specified playbook is valid with the current license.
+func (e *LicenseChecker) PlaybookAllowed(isPlaybookPublic bool) bool {
+	// Private playbooks are E20-only
+	return e.isAtLeastE20Licensed() || isPlaybookPublic
+}
+
+// RetrospectiveAllowed returns true if the retrospective feature is allowed with the current license.
+func (e *LicenseChecker) RetrospectiveAllowed() bool {
```
lieut-data, line 40: For completeness, should we add some new server-side checks for the timeline functionality as well? (Presumably this was just enforced client-side before.)

lieut-data (approved):
Thanks, @agarciamontoro!

cpoile (approved):

agarciamontoro:
@lieut-data, I added checks for the timeline, and found out that we can also add them for the stats. Can you take a second look? Thanks!

agarciamontoro:
I don't know what shortcut I pressed, but I somehow closed the PR. Sorry for the noise, I did not mean to :sweat_smile:

itao:
I can't find an option to approve, might be because it was closed and re-opened, so approved!

lieut-data (approved):
Thanks again, @agarciamontoro :)

PR #1000, opened 2022-01-25
MM-41212 Fix run list filters not persisting.

crspeller:
#### Summary
Fixes run list filters so they update properly. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41212

### DIFF
--- a/webapp/src/hooks/general.ts
+++ b/webapp/src/hooks/general.ts
@@ -442,7 +442,7 @@ export function useNow(refreshIntervalMillis = 1000) {
 }
 
 const combineQueryParameters = (oldParams: FetchPlaybookRunsParams, searchString: string) => {
-    const queryParams = qs.parse(searchString, {ignoreQueryPrefix: true, parseArrays: true});
+    const queryParams = qs.parse(searchString, {ignoreQueryPrefix: true});
     return {...oldParams, ...queryParams};
 };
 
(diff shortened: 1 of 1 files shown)


webapp/src/hooks/general.ts
```diff
@@ -487,7 +491,7 @@ export function useRunsList(defaultFetchParams: FetchPlaybookRunsParams):
         const newFetchParams: Record<string, unknown> = {...fetchParams};
         delete newFetchParams.page;
         delete newFetchParams.per_page;
-        history.replace({search: qs.stringify(newFetchParams, {addQueryPrefix: false, indices: false})});
+        history.replace({search: qs.stringify(newFetchParams, {addQueryPrefix: false, arrayFormat: 'brackets'})});
```
agarciamontoro, line 494: Do we want the indices in the query string?

webapp/src/hooks/general.ts
```diff
@@ -441,19 +441,23 @@ export function useNow(refreshIntervalMillis = 1000) {
     return now;
 }
 
+const combineQueryParameters = (oldParams: FetchPlaybookRunsParams, searchString: string) => {
+    const queryParams = qs.parse(searchString, {ignoreQueryPrefix: true, parseArrays: true});
```
agarciamontoro, line 445: I think `parseArrays` default to `true`, right? I'm not opposed to explicitly specifying default values, though, I think it's a good practice.

agarciamontoro (approved):
Looks good :tada: I left a couple of non-blocking questions below, nothing important.

gbochora (approved):
Looks nice! I've learned about react hooks thanks to this PR : )

webapp/src/hooks/general.ts
crspeller: I didn't think they where necessary. ArrayFormat overrides indices so the result here is no indices but it will also correctly parse as an array.

webapp/src/hooks/general.ts
crspeller: It does!

PR #1002, opened 2022-01-25
MM-40928 - Add metrics (frontend)

cpoile:
#### Summary
- Add key metrics
- Validate when pressing the "Add" button, and transition to the view component for the metric
- When pressing the "Add Metric" or "edit pencil" outside the editing component, remember the requested command, ask for the validation, if validation works then proceed with the request (scrap the request if the validation failed)
- the e2e tests are a good example of everything that's supported (the video below doesn't show all the combinations)

Example 

https://user-images.githubusercontent.com/1490756/151065037-57864cd2-0b1e-41f0-b0ff-1e5fd5859e19.mp4

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40928

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/backstage/playbook_edit/metrics/metric_edit.tsx
```diff
@@ -0,0 +1,241 @@
...
+    const [curState, setCurState] = useState(metric);
+    const [curTargetString, setCurTargetString] = useState(targetToString(metric.target, metric.type));
+    const [curSaveToggle, setCurSaveToggle] = useState(saveToggle);
+    const [titleError, setTitleError] = useState('');
+    const [targetError, setTargetError] = useState('');
+
+    const errorTitleDuplicate = formatMessage({defaultMessage: 'A metric with the same name already exists. Please add a unique name for each metric.'});
+    const errorTitleMissing = formatMessage({defaultMessage: 'Please add a title for your metric.'});
+    const errorTargetCurrencyInteger = formatMessage({defaultMessage: 'Please enter a number, or leave the target blank.'});
+    const errorTargetDuration = formatMessage({defaultMessage: 'Please enter a duration in the format: dd:mm:ss (e.g., 12:00:00), or leave the target blank.'});
```
gbochora, line 34: It seems there is a typo in the message, should be dd:hh:mm instead of dd:mm:ss

cpoile: Thanks, fixed!

gbochora (approved):
Looks nice! Just two minor issues:
- Retrospective disable doesn't disable metrics UI
- If I press on Add Metric button, then navigate to the `Status Updates` tab and then back to the `Retrospective` metric with an empty name is added to the list.

webapp/src/components/backstage/playbook_edit/metrics/metric_edit.tsx
```diff
@@ -0,0 +1,241 @@
...
+    onAdd: (metric: Metric) => void;
+    saveToggle: boolean;
+    saveFailed: () => void;
+}
+
+const MetricEdit = ({metric, otherTitles, onAdd, saveToggle, saveFailed}: Props) => {
+    const {formatMessage} = useIntl();
+    const [curState, setCurState] = useState(metric);
+    const [curTargetString, setCurTargetString] = useState(targetToString(metric.target, metric.type));
+    const [curSaveToggle, setCurSaveToggle] = useState(saveToggle);
```
crspeller, line 27: I don't get what is is about.

webapp/src/components/backstage/playbook_edit/metrics/metrics.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    const requestAddMetric = (addType: MetricType) => {
+        // Only add a new metric if we aren't currently editing.
+        if (curEditingIdx === -1) {
+            addMetric(addType);
+            return;
+        }
+
+        // We're editing. Try to close it, and if successful add the new metric.
+        setNextTask({type: TaskType.add, addType});
```
crspeller, line 46: This next task thing is quite confusing to me. Could we just set the appropriate state here?

crspeller (changes requested):
Looks good. Little confused by the state handling in the Metrics component. 

Also I noticed when trying it out that after you click add you are committed to adding a metric. There is no cancel button. Should we add a cancel button?

webapp/src/components/backstage/playbook_edit/metrics/metric_edit.tsx
cpoile: @crspeller It's a toggle to trigger the current editing component to try to save. Once triggered, it validates itself, and if valid calls save with its `curState`, and then the save does the `nextTask` (if any) (which is your question below). 
The alternative is to do all the validation and `curTask` (what we're currently editing) state in the `Metrics` component, but then typing a single character forces the entire Metrics component to redraw (all viewing and editing metric children), which is annoying. I'm not sure what the best design is for this. And even lifting the `curState` into the Metrics component doesn't fix the bug that @gbochora found -- switching to a different tab erases the `curState` (the edited data)... so, what, we lift that into the PlaybookEdit component? Kind of annoying. Would love your insight here. We can zoom about it too if you want (that might be quicker than a thread in github).

cpoile: Nope, it didn't work. I think we have to lift the state so that we can move between tabs and maintain the currently editing metric. 

The only little UX problem (that I can see) is that you can click the overall "Save" button while you're in the middle of editing, and it will save everything in the playbook /except/ the metric you are currently editing.  @abhijit-singh Is this acceptable? Sort of like a "look, you didn't add it, so you lost it" kind of behavior. If we want to force the user to save the currently editing metric, we need to either grey out the `Save` button until they're done editing (which could be confusing), or do that and add a tooltip, or add the verification-check-save flow (complicated). Your choice.

webapp/src/components/backstage/playbook_edit/metrics/metric_view.tsx
```diff
@@ -0,0 +1,143 @@
...
+import {targetToString} from 'src/components/backstage/playbook_edit/metrics/shared';
+
+interface Props {
+    metric: Metric;
+    editClick: () => void;
+}
+
+const MetricView = ({metric, editClick}: Props) => {
+    const {formatMessage} = useIntl();
+    const noTarget = formatMessage({defaultMessage: ': No target set.'});
```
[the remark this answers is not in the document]
cpoile: Good catch, thanks.

webapp/src/components/backstage/playbook_edit/metrics/metric_edit.tsx
```diff
@@ -0,0 +1,241 @@
...
+                    value={curState.title}
+                    onChange={(e) => {
+                        const title = e.target.value;
+                        setCurState((prevState) => ({...prevState, title}));
+                        setTitleError('');
+                    }}
+                    autoFocus={true}
+                />
+                {
+                    titleError !== '' &&
```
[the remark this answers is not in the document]
cpoile: I did it. It's cleaner, agreed.

webapp/src/components/backstage/playbook_edit/metrics/metrics.tsx
cpoile: Hopefully the above cleared it up for you? If not: we can't set the state directly here, it has to be set after the child edit component saves; this is the state that we'll move to /if/ the child component validates and saves successfully.

cpoile:
> Also I noticed when trying it out that after you click add you are committed to adding a metric. There is no cancel button. Should we add a cancel button?

The delete button basically does that. What do you think @abhijit-singh ?

webapp/src/components/backstage/playbook_edit/metrics/metric_view.tsx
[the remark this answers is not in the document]
cpoile: oops, did this in #1009 by accident https://github.com/mattermost/mattermost-plugin-playbooks/pull/1009/commits/aa418ecdd5d5244e0fcce9fd63fdd940bbad9a59

webapp/src/components/backstage/playbook_edit/metrics/metric_edit.tsx
abhijit-singh: That's not ideal, but I think that's okay as we'll soon be updating the editor to save section-wise and that should solve the UX issue. 

But just to double-check, can we not save the metric being edited currently even if it has all the details filled in? In case there are validation errors, then it is completely okay to discard the data imo.

abhijit-singh (changes requested):
Looks good @cpoile! One doubt and a few requests:
1. Should we add a maximum length restriction for the metric name? There could be layout issues as we display the metric value in multiple places and if we just truncate it everywhere, it'll anyways not add any value. I'm thinking as low as 24 or 32 characters.
image
2. Can we remove the target/description from the collapsed state if left blank?
image
3. Can we increase field height to 40px ([medium-size](https://zeroheight.com/29be2c109/p/937efe-text-inputs/t/77431c))? It was smaller in the designs but it'll be better to have it consistent with other fields. Updated the designs now. Note that the icon size will also change to 16px with this. 
4. Add a red border to the field for the error state. It is there for the `name` field but not the `target` field.
image


And yes, the delete button is the same functionality as `Cancel`. Don't think we need a separate cancel button which might end up adding more complexity to the add/edit/delete flows.

webapp/src/components/backstage/playbook_edit/metrics/metric_edit.tsx
cpoile: As discussed in the meeting -- this whole component hierarchy is going to be rewritten soon, so the decision was to merge with this rough edge, with the idea that the rewrite should be able to fix it with less engineering effort.

crspeller (approved):

cpoile:
@abhijit-singh Addressed all the comments -- for 1 I made the limit 64 characters (which is still only about 3/4 of the width), does it look ok?

image

abhijit-singh (approved):
Thank you @cpoile! 

There are a couple more extremely minor things, non-blocking so approving the PR. 

1. Can we add a tooltip when on the `Add metric` button when there are [already 4 metrics added](https://www.figma.com/file/1u5HZZMamACauYfR8121Uj/Retrospective-Report?node-id=1062%3A342923)? 
image
2. Is it possible to vertically centre align the text and icons in this container? So that the spacing above and below is equal? The text especially looks misaligned right now.
image

cpoile:
@abhijit-singh Thanks! The alignment is correct in #1009, and I'll add in the tooltip there too. Both because it's easier to modify that one than to modify 1002 and rebase 1009 onto it.

PR #1004, opened 2022-01-26
MM-40153: Fix wrapping in run overview

agarciamontoro:
#### Summary
Flex children use `min-width: auto` by default, so setting `min-width: 0` avoids that they expand the width of their parents.

I also deleted some unused code I found.

Before:
before

And after:
after

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40153

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


iomodo (approved):
Nice!

calebroseland (approved):
Good stuff!

PR #1006, opened 2022-01-26
MM-40499 - Fix for: categorization default setting

cpoile:
#### Summary
- Original bug in ticket was fixed at some previous point
- But found a new bug -- @agarciamontoro was this meant to be set to true by default in the templates? (I don't think it should be -- there's no category name set, so it should default to false) -- tagging you because looks like you touched it last

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40499

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


agarciamontoro (approved):
Huh, I don't remember modifying these. Don't these come from https://github.com/mattermost/mattermost-plugin-playbooks/pull/652? Anyway, I think it's ok to set them to false.

cpoile:
Closing because it looks like the problem was solved in another PR>

PR #1007, opened 2022-01-26
MM-41244: Simplify run status post

lieut-data:
#### Summary
Further simplifications to the run status post to optimize for threads view:

CleanShot 2022-01-26 at 13 16 43@2x

CleanShot 2022-01-26 at 13 17 03@2x

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41244

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


iomodo (approved):

gbochora (approved):
Just one question. It seems status posts will not contain a public channel link anymore. 
Is there another way to join the channel from threads, or there is no need for such link?

abhijit-singh (changes requested):
It personally throws me off when I mean to check an update and the text I see is `@someone ran this playbook`. I just checked and if we make it an h5 (#####), it has almost the same sizing but doesn't show the second line of text in the thread list view. I'd prefer it and it also looks less cluttered like that.
image


@gbochora, I think it is okay to let the users go to the `run page` from the message and then join the channel from there.

gbochora:
> @gbochora, I think it is okay to let the users go to the `run page` from the message and then join the channel from there.
Cool, thanks for the answer : )

abhijit-singh (approved):

PR #1008, opened 2022-01-26
MM-41036: Render update request post in threads

lieut-data:
#### Summary
The update request custom post failed to render in the threads view because it depended on `currentPlaybookRun`, of which there isn't one outside of a playbook run channel.

**Before:**
CleanShot 2022-01-26 at 16 27 15@2x

**After:**
CleanShot 2022-01-26 at 16 27 31@2x

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-41036

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


cpoile (approved):
Cool!

calebroseland (approved):

tests-e2e/cypress/support/api/user.js
```diff
@@ -247,6 +247,7 @@ Cypress.Commands.add('apiCreateUser', ({
         if (hideOnboarding) {
             cy.apiSaveOnboardingPreference(createdUser.id, 'hide', 'true');
             cy.apiSaveOnboardingPreference(createdUser.id, 'skip', 'true');
+            cy.apiSaveUserPreference([{user_id: createdUser.id, category: 'crt_tutorial_triggered', name: createdUser.id, value: '999'}], createdUser.id);
```
Phrynobatrachus, line 250: Can you put this in `apiDisableTutorials` instead please? Or if there's enough overlap between `hideOnboarding` and `bypassTutorial`, maybe all of that could be combined?

Phrynobatrachus (changes requested):

tests-e2e/cypress/support/api/user.js
lieut-data: Ooh, good catch -- much cleaner! I opted to inline into `apiDisableTutorials` for now.

Phrynobatrachus (approved):

PR #1009, opened 2022-01-26
MM-40926 - Playbook delete metric

cpoile:
#### Summary
- Delete metric
- The delete button does the same verification before switching to the deleting state (so that a metric that is being edited isn't persisted in an invalid state).
- But if the metric deleted is the one currently being edited, we're allowed to delete even if it's invalid.
- @abhijit-singh I adjusted the confirmation text: it doesn't mention `You will still be able to access historical data for this metric.` if the metric has not been persisted yet. Let me know if you agree (I think it's more user-friendly this way.)

- Looks like: (the three metrics here are already saved; this shows the validation and the different confirmation text for saved vs. new metrics)

https://user-images.githubusercontent.com/1490756/151259752-9acac0fa-d551-4243-bb9d-45d7728f8a00.mp4

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-40929

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/backstage/playbook_edit/metrics/metrics.tsx
```diff
@@ -55,7 +62,19 @@ const Metrics = ({playbook, setPlaybook, setChangesMade}: Props) => {
...
+
+    const requestDeleteMetric = (idx: number) => {
+        // Confirm delete immediately if we aren't currently editing, or editing the requested idx.
+        if (curEditingIdx === -1 || curEditingIdx === idx) {
+            setDeletingIdx(idx);
+            return;
+        }
+
+        // We're editing. Try to close it, and if successful delete the metric.
+        setNextTask({type: TaskType.delete, index: idx});
```
gbochora, line 77: Why do we save the editing metric before delete? What if we leave it as it is and just remove the item from the list?

gbochora (approved):
Looks great!

webapp/src/components/backstage/playbook_edit/metrics/metrics.tsx
cpoile: In this case we're editing a different metric -- we need to save it first, then delete the requested metric. I've updated the comment to be more clear on this. Thanks!

gbochora: Sorry, my question was not clear : ) I meant, what if we don't save metric and leave it in editing mode? Nor make fields validation. Just delete an item on which user clicked. After deleting, a user will continue editing.

abhijit-singh (commented):
Looks good to me functionally! Will we be moving the delete icon inside the box as per @gbochora's suggestion [here](https://community.mattermost.com/core/pl/pgeh8a86bibx9xwb5nm3jeo69y)?
Also, a couple of other requests:
1. Can we make the icon button size for edit and delete consistent? I'd used 28x28 icon buttons in the designs.
image
2. Can we change the button colour in the delete modal to red as it is a destructive action. This was incorrect in the [designs, just updated](https://www.figma.com/file/1u5HZZMamACauYfR8121Uj/Retrospective-Report?node-id=1062%3A342475).
image

gbochora:
One more minor issue, Text and icon are not centered in input fields.
Screenshot 2022-02-01 at 11 07 12

webapp/src/components/backstage/playbook_edit/metrics/metrics.tsx
cpoile: @gbochora Sorry, I'm not really sure what flow you are describing. Could you do a step-by-step description?

cpoile:
@gbochora I'm not sure how you're getting the text and icon to be not centered. What browser are you using? And are you using a zoom level other than 100%?

gbochora:
Safari, zoom level is 100%. I can reproduce it only when height is 36px, on 40px content is centred.

crspeller (approved):

cpoile:
@abhijit-singh Both are addressed, thanks!

abhijit-singh (approved):
💯

crspeller (approved):

PR #1016, opened 2022-02-02
[MM-41261] Add toast functionality

iomodo:
#### Summary
This PR adds Toast banner functionality. 
You can use like 
```js
const {add, remove} = useToast();

add('some text', some_duration);
```
Toast will be removed after `some_duration` automatically. If duration is less than or equal to 0, toast will not be removed, you can remove it by calling the `remove` function or by clicking the `X` in the UI.
Multiple toasts are supported at the same time.
Current PR only addresses the "duplicate" playbook flow. If we need toasts on other places I'll push separate PR for those.

https://user-images.githubusercontent.com/2340127/152172119-a919f1cc-880c-4fa1-b0b7-1deae54c5534.mp4

#### Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-41261

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


webapp/src/components/backstage/toast_banner.tsx
```diff
@@ -0,0 +1,136 @@
...
+    }, [toasts]);
+
+    const addHelper = (content: string, duration: number) => {
+        const id = toastCount++;
+        const toast = {id, content, duration};
+        setToasts((ts) => [...ts, toast]);
+    };
+
+    // do not recreate every time
+    const add = useCallback(
```
crspeller, line 101: Seems like over-optimization maybe?

webapp/src/components/backstage/toast_banner.tsx
```diff
@@ -0,0 +1,136 @@
...
+            return i;
+        }
+    }
+    return -1;
+};
+
+export const ToastProvider = (props: Props) => {
+    const [toasts, setToasts] = useState<ToastType[]>([]);
+
+    useEffect(() => {
```
crspeller, line 81: Instead of using useEffect maybe we can create a helper for setToasts that calls this? That way we don't get a double re-render.

webapp/src/components/backstage/backstage.tsx
```diff
@@ -110,96 +112,98 @@ const Backstage = () => {
...
+                <Switch>
+                    <Route path={`${match.url}/error`}/>
+                    <Route>
+                        <BackstageNavbar className='flex justify-content-between'>
+                            <div className='d-flex items-center'>
                                 <BackstageTitlebarItem
-                                    onClick={promptForFeedback}
-                                    to={`/${teams[0].name}/messages/@surveybot`}
-                                    data-testid='giveFeedbackButton'
+                                    to={`${match.url}/runs`}
```
crspeller, line 122: Are these extra changes here intentional?

webapp/src/hooks/crud.ts
```diff
@@ -125,6 +130,7 @@ export function usePlaybooksCrud(
     const duplicatePlaybook = async (playbookId: Playbook['id']) => {
         await clientDuplicatePlaybook(playbookId);
         await fetchPlaybooks();
+        addToast('successfully duplicated playbook', DuplicateBannerTimeout);
```
crspeller, line 133: This API makes me happy 😄

crspeller (changes requested):
This is great! Just a few nits.

abhijit-singh (commented):
This is awesome, thanks @iomodo! Would it be possible to [animate](https://www.figma.com/proto/kAV9UFVPRUXknFrydMc3qK/3.0-Ideas-%26-Mocks?page-id=310%3A184862&node-id=310%3A184917&viewport=489%2C48%2C0.31&scaling=min-zoom&starting-point-node-id=310%3A184917&show-proto-sidebar=1) the toast up and then back down? I think it should be possible with CSS transitions and not need anything fancier. 

Also, can we capitalize the 'S' in the message `Successfully duplicated playbook`. Thanks!

webapp/src/components/backstage/toast_banner.tsx
iomodo: Without `useEffect` I'm getting wired concurrency bug since `setToasts` is not a synchronous operation. When we have 2 active toasts, `toasts` array might not be up to date and `setToasts` leaves inconsistent state by not removing the toast at all or seemingly removing several toasts at once.

webapp/src/components/backstage/backstage.tsx
iomodo: there are no changes here except wrapping everything inside `<ToastProvider>`, just a wired rendering

iomodo:
Added animation

https://user-images.githubusercontent.com/2340127/152764476-9795c387-a9c2-4ae3-a775-ce774ce47ff0.mp4

webapp/src/components/backstage/backstage.tsx
crspeller: @iomodo Are you sure? For example it shows the icon size changing from `1` to `1.4`.

iomodo: Yep sure, still the github rendering thing. That icon size actually is `1.4`

webapp/src/components/backstage/toast_banner.tsx
crspeller: Using `setToasts((toasts) => toasts.splice(i, 1))` should fix that? That's supposed to be able to be called asynchronously if you use the function form.

crspeller (changes requested):
Also not sure if I like that animation. I feel like it's worse then no animation. Maybe a quick fade in?

webapp/src/components/backstage/toast_banner.tsx
```diff
@@ -0,0 +1,136 @@
...
+const StyledClose = styled.i`
+    color: var(--sidebar-text-60);
+`;
+
+let toastCount = 0;
+
+interface ToastType {
+    id: number;
+    content: string;
+    duration: number;
```
calebroseland, line 76: Post-MVP
```suggestion
    id: number;
    content: React.ReactNode;
    duration: number;
    type: 'success' | 'warning' | 'error' | 'info'
```

webapp/src/components/backstage/toast_banner.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    toasts.map(({content, id, ...rest}) => (
+                        <CSSTransition
+                            key={id}
+                            classNames='fade'
+                            timeout={2000}
+                        >
+                            <StyledToast
+                                {...rest}
+                            >
+                                <StyledCheck className={'icon icon-check'}/>
```
calebroseland, line 139: Post-MVP; icon according to type
```suggestion
                                <StyledIcon className={`icon icon-${icons[type]}`}/>
```

webapp/src/components/backstage/toast_banner.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+        const index = getIndexOfFirstRemovableToast(toasts);
+        if (index >= 0) {
+            window.setTimeout(() => {
+                setToasts((ts) => [...ts.slice(0, index), ...ts.slice(index + 1)]);
+            }, toasts[index].duration);
+        }
+    }, [toasts]);
+
+    const add = (content: string, duration: number) => {
```
calebroseland, line 113: Post-MVP; perhaps a default duration for ease &/ consistency
```suggestion
    const add = (content: React.ReactNode, type: ToastType['type'], duration = DEFAULT_DURATION) => {
```

webapp/src/components/backstage/backstage.tsx
calebroseland: 👍 Yep, shows correctly in an IDE diff

webapp/src/components/backstage/toast_banner.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+  
+    &.fade-exit {
+        transform: translateY(0);
+    }
+  
+    &.fade-exit-active {
+        transform: translateY(280px);
+        transition: transform 1s cubic-bezier(1, -0.4, 0.5, 1);
+    }
```
calebroseland, line 36: I agree linear on the entrance probably isn't the best, but not a huge deal. Here's a possible alt. set of timings.

https://user-images.githubusercontent.com/11724372/152876957-9714ac9e-8e5d-4a84-bd66-3c6359e6a8b1.mp4

```suggestion
    &.fade-enter {
        transform: translateY(80px);
    }

    &.fade-enter-active {
        transform: translateY(0);
        transition: transform 0.5s cubic-bezier(.44,.13,.42,1.43)
    }

    &.fade-exit {
        transform: translateY(0);
    }

    &.fade-exit-active {
        transform: translateY(280px);
        transition: transform 0.75s cubic-bezier(.59,-0.23,.42,1.43);
    }
```

calebroseland (approved):
Long-term, I might prefer to see something like [react-toastify](https://github.com/fkhadra/react-toastify) at the suite or platform level, but for a minimalist no-deps implementation, this is really great!

Optional suggestions: 
- alternate animation timings
- some ideas for possible changes that could be done later to enhance reusability (e.g. not hardcoded as `success`-like)

webapp/src/components/backstage/toast_banner.tsx
iomodo: Aah you are so correct, was not using the functions form here 🧑‍🎓

iomodo:
Thanks for the suggestions @calebroseland <3
I added the default duration and used your animation consts :)
Skipping the `type` for now, It might need some additional ui changes(colours and such). Let's defer it for future.

abhijit-singh (approved):
💯

crspeller (approved):

PR #1020, opened 2022-02-03
MM-41515 - Fix for: Cannot paginate playbooks

cpoile:
#### Summary
- There was a change made in https://github.com/mattermost/mattermost-webapp/pull/9639 where the height was changed from `height: calc(100% - 40px)` to `height: 100%`, which I don't totally understand. Boards made a fix in https://github.com/mattermost/focalboard/pull/2217 to change it back, and we're doing similar here. I'm not totally pleased with this solution, but I don't understand why the original change was made in webapp (waiting on an explanation) so I don't want to mess with webapp. And we're broken now, so this fixes our current problem. Hopefully in the future we can change it back.
- While working on this I also fixed the scrollbar problem I've been seeing for awhile:

Previous:
image

Fixed:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-41515

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/assets/right_dots.tsx
+++ b/webapp/src/components/assets/right_dots.tsx
@@ -1,12 +1,12 @@
 // Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
 // See License for license information.
 
-import React, {FC} from 'react';
+import React from 'react';
 import styled from 'styled-components';
 
 const Icon = styled.svg`
     position: absolute;
-    right: 0;
+    right: 8px;
     bottom: 0;
     max-height: 70%;
     height: auto;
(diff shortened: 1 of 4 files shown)


crspeller (approved):
This doesn't seem right, but I guess we need to deal with what the webapp is doing.

calebroseland:
Indeed, mattermost-webapp should give us a stable container—we shouldn't have to account for the height of global header , which isn't in our repo.

ref-link: https://github.com/mattermost/mattermost-plugin-playbooks/pull/962

calebroseland (approved):

PR #1026, opened 2022-02-07
Mm 40932 enter run metrics

gbochora:
#### Summary
This PR adds metrics input fields to the retrospective tab based on the playbook configuration. Users can enter metrics values and publish them together with report text.

Changes:
- Add `MetricData` component for entering metric values. 
- Add live error validation to the `MetricData` component.
- Add metric values auto-save functionality to the `MetricData` component.
- Add `MetricInput` component to reuse it in metric_edit and metric_data files. 
- Move retrospective `Publish` button from report to retrospective file.
- Move metrics values validation function to the shared file.
- Move styles to the shared file.
- Change title color from blue to black.

https://user-images.githubusercontent.com/4368372/152859126-62bfbb50-811f-4574-ab91-52626b59f800.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40932

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated


cpoile:
Hey @gbochora, two initial comments after trying it out:
1. If the target is blank or zero, I have assumed that the target is "not entered". We might change that in the future to allow `0` as a valid value, but to do so we would need another bit per entry that would say whether or not the field was "entered" or not. What do you think? If that sounds like a reasonable assumption, we would need to assume that a blank or `0` entry is not filled out. (So a `0` or blank entry would be considered incomplete, and for the target label: instead of `Target: 0`, it would be blank.)
2. I think Abhijit and Tao would want the metrics to be filled in, if they are present. @abhijit-singh what do you think about making the `Publish` button disabled until each metric has a valid value? And when the button is disabled: have a tooltip with a message like `Please fill in the metrics before publishing the retrospective`?

gbochora:
1. In some cases, 0 might be a valid number, so we should allow it. I think an issue with the current version is that it automatically stores blank values as zeros. 
I've changed logic to this:
- If a value is a blank, do not display an error message and do not save it(in such a way blank values are not stored as zero values)
- If a value is a valid number or duration, save it, even if it's zero.
- If a current metric value was changed to blank, display error(`Please fill in the metric value.`). Keep current valid value, do not save blank value.
What do you think?

2. You right! I forgot to add this check.

cpoile:
Cool. We'll talk about 1 in the meeting today -- for background: for the metric editor I've assumed a 0 value for a target means the target is not set. So when we have a `Target: 0` in the UI, it should read: `Target not set` or just be blank. If we want to change that, we're going to have to add a field that says `the target is not set`.

cpoile:
Probably should include changes from #1029 to this PR (either wait for it to get on master, or rebase onto it, your choice). Sorry for the churn. But that makes its easy to remove the `Target: __` info when it's null. :)

abhijit-singh (changes requested):
Thanks, @gbochora! This is great!

 A few requests:
 1. Can we add an `*` to all the metric labels to indicate they're required fields. [Figma file](https://www.figma.com/file/1u5HZZMamACauYfR8121Uj/Retr
image
ospective-Report?node-id=1809%3A334349) 
 2. If the user clicks on `publish` and any of the metrics does not have a value, show the error state for that field and do not trigger the publish confirmation modal.
 3. Change the font-weight for the help text of the metric fields to normal/400. Seems to be 600 currently. 
 4. For the published state, can we add a grey background to the fields to make it clear that they are disabled/read-only. 
 
image

cpoile:
A comment Abhijit made on #1028 -- we'll have to remove the metrics fields (if there are any metrics) when the user is in a version lower than Enterprise.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+            validateInputs() {
+                const errors = verifyInputs(inputsValues, true);
+                setInputsErrors(errors);
+
+                return !hasErrors(errors);
+            },
+        }),
+    );
+
+    const errorCurrencyInteger = formatMessage({defaultMessage: 'Please enter a number, or leave the target blank.'});
```
cpoile, line 56:
```suggestion
    const errorCurrencyInteger = formatMessage({defaultMessage: 'Please enter a number.'});
```
Because they aren't allowed to enter a blank metric. :)

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+    const errorDuration = formatMessage({defaultMessage: 'Please enter a duration in the format: dd:hh:mm (e.g., 12:00:00), or leave the target blank.'});
```
cpoile, line 57:
```suggestion
    const errorDuration = formatMessage({defaultMessage: 'Please enter a duration in the format: dd:hh:mm (e.g., 12:00:00).'});
```

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+const MetricsData = forwardRef(({metricsData, metricsConfigs, isPublished, onEdit, flushChanges}: MetricsProps, ref) => {
+    const {formatMessage} = useIntl();
+
+    const initialValues = new Array(metricsConfigs.length);
+    metricsConfigs.forEach((mc, index) => {
+        const md = metricsData.find((metric) => metric.metric_config_id === mc.id);
+        initialValues[index] = md ? metricToString(md.value, mc.type) : '';
+    });
+    const [inputsValues, setInputsValues] = useState(initialValues);
```
cpoile, line 34: This is interesting. You could also use a function initializer, and create the initialValues inside that. This way the logic is only run once. Right now, it's run on every render.
So, something like:
```suggestion
    const [inputsValues, setInputsValues] = useState(() => {
        ...logic that returns an array of initial values
    });
```

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+    const [inputsErrors, setInputsErrors] = useState(new Array(metricsConfigs.length).fill(''));
+
+    // Handles click outside of metrics inputs to save changes
+    const inputRef = useRef(null);
+    useClickOutsideRef(inputRef, () => {
+        flushChanges();
+    });
+
+    //  validateInputs function is called from retrospective component on publish button click, to validate metrics inputs
+    useImperativeHandle(
```
cpoile, line 44: Ah, this is better than the trigger code I was using in the playbook metrics edit tab. Nicely done! We should probably rewrite the playbook metrics edit tab to use this.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+                errors[index] = metricsConfigs[index].type === MetricType.Duration ? errorDuration : errorCurrencyInteger;
+            }
+        });
+        return errors;
+    };
+
+    function stringsToMetricsData(values: string[], errors: string[]) {
+        const newMetricsData = new Array<RunMetricData>();
+        errors.forEach((error, index) => {
+            // When input value is invalid, remain existing metric value
```
cpoile, line 77: I'm not sure what this comment means -- do you mean "When input value is invalid, retain existing metric value"?

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+            if (error) {
+                const metric = metricsData.find((m) => m.metric_config_id === metricsConfigs[index].id);
+                if (metric) {
+                    newMetricsData.push(metric);
```
cpoile, line 81: What happens if the metric is not found? Is that possible? 
If it's not possible, is there a different way of programming this?

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    return (
+        <div>
+            {
+                metricsConfigs.map((mc, idx) => {
+                    let placeholder = formatMessage({defaultMessage: ' Add value'});
+                    let inputIcon = <DollarSign sizePx={18}/>;
+                    if (mc.type === MetricType.Integer) {
+                        placeholder = formatMessage({defaultMessage: ' Add value'});
```
cpoile, line 108: nit: we don't need to repeat this assignment

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
```diff
@@ -0,0 +1,148 @@
...
+    );
+});
+
+function hasErrors(errors: string[]) {
+    for (let i = 0; i < errors.length; i++) {
+        if (errors[i] !== '') {
+            return true;
+        }
+    }
+    return false;
```
cpoile, line 145: nit: There's an easier way to do this in JS. :)
(e.g., `errors.some(...)`)

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/retrospective/retrospective.tsx
```diff
@@ -1,33 +1,77 @@
...
 
 import Report from './report';
+
 import TimelineRetro from './timeline_retro';
 
+const editDebounceDelayMilliseconds = 2000;
+
 interface Props {
     playbookRun: PlaybookRun;
+    metricsConfigs: Metric[] | null;
```
cpoile, line 40: I make a suggestion below in `playbook_run_backstage` about the `||`, that lets you do this:
```suggestion
    metricsConfigs?: Metric[];
```

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
```diff
@@ -433,10 +440,12 @@ const PlaybookRunBackstage = () => {
                         <Route path={`${match.url}/retrospective`}>
                             <Retrospective
                                 playbookRun={playbookRun}
+                                metricsConfigs={playbook?.metrics || null}
```
cpoile, line 443:
```suggestion
                                metricsConfigs={playbook?.metrics}
```

cpoile (approved):
Nice work @gbochora !  Nothing major I think, just little comments. I won't block on them.

Great job!

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
gbochora: Nice catch : )

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
gbochora: Yes, that's right. When the input value is invalid, I look for the last saved value and retain it.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
gbochora: Thanks, much better way : )

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/metrics/metrics_data.tsx
gbochora: Yes, it is possible. When a metric value is empty, and a user enters an invalid value. In this case, `error` value is not empty, and `metricsData` array does not contain a value for this metric.

I've changed the code, still not clear enough, but I think a little better.

abhijit-singh (approved):
Looks amazing, thanks @gbochora!

One minor change not directly related to metrics, but maybe can be a part of this PR as it is on the retro page. Can be done later as well, not high priority. For the `Published X hours ago` text and icon :
- Can we change the color to `--center-channel-color` with 72% opacity.
- Make font-weight 400/normal
- And lastly add about 16px spacing between the text and the `Publish` button. 
image

gbochora:
@abhijit-singh Thanks. Changed timestamp icon and text styles.

cpoile:
@calebroseland Gentle ping on this -- we need this in asap for next release. Thanks!

calebroseland (approved):
LGTM! Everything's working well. A couple of small things to consider, possibly deferrable to a polish PR later on, but nothing blocking.

Also, re: the nits and minor ones, don't feel the need to adjust the stylistic/spacing things—those are not important right now.

webapp/src/components/backstage/playbook_edit/metrics/shared.ts
```diff
@@ -43,3 +43,18 @@ export const stringToTarget = (target: string, type: MetricType) => {
         minutes: ddmmss[2],
     }).as('milliseconds');
 };
+
+export const isMetricValueValid = (type: MetricType, value: string) => {
+    if (type === MetricType.Duration) {
+        const regex = /(^$|^\d{1,2}:\d{1,2}:\d{1,2}$)/;
```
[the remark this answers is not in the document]
calebroseland: Alt possibility—reuse the duration input from the status update reminder?

CleanShot 2022-02-20 at 00 14 38@2x

gbochora:
@calebroseland thanks a lot for the detailed feedback!

webapp/src/components/backstage/playbook_edit/metrics/shared.ts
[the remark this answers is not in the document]
gbochora: It seems 99 days should be enough in this case. @abhijit-singh  what do you think?

PR #1029, opened 2022-02-08
MM-41684 - Support null targets and values

cpoile:
#### Summary
- I'm proposing to use the null.v4 pkg to support storing nulls in sql and json. Here I'm using it for the `int64` target and value fields, but the pkg supports strings and floats etc. It's pretty lightweight, simple to use and seamless (unless you want to do calculations in go, but then you'd have the same problem with using pointers I'd think).
- Here's the docs: https://pkg.go.dev/gopkg.in/guregu/null.v4 and the githup repo: https://github.com/guregu/null
- @lieut-data for visibility and if you want to comment or nix this addition.
- Added e2e tests to lock down our ability to remove target fields or add a 0 value for a target.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-41684

#### Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,5 +28,6 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.7.0
 	github.com/writeas/go-strip-markdown v2.0.1+incompatible
+	golang.org/x/tools v0.1.9 // indirect
 	gopkg.in/guregu/null.v4 v4.0.0 // indirect
 )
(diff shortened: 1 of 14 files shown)


gbochora (approved):
This is great! 🎉

crspeller (approved):
I like how explicit it is. Makes it hard to screw up. A little inconvenient though.

PR #1036, opened 2022-02-14
MM-41265: Welcome message action backend

agarciamontoro:
#### Summary
This is the first PR for the Actions theme, focused on the backend work. I'm opening one for the frontend in a bit.

For a bit of context before you go and review this +1.5k lines PR:

- This first iteration focuses on a single action: send a welcome message when a new member joins a channel. This action was already defined in playbooks, and this PR generalizes that logic to all channels.
- I've tried to keep it simple, but ended up being a bit generic, as the next PR I'm working on already adds a new action. That's why the action data model is flexible: in the database, the payload is a JSON blob to let each action carry any information it needs there, and that's why the `Payload` field of the action struct in the server is an `interface{}`.
- Because of that need of being generic, I have added a new dependency: [mapstructure](https://pkg.go.dev/github.com/mitchellh/mapstructure). This let us decode the JSON we get from the client in steps: depending on the value of the `action_type` field, the `payload` will have to adhere to a specific data model (in this case we only have one, `WelcomeMessagePayload`). `mapstructure` let us decode a generic map into a specific structure. If you have ideas to solve this in a more simple way, please tell me!
- The trigger for the welcome action is when a new member joins, but we need it to happen when the user _first_ views a channel. This was already solved by @chris.poile in the playbook action, but this PR generalizes that structure to every channel that has a welcome message action enabled. All the logic around that that lived under the playbook run service/store is now under the action service/store.
- For existing runs that have a welcome message configured, we need to get their MessageOnJoin column and create a new action for the run's channel with that message. This is handled in the new migration.
- For new runs that have a welcome message configured, we now need to create a new action for them and attach it to the run's channel. This is done in the playbook run service, right after creating the channel for the new run.

And I think that's mostly it. I've prepared the commits so you can review them one by one, describing any subtleties in the commit message and description if needed. I hope reviewing this is not too annoying, ask me anything if something's not clear.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41265

#### Checklist

- [ ] Telemetry updated - Added [a ticket](https://mattermost.atlassian.net/browse/MM-41801) for tackling this holistically once everything's merged
- [X] Unit tests updated


server/app/actions_service.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// setViewedChannelForEveryMember mark channelID as viewed for all its existing members
+func (a *channelActionServiceImpl) setViewedChannelForEveryMember(channelID string) error {
+	// TODO: this is a magic number, we should load test this function to find a
+	// good threshold to share the workload between the goroutines
+	perPage := 200
+
+	page := 0
+	var wg sync.WaitGroup
```
agarciamontoro, line 37: This is literally premature optimization. But I was concerned that enabling a welcome message in a channel with thousands of users would clog the server. If you think this is not needed, we can err on the more simple solution of simply iterating through all users sequentially.

server/api/actions.go
```diff
@@ -0,0 +1,147 @@
...
+
+	ReturnJSON(w, &actions, http.StatusOK)
+}
+
+// checkAndSendMessageOnJoin handles the GET /run/{id}/check_and_send_message_on_join/{channel_id} endpoint.
+func (a *ActionsHandler) checkAndSendMessageOnJoin(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	channelID := vars["channel_id"]
+	userID := r.Header.Get("Mattermost-User-ID")
+
```
srkgupta, line 113: @agarciamontoro This function is missing a validation to check if user has access to the specified ChannelID or not. 

Suggested check:
```
	if !a.PermissionsCheck(w, a.permissions.ChannelActionView(userID, channelID)) {
		return error....
	}
```

server/app/action.go
```diff
@@ -0,0 +1,80 @@
...
+	GenericChannelActionWithoutPayload
+	Payload WelcomeMessagePayload `json:"payload"`
+}
+
+const (
+	// Action types
+	ActionTypeWelcomeMessage = "send_welcome_message"
+
+	// Trigger types
+	TriggerTypeNewMemberJoins = "new_member_joins"
```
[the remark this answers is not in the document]
iomodo: Would it make sense to decouple triggers and actions, by having the pairing mechanism?
I imagined future architecture like this:
1. Have triggers registered
2. Have actions registered
3. Have the trigger-action config(what trigger causes what action)
4. When configured a trigger action initialize data accordingly.
5. When triggered fire an event and save the event in the trigger queue(channel), with the payload
6. In dedicated go routine get the events from the queue and check from config which actions do they correspond. Parse the payload and run the action.

iomodo (approved):
Great work @agarciamontoro !
Couple of non-blocking questions/discussions below.

agarciamontoro:
Latest commit changes the welcome message author from the Playbooks bot to the System. It looks like this now:

image

server/app/action.go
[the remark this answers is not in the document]
agarciamontoro: > Just to clarify, when we'll move towards many runs in a single channel, we will have to implement different trigger-action for showing up the playbook welcome messages, right?

The welcome message is now detached from Playbooks, actually. That the code for handling it lives under the Playbooks repo is just a temporary implementation detail, but they're tied to the channels: one channel, one welcome message. When we get multiple runs in a channel, we'll need to decide how to manage the welcome message defined in the Playbook that gets transferred to the created channel (maybe we should simply not allow that, but that's something we can discuss when we design that)

> Would it make sense to decouple triggers and actions, by having the pairing mechanism?

What you list is more or less the ideas I had in mind, but we're right now testing if this whole thing makes sense with a predefined set of hardcoded triggers and actions. When we validate this and generalize the framework, we'll get to something similar to that.

server/app/actions_service.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	default:
+		return fmt.Errorf("trigger type %q not recognized", action.TriggerType)
+	}
+
+	// Validate the payload depending on the action type
+	switch action.ActionType {
+	case ActionTypeWelcomeMessage:
+		var payload WelcomeMessagePayload
+		if err := mapstructure.Decode(action.Payload, &payload); err != nil {
```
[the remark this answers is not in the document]
agarciamontoro: How would you use `gob` here? I'm not sure how it works, but given it's part of the standard library, it may be a better solution.

server/sqlstore/actions.go
```diff
@@ -0,0 +1,288 @@
...
+
+	var action app.GenericChannelAction
+	err := c.store.getBuilder(c.store.db, &action, c.channelActionSelect.Where(sq.Eq{"c.ID": id}))
+	if err == sql.ErrNoRows {
+		return app.GenericChannelAction{}, errors.Wrapf(app.ErrNotFound, "action does not exist for id %q", id)
+	} else if err != nil {
+		return app.GenericChannelAction{}, errors.Wrapf(err, "failed to get action by id %q", id)
+	}
+
+	return action, nil
```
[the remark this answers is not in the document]
agarciamontoro: The idea is that we unmarshal the payload in the app layer, not in the SQL store. Does that make sense?

server/api/actions.go
```diff
@@ -0,0 +1,147 @@
...
+	actions, err := a.channelActionsService.GetChannelActions(channelID, triggerType)
+	if err != nil {
+		a.HandleErrorWithCode(w, http.StatusInternalServerError, fmt.Sprintf("unable to retrieve actions for channel %s", channelID), err)
+		return
+	}
+
+	ReturnJSON(w, &actions, http.StatusOK)
+}
+
+// checkAndSendMessageOnJoin handles the GET /run/{id}/check_and_send_message_on_join/{channel_id} endpoint.
```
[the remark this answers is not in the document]
agarciamontoro: Good eye, thanks! Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1036/commits/5983a2344f30feebb99bc613628011448fba4a59

server/app/action.go
```diff
@@ -0,0 +1,80 @@
...
+type GenericChannelAction struct {
+	GenericChannelActionWithoutPayload
+	Payload interface{} `json:"payload"`
+}
+
+type WelcomeMessagePayload struct {
+	Message string `json:"message" mapstructure:"message"`
+}
+
+type WelcomeMessageAction struct {
```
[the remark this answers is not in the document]
agarciamontoro: Whoops, good catch! This was part of an early iteration of the code, I think. Removed in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1036/commits/529448bad4cde80edf9221eeea7718a3d53484b3

server/app/actions_service.go
agarciamontoro: @iomodo, can we get all members at once? The ListMembers function is paginated, and we cannot know how many members the channel has.
What comes to mind is that we can create a new `SetViewedChannelForAllMembers(userIDs, channelID)` that is in charge of both retrieving the channel members and setting the channel viewed for all of them at once. Does that make sense?

server/api/actions.go
```diff
@@ -0,0 +1,147 @@
...
+	ReturnJSON(w, &actions, http.StatusOK)
+}
+
+// checkAndSendMessageOnJoin handles the GET /actions/channels/{channel_id}/check_and_send_message_on_join endpoint.
+func (a *ActionsHandler) checkAndSendMessageOnJoin(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	channelID := vars["channel_id"]
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	hasViewed := a.channelActionsService.CheckAndSendMessageOnJoin(userID, channelID)
```
srkgupta, line 114: Hi @agarciamontoro 

This API `check-and-send-message-on-join` is missing an authorization check to verify if the user has access to the specified channelID or not. As a result, the user can specify any channelID (which the user is not a member of) and can get a websocket event from the server containing the welcome message configured for that channel. Can you please add this check here.

agarciamontoro: Good catch, @srkgupta, thank you! Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1036/commits/31fddc8a6b217cd85083bfbe821287606b4da798

server/api/actions.go
```diff
@@ -0,0 +1,151 @@
...
+	if !a.PermissionsCheck(w, a.permissions.ChannelActionUpdate(userID, channelID)) {
+		return
+	}
+
+	var newChannelAction app.GenericChannelAction
+	if err := json.NewDecoder(r.Body).Decode(&newChannelAction); err != nil {
+		a.HandleErrorWithCode(w, http.StatusBadRequest, "unable to parse action", err)
+		return
+	}
+
```
srkgupta, line 137: The channelID in the body params of the PUT request might be different from the one specified in the Query Param. For example something like this:

```
PUT /channels/{channel_id}/{action_id}

{"channel_id": "different_channel_id".........}
```

In this case, the validation to check if user has write permissions to channel will pass. However when the server calls `a.channelActionsService.Update(newChannelAction)` it might actually be updating a different channel ID. 

Hence, a suggested fix is to include this:

```
	if newChannelAction.ChannelID != channelID {
		newChannelAction.ChannelID = channelID
	}
```

server/app/permissions_service.go
```diff
@@ -396,11 +396,50 @@ func (p *PermissionsService) RunViewByChannel(userID, channelID string) error {
...
+	}
+
+	permission := model.PermissionManagePublicChannelProperties
+	if channel.Type == model.ChannelTypePrivate {
+		permission = model.PermissionManagePrivateChannelProperties
+	}
+
+	return pluginAPI.User.HasPermissionToChannel(userID, channelID, permission)
+}
+
```
srkgupta, line 442: Just a suggestion to prevent future issues:

The function name indicates to check if user is the channelAdmin, however the actual checks performed here is to verify if user has permissions to manage channel properties or not. `Manage Channel Settings` permission might have been enabled for all members too which does not makes them channelAdmin. These two are different and might cause issues if this function is reused in future instances. 

Suggested Fix: 
Rename function to check if user has manage channel settings permissions instead of check isChannelAdmin.

agarciamontoro:
@srkgupta, thank you again for the thorough review! I addressed all your comments in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1036/commits/9d18e3c00fb5a58055297072a1205695288a4f73

server/api/actions.go
agarciamontoro: Instead of silently changing the body contents, I returned a 400 Bad Request Error, just to make the life of client developers easier: if the client has a bug where the URL and the body are different, I believe it's better to fail early and loudly.

srkgupta: @agarciamontoro can we make the behavior consistent in `createChannelAction` function as well. There we override and proceed further - like this:

```
	if channelAction.ChannelID != channelID {
		channelAction.ChannelID = channelID
	}
```

agarciamontoro: Whoops, didn't see that. Let me change it, thanks again!

agarciamontoro: Heh, changing this surfaced a bug in the tests, so I made *my* life easier :joy: Fixed in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1036/commits/ba08567bc8ef96f4717b076fe70a66eaf991f680 Thanks :)

srkgupta (approved):
Thanks @agarciamontoro. The final changes LGTM.

calebroseland (approved):
LGTM! Tested alongside the other PR and it appears to be working nicely. :)

PR #1046, opened 2022-02-17
Mm 41664 playbook key metrics tab backend

gbochora:
#### Summary
This PR Adds backend functionality to spit out metrics statistics info. Front-end will display statistics info on the  `Playbook Key metrics` tab.
PR adds these fields to the `Playbook` stats API:

- `MetricOverallAverage []int64` - Average value for each metric, indexed by metric, in the same order as `playbook.metrics`. Only published metrics are included in the calculations. 
- `MetricRollingAverage []int64` - Average values for the last published `x` runs.
- `MetricRollingAverageChange []int64` - Last 10 days average diff compared to previous 10 days average. For example, value +15 means that the last 10 days value went up by 15% compared to the previous period.
- `MetricValueRange [][]int64` - Each array contains two values, min and max values for each metric.
- `MetricRollingValues [][]int64` - Each array is a list of the last `x` published values. The first element in the list is the most recent.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41664

#### Checklist

- ~~ [ ] Telemetry updated~~ 
- ~~ [ ] Gated by experimental feature flag~~ 
- [x] Unit tests updated

### DIFF
--- a/server/sqlstore/stats.go
+++ b/server/sqlstore/stats.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"gopkg.in/guregu/null.v4"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/bot"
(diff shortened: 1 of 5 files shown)


server/sqlstore/stats.go
```diff
@@ -302,6 +304,131 @@ func (s *StatsStore) ActiveParticipantsPerDayLastXDays(x int, filters *StatsFilt
...
+// Returns list with null values, if Playbook has metrics but there are no any published values
+func (s *StatsStore) MetricRollingValuesLastXRuns(x int, offset int, filters *StatsFilters) [][]int64 {
+	// retrieve metric configs metricsConfigsIDs for playbook
+	metricsConfigsIDs, err := s.retrieveMetricConfigs(filters.PlaybookID)
+	if err != nil {
+		s.log.Warnf("Error retrieving metrics configs ids for playbook %w", err)
+		return [][]int64{}
+	}
+	metricsValues := make([][]int64, 0)
+	for _, id := range metricsConfigsIDs {
```
cpoile, line 382: We could do this in one statement for performance, but maybe it's not necessary?  Maybe we can add a comment here, something like: "It would be possible to turn this into a single statement; keep in mind if the playbookStats call becomes slow"

cpoile (approved):
Awesome awesome George! Let's get this (and the other PRs) in! Great work!

crspeller (approved):

PR #1059, opened 2022-02-24
MM-42084 - No metrics blank state

cpoile:
#### Summary
- No metrics blank state with a link to enter metrics in the edit/retro tab:
- @justinegeffen Can you let me know if the text needs editing? Thanks!

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42084

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):

abhijit-singh (approved):
💯

justinegeffen:
Thanks, @cpoile! I see there are two different sentences in the files - are they both going to be used, or just this one? 

I'm hesitant to use "concretely demonstrate performance" because it sounds like a vague claim (no offense!). I'm guessing we want to emphasize that you can accurately gauge performance improvements from these metrics? If so, we can say something like:

"Use metrics to understand patterns and progress across runs and track key performance areas."

Thoughts?

cpoile:
Thanks @justinegeffen! I realize the sentence is basically the same as the one you had edited before (the one in the upgrade banner), so I'll use the same wording (and that means the existing translations will be applied). 👍

PR #1060, opened 2022-02-25
e2e: enable retries on flaky viewport test

Phrynobatrachus:
This test fails some low percentage of the time in CI but still seems useful to keep, so retry on failure for now. This could be a Cypress issue (I've seen a few different viewport related failures on various tests) or an actual UI inconsistency, I haven't been able to repro locally so I'm not sure which.


gbochora (approved):
🎉

PR #1062, opened 2022-02-28
Mm 41031 add prometheus metrics

gbochora:
#### Summary
This PR adds collecting and exposing Prometheus via /metrics on the 9092 port. After merging, the SRE team will configure the Prometheus server to pull metrics from the 9092 port and monitor them.

Changelog:

- Add the Prometheus metrics (`metrics.metrics.go`)
- Add server to expose Prometheus metrics (`metrics.service.go`)
- Add scheduler from `focalboard` project, which helps to run recurring tasks (`scheduler.scheduler.go`)
- Add running a recurring task to update metrics values and starting a server to expose metrics (`plugin.go`)
- Add functions to the store layer to retrieve: 
	- Total number of active playbooks
	- Total number of active runs
	- Total number of overdue updates runs
	- Total number of completed runs without retro published

- Generate new store mock files because of changes in `PlaybookStore` and `PlaybookRunStore` interfaces.
- Add tracking created and finished runs count (`playbook_run_service.go`)
- Add tracking created, archived, and restored playbooks count (`playbook_service.go`)
- Add tracking errors count to bot logger (`bot.go`, `logger.go`, `api.api.go`) In the HandleErrorWithCode function, replace `Warnf` with `Errorf`, it helps to track the total error count and also seems right.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41031

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated


hanzei:
CC @levb @catalintomai as this might be interesting for apps (mostly for metrics about bindings)

server/plugin.go
```diff
@@ -272,3 +298,65 @@ func (p *Plugin) UserHasLeftChannel(c *plugin.Context, channelMember *model.Chan
...
+
+func (p *Plugin) runMetricsUpdaterTask(playbookStore app.PlaybookStore, playbookRunStore app.PlaybookRunStore, updateMetricsTaskFrequency time.Duration) {
+	metricsUpdater := func() {
+		if playbooksActiveTotal, err := playbookStore.GetPlaybooksActiveTotal(); err == nil {
+			p.metricsService.ObservePlaybooksActiveTotal(playbooksActiveTotal)
+		} else {
+			p.pluginAPI.Log.Error("error updating metrics, playbooks_active_total", err)
+		}
+
+		if runsActiveTotal, err := playbookRunStore.GetRunsActiveTotal(); err == nil {
```
crspeller, line 330: This kinda makes it awkward in HA because this is cross cluster while the stuff like runs created would be per-server. I guess the server metrics already do this so admins are aware?

server/scheduler/scheduler.go
```diff
@@ -0,0 +1,75 @@
...
+type ScheduledTask struct {
+	Name      string        `json:"name"`
+	Interval  time.Duration `json:"interval"`
+	Recurring bool          `json:"recurring"`
+	function  func()
+	cancel    chan struct{}
+	cancelled chan struct{}
+}
+
+func CreateTask(name string, function TaskFunc, timeToExecution time.Duration) *ScheduledTask {
```
crspeller, line 22: I might want to put a warning here warning people to not mess up and use this incorrectly for HA.

crspeller (changes requested):
Good stuff! Just a couple comments and questions.

server/metrics/metrics.go
```diff
@@ -0,0 +1,251 @@
...
+		Subsystem:   MetricsSubsystemRuns,
+		Name:        "participants_active_total",
+		Help:        "Total number of active participants (i.e. members of the playbook run channel when the run is active).",
+		ConstLabels: additionalLabels,
+	})
+	m.registry.MustRegister(m.participantsActiveTotal)
+	return m
+}
+
+func (m *Metrics) IncrementPlaybookCreatedTotal(num int) {
```
agarciamontoro, line 181: We should probably rename all these functions to use Count instead of Total, too
```suggestion
func (m *Metrics) IncrementPlaybookCreatedCount(num int) {
```

server/bot/logger.go
```diff
@@ -84,6 +84,8 @@ func (b *Bot) Errorf(format string, args ...interface{}) {
 	if level(b.configService.GetConfiguration().AdminLogLevel) >= 1 {
 		b.logToAdmins("ERROR", message)
 	}
+
+	b.metricsService.IncrementErrorsTotal(1)
```
agarciamontoro, line 88: As we discussed in the call: what if we use a middleware to inspect the status code returned by the endpoint handler, instead of plugging this in the Errorf call, which may not be called? Maybe `Router.Use`, which we already use for checking authorization in all requests, is helpful here https://pkg.go.dev/github.com/gorilla/mux?utm_source=godoc#Router.Use

server/sqlstore/playbook_run_test.go
```diff
@@ -830,6 +830,382 @@ func TestTasksAndRunsDigest(t *testing.T) {
...
+		t.Run("few active runs with overdue", func(t *testing.T) {
+			actual, err := playbookRunStore.GetOverdueUpdateRunsTotal()
+			require.NoError(t, err)
+			require.Equal(t, int64(9), actual)
+		})
+
+		// add active runs with due reminder
+		createRuns(store, playbookRunStore, 4, app.StatusInProgress, -1)
+
+		t.Run("few active runs with overdue", func(t *testing.T) {
```
agarciamontoro, line 929: This test's name is the same as the previous one. Something like the following?
```suggestion
		t.Run("few active runs with due", func(t *testing.T) {
```

server/sqlstore/playbook_run_test.go
```diff
@@ -830,6 +830,382 @@ func TestTasksAndRunsDigest(t *testing.T) {
...
+			actual, err := playbookRunStore.GetOverdueUpdateRunsTotal()
+			require.NoError(t, err)
+			require.Equal(t, int64(9), actual)
+		})
+	}
+}
+
+func TestGetOverdueRetroRunsTotal(t *testing.T) {
+	// overdue: 0 means no reminders at all. -1 means set only due reminders. 1 means set only overdue reminders.
```
agarciamontoro, line 938: This comments does not seem to apply here.

server/sqlstore/playbook_run_test.go
```diff
@@ -830,6 +830,382 @@ func TestTasksAndRunsDigest(t *testing.T) {
...
+		// create two playbooks
+		playbook1ID, err := playbookStore.Create(playbook1)
+		require.NoError(t, err)
+		playbook2ID, err := playbookStore.Create(playbook2)
+		require.NoError(t, err)
+
+		// create active runs without participants
+		createRuns(store, playbookRunStore, "", nil, team1ID, 2, app.StatusInProgress)
+		// create finished runs with participants
+		createRuns(store, playbookRunStore, playbook1ID, nil, team1ID, 3, app.StatusInProgress)
```
agarciamontoro, line 1188: The comment is not synced with the code: the runs are not finished and they do not have participants.

agarciamontoro (changes requested):
Impressive work, Giorgi! I left some nits and questions below.

server/app/playbook_run_service.go
```diff
@@ -284,6 +288,7 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
 	}
 
 	s.telemetry.CreatePlaybookRun(playbookRun, userID, public)
+	s.metricsService.IncrementRunsCreatedTotal(1)
```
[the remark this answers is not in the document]
gbochora: Yep, but they have a different purpose. I couldn't figure out how we can avoid this, any suggestions?

server/plugin.go
```diff
@@ -22,6 +26,12 @@ import (
 	"github.com/mattermost/mattermost-plugin-api/cluster"
 )
 
+const (
+	updateMetricsTaskFrequency = 15 * time.Minute
+
+	metricsExposePort = ":9092"
```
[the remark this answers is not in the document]
gbochora: I've updated the port to 9093, to avoid conflict with Focalboard, checked with the SRE team, and they are ok.

server/sqlstore/playbook_run_test.go
gbochora: Great catch! The comment was correct, code not.

server/bot/logger.go
gbochora: A much better solution, thanks, Alejandro! In addition, it turned out that `Plugin` implements the `ServeHTTP` function, another place to put status code check. But I like your version.

server/plugin.go
gbochora: On the server, only the leader node does this work. So we need to expose IsLeader to the plugin API to do it the same way. The downsides are it will break backward compatibility and take longer to merge changes.
After discussion with @lieut-data , we've decided that a more pragmatic way would be the current implementation. Data is collected once in 15 minutes, so it should not affect performance.

agarciamontoro (approved):
Great job, @gbochora, this was a tricky PR. Thanks! :tada:

lieut-data:
Nice work, @gbochora! Excited to see this live on community soon :)

PR #1064, opened 2022-03-01
Fix/weblate conflicts new

spirosoik:
#### Summary
Fix conflicts of Weblate


lieut-data (approved):
Thanks, @spirosoik! Is there anything we should be doing to help avoid these conflicts?

spirosoik:
@lieut-data as far I know we cannot but I need to check this a bit more.

DSchalla (approved):

lieut-data (approved):

PR #1067, opened 2022-03-02
MM-41862: Fix infinite scroll column sorting in runs list

agarciamontoro:
#### Summary

The `useRunsList` appends the newly fetched runs to the existing array, so pagination is transparent to the caller. You just set the page to the next one, and the returned `playbookRunsList` contains all the runs for all the pages you've already fetched. 

That's why we need to manually reset the page to 0 when changing the sorting strategy (either using a different field or setting the opposite direction). Otherwise, the `useRunsList` hook will simply add the newly fetched runs to the existing array, which will now contain elements from two different sorting strategies. Thus, potentially containing duplicate elements. This, in turns, makes the InfiniteScroll component buggy, as it ends up containing elements with the same key value.

This bug affected the main list of runs and the list of runs in the playbook usage (affected both by sorting by the columns and by a specific point in time). All those cases are shown in the video below:

https://user-images.githubusercontent.com/3924815/156462447-0d08547c-ac48-4884-8f6c-5d24f5ba364d.mp4

An alternative solution would be to refactor the `useRunsList` hook so that it simply returns what the caller asked for, not appending anything to the existing data. I think that solution is better, but it implies refactoring all callers of `useRunsList`, and I did not want to risk adding more bugs here and there trying to solve this specific one. If you agree this alternative solution is better, we can create a ticket a fix it afterwards to avoid risks in v1.25.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41862

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbooks/playbook_usage.tsx
+++ b/webapp/src/components/backstage/playbooks/playbook_usage.tsx
@@ -31,7 +31,7 @@ const PlaybookUsage = (props: Props) => {
 
     useEffect(() => {
         setFetchParams((oldParams) => {
-            return {...oldParams, playbook_id: props.playbook.id};
+            return {...oldParams, playbook_id: props.playbook.id, page: 0};
         });
     }, [props.playbook.id, setFetchParams]);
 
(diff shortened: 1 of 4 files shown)


agarciamontoro:
@Phrynobatrachus, I added you as a reviewer mainly to let you know when this fix is merged, but a second pair of eyes looking into whether this is actually fixed wouldn't hurt.

lieut-data (approved):
Nice, pragmatic solution! 0/5 on whether we restore as you propose.

crspeller (approved):
Makes sense.

PR #1068, opened 2022-03-02
MM-42210 Fix import of private playbooks.

crspeller:
#### Summary
Fix import not working for unlicenced servers because licencing doesn't allow private playbooks.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42210


lieut-data (approved):
Makes sense! Are there other import steps that might fail, eg having retrospective enabled?

crspeller:
@lieut-data Public / Private is the only one checked at creation time. Everything else is checked elsewhere.

Phrynobatrachus (approved):

PR #1071, opened 2022-03-03
MM-41950 New playbooks permissions

crspeller:
#### Summary
Adding some permissions for playbooks. See details in the linked ticket. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41950


webapp/src/components/backstage/select_users_below.tsx
```diff
@@ -143,7 +147,7 @@ const UserLine = (props: UserLineProps) => {
         </IconWrapper>
     );
 
-    if (props.hasPermissionsToEditMembers) {
+    if (props.hasPermissionsToEditRoles && props.member.user_id !== props.currentUserId) {
```
gbochora, line 150: Why doesn't UI allow a user to edit his role? If I'm not mistaken, the backend doesn't have such restriction.

server/app/permissions_service.go
```diff
@@ -224,6 +225,21 @@ func (p *PermissionsService) PlaybookModifyWithFixes(userID string, playbook *Pl
 		if err := p.PlaybookManageMembers(userID, oldPlaybook); err != nil {
 			return errors.Wrap(err, "attempted to modify members without permissions")
 		}
+
```
gbochora, line 228: Is it possible that a user has `PlaybookManageRoles` permission and doesn't have `PlaybookManageMembers` permission?
If yes, it seems such users will not be able to edit roles.

gbochora (approved):
Looks great! Just a few questions.

server/app/permissions_service.go
```diff
@@ -224,6 +225,21 @@ func (p *PermissionsService) PlaybookModifyWithFixes(userID string, playbook *Pl
...
+		oldMemberRoles := map[string]string{}
+		for _, member := range oldPlaybook.Members {
+			oldMemberRoles[member.UserID] = strings.Join(member.Roles, ",")
+		}
+
+		// Also need to check if roles changed. If so we need to check manage roles permission.
+		for _, member := range playbook.Members {
+			if oldRoles, ok := oldMemberRoles[member.UserID]; !((!ok && len(member.Roles) == 1 && member.Roles[0] == PlaybookRoleMember) || (ok && strings.Join(member.Roles, ",") == oldRoles)) {
```
agarciamontoro, line 236: Nit: can we move this condition to a named variable to understand what it is without having to parse it in one's head?

server/api_playbooks_test.go
```diff
@@ -762,6 +762,17 @@ func TestPlaybooksPermissions(t *testing.T) {
...
+		t.Run("public with permissions", func(t *testing.T) {
+			defaultRolePermissions := e.Permissions.SaveDefaultRolePermissions()
+			defer func() {
+				e.Permissions.RestoreDefaultRolePermissions(defaultRolePermissions)
+			}()
+			e.Permissions.AddPermissionToRole(model.PermissionPublicPlaybookManageProperties.Id, model.PlaybookMemberRoleId)
+
+			err := e.PlaybooksClient.Playbooks.Update(context.Background(), *e.BasicPlaybook)
+			assert.NoError(t, err)
+		})
```
agarciamontoro, line 774: Not sure if I'm parsing this right, but there it goes:
When this test case finishes, `e.BasicPlaybook` has been successfully updated, so the following test cases do not attempt to change anything in the playbook. If that's right, I'm not sure whether the following test cases function properly.
I'd be in favor of making the tests completely independent of each other, if that's possible.

server/api_playbooks_test.go
```diff
@@ -773,7 +784,77 @@ func TestPlaybooksPermissions(t *testing.T) {
...
+			defer func() {
+				e.Permissions.RestoreDefaultRolePermissions(defaultRolePermissions)
+			}()
+			e.Permissions.AddPermissionToRole(model.PermissionPublicPlaybookManageProperties.Id, model.PlaybookMemberRoleId)
+			e.Permissions.AddPermissionToRole(model.PermissionPublicPlaybookManageMembers.Id, model.PlaybookMemberRoleId)
+
+			err := e.PlaybooksClient.Playbooks.Update(context.Background(), *e.BasicPlaybook)
+			assert.NoError(t, err)
+		})
+	})
```
agarciamontoro, line 826: The same may happen here (and in the following one) if in the future we add test cases below the one that successfully edit the playbook.

agarciamontoro (changes requested):
Looks great :) Thank you, Christopher!

I am a bit confused about some tests, and left some nits below.

webapp/src/components/backstage/select_users_below.tsx
crspeller: Correct. This actually matches the behavior of channels. The idea is you want to prevent the user from demoting themselves.

server/app/permissions_service.go
crspeller: It's technically possible but there isn't any way to make that happen in the UI. I also don't think that configuration makes a whole lot of sense.

server/api_playbooks_test.go
crspeller: These work fine because there are failures than successes. I agree it might not be totally clear what is happening. We could just make everything a separate test if you think that would clear it up.

server/api_playbooks_test.go
agarciamontoro: Aaaaah, I see now. Thanks for the clarification.

server/app/permissions_service.go
agarciamontoro: It definitely is, thanks!

server/api_playbooks_test.go
agarciamontoro: Yup, it works now. I'm concerned that we inadvertently break them (as in making them useless because they always pass) in the future. We can solve it then if it happens, though, I'm not proposing to do it now preemptively.

agarciamontoro (approved):
:rocket:

crspeller:
Not sure why errcheck from golangci-lint decided that this PR needed to make that change. But I have made it happy.

PR #1072, opened 2022-03-05
[I18N] playbook.tsx: Small grammatical typo correction

tsabi:
#### Summary
Small, one-line correction of a grammatical error in an English text

#### Ticket Link
We talked this in chat, no issue ticket

#### Checklist
- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


cwarnermm (approved):
Thanks, @tsabi! Great catch!

calebroseland (changes requested):
Thanks for submitting this, @tsabi! Just a quick `make i18n-extract` needed to update `en.json`

> Please run "make i18n-extract" and commit the changes in webapp/i18n/en.json.

tsabi:
> Thanks for submitting this, @tsabi! Just a quick `make i18n-extract` needed to update `en.json`
> 
> > Please run "make i18n-extract" and commit the changes in webapp/i18n/en.json.

Ok, i did.

I would like to point out that the current use of translation json files is not ideal, as all text is referenced by a dynamically generated (I guess the hash of the text to be translated) identifier.  This current situation will be interpreted by Weblate, for example, as deleting a text and inserting a new text, so the translations of the old text will be deleted.
On the other hand, if you use a static identifier for your texts - which is also possible with `formatMessage` - Weblate will interpret it as if the text has remained, but only a few characters have been changed inside, and this is much more convenient for translators.

I know it's a lot of work to go through the source code and burn in the current hash as a static identifier, but as a translator I feel it's worth it.

calebroseland (approved):
Thanks @tsabi—this is great feedback and there've been a few conversations about this topic and how we can improve things.

Also, sorry for not seeing this. It's often best to re-request a review once it's ready again, that'll notify us directly.

Thanks again!

PR #1079, opened 2022-03-09
[MM-40860] Fix long translations in status update modal

iomodo:
#### Summary
This PR fixes longer translations in the status update modal

Screen Shot 2022-03-09 at 14 43 43

#### Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-40860

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/webapp/src/components/assets/buttons.tsx
+++ b/webapp/src/components/assets/buttons.tsx
@@ -39,6 +39,7 @@ export const PrimaryButton = styled(Button)`
     background: var(--button-bg);
     color: var(--button-color);
     transition: background 0.15s ease-out;
+    white-space: nowrap;
 
     &:active:not([disabled])  {
         background: rgba(var(--button-bg-rgb), 0.8);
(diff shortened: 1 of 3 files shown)


crspeller (approved):

calebroseland (commented):
I think we may need to apply this `nowrap` to just the update post dialog, this is happening as of now in RHS About.


CleanShot 2022-03-11 at 14 27 01@2x

iomodo:
@calebroseland `nowrap` will not fix that. This will still happen because there is not enough real estate, when we compress the page.
Screen Shot 2022-03-15 at 13 03 06

Without compression it looks good either way.  
Screen Shot 2022-03-15 at 13 06 50

Screen Shot 2022-03-15 at 13 07 41

Thoughts what could be done?

calebroseland (approved):
Doh, sorry for the long delay, Shota. I was brewing on it a bit, but what I meant was possibly limiting `white-space: nowrap` to just the buttons in `update_run_status_modal`.

```tsx
const FooterContainer = styled.div`
    // ...
    ${PrimaryButton}, ${TertiaryButton} {
        white-space: nowrap;
    }
`;
```

CleanShot 2022-03-22 at 22 30 28@2x

This would've keep the modal buttons one-line, but allow other buttons to wrap as the currently do, but that has trade-offs and probably doesn't make much sense.


(Separately; during mobile-view support efforts) perhaps we should just tweak/refine the that RHS section to intentionally wrap vertically at narrow widths
image

PR #1083, opened 2022-03-15
e2e: channel action tests

Phrynobatrachus:
Will be adding more for the keywords and other actions as they come in.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41516


agarciamontoro:
I think we can merge and add the new ones later. Thoughts, @Phrynobatrachus?

PR #1084, opened 2022-03-15
GraphQL Prototype

crspeller:
This is a prototype PR for what a GraphQL implementation for playbooks could look like. Just to get the conversation started. 

The backend is based on https://github.com/mattermost/mattermost-server/pull/19486 the mattermost-server implementation. Eventually we would be able to reuse that portion when we move over and the multi product architecture allows for combining our schemas. In turn this is based on https://github.com/graph-gophers/graphql-go

The important parts of the backend:
- [The graphql schema](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1084/files#diff-f17ebf9f1d405db89bb404274c3717011310b388a97387b8b35362d9c3daad28) which powers a lot of the automation and is hopefully the single source of truth. 
- The "resolvers" in graphql_root and graphql_playbook. The playbook one isn't much right now that code just resolves type incompatibilities in how graphql-go works. The root resolver implements the basics of how fetch and update work. Revolvers can be [chained ](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1084/files#diff-3bb00d44625dd2e0c20943ac9e3d313f0307bf56acd6db3c6c7bd89c9f20ac24R31) and reference each other. 

On the client side we have no current implementation to copy so I decided to use the simplest to use graphql client Apollo. I debated between Apollo and Relay. They both solve the caching problem for you. Relay is more complex but it allows for fragmented queries to allow sub-components to request exactly what data they need and be forced to only use that. It's a really nice architecture however it very much encompasses your code with Relay code which I didn't want to commit to. It also uses unreleased React features which would force us to upgrade to unstable React versions. (Relay is developed by FB) Apollo is simple and straightforward to understand.

Client types are quite the thing with graphql. I decided to try and generate them using `graphql-codegen`. You can see the generated types [here](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1084/files#diff-844a6674ff263d4deea57152f0e0c7bf9d1ae30ac2cef9857f737cf14ccf581d). It generates them from the graphql schema the server uses. 
In addition it is able to [generate hooks compatible with Apollo](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1084/files#diff-844a6674ff263d4deea57152f0e0c7bf9d1ae30ac2cef9857f737cf14ccf581d). It does this by reading [queries from files](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1084/files#diff-55cd529d2e1190631c17c838789b1bd753f359ed7d8174b64aeb4cb90bd833c7) that we write. Then you can easily use that in [your own hooks](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1084/files#diff-3c602d7e98bae08f51f476456f52b3f8349ad62bba4f379d4225b9d2c4e6afe2R110).

Love to have everyone's feedback on the direction this is headed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,7 @@ endif
 .PHONY: webapp
 webapp: webapp/node_modules
 ifneq ($(HAS_WEBAPP),)
+	cd webapp && $(NPM) run graphql;
 ifeq ($(MM_DEBUG),)
 	cd webapp && $(NPM) run build;
 else
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/golang/mock v1.6.0
 	github.com/gorilla/mux v1.8.0
+	github.com/graph-gophers/graphql-go v1.3.0
 	github.com/jmoiron/sqlx v1.3.4
 	github.com/lib/pq v1.10.4
 	github.com/mattermost/mattermost-plugin-api v0.0.22-0.20211207232216-3faec618d311
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -17,10 +17,11 @@ import (
 // Handler Root API handler.
 type Handler struct {
 	*ErrorHandler
-	pluginAPI *pluginapi.Client
-	APIRouter *mux.Router
-	root      *mux.Router
-	config    config.Service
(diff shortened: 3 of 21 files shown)


crspeller:
Closing prototype in favor of actual implementation.

PR #1085, opened 2022-03-16
MM-42550: Edit prompt post once the run is actually created

agarciamontoro:
#### Summary

Until now, the keywords flow was like this:
1. A user posts a message containing some keywords
2. The Playbooks bot sends an automatic post containing two buttons: "Yes, run playbook" and "No, ignore thread"
3. The user clicks on "Yes, run playbook"
4. The run creation interactive dialog is shown, letting the user add a run name and start the run.
5. At the same time as 4, the automatic post sent in 2 is *immediately* edited to "@user ran the playbook <playbook_name>".

The important part here is that the last step happened at the same time the interactive dialog was shown, before that dialog was confirmed, so even if the user decided to not run the playbook, the automatic post would still show "@user ran the playbook <playbook_name>".

This PR changes that last step with the following one:

5. If, and only if, the user does start the run, the automatic post sent in 2 is edited to "@user started the run <run_name> using the <playbook_name> playbook". If the user decides, in step 4, to not run the playbook, the automatic post will not be edited.

The last commit in this PR is somehow isolated, as it just adds some copy I forgot to add in the main PR, #1081.

As usual, review by commit, it should be easier :)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42550

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):

gbochora (approved):
Nice!

PR #1086, opened 2022-03-16
MM-42394: Sidebar categorization action

agarciamontoro:
#### Summary

This PR adds the third and last action from this effort to refactor playbook actions into channel actions. This time, the action to automatically categorize a channel in the sidebar for a user when they join a new channel. Until now, this was a feature only available to run channels, but this PR generalizes it to every channel.

As usual, please review by commit and read the commit descriptions. I tried to explain everything there, and I think it will make your reviewer life easier.

#### Demo

This video shows the new action setting, "Add the channel to a sidebar category for the user", and how the modal behaves when the scrolling height is higher than the threshold set:

https://user-images.githubusercontent.com/3924815/158659164-1e0df322-39c2-4fd9-8c10-7869acb236a2.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42394

#### Checklist

- [X] Telemetry updated
- [X] Unit tests updated


agarciamontoro:
The failure in the E2E tests seem to be caused by some changes upstream. See [this conversation](https://community-daily.mattermost.com/core/pl/ept6ccao9ifwpcc5adqrrqysow) for more context.

abhijit-singh (approved):
Perfect! 💯

server/sqlstore/migrations.go
```diff
@@ -1899,6 +1899,54 @@ var migrations = []Migration{
...
+	{
+		fromVersion: semver.MustParse("0.50.0"),
+		toVersion:   semver.MustParse("0.51.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			// Retrieve the channel ID and category name of every run
+
+			selectQuery := sqlStore.builder.
+				Select("ChannelID", "CategoryName").
+				From("IR_Incident").
+				Where(sq.NotEq{"CategoryName": ""})
```
gbochora, line 1914: Do we need to create actions for finished runs too?

agarciamontoro: Good question and good eye, I asked the same thing when we were designing this. We do want it for finished runs too, as they can be restarted at anytime, and we still want their configuration to be the same as before.

webapp/src/components/actions_modal.tsx
```diff
@@ -74,6 +86,21 @@ const ActionsModal = () => {
...
+            Object.entries(record).forEach(([triggerString, actionsInRecord]) => {
+                const trigger = triggerString as ChannelTriggerType;
+                const finalActions = [] as ChannelAction[];
+                defaultActions[trigger].forEach((defaultAction: ChannelAction) => {
+                    const actionFetched = actionsInRecord.find((actionInRecord) => equalActionType(actionInRecord, defaultAction));
+                    if (actionFetched) {
+                        finalActions.push(actionFetched);
+                    } else {
+                        finalActions.push(defaultAction);
+                    }
```
gbochora, line 99: It seems if there is a `KeywordsPosted` action in the fetched list, `find` will not be able to return it because keywords will be different. So it will add default action instead of fetched one.

gbochora (commented):
Great work! Two possible issues:

1. Clicking on the toggle button(`Channel Action Modal`) does not change the state. It might not be clear that clicking on the `action description enables it.
2. Can't add `keywords` action, after page refresh information is lost. 
Console logs: `(in promise) Error: {"error":"invalid action"}`

webapp/src/components/actions_modal.tsx
```diff
@@ -32,6 +34,15 @@ const defaultActions: ActionsByTrigger = {
                 message: '',
             },
         },
+        {
+            channel_id: '',
+            enabled: false,
+            action_type: ChannelActionType.CategorizeChannel,
+            trigger_type: ChannelTriggerType.NewMemberJoins,
```
calebroseland, line 41: (For potential refinement later on)

Does `defaultActions` need to be modeled with mixed object+array data? I just feel that this may complicate the fetch+merge code. With each `ChannelAction` already having a `trigger_type`, could it just be a simple array?

```ts
const defaultActions: Record<ChannelTriggerType, ChannelAction[]> = {t1: [a1, a2], t2: [a3]}

```

```ts
const defaultActions: `ChannelAction[] = [a1, a2, a3]

// if needed for rendering/etc., back to `Record<ChannelTriggerType, ChannelAction[]>`
import groupBy from 'lodash/groupBy';
const defaultActionsByTriggers = groupBy(defaultActions, 'trigger_type')
```

webapp/src/components/actions_modal.tsx
```diff
@@ -74,6 +86,21 @@ const ActionsModal = () => {
                 }
             });
 
+            // Merge the fetched actions with the default ones
+            Object.entries(record).forEach(([triggerString, actionsInRecord]) => {
```
calebroseland, line 90: If `defaultActions` were `ChannelAction[]`,  could something like `const record = defaultActions.reduce(/* ... merge fetchedActions ... */)` potentially increase clarity here? (Not a big deal for now, but maybe the complexity could be refined later on.)

calebroseland (commented):
Lot's of good stuff! Tested the functionality and `channel join` -> `do categorization` is working well.

> Clicking on the toggle button(Channel Action Modal) does not change the state

Also noticed this—there's a conflicting double `onChange` handler in `actions_modal_action.tsx` from a previous PR that would be good to address.

`actions_modal` has some amount of complexity in the way it handles default/fetched data, and some potential (reading clarity) refactoring might be nice, but that could wait for later iterations.

webapp/src/components/actions_modal.tsx
agarciamontoro: Great catch, thank you! I just got rid of all the unneeded complexity in the equal function. See https://github.com/mattermost/mattermost-plugin-playbooks/pull/1086/commits/8c61cc0e81e10970dded2346e89f101cff0cbe0f This fixes it.

agarciamontoro:
Thank you, @calebroseland and @gbochora for the thorough reviews :)

I fixed the two bugs you pointed out, and created this tech debt ticket https://mattermost.atlassian.net/browse/MM-42620 to address your suggestions, @calebroseland

webapp/src/components/actions_modal.tsx
agarciamontoro: Yes, the motivation for having an array for each trigger was to simplify the rendering logic. But it's true that it made this fetch+merge code very complex. I'll address this in https://mattermost.atlassian.net/browse/MM-42620, thanks!

webapp/src/components/actions_modal.tsx
agarciamontoro: Good point, I'll add these two comments to the ticket. Thanks!

gbochora (approved):

calebroseland (approved):

PR #1088, opened 2022-03-18
MM-42437: Do not broadcast creation if broadcast is disabled

agarciamontoro:
#### Summary
When a playbook run is being created, and only then, we can assume that its configuration is an exact copy of the configuration of its playbook.

While that assumption is correct, its interpretation in this piece of code was not right. The configuration of the playbook is not only `playbook.BroadcastChannelIDs`, but that value along with `playbook.BroadcastEnabled`. If the broadcast is enabled in the
playbook (regardless of the value of `playbook.BroadcastChannelIDs`, which may contain some non-empty value that was stored from a previous configuration), then `playbookRun.BroadcastChannelIDs` is guaranteed to be empty.

The values in the `playbookRun` struct are always valid because, on creation, we do check if the playbook corresponding configuration is enabled or not, only copying the values to the run if it is enabled.

That's why we need to use `playbookRun.BroadcastChannelIDs` when broadcasting the announcement of the playbook run, and not the one in the playbook struct.

@lieut-data, the original ticket talked generally about "broadcasting", but I only found buggy behaviour in the broadcast of the announcement post, not on the subsequent broadcasted status updates. Is that what the ticket means to solve?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42437

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


iomodo (approved):
Good catch!

agarciamontoro:
> Any way we can test this?

E2E test done :)

PR #1093, opened 2022-03-21
e2e: CircleCI housekeeping

Phrynobatrachus:
I added the license upload success/fail check back in now that the server issue there is fixed, and also updated legacy convenience images to new versions as [recommended by Circle](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,7 +295,7 @@ jobs:
         environment:
           POSTGRES_USER: mmuser
           POSTGRES_DB: mattermost_test
-      - image: circleci/mariadb:latest
+      - image: cimg/mariadb:10.6
         environment:
           MARIADB_ROOT_PASSWORD: mostest
           MARIADB_DATABASE: mattermost_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -288,7 +288,7 @@ jobs:
     steps:
       - test-with-db
 
-  test-MariaDB-Postgres12:
+  test-MariaDB10-Postgres12:
     docker:
       - image: docker.io/mattermost/builder:go-1.16.5-node-16.4.0
       - image: cimg/postgres:12.9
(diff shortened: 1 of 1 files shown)


Phrynobatrachus:
Yeah I'm not sure what's up with that, I'm re-running it to see if it'll complete this time. The postgres logs for the Cypress jobs looked normal otherwise, which should be the only thing affected on those particular ones.

.circleci/config.yml
```diff
@@ -308,12 +309,12 @@ jobs:
   test-MySQL-Postgres-latest:
```
lieut-data, line 309: Seems like this target was exclusively for `-latest`? Is there an equivalent for `cimg`?

.circleci/config.yml
```diff
@@ -290,11 +291,11 @@ jobs:
   test-MariaDB-Postgres12:
     docker:
       - image: docker.io/mattermost/builder:go-1.16.5-node-16.4.0
-      - image: circleci/postgres:12-alpine
+      - image: cimg/postgres:12.9
         environment:
           POSTGRES_USER: mmuser
           POSTGRES_DB: mattermost_test
-      - image: circleci/mariadb:latest
+      - image: cimg/mariadb:10.6
```
lieut-data, line 298: I like the idea of pinning MariaDB here. Should we rename the target as `test-MariaDB10-Postgres12`?

.circleci/config.yml
```diff
@@ -272,11 +273,11 @@ jobs:
   test-MySQL8-Postgres11:
     docker:
       - image: docker.io/mattermost/builder:go-1.16.5-node-16.4.0
-      - image: circleci/postgres:11-alpine
+      - image: cimg/postgres:11.13
```
lieut-data, line 276: Is there a way to avoid the minimum version specification, and just target the latest `11.x`? (here, and elsewhere)

lieut-data (changes requested):
Thanks, @Phrynobatrachus! Just a few thoughts on image version selection below.

.circleci/config.yml
Phrynobatrachus: They appear to have gotten rid of latest tagging in the new batch of images.

lieut-data:
It might be in GitHub, yes. @crspeller, do you have admin access here? Can you take a look?

crspeller:
@lieut-data @Phrynobatrachus I removed that one as a required check.

lieut-data (approved):
Thanks, @Phrynobatrachus!

PR #1095, opened 2022-03-22
Add overdue tasks filter

gbochora:
#### Summary
This PR adds `overdue tasks` filter to the RHS checklist. A special indicator is shown near the `Checklist` title if there are overdue tasks. Clicking on the indicator turns on/off the filter. Overdue tasks filter works together with existing filters.

https://user-images.githubusercontent.com/4368372/159499097-b3bd1c25-9743-4b5b-a043-2b358c9bdc7d.mp4

I'll add e2e tests after merging with the checklist new UI.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42399

#### Checklist

- ~~[ ] Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/rhs/rhs_checklist_list.tsx
```diff
@@ -394,13 +405,39 @@ const StyledPrimaryButton = styled(PrimaryButton)`
...
+    color: ${(props) => (props.toggled ? 'var(--button-color)' : 'var(--dnd-indicator)')};
+`;
+
 export default RHSChecklistList;
 
+const overdueTasks = (checklists: Checklist[]) => {
+    let count = 0;
+    for (const list of checklists) {
+        for (const item of list.items) {
+            if (item.due_date > 0 && DateTime.fromMillis(item.due_date) <= DateTime.now()) {
```
crspeller, line 428: Nit: Might be just a bit nicer if this used a constant `DateTime.now()` instead of smearing it.

crspeller (approved):
Nice.  
FYI @iomodo since you are working on this component.

webapp/src/components/rhs/rhs_checklist_list.tsx
```diff
@@ -212,6 +214,15 @@ const RHSChecklistList = (props: Props) => {
             <MainTitleBG numChecklists={checklists.length}>
                 <MainTitle>
                     {formatMessage({defaultMessage: 'Checklists'})}
+                    {
+                        overdueTasksNum > 0 &&
+                        <OverdueTasksToggle
+                            toggled={checklistItemsFilter.overdueOnly}
+                            onClick={() => selectOption('overdueOnly', !checklistItemsFilter.overdueOnly)}
+                        >
+                            {formatMessage({defaultMessage: '{num} tasks overdue'}, {num: overdueTasksNum})}
```
agarciamontoro, line 223: You can use the [plural format](https://formatjs.io/docs/core-concepts/icu-syntax#plural-format) here to cover the case where there is a single task overdue:
```suggestion
                            {formatMessage({defaultMessage: '{num} {num, plural, =1 {task} other {tasks}} overdue'}, {num: overdueTasksNum})}
```

agarciamontoro (changes requested):
Looks great! :tada:

A couple of suggestions below.

webapp/src/components/rhs/rhs_checklist_list.tsx
```diff
@@ -394,13 +405,40 @@ const StyledPrimaryButton = styled(PrimaryButton)`
...
+`;
+
 export default RHSChecklistList;
 
+const overdueTasks = (checklists: Checklist[]) => {
+    let count = 0;
+    const now = DateTime.now();
+    for (const list of checklists) {
+        for (const item of list.items) {
+            if (item.due_date > 0 && DateTime.fromMillis(item.due_date) <= now) {
```
agarciamontoro, line 429: The second suggestion, that I somehow forgot to add before:
Should we count only unchecked items here?
```suggestion
            if (item.state !== ChecklistItemState.Open && item.state !== ChecklistItemState.InProgress && item.due_date > 0 && DateTime.fromMillis(item.due_date) <= now) {
```

gbochora: Yes, we are not interested in finished tasks. We want to show number of tasks that are open/in progress and are overdue.

webapp/src/components/rhs/rhs_checklist_list.tsx
gbochora: Great catch! Also didn't know about plural support, thanks ;)

webapp/src/components/rhs/rhs_checklist_list.tsx
agarciamontoro: Ugh, sorry, my suggestion above was wrong, I should have used `===` instead of `!==`. I tested the PR and overdue *checked* tasks are still counted, so I think we still need to fix this, right?
This is what I meant before, sorry for the confusion:
```suggestion
            if ((item.state === ChecklistItemState.Open || item.state === ChecklistItemState.InProgress) && item.due_date > 0 && DateTime.fromMillis(item.due_date) <= now) {
```

agarciamontoro (approved):
Perfect, thank you! :tada:

webapp/src/components/rhs/rhs_checklist_list.tsx
gbochora: Oh, now I see what you meant. Nice catch! Fixed it.

abhijit-singh (approved):
Looks great @gbochora! Just one small nit - when just 1 task is overdue, can the text be `1 task overdue` instead of `1 tasks overdue`? Just have to remove the `s`.

PR #1098, opened 2022-03-23
MM-42700: Fix duplicated actions

agarciamontoro:
#### Summary
When saving an action in the client, the `saveChannelAction` function decides whether to send a `PUT` or a `POST` request depending on the ID of the action. If it exists, it means that the action is already in the server, and we need to update it, so it sends a `PUT`. If it is not, the action does not exist in the server yet, so it sends a `POST`.

Not having the ID set in the action, though, means that we can end up with duplicated actions every time we save. This PR makes sure that right after the action is saved, the ID is set, so we don't send POST requests with that action anymore.

It also prevents this in the API level in the server, making sure that only a single action with a specific pair (action type, trigger type) exists in the database. These changes in the API made me change the action tests, which were creating multiple actions of the same type :)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42700

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


server/api_actions_test.go
```diff
@@ -32,6 +32,32 @@ func TestActionCreation(t *testing.T) {
...
+		// Verify that the API succeeds
+		assert.NoError(t, err)
+		assert.NotEmpty(t, actionID)
+
+		// Try to create the same action again
+		_, err = e.PlaybooksClient.Actions.Create(context.Background(), e.BasicPublicChannel.Id, action)
+
+		// Verify that the API fails with a 500 error
+		requireErrorWithStatusCode(t, err, http.StatusInternalServerError)
+	})
```
agarciamontoro, line 59: This is the only new test. The other changes in this file are just adapting the tests so that they don't rely on creating multiple actions of the same type.

agarciamontoro:
@crspeller, server tests are failing in the `TestTestFramework`. Any ideas about what could be happening?

server/api_actions_test.go
```diff
@@ -237,8 +279,11 @@ func TestActionUpdate(t *testing.T) {
 	}
 
 	t.Run("valid update", func(t *testing.T) {
+		e := Setup(t)
```
crspeller, line 282: I wonder if this change is the cause of the too many clients failure.

crspeller (approved):
I have not seen this error before. Maybe somewhere we are leaking connections? Not against a temporary solution of increasing the max connections but we should probably have a look to see where we are leaking connections.

gbochora (approved):
Nice!

agarciamontoro:
Some unrelated flaky tests are failing, and I don't seem to have the permissions to click the Rerun workflow from failed button. @Phrynobatrachus, do you have that power?

agarciamontoro:
Thanks, @Phrynobatrachus :) Merging now :tada:

PR #1101, opened 2022-03-24
MM-42437: Create a new channel for the do-not-broadcast test

agarciamontoro:
#### Summary
This is a fix for a test-bug introduced in #1088. The test added there relies on the channel being completely empty, so we need to create a brand new channel instead of relying on one that was used by other tests before.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42437

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


Phrynobatrachus:
I'm going to merge this for CI, but tagging @iomodo here in case you still wanted to look.

PR #1102, opened 2022-03-24
e2e: update and add more action tests

Phrynobatrachus:
I revised the setup slightly so that each action test happens in its own channel, and added a couple new tests.

### DIFF
--- a/tests-e2e/cypress/integration/channels/general_actions_spec.js
+++ b/tests-e2e/cypress/integration/channels/general_actions_spec.js
@@ -134,9 +134,10 @@ describe('channels > general actions', () => {
                 cy.findByRole('button', {name: /save/i}).click();
 
                 // # Post the trigger phrase
+                cy.uiPostMessageQuickly('error detected red alert!');
+
                 // * Verify that the bot posts the expected prompt
                 // # Open the playbook run modal
-                cy.uiPostMessageQuickly('error detected red alert!');
                 cy.getLastPostId().then((postId) => {
                     cy.get(`#post_${postId}`).within(() => {
                         cy.contains('trigger for the Public Playbook').should('exist');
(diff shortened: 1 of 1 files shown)


tests-e2e/cypress/integration/channels/general_actions_spec.js
```diff
@@ -33,33 +34,128 @@ describe('channels > general actions', () => {
...
+                title: 'Public Playbook',
+                memberIDs: [],
+                retrospectiveTemplate: 'Retro template text',
+                retrospectiveReminderIntervalSeconds: 60 * 60 * 24 * 7 // 7 days
+            });
 
-        // * Verify the welcome message is shown
-        cy.verifyEphemeralMessage('test ephemeral welcome message');
+            // # Use the basic user to create their own channel,
+            // # so they only see one playbook in the list
```
agarciamontoro, line 111: I don't understand this comment. Why does creating a new channel make the list to contain a single playbook?

agarciamontoro (approved):
Neat! I left some non-blocking comments below, but this looks fantastic, thank you!

tests-e2e/cypress/integration/channels/general_actions_spec.js
Phrynobatrachus: Since the user created by `apiInitSetup` is a sysadmin, they potentially have access to other teams and playbooks created by other tests that run prior to this spec file, which means there could be a dozen playbooks in the action dropdown already. The user created by `apiCreateUser` is not a sysadmin, so I don't need to explicitly handle that in the dropdown selection logic when logged in with them.

agarciamontoro: Ah, I see, that makes sense! Thanks for the context :)

PR #1103, opened 2022-03-25
[MM-42080] Implement checklist item new design

iomodo:
#### Summary
This PR implements new design of the checklist item.

https://user-images.githubusercontent.com/2340127/160129592-a89e01fa-9356-405c-bf55-5185a02d668b.mp4

https://user-images.githubusercontent.com/2340127/160129601-bbf3a9ac-0ac2-4559-ae47-f71676b4fe99.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-42080

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/checklist_item/command.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+    const commandButton = (
+        <CommandText
+            onClick={() => {
+                setWasOpened(true);
+                setCommandOpen(true);
+            }}
+        >
+            <TextWithTooltipWhenEllipsis
+                id={props.checklistNum.toString(10)}
```
crspeller, line 95:
```suggestion
                id={`checklist-command-button-tooltip-${props.checklistNum}`}
```

webapp/src/components/checklist_item/command.tsx
```diff
@@ -0,0 +1,274 @@
...
+    const {formatMessage} = useIntl();
+    const commandRef = useRef(null);
+    const [running, setRunning] = useState(false);
+    const [lastRun, setLastRun] = useState(props.command_last_run);
+    const dispatch = useDispatch();
+
+    const [commandOpen, setCommandOpen] = useState(false);
+    const [wasOpened, setWasOpened] = useState(false);
+
+    // Immediately stop the running indicator when we get notified of a more recent execution.
```
crspeller, line 40: How does this work? Is this actually effective? 

Why do we need this? The whole running spinner thing is a fake anyway.

webapp/src/components/checklist_item/description.tsx
```diff
@@ -0,0 +1,148 @@
...
+    showDescription: boolean;
+}
+
+const ChecklistItemDescription = (props: DescriptionProps) => {
+    const {formatMessage} = useIntl();
+    const placeholder = formatMessage({defaultMessage: 'Add a description(optional)'});
+
+    const channelNamesMap = useSelector<GlobalState, ChannelNamesMap>(getChannelsNameMapInCurrentTeam);
+    const team = useSelector<GlobalState, Team>(getCurrentTeam);
+    const [editedValue, setEditedValue] = useState(props.value);
```
crspeller, line 32: Why do we need this state? It appears like our parent handles this for us.

crspeller (changes requested):
Nice!  A few comments.

abhijit-singh (changes requested):
Thanks @iomodo. A few points after playing around with it for a bit:
1. In the RHS, the icons on the right that appear on-hover should be in a container overlayed on top of the content and not inline. 
image

2. In case the task title takes multiple lines, the checkbox should still be aligned to the top. 

3. In the view-state, the description should only be shown if it has been added by the user. We don't need to show the placeholder.

4. The user should be able to click-to-expand and collapse a task. 

5. In the edit state, the title text should wrap.

6. The add assignee popup should show channel members as well as non-channel members, as it does today. It can have the same dimensions and details shown as in the current implementation.

7. The add assignee and slash command popup overlap if clicked on one after the other. The first one should close as soon as the user clicks outside. 

image
8. We should not hide the assignee name in the edit mode. 
image

webapp/src/components/checklist_item/assign_to.tsx
```diff
@@ -0,0 +1,214 @@
...
+    user: UserProfile;
+}
+
+interface AssignedToProps {
+    playbookRunId: string;
+    checklistNum: number;
+    itemNum: number;
+    assignee_id: string;
+    editable: boolean;
+    withoutName?: boolean;
```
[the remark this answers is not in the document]
iomodo: When rendered in the checklist_item(not in the hover menu) `<AssignTo>` might be just  a ProfilePic(without name) or a pic with the name, depending on the real estate available.
If there is command to be rendered we render this component without name. So this prop is needed.

webapp/src/components/checklist_item/command.tsx
iomodo: It's hacky but it works because 
```js 
const [lastRun, setLastRun] = useState(props.command_last_run);
``` 
is not assigning data synchronously.

Idea is to remove spinner when playbook run is updated. In a sense spinner it's not that fake :)
We could converge to just removing the spinner in a second every time.



Code was introduced here:
https://github.com/mattermost/mattermost-plugin-playbooks/commit/2d2884ee1473243589206f14c556868b6a77adcc#diff-5f1e752b29672a0bc877707894b8bb1f5cdbe22b260c604b73ba19aad58023c3R161

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -0,0 +1,403 @@
...
+        cursor: pointer;
+        margin-right: 8px;
+        margin-top: 2px;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        width: 16px;
+        min-width: 16px;
+        height: 16px;
+        border: 1px solid rgba(var(--center-channel-color-rgb),0.24);
```
gbochora, line 308: Nit: space
```suggestion
        border: 1px solid rgba(var(--center-channel-color-rgb), 0.24);
```

webapp/src/components/checklist_item/description.tsx
```diff
@@ -0,0 +1,142 @@
...
+        singleline: true,
+        mentionHighlight: false,
+        atMentions: true,
+        team,
+        channelNamesMap,
+    };
+
+    const computeHeight = (e: React.FocusEvent<HTMLTextAreaElement>) => {
+        e.target.style.height = '5px';
+        e.target.style.height = (e.target.scrollHeight) + 'px';
```
gbochora, line 43: Do we need this line: `e.target.style.height = '5px';`? 
new value is assigned on the next line.

webapp/src/hooks/general.ts
```diff
@@ -581,3 +581,14 @@ export const usePrevious = (value: any) => {
...
+// Create a portal to render while dragging
+export const usePortal = (parent: HTMLElement) => {
+    const [portal] = useState(document.createElement('div'));
+
+    useEffect(() => {
+        parent.appendChild(portal);
+    }, [parent, portal]);
+
+    return portal;
+};
```
gbochora, line 594: Nit: new line at end of file.

webapp/src/components/checklist_item/inputs.tsx
```diff
@@ -0,0 +1,115 @@
...
+        <CancelButton
+            onClick={props.onCancel}
+        >
+            {'Cancel'}
+        </CancelButton>
+        <SaveButton
+            onClick={props.onSave}
+        >
+            {'Save'}
+        </SaveButton>
```
gbochora, line 92: Do we need i18n for `Cancel` and 'Save'?

gbochora (approved):
Great work! Just a few nits.

abhijit-singh (commented):
Thank you for addressing the comments @iomodo! It seems one of them is still persisting:

5. In the edit state, the title text should wrap.

There are a few other minor requests but they're non-blocking and can be done in a follow-up.

PR #1104, opened 2022-03-28
MM-42845: Move the bar from the header to the modal body

agarciamontoro:
#### Summary
I could not reproduce the bug locally, so let's see if moving it from right after the header to right before the modal body (which ends up being the exact same spot) works.
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42845

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


agarciamontoro:
Do cloud test servers work here? Let's see :eyes:

agarciamontoro:
No, it doesn't, I always forget we cannot update plugins there :man_facepalming:

agarciamontoro:
@Phrynobatrachus, is there an easy way to test this PR in a test server before merging? It works locally, but the bug seems to only happen on community and test servers, so I'd like to check there first.

Phrynobatrachus (approved):

calebroseland (approved):

PR #1106, opened 2022-03-29
MM-42854: Use div instead of span to fix a bug in Safari

agarciamontoro:
#### Summary
Webkit-based browsers seem not to apply `flex-direction: column` when the children are `span`, so this PR changes the trigger header elements to `div` instead to fix the bug. Firefox and Chrome render both implementations in the same way.

I was only able to test this using Midori, which renders the whole webapp in a terrible way, but I was able to reproduce the bug and to confirm that this solution fixes it. However, I still need someone to take a look in Safari.

Screenshot of the solution in Midori:

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42854

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/actions_modal_trigger.tsx
+++ b/webapp/src/components/actions_modal_trigger.tsx
@@ -121,12 +121,12 @@ const Legend = styled.legend`
     margin: 0;
 `;
 
-const Label = styled.span`
+const Label = styled.div`
     font-size: 11px;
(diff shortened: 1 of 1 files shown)


lieut-data:
Looks good here! 
CleanShot 2022-03-29 at 14 41 07@2x

iomodo (commented):
Great!

PR #1109, opened 2022-03-29
MM-42728: fix every other chart label

lieut-data:
#### Summary
The labelling of the x-axis intentionally omits every other tick so as not to clutter the UI, but this unintentionally results in that data not displaying on hover. Reference the original labels prop when rendering instead.

Before:
CleanShot 2022-03-29 at 17 42 46

After:
CleanShot 2022-03-29 at 17 48 02

I haven't written a unit test for this, since it involves hovering over a graph and I don't know how to do that.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-42728

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbooks/bar_graph.tsx
+++ b/webapp/src/components/backstage/playbooks/bar_graph.tsx
@@ -72,9 +72,15 @@ const BarGraph = (props: BarGraphProps) => {
                     tooltips: {
                         callbacks: {
                             title(tooltipItems: any) {
-                                if (props.tooltipTitleCallback) {
-                                    return props.tooltipTitleCallback(tooltipItems[0].xLabel);
+                                if (props.labels) {
+                                    const label = props.labels[tooltipItems[0].index];
+                                    if (props.tooltipTitleCallback) {
+                                        return props.tooltipTitleCallback(label);
+                                    }
+
+                                    return label;
                                 }
+
                                 return tooltipItems[0].xLabel;
                             },
                             label(tooltipItem: any) {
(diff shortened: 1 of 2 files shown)


agarciamontoro (approved):
Looks good!

> I haven't written a unit test for this, since it involves hovering over a graph and I don't know how to do that.

The only thing I can think of is to write an E2E test, but even then [hovering is not trivial](https://docs.cypress.io/api/commands/hover). I'm ok merging this without a test, though.

webapp/src/components/backstage/playbooks/bar_graph.tsx
```diff
@@ -72,9 +72,15 @@ const BarGraph = (props: BarGraphProps) => {
                     tooltips: {
                         callbacks: {
                             title(tooltipItems: any) {
-                                if (props.tooltipTitleCallback) {
-                                    return props.tooltipTitleCallback(tooltipItems[0].xLabel);
+                                if (props.labels) {
+                                    const label = props.labels[tooltipItems[0].index];
```
crspeller, line 76: Is `tooltipItems` guaranteed to have at least one element?

crspeller (approved):

webapp/src/components/backstage/playbooks/bar_graph.tsx
lieut-data: In my inspection, it's always exactly one element. I suspect it's a consequence of how we've configured this chart, always configuring just one tooltip and getting that later in this callback.

PR #1113, opened 2022-03-30
MM-42509: Change initial fetch to get only in progress runs

agarciamontoro:
#### Summary
In community, in my case, the first request to retrieve runs gets 86 of them, of which only 6 are in progress. This request happens every time the team changes, to populate the Redux store with all the team runs. However, I suspect`*` we only need the runs in progress, which are the ones where we need to automatically open the RHS on channel switch.

This PR changes that initial request to get only in progress runs.

`*` I say suspect because the logic there is a bit complex, and I'm not 100% sure of all the implications, but I'm confident enough to at least try this in community for a while and see how it behaves. I tested it thoroughly locally, but only time, in community, will tell.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42509

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/channels/rhs_spec.js
+++ b/tests-e2e/cypress/integration/channels/rhs_spec.js
@@ -370,6 +370,35 @@ describe('channels > rhs', () => {
                 });
             });
         });
+
+        it.only('when navigating directly to a finished playbook run channel and clicking on the button', () => {
+            // # Run the playbook
+            const now = Date.now();
+            const playbookRunName = 'Playbook Run (' + now + ')';
+            const playbookRunChannelName = 'playbook-run-' + now;
+            cy.apiRunPlaybook({
+                teamId: testTeam.id,
+                playbookId: testPlaybook.id,
+                playbookRunName,
+                ownerUserId: testUser.id,
+            }).then((playbookRun) => {
+                // # End the playbook run
+                cy.apiFinishRun(playbookRun.id);
+            });
+
+            // # Navigate directly to the application and the playbook run channel
+            cy.visit(`/${testTeam.name}/channels/${playbookRunChannelName}`);
+
+            // # Click the icon
+            cy.get('#channel-header').within(() => {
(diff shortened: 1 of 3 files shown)


lieut-data (approved):
Makes sense to me! Thanks, @agarciamontoro :)

(I think the existing opener e2e tests cover this case?)

agarciamontoro:
> I think the existing opener e2e tests cover this case?

Yup, they do :)

crspeller (approved):
I think this is fine as long as we don't want to open the RHS for finished runs.

agarciamontoro:
> I think this is fine as long as we don't want to open the RHS for finished runs.

Yes, I've checked and finished runs no longer show the run RHS when clicking on `Toggle run details`, but the generic "This channel is not running any playbook" RHS (the one in the screenshot). This happens only after reloading, though, when you manually finish the run (clicking on the `Finish run` button), you still see the RHS with the finished state.
image

@itao, what do you think about this?

If we want to fix it (I'm not sure it's bad), we could check if the current channel has a finished run when clicking on the `Toggle run details` button).

lieut-data:
> Yes, I've checked and finished runs no longer show the run RHS when clicking on Toggle run details

Oh, I missed this when thinking through the problem. We definitely don't want to hide the ability to see a finished run's RHS state -- at least the way the product works today, that means the checklist data is entirely unavailable.

Can we still support querying this data on demand when opening the RHS?

agarciamontoro:
> at least the way the product works today, that means the checklist data is entirely unavailable.

Hm, that's true, I hadn't thought about that.

> Can we still support querying this data on demand when opening the RHS?

Yup, that'd be a good solution. Let me try it.

agarciamontoro:
Sorry for the double review, but can you guys take another look to the new logic and test? Thank you!

crspeller (approved):

lieut-data (approved):
Nice work!

PR #1117, opened 2022-04-04
[GH-19356] MM-41063, MM-40565: Increase new task dialog description length

ryoarmanda:
#### Summary

Increases the maximum length of the Add New Task dialog to match with the default character limit in webapp, which is 4000.

Also changes the description input type to `textarea` to support easier entry for long texts.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19356

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -24,6 +24,8 @@ import (
 	pluginapi "github.com/mattermost/mattermost-plugin-api"
 )
 
+const checklistItemDescriptionCharLimit = 4000
+
 const (
 	// PlaybookRunCreatedWSEvent is for playbook run creation.
 	PlaybookRunCreatedWSEvent = "playbook_run_created"
@@ -584,9 +586,10 @@ func (s *PlaybookRunServiceImpl) OpenAddChecklistItemDialog(triggerID, playbookR
 			{
 				DisplayName: "Description",
 				Name:        DialogFieldItemDescriptionKey,
-				Type:        "text",
+				Type:        "textarea",
 				Default:     "",
 				Optional:    true,
+				MaxLength:   checklistItemDescriptionCharLimit,
 			},
 		},
 		SubmitLabel:    "Add task",


server/app/playbook_run_service.go
```diff
@@ -584,9 +586,10 @@ func (s *PlaybookRunServiceImpl) OpenAddChecklistItemDialog(triggerID, playbookR
 			{
 				DisplayName: "Description",
 				Name:        DialogFieldItemDescriptionKey,
-				Type:        "text",
+				Type:        "textarea",
 				Default:     "",
 				Optional:    true,
+				MaxLength:   checklistItemDescriptionCharLimit,
```
agarciamontoro, line 592: As far as I'm aware, this is already handled by the interactive dialog logic here: https://github.com/mattermost/mattermost-webapp/blob/master/components/interactive_dialog/dialog_element/dialog_element.tsx#L22 I'd be fine removing it from here and letting the default `maxlength` apply

agarciamontoro (approved):
Thanks, @ryoarmanda, I tested it and it looks great! I think we can remove the `MaxLength`, but I'm ok either way, honestly.

server/app/playbook_run_service.go
ryoarmanda: Yeap, found it during my read through as well! However, it does not exactly match with the default character limit in the Edit Task flow, specified in the MarkdownTextbox component: https://github.com/mattermost/mattermost-plugin-playbooks/blob/aa888726e55f4cb26024f90d84630e5f7e7532a5/webapp/src/components/markdown_textbox.tsx#L17 

I took it from the issue that it's best for the limits on Add Task and Edit Task to exactly match, so I explicitly set `maxlength` to be equal to above. If I let the default `maxlength`, the issue can still happen due to the mismatch, just on a farther cutoff point.

calebroseland (approved):
LGTM!

PR #1122, opened 2022-04-05
[MM-42080] Support item addition

iomodo:
#### Summary
This PR adds support for item addition.

https://user-images.githubusercontent.com/2340127/161836476-cba470cc-47b6-40d6-8689-f17da680d650.mp4

#### Ticket Link
  Fixes: https://github.com/mattermost/mattermost-server/issues/XXXXX

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


abhijit-singh (changes requested):
This is great @iomodo! A few minor requests:
1. Can we make the font-weight for `Add a task` normal or 400? It doesn't have to be bold.
2. In the hover state for `Add a task`, can we change the background color to --button-bg with 8% opacity?
3. When a new task is added, the description area seems to have some extra space and then shrinks on focus. The extra space should never be there. 

4. When switching to the description field by pressing `tab`, it seems to have a border slightly visible. It will be ideal if we can hide it.

webapp/src/components/backstage/playbooks/playbook_preview_checklists.tsx
```diff
@@ -67,6 +67,9 @@ const PlaybookPreviewChecklists = (props: Props) => {
                                     dragging={false}
                                     disabled={true}
                                     collapsibleDescription={false}
+                                    newItem={false}
+                                    // eslint-disable-next-line @typescript-eslint/no-empty-function
+                                    cancelAddingItem={() => {}}
```
crspeller, line 72: Maybe we can just make this parameter optional instead of having to pass an empty function here?

crspeller (changes requested):
Great! Just a couple nits.

crspeller (approved):
Great!

webapp/src/components/rhs/rhs_checklist.tsx
```diff
@@ -169,7 +170,7 @@ const AddTaskLink = styled.button`
     color: var(--center-channel-color-64);
 
     &:hover {
-        background-color: rgba(var(--center-channel-color-rgb), 0.04);
+        background-color: rgba(var(--center-channel-color-rgb), 0.08);
```
abhijit-singh, line 173: This should be `background-color: var(--button-bg-08);`

abhijit-singh (approved):
Just added one small comment. Else looks great!

crspeller (approved):

webapp/src/components/checklist_item/assign_to.tsx
```diff
@@ -192,3 +198,7 @@ const ControlComponentAnchor = styled.a`
     position: relative;
     top: -4px;
 `;
+
+export const DropdownArrow = styled.i`
+    color: var(--center-channel-color-32);
+`;
```
gbochora, line 204: Nit: newline.

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -140,48 +158,70 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
-                    onSelectedChange={onDueDateChange}
-                />}
-            </Row>
+            {(assigneeID !== '' || command !== '' || props.checklistItem.due_date > 0 || isEditing) &&
+                <Row>
+                    {(assigneeID !== '' || isEditing) &&
+                        <AssignTo
+                            assignee_id={assigneeID || ''}
+                            editable={isEditing}
+                            withoutName={command !== '' && !isEditing}
```
gbochora, line 167: Do we hide the `Assign` button name when the second button in the row is the `Due date` button?
If yes, then here we need `|| props.checklistItem.due_date > 0` condition. It seems you removed it during the merge.

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -140,48 +158,70 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
                         setIsEditing(false);
-                        clientEditChecklistItem(props.playbookRunId, props.checklistNum, props.itemNum, {
-                            title: titleValue,
-                            command: props.checklistItem.command,
-                            description: descValue,
-                        });
+                        if (props.newItem) {
+                            if (props.cancelAddingItem) {
+                                props.cancelAddingItem();
+                            }
```
gbochora, line 208: Same is here.

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -140,48 +158,70 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
-                }
-                {(props.checklistItem.due_date > 0 || isEditing) &&
-                <DueDateButton
-                    editable={isEditing}
-                    date={props.checklistItem.due_date}
-                    mode={Mode.DateTimeValue}
-                    onSelectedChange={onDueDateChange}
-                />}
-            </Row>
+            {(assigneeID !== '' || command !== '' || props.checklistItem.due_date > 0 || isEditing) &&
```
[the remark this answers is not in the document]
iomodo: Empty `Row` is rendered and it messes up the margins

webapp/src/components/checklist_item/checklist_item.tsx
iomodo: Automatic merge failure

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -140,48 +158,70 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
+                            if (props.cancelAddingItem) {
+                                props.cancelAddingItem();
+                            }
+                            clientAddChecklistItem(props.playbookRunId, props.checklistNum, {
+                                title: titleValue,
+                                command,
+                                description: descValue,
+                                state: ChecklistItemState.Open,
+                                command_last_run: 0,
+                                due_date: 0,
```
gbochora, line 215: I guess due date needs a state like `assigneID`, which should be set here instead of 0, and also, we will need to update the state in the `onDueDateChange`.

iomodo: correct, already pushed the changes

webapp/src/components/checklist_item/duedate.tsx
```diff
@@ -363,11 +360,11 @@ const CalendarIcon = styled.div<{overdueOrDueSoon: boolean}>`
...
-    font-weight: 400;
-    font-size: 14.4px;
-    line-height: 14px;
-    &:margin-left: 4px;
-    &:margin-right: 4px;
+    font-size: 14px;
+    &{
+        margin-left: 4px;
+    }
+    color: var(--center-channel-color-32);
```
gbochora, line 367: if task is overdue or due soon, color should be `rgba(var(--center-channel-color-rgb)`, like calendar icon on the 358 line.

gbochora (changes requested):
I left some comments. 

Few more:
- Hover menu persists after exiting from the checklist item edit mode
https://user-images.githubusercontent.com/4368372/163035828-7783d992-b030-48d5-ba4f-32f28fe1deb7.mp4

- Checklist item(any) drag causes errors in the console when edit mode is enabled. 
image

gbochora (approved):
Nice!

abhijit-singh (approved):
Thanks for making the changes @iomodo! Looks so much better 🎉 

I noticed that a few things out of the Figma comments were not resolved. I've marked them in Figma and also listed them here. Approving this PR as they are minor and non-blocking. 

1. The task hover state background should be `center-channel-text` with 4% opacity and not 8%.
2. In the hover actions, the calendar icon button is larger than the rest. The class applied is icon-16 when it should be icon-12 (which makes font-size: 14.4px). The 3 dot icon also looks larger than the rest.
3. On hover, the icon colour should change from `center-channel-text 56% opacity` to `center-channel-text 72% opacity
4. In the edit state, colour of the icon in each chip should be the same — `center-channel-text 56% opacity`

PR #1127, opened 2022-04-07
undo summary capital letter in Run Summary

trilopin:
#### Summary
In playbook_edit, "Run Summary" copy is shown.
Replace it with "Run summary" and use correct capitalization.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42857

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


assets/i18n/en.json
```diff
@@ -67,6 +67,13 @@
     "id": "app.user.digest.tasks.heading",
     "translation": "Your assigned tasks"
   },
+  {
```
lieut-data, line 70: This change surprises me -- is there some non-determinism between platforms that explains why the tool "moved" this?

trilopin: Yeah, I noticed that, and also surprised me.

lieut-data: I'll defer to @calebroseland since I /think/ he has the most context here.

calebroseland (approved):

assets/i18n/en.json
calebroseland: (I'll follow up on this separately—will go ahead and merge)

PR #1129, opened 2022-04-14
[MM-42080] Support checklist addition

iomodo:
#### Summary
This PR supports checklist addition and is the last PR in the RHS checklist redesign(except fit and polish work left)

https://user-images.githubusercontent.com/2340127/163439389-03fbd1ca-85df-4fd8-b821-b0ae46ffea03.mp4

#### Ticket Link
  Fixes: https://github.com/mattermost/mattermost-server/issues/MM-42080

#### Checklist

- ~~[ ] Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -103,6 +103,7 @@ func NewPlaybookRunHandler(
 	checklistRouter.HandleFunc("/add-dialog", handler.addChecklistItemDialog).Methods(http.MethodPost)
 	checklistRouter.HandleFunc("/skip", handler.checklistSkip).Methods(http.MethodPut)
 	checklistRouter.HandleFunc("/restore", handler.checklistRestore).Methods(http.MethodPut)
+	checklistRouter.HandleFunc("/duplicate", handler.duplicateChecklist).Methods(http.MethodPost)
 
 	checklistItem := checklistRouter.PathPrefix("/item/{item:[0-9]+}").Subrouter()
 	checklistItem.HandleFunc("", handler.itemDelete).Methods(http.MethodDelete)
@@ -1301,6 +1302,24 @@ func (h *PlaybookRunHandler) removeChecklist(w http.ResponseWriter, r *http.Requ
 	userID := r.Header.Get("Mattermost-User-ID")
 
 	if err := h.playbookRunService.RemoveChecklist(id, userID, checklistNum); err != nil {
+		h.HandleError(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusCreated)
+}
+
+func (h *PlaybookRunHandler) duplicateChecklist(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	playbookRunID := vars["id"]
+	checklistNum, err := strconv.Atoi(vars["checklist"])
+	if err != nil {
+		h.HandleErrorWithCode(w, http.StatusBadRequest, "failed to parse checklist", err)
+		return
+	}
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	if err := h.playbookRunService.DuplicateChecklist(playbookRunID, userID, checklistNum); err != nil {
 		h.HandleError(w, err)
 		return
 	}
--- a/webapp/src/components/collapsible_checklist_hover_menu.tsx
+++ b/webapp/src/components/collapsible_checklist_hover_menu.tsx
@@ -9,6 +9,7 @@ import {DraggableProvidedDragHandleProps} from 'react-beautiful-dnd';
 import {
     clientSkipChecklist,
     clientRestoreChecklist,
+    clientDuplicateChecklist,
 } from 'src/client';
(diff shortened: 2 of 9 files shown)


abhijit-singh (changes requested):
Thanks @iomodo! A few minor change requests:

1.  Can we make the border-radius for the `Add a checklist` box 4px. This was incorrect in the design, updated it now. 
2. Can we make the border colour in the default state `center-channel-text` color with 16% opacity as in the design. 
3. On hover, the background colour should be `--button-bg-08`, just like `add task`.
4. When a new checklist is added, for the placeholder text `Add checklist name`, the font-weight should be 400 and font-style should be italic. 
5.  When a new checklist is added, for the `Cancel` and `Save` buttons, the font-size should be 12px (currently 14px) and the spacing between the two buttons and also between the input box and the cancel button should be 4px (currently 8px). 
6. The `Cancel` button should be styled as a [tertiary button](https://zeroheight.com/29be2c109/p/40e456-buttons/t/896a98). 
7. When a checklist is added, the border-radius for the checklist container should be 4 px. Same for collapsed checklists. (I just noticed that the bottom two corners don't have the radius).

Also wanted to call out that the keyboard shortcuts are working perfectly! ENTER to save, ESC to cancel, and Tab to change focus! Thank you! 🙌

abhijit-singh (changes requested):
Thanks @iomodo. A couple of issues are still there:

1. For this point:
>  Can we make the border colour in the default state center-channel-text color with 16% opacity as in the design.

I think the text-color has also changed because of this. It should be as it was earlier (-center-chanel-text with 64% opacity)
image

2. When a new checklist is added, for the Cancel and Save buttons, the font-size should be 12px (currently 14px) and the spacing between the two buttons and also between the input box and the cancel button should be 4px (currently 8px).

iomodo:
Thanks for your patience @abhijit-singh <3

https://user-images.githubusercontent.com/2340127/164032455-82373cb9-353b-43f4-b88e-9463c0d49df6.mp4

crspeller (approved):

webapp/src/components/collapsible_checklist_hover_menu.tsx
```diff
@@ -19,7 +20,6 @@ export interface Props {
     checklistIndex: number;
     checklistTitle: string;
     onRenameChecklist: () => void;
-    onDeleteChecklist: () => void;
```
trilopin, line 22: silly question, delete has been moved to another place or is not possible anymore?

server/api_runs_test.go
```diff
@@ -786,84 +785,84 @@ func TestChecklistManagement(t *testing.T) {
 		},
 		{
 			"Swap two checklists, moving the first one",
-			[]string{"0", "1"},
+			[]string{"1", "0"},
```
trilopin, line 788: I might misunderstand this but, why swap is not done after your change?

trilopin (commented):
Thanks, @iomodo! 

I couldn't follow the issue link from the PR description, please provide a valid one so I can better understand the motivation of the task.

Most of my comments are optional suggestions or questions.

server/api_runs_test.go
iomodo: Before checklists where added in the front of an array, now they are added at the end, hence the change

webapp/src/components/collapsible_checklist_hover_menu.tsx
iomodo: You can't remove a checklist(or a task) from the RHS anymore. You will be able to remove checklists(and tasks) while creating the Playbook of course.

server/api/playbook_runs.go
```diff
@@ -1308,6 +1309,24 @@ func (h *PlaybookRunHandler) removeChecklist(w http.ResponseWriter, r *http.Requ
 	w.WriteHeader(http.StatusCreated)
 }
 
+func (h *PlaybookRunHandler) duplicateChecklist(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	playbookRunID := vars["id"]
+	checklistNum, err := strconv.Atoi(vars["checklist"])
+	if err != nil {
```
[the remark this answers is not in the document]
iomodo: Yea this is not consistent. We check checklist params in the App layer here https://github.com/mattermost/mattermost-plugin-playbooks/blob/8a57e6e88ca6141fb80ae9f10e38474d04fbd7dd/server/app/playbook_run_service.go#L1956

iomodo:
Sorry for the confusion @trilopin. Task is to update the checklist design in the RHS. Here are the actual designs https://www.figma.com/file/DtVfHwxbUIXC6uSQXKgkb6/✍%EF%B8%8FBetter-Editor?node-id=699%3A213350

server/api/playbook_runs.go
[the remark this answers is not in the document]
trilopin: Thanks, I was more worried of not checking it than the place!

trilopin (approved):
Awesome!

PR #1135, opened 2022-04-15
Mm 43165 update run data model to include actions flags

gbochora:
#### Summary
This PR updates run data model to include boolean flags for actions. Flags will enable users to enable/disable actions mid-run.

Change-log:
- Added columns to `IR_Incident` table
  - `StatusUpdateBroadcastChannelsEnabled` - indicates that channels broadcast action is enabled on status update 
  - `StatusUpdateBroadcastFollowersEnabled` - indicates that followers broadcast action is enabled on status update 
  - `StatusUpdateBroadcastWebhooksEnabled` - indicates that webhooks action is enabled on status update 
- Add migration code that enables webhooks and channels broadcast actions only if corresponding list is not empty. Migration code also modifies archived runs so that they will have set flag correctly after restoration.
- Add copying webhooks and channels broadcast actions flag values from the playbook
- Broadcast to followers action is enabled by default
- Add action flag checks to the `CreatePlaybookRun`,  `UpdateStatus`, `FinishPlaybookRun`,  `RestorePlaybookRun` and `PublishRetrospective` methods. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43165

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated


server/api/playbook_runs.go
```diff
@@ -484,19 +484,18 @@ func (h *PlaybookRunHandler) createPlaybookRun(playbookRun app.PlaybookRun, user
 			playbookRun.DefaultOwnerID = pb.DefaultOwnerID
 		}
 
-		if pb.BroadcastEnabled {
-			playbookRun.BroadcastChannelIDs = pb.BroadcastChannelIDs
-		}
+		playbookRun.StatusUpdateBroadcastFollowersEnabled = true
+
+		playbookRun.StatusUpdateBroadcastChannelsEnabled = pb.BroadcastEnabled
```
trilopin, line 489: Optional comment, maybe out of scope of this pr.

Boolean flags have different names in playbook and run `BroadcastEnabled`vs `StatusUpdateBroadcastChannelsEnabled`.  would make sense to have a consistent naming? 

Are these actions only triggered when the status is updated? If so, the new one created for the run seems more relevant to me.

server/sqlstore/migrations.go
```diff
@@ -1947,6 +1947,55 @@ var migrations = []Migration{
...
+
+			// enable channels broadcast where channels ids list is not empty
+			channelsBroadcast := sqlStore.builder.
+				Update("IR_Incident").
+				Set("StatusUpdateBroadcastChannelsEnabled", true).
+				Where(sq.NotEq{"ConcatenatedBroadcastChannelIDs": ""})
+
+			if _, err := sqlStore.execBuilder(e, channelsBroadcast); err != nil {
+				return errors.Wrapf(err, "failed updating the StatusUpdateBroadcastChannelsEnabled field")
+			}
```
trilopin, line 1987: Don't we need an additional migration for the broadcast-followers boolean flag?

server/app/playbook_run_service_test.go
```diff
@@ -518,6 +590,161 @@ func TestUpdateStatus(t *testing.T) {
...
+		type webhookPayload struct {
+			app.PlaybookRun
+			ChannelURL   string                  `json:"channel_url"`
+			DetailsURL   string                  `json:"details_url"`
+			StatusUpdate app.StatusUpdateOptions `json:"status_update"`
+		}
+
+		webhookChan := make(chan webhookPayload)
+
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
trilopin, line 617: why do we need a server to test the service layer?

trilopin (commented):
Thanks @gbochora , I added some questions in some parts I don't fully understand.

server/api/playbook_runs.go
gbochora: Yes, only the status update event triggers these actions. Maybe the solution here is to change playbook field names.

trilopin: Thanks for the info. There's no need to do it in this PR. 

I'll push to improve that.

server/sqlstore/migrations.go
gbochora: Good question! The playbook doesn't have a flag indicating whether a broadcast to the followers is enabled. It's always enabled, so I set `StatusUpdateBroadcastFollowersEnabled` column default value to true.

server/app/playbook_run_service_test.go
gbochora: We need a server to test if the status update triggered a webhook call. When a status update is posted, the service layer calls the corresponding webhooks, and this server receives those calls.

server/app/playbook_run_service_test.go
```diff
@@ -518,6 +590,161 @@ func TestUpdateStatus(t *testing.T) {
...
+		err := s.UpdateStatus(playbookRun.ID, "user_id", statusUpdateOptions)
+		require.NoError(t, err)
+		require.False(t, broadcastChannel1)
+		require.False(t, broadcastChannel2)
+		require.False(t, broadcastFollower1)
+		require.False(t, broadcastFollower2)
+
+		select {
+		case <-webhookChan:
+			require.Fail(t, "should not receive webhook on status update")
```
agarciamontoro, line 743: Really really nice test!

agarciamontoro (approved):
Looks awesome! Thanks, Giorgi :)

server/app/playbook_run_service_test.go
trilopin: Thanks for the info!

trilopin (approved):
Well done!

PR #1137, opened 2022-04-16
MM-42374: Playbook Editor, layout overhaul and header content

calebroseland:
#### Summary
- header contents
-  a large reworking of the overall layout, some design pivots as discussed (all open to change, feedback, and refinement)
  - integrated title bar with simplified options (`Back and Run` in the corners)
  - Private label, members, auto-follow brought under the main heading (very open to feedback)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42374

#### Checklist

Item | Y / N / NA [^1]
---|---
Telemetry updated? | N
Gated by experimental feature flag? | Special route
Unit/E2E tests updated? | N


webapp/src/components/backstage/backstage.tsx
```diff
@@ -86,6 +89,9 @@ const BackstageBody = styled.div`
 
 const Backstage = () => {
     const {formatMessage} = useIntl();
+    const containerRef = useRef<HTMLDivElement>(null);
+
+    const {y} = useScroll(containerRef);
```
calebroseland, line 94: From [React-Use](https://github.com/streamich/react-use), let me know your thoughts—I think this would be a good addition, but can be removed if needed.

iomodo (approved):
Great work @calebroseland !
Approving to keep the ball rolling. As discussed, deferring changes to the next PRs.

crspeller (approved):
Looks good! 

From our discussion the `...attrs` pattern seems like overkill. Let's remove that. 

I understand the blank styled components now.

Going to approve now you can decide to address these now or in the follow up.

abhijit-singh (commented):
Thanks @calebroseland! This is coming together well! A few thoughts (I don't mind if they're addressed in a follow-up PR if this is blocking further development):

1. Having the `Auto-follow runs` button below the title isn't working too well in my opinion. I'd prefer something closer to this - 
image Happy to discuss this offline and iterate later though. 

2. Should we limit the max-width of the center content to something close to 960px? It opened up like this for me and having content span such a long width is not recommended. image

3. There seems to be an overlap of the sticky title and the description on-scroll. We could delay showing the sticky title row even more. image
4. I'm hoping we can fix the alignment issues in the [icon buttons](https://zeroheight.com/29be2c109/p/00b099-icon-buttons/t/67408e) and make the icon and the number vertically and horizontally center aligned.
 image

5. The link icon button should not be rounded and can be styled like our standard [`medium` icon-button](https://zeroheight.com/29be2c109/p/00b099-icon-buttons/t/792ae3). 
image

calebroseland:
(Follow-up PR in progress to address review comments and a number of other things—cleanup, simplification, etc.)

PR #1138, opened 2022-04-16
Fix private/public icon update issue

gbochora:
#### Summary
On the `Playbook preview` page, public/private icon did not update automatically when converting a playbook to private. 
A similar case was with the number of members, addin/removing a member did not update the number.
`Playbook access modal` and `Playbook preview` have their own copies of playbook object, so changes in the modal didn't reflect on the preview page. This PR fixes both issues by updating the preview page state from the modal.

Note: New Playbook editor has the same issues. I can't fix it now because it's under development.

https://user-images.githubusercontent.com/4368372/163686074-e9cd6a08-26cd-4c01-bc79-f41ad5e66f16.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41580

#### Checklist

- ~~[ ] Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_access_modal.tsx
```diff
@@ -88,6 +90,12 @@ const PlaybookAccessModal = ({
         }
     };
 
+    useEffect(() => {
```
crspeller, line 93: This can cause double renders. Maybe intercepting uses of updatePlaybook would be better?

crspeller (changes requested):

crspeller (approved):
Great!

iomodo (approved):

PR #1146, opened 2022-04-25
supporting e2e tests for webapp changes

lieut-data:
#### Summary
Supporting e2e changes for https://github.com/mattermost/mattermost-webapp/pull/10230.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-42755

#### Checklist
- [ ] ~Telemetry updated~
- [x] Unit tests updated


tests-e2e/cypress/integration/playbooks/overview_spec.js
```diff
@@ -78,16 +97,56 @@ describe('playbooks > overview', () => {
...
+                // # Open Playbooks
+                cy.get('[aria-label="Select to open product switch menu."]').click({force: true});
+                cy.get('a[href="/playbooks"]').click({force: true});
 
-        // # Click Run Playbook
-        cy.findByTestId('run-playbook').click({force: true});
+                // Click through to open the playbook
+                cy.findByTestId('playbooksLHSButton').click({force: true});
+                cy.get('[placeholder="Search for a playbook"]').type(testPublicPlaybook.title);
+                cy.findByTestId('playbook-title').click({force: true});
```
Phrynobatrachus, line 112: There's an issue with this logic in the context of the overall spec file. When the `it('should duplicate playbook')` test runs it creates `Copy of Public Playbook`, so the search filters down to 2 results instead of 1, and the `cy.findByTestId` fails since there's multiple valid selections. 

Could be another point in favor of isolating `it` block state entirely, see #1101 and #1102 for similar examples of gotchas.

Phrynobatrachus (changes requested):
There's an issue here,

agarciamontoro (approved):
Tests look good to me (modulo Cass' comment).

tests-e2e/cypress/integration/playbooks/overview_spec.js
lieut-data: Ooh, excellent catch @Phrynobatrachus! Serves me right for only running `describe.only` and then pushing :p -- pushed a fix to use a separate playbook for now.

Phrynobatrachus (approved):
LGTM, thanks!

(I assume cloud tests will pass on Monday's FF, 0/5 if you want to merge now or hold for those)

lieut-data:
Going to merge now, and I'll own anything that breaks next week :)

PR #1147, opened 2022-04-25
DX, Deps: fix console warnings, npm use exact version, better displayName support

calebroseland:
#### Summary

1) https://github.com/mattermost/mattermost-plugin-playbooks/commit/9fca2d10f878a0e5d3328b9a064fab72533c903f: fix key warning
```
Warning: Each child in a list should have a unique "key" prop.
```
2) https://github.com/mattermost/mattermost-plugin-playbooks/commit/e1657b79dea2aec5cc6af07748dab00bd05680df: fix modalHeaderText warning
```
Warning: Failed prop type: The prop `modalHeaderTeequired in `zu`, but its value is `undefined`.
```
3) https://github.com/mattermost/mattermost-plugin-playbooks/commit/a541f4cc179903d7f84ba67b804b0eea0cd9c8fb: save exact version when adding webapp dependencies
4) https://github.com/mattermost/mattermost-plugin-playbooks/commit/abe5c0e729a11a5478a2b1d9f0172ecd79c46a81:  transform `displayName` for anonymous functional components à la
```tsx
const ThingComponent = (props: Props) => <div/>;
```
CleanShot 2022-04-25 at 16 24 31@2x
CleanShot 2022-04-25 at 16 23 55@2x

#### Checklist

Item | Y / N / NA [^1]
---|---
Telemetry updated? | N
Gated by experimental feature flag? | N
Unit/E2E tests updated? | N


lieut-data (approved):
Sweet!

webapp/src/components/widgets/generic_modal.tsx
```diff
@@ -136,7 +136,9 @@ export default class GenericModal extends React.PureComponent<Props, State> {
                     className='GenericModal__header'
                     closeButton={true}
                 >
-                    <ModalHeading id={`${this.props.id}_heading`}>{this.props.modalHeaderText}</ModalHeading>
```
trilopin, line 139: I've always used extensively the ternary operator `?` instead of the logical one '`&&` because I remembered safer in some cases.

That was so internalized that I had to search why and found this https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx. Aren't we doing the case undefined && anything that the article refers to?

calebroseland: I actually referenced the same article [here](https://github.com/mattermost/mattermost-plugin-playbooks/pull/634/files/7639ded6dffaad62b675b6a9f7431d438afd5390#r673442336) in a very similar comment. 

I guess it came down to that `&&` is a bit more common in playbooks. I'll wrap it in a bool to make it more render-safe; I guess it might be a good discussion topic to bring up in the dev meeting.

PR #1148, opened 2022-04-26
update run status ask for confirmation if accidental cancel

trilopin:
#### Summary
If the user leaves without saving changes in update_run_status modal, a confirmation modal will be shown.
E2E tests added

This modal has a mix of visibility state management that makes it a bit harder to make changes: inside the component uses useState(showModal), and from the outside is controlled by modalProps.onHide().  That is the main reason why I had to replace some setShowModal with modalProps.onHide().

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-43513

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/modals/update_run_status_modal.tsx
+++ b/webapp/src/components/modals/update_run_status_modal.tsx
@@ -225,8 +225,6 @@ const UpdateRunStatusModal = ({
                 confirmButtonText={hasPermission ? formatMessage({defaultMessage: 'Post update'}) : formatMessage({defaultMessage: 'Ok'})}
                 handleCancel={() => true}
                 {...modalProps}
-
-                // handleCancel={onTentativeHide}
                 onHide={onTentativeHide}
                 handleConfirm={hasPermission ? onSubmit : null}
                 autoCloseOnConfirmButton={false}
--- a/webapp/src/components/modals/update_run_status_modal.tsx
+++ b/webapp/src/components/modals/update_run_status_modal.tsx
@@ -13,6 +13,8 @@ import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
 import {getChannel} from 'mattermost-redux/selectors/entities/channels';
 
 import GenericModal, {Description, Label} from 'src/components/widgets/generic_modal';
+import UnsavedChangesModal from 'src/components/widgets/unsaved_changes_modal';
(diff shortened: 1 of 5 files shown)


trilopin:
There is a failing cypress test where I would love to have some of your help @Phrynobatrachus

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -102,20 +104,35 @@ const UpdateRunStatusModal = ({
...
 
+    const onTentativeHide = () => {
+        if (providedMessage === message || message === defaultMessage || message === '') {
+            onActualHide();
+        } else {
+            setShowUnsaved(true);
+        }
+    };
+
+    const onActualHide = () => {
```
iomodo, line 115: Shouldn't we hide the modal here like `setShowModal(false)` ?

iomodo, line 115: Shouldn't we hide the modal here like `setShowModal(false)` ?

iomodo (commented):
Looks great @trilopin !

Just one question below

webapp/src/components/modals/update_run_status_modal.tsx
trilopin: we have some challenges managing visibility state through two different places: parent through modalProps and the component itself through setShowModal useState hook. I feel that we should have one only source of truth.

If I manage everything internally through setShowModal, it gets to a point where I can not post run-updates twice in a row because we mixed modal visibility.

iomodo:
Ahh makes sense, thanks for clarification @trilopin.

iomodo (approved):

trilopin:
@iomodo I'm not fully satisfied with the mixed state management. If you had any idea how to make this cleaner I would be super happy.

Phrynobatrachus (approved):
Sorry for the commit mess, made a couple silly errors and hit some annoying test flakes. :grimacing: 

In general, my impression has been that `findByRole` is more reliable than a plain `get`, particularly on these popover modals, so that's why I changed to it.

Fingers crossed this CI run should pass now. Manual exercise LGTM as well.

calebroseland (approved):

PR #1150, opened 2022-04-26
Add skipping daiy digest when there are only tasks due after today

gbochora:
#### Summary
This PR excludes summary info from daily digest message when there are only tasks due in the future and returns empty text. As a result, the daily digest message is skipped when we only have tasks due in the future.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43586

#### Checklist

- ~~[ ] Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -2829,16 +2829,21 @@ func buildAssignedTaskMessageSummary(runs []AssignedRun, locale string, timezone
 		}
 	}
 
-	// omit assigned tasks header if runs info is empty
-	if runsInfo.String() != "" {
-		if onlyDueUntilToday {
-			msg.WriteString(T("app.user.digest.tasks.num_assigned_due_until_today", total-tasksDoAfterToday))
-		} else {
-			msg.WriteString(T("app.user.digest.tasks.num_assigned", total))
-		}
-		msg.WriteString("\n\n")
-		msg.WriteString(runsInfo.String())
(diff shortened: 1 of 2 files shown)


server/app/playbook_run_service.go
```diff
@@ -2829,17 +2829,22 @@ func buildAssignedTaskMessageSummary(runs []AssignedRun, locale string, timezone
...
-		}
-		msg.WriteString("\n\n")
-		msg.WriteString(runsInfo.String())
+	// if there are only tasks that are due after today and we need to show only tasks due now, skip a message
+	if onlyDueUntilToday && tasksDoAfterToday == total {
+		return ""
 	}
 
+	// add title
+	if onlyDueUntilToday {
```
crspeller, line 2838: Logic is a bit hard to follow here. We seem to have gotten rid of the behavior where if the `runsInfo` is empty then we skip writing this stuff. Was that intentional?

crspeller (changes requested):

server/app/playbook_run_service.go
gbochora: Yep, it's pretty complicated. 
Instead of `runsInfo` is empty I've added `tasksDoAfterToday == total` condition, lines 2832-2835. 
The logic here is: If we need only tasks that are due until today and all tasks are due after today(`tasksDoAfterToday == total`), we return an empty message.

agarciamontoro (approved):
Looks good!

crspeller (approved):
Looks good!

PR #1151, opened 2022-04-26
[MM-42371] Add editable checklist to the backstage

iomodo:
#### Summary
This PR adds editable checklists to the backstage

https://user-images.githubusercontent.com/2340127/165367915-3edb515e-9301-4484-ae7a-395e9c53a42d.mp4

#### Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-42371

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


iomodo:
Merged with master, here's the new ui:

https://user-images.githubusercontent.com/2340127/166259548-f854eb2a-5e3d-4bf2-a497-82578b77aea4.mp4

calebroseland (changes requested):
Awesome work! Great seeing this in ~~backstage~~ _Playbooks_ 🤓

I'm a little confused about where you'd like `<ChecklistList playbook={props.playbook}/>` to be rendered—it's in 3 places atm. 

- `playbook_editor/checklists` (not referenced/imported)
- `playbook_editor/outline/section_checklists` (not referenced/imported)
- `playbook_editor/outline/outline` (is rendered here in outline root atm)

`section_checklists` matches the pattern the other sections are in, but wherever you think is best.

I'm not seeing any polish-level issues not mentioned elsewhere or that need to be done now (can be done in final polish later).

There's one piece of [existing CSS](https://github.com/mattermost/mattermost-plugin-playbooks/blob/MM-42371-checklist-full-edit-component/webapp/src/components/backstage/playbook_editor/playbook_editor.tsx#L327-L336) not kicking in, oddly causing:
CleanShot 2022-05-02 at 22 26 01

Not sure what is causing that; not a blocker though—that code will be cleaned up later this week.

iomodo:
Sorry for the confusion @calebroseland, forgot to remove 2 files.
Also moving `teamselection` in the `playbook_editor`

calebroseland (approved):

abhijit-singh (dismissed):
Awesome @iomodo! I have a few doubts and a couple change requests:
1. For the `assign to` option, I had it in the design for custom positions, but it probably doesn't make sense to assign it to specific people at the playbook level. Can we remove that until we introduce custom positions?
2. For the `due date` attribute, we should only be allowing relative times in there, similar to the `default update timer` field we have under status updates today. The default options  here can be `4 hours`, `1 day`, and `7 days` but these might change.
3. After reordering a checklist, it seems like the hover actions on the tasks stop appearing. I just checked and this is happening in the RHS as well.

4. There's seems to be some extra space between the task title and the attributes. It can be reduced by at least 4px. This also is there in the RHS as well.

iomodo:
@abhijit-singh thanks for the review
1. Removed (though imo it might have been valuable having pre-assigned people to the tasks)
2. Fixed
3. I believe this was introduced [here](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1161). I'll create a ticket.
4. Fixed

tests-e2e/cypress/integration/channels/rhs/checklist_spec.js
```diff
@@ -285,19 +285,21 @@ describe('channels > rhs > checklist', () => {
             });
         });
 
-        it('can set due date, from hover menu', () => {
+        it.only('can set due date, from hover menu', () => {
```
crspeller, line 288: Forgot to remove only.

crspeller (approved):
Good stuff! Just need to remove e2e only.

iomodo:
I'm pushing this. Let's defer UI changes(if there are any) to the next PR.
cc @abhijit-singh

PR #1153, opened 2022-04-27
MM-43132: Prior work for Run Actions modal

agarciamontoro:
#### Summary
This PR lays the groundwork for the run actions modal, so this is the first of two PRs for MM-43132.

You can review by commit, the commit messages should give enough context, but in short this PR
- Adds the redux boilerplate to accomodate both channel and run actions. For example, the redux actions were called simply `ShowActionsModal` until now, so the first commit changes that to `ShowChannelActionsModal` for being able to also add `ShowRunActionsModal`. This boilerplate is not used yet, that will come in the next PR.
- Refactors a couple of components: mainly the `ActionsModal`, which is now called `ChannelActionsModal`, but also the `PatternedTextArea`, which will be used in the `RunActionsModal` in the next PR.

This PR is not meant to change any logic, so I'm relying on the existing E2E tests for validating it. I also manually checked and everything seems to work as expected.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43132

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/backstage/playbook_edit/automation/webhook_setting.tsx
```diff
@@ -0,0 +1,42 @@
...
+import PatternedTextArea from 'src/components/patterned_text_area';
+
+interface Props {
+    enabled: boolean;
+    onToggle: () => void;
+    textOnToggle: string;
+    placeholderText: string;
+    errorText: string;
+    input: string;
+    pattern: string;
```
trilopin, line 17: this could improve a bit by passing a regexp obj instead of the string:
- provide a regex better than the current one (which is currently flagging invalid URLs as valid)
- preserve perf in case regex is complex (compiled once via RegExp constructor in a module constant)

agarciamontoro:
> Why these modals are managed with their own state at the plugin store instead of using openModal like the status-update-modal?

@trilopin, that's a good question. To be honest, `openModal` is something we imported relatively recently from webapp and I had completely forgotten it existed, so I just added the state "manually" to the store. I don't know which strategy I like better, though. Thoughts on keeping this as it is today and tackle this holistically later?

webapp/src/components/backstage/playbook_edit/automation/webhook_setting.tsx
agarciamontoro: https://mattermost.atlassian.net/browse/MM-43888

agarciamontoro: By the way, you said that invalid URLs are flagged as valid. Could you give an example?

trilopin: For example 

https://fake
https://something.faketld
https://something.com.morefake

trilopin: Anyway, I agree with doing it in a separate pr

trilopin:
It was more curiosity and topic for guild than this pr.

trilopin:
It was more curiosity and topic for guild than this pr.

gbochora (commented):
Great work @agarciamontoro!
A few issues(some of them may have existed before):
- When a user tries to save actions with invalid data, it causes an error in the backend, and the user is not notified about it. So it seems that everything was saved. Only after page refresh, we can see that actions were not saved.

https://user-images.githubusercontent.com/4368372/166209101-b3dfec33-5afe-45b7-99cd-62e0623dfb8b.mp4


- Can't clear keywords input, on save the backend logs an error: `public error message: invalid action; internal details: payload field 'keywords' must contain at least one keyword`

https://user-images.githubusercontent.com/4368372/166209752-987b7395-1d39-49e4-a9e9-e758dada6000.mp4

- Actions dialog shows configs for the channel that I visited before.

https://user-images.githubusercontent.com/4368372/166211272-71e65c1d-e7e7-4c5a-94c8-ec9587c1a3bb.mp4

- When the `category` action is disabled, saving actions settings causes an error in the backend:
`public error message: invalid action; internal details: payload field 'category_name' must be non-empty`

trilopin:
@agarciamontoro will you solve the issues in this PR or in the Run actions one?

agarciamontoro:
@trilopin, I'm solving them here so the foundations are good.

trilopin:
> @trilopin, I'm solving them here so the foundations are good.

pls ping me when you send the fixes to do the final review

agarciamontoro:
@trilopin, here's the ping you requested :)

server/api_actions_test.go
```diff
@@ -12,10 +12,10 @@ import (
 )
 
 func TestActionCreation(t *testing.T) {
-	e := Setup(t)
-	e.CreateBasic()
-
 	t.Run("create valid action", func(t *testing.T) {
+		e := Setup(t)
+		e.CreateBasic()
```
trilopin, line 17: Why did you need to set a server per subtest?

server/api_actions_test.go
```diff
@@ -289,6 +308,47 @@ func TestActionUpdate(t *testing.T) {
...
+		// Verify that the API succeeds
+		assert.NoError(t, err)
+		assert.NotEmpty(t, actionID)
+
+		// Remove the keywords from the payload in the action
+		payload.Keywords = []string{}
+		newAction.Payload = payload
+
+		// Make the Update request with the new action
+		err = e.PlaybooksClient.Actions.Update(context.Background(), newAction)
```
trilopin, line 346: it would be great if we additionally check that the action is successfully updated. Could we get that and assert it?

tests-e2e/cypress/integration/channels/general_actions_spec.js
```diff
@@ -298,4 +298,45 @@ describe('channels > general actions', () => {
...
+            // # Enable the categorization action and set the name
+            const categoryName = 'example category ' + Date.now();
+            cy.contains('sidebar category').click();
+            cy.contains('Enter category name').click().type(categoryName + '{enter}');
+
+            // # Save action
+            cy.findByRole('button', {name: /save/i}).click();
+
+            // # Switch to the additional channel
+            cy.get('#sidebarItem_' + channel.name).click();
```
trilopin, line 327: oh, nice shortcut!

server/api_actions_test.go
agarciamontoro: I explained that in the commit:
There can be only one action of each type, so we need to create a whole new test environment for each unit test. That's why this commit copies the Setup and CreateBasic functions to each individual test. Otherwise, some of the tests fail because a previous one already created an action of the same type.

server/api_actions_test.go
agarciamontoro: Great suggestion, I'll do it!

server/api_actions_test.go
trilopin: Sorry @agarciamontoro, I jumped directly into the code.

Wouldn't it be better if we create a fresh channel without actions than a completely new server with all new resources?

trilopin (approved):
Some optional comments, questions and suggestions. Code and feature seems fine to me. 🎉

gbochora:
@agarciamontoro I can still reproduce the issue with the action dialog showing an incorrect state.

https://user-images.githubusercontent.com/4368372/166893529-566f4b5d-8cdd-460f-b010-1c47a05dfb0b.mp4

server/api_actions_test.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1153/commits/1b0408fc9fb399cb9a4d41e204dccbb610562e8d

server/api_actions_test.go
agarciamontoro: You're perfectly right! This is much better. Done in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1153/commits/1b0408fc9fb399cb9a4d41e204dccbb610562e8d. Thanks!

agarciamontoro:
@gbochora, I'm trying to reproduce what you see but I'm unable. What OS/browser are you using? Could you give me some specific steps?

gbochora:
I'm using Chrome Version 100 / MacOS.  Can you try adding a small delay(1-2 sec) to the  `await fetchChannelActions(id);` call? I think it should help to reproduce the issue.

agarciamontoro:
@gbochora: Oh, I see. Let me reset to the empty state as soon as the channel switches. If the request takes too long, at least we have an empty state, and not the state from before.
Ideally, we should cache this somehow (in the store, maybe), but let me know if this fixes things in your end.

agarciamontoro:
@gbochora, @trilopin, @Phrynobatrachus: I think I fixed the issue. Long story short: the underlying `GenericModal` was incorrectly calling `onHide` right after `onSave`, which caused a race condition in the state management. There's a longer explanation in the commit message: https://github.com/mattermost/mattermost-plugin-playbooks/pull/1153/commits/34419e2a0a2d06320d42b7563935b06371c21c1c

The second commit, https://github.com/mattermost/mattermost-plugin-playbooks/pull/1153/commits/ea1db7b4a841b4690a45821ff6ce0c7ac9b8183c, is just some code hygiene. Without it, the solution still works. However, I found out that those functions were being regenerated on every render and I decided to wrap them on `useCallback`, apart from moving the constants outside of the component. As said, just improving code, but not changing the logic.

Let's see if CI is happy now, the test passed locally for me.

Phrynobatrachus (approved):

agarciamontoro:
Failing tests are expected, merging.

PR #1155, opened 2022-04-28
Mm 43130 status update modal broadcast info

trilopin:
#### Summary
Screenshot 2022-04-28 at 14 08 37

Brand new description in update run status modal.
This PR covers the new description we want to set in the modal for Run "Post update". The new description informs the user about how many channels and DM will be broadcasted with the update. A tooltip will be shown if the user does hover on the counters. Tooltips show up to 5 names and "and N others" in case we have more than 5.

If there are no channels or followers (DM), a different message will be shown in the description.

From the pure tech perspective, a new hook has been implemented to extract the names of the users from the userIDs. Thanks @calebroseland for your help :)

Additionally, the navigation is now done in the same tab (we force modal close if the user navigates). In a different PR, a mechanism to prevent user from losing unsaved changes will be implemented. https://github.com/mattermost/mattermost-plugin-playbooks/pull/1148 has to be merge first.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43130

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~
- [ ] E2E tests


webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -91,8 +103,16 @@ const UpdateRunStatusModal = ({
...
             return result;
-        }, [])?.join(', ');
-    });
+        }, []);
+    })?.slice(0, NAMES_ON_TOOLTIP) || [];
+    const followerNames = useFormattedUsernames(runMetadata?.followers?.slice(0, NAMES_ON_TOOLTIP));
+
+    const generateTooltipText = (names: string[], total: number) => {
+        return total ? formatMessage({defaultMessage: '{text}{rest, plural, =0 {} one { and other} other { and {rest} others}}'}, {
+            text: total > NAMES_ON_TOOLTIP ? names.join(', ') : formatList(names, {type: 'conjunction'}),
```
agarciamontoro, line 112: I didn't know `formatList`, that's cool.

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -133,33 +153,53 @@ const UpdateRunStatusModal = ({
         }
     };
 
+    const description = () => {
+        const broadcastChannelCount = run?.broadcast_channel_ids.length ?? 0;
```
agarciamontoro, line 157: We also need to check here whether the setting is enabled or not with the new `status_update_broadcast_channels_enabled`. So maybe something like
```suggestion
        const broadcastChannelCount = 0;
        if (run?.status_update_broadcast_channels_enabled) {
            broadcastChannelCount = run?.broadcast_channel_ids.length ?? 0;
        }
```

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -133,33 +153,53 @@ const UpdateRunStatusModal = ({
...
+
+        const OverviewLink = (...chunks: string[]) => (
+            <Link to={pluginUrl(`/runs/${playbookRunId}/overview`)}>
+                {chunks}
+            </Link>
+        );
+
+        if ((broadcastChannelCount + followersChannelCount) === 0) {
+            return formatMessage({
+                defaultMessage: 'This update will be saved to <OverviewLink>overview page</OverviewLink> and broadcasted to <OverviewLink>0 channels</OverviewLink>',
```
agarciamontoro, line 168: I'd lean towards removing the final part of the sentence altogether if we are not going to broadcast. Thoughts, @abhijit-singh?
```suggestion
                defaultMessage: 'This update will be saved to <OverviewLink>overview page</OverviewLink>',
```

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -133,33 +153,53 @@ const UpdateRunStatusModal = ({
...
+            }, {OverviewLink});
+        }
+
+        return formatMessage({
+            defaultMessage: 'This update will be broadcasted to {hasBroadcast, select, true {<OverviewLink><ChannelsTooltip>{broadcastChannelCount, plural, =1 {one channel} other {{broadcastChannelCount, number} channels}}</ChannelsTooltip></OverviewLink>} other {}}{hasFollowersAndChannels, select, true { and } other {}}{hasFollowers, select, true {<OverviewLink><FollowersTooltip>{followersChannelCount, plural, =1 {one direct message} other {{followersChannelCount, number} direct messages}}</FollowersTooltip></OverviewLink>} other {}}.',
```
agarciamontoro, line 173: Did we decide to link to the overview page from the broadcasted channels and followers text? I'm ok with that, just curious what the final decision was.

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -133,33 +153,53 @@ const UpdateRunStatusModal = ({
...
+                <Tooltip
+                    id={`${ID}_broadcast_followers_tooltip`}
+                    content={generateTooltipText(followerNames, followersChannelCount)}
+                >
+                    <span tabIndex={0}>{chunks}</span>
+                </Tooltip>
+            ),
+            hasFollowersAndChannels: Boolean(broadcastChannelCount && followersChannelCount).toString(),
+            hasBroadcast: Boolean(broadcastChannelCount).toString(),
+            hasFollowers: Boolean(followersChannelCount).toString(),
```
agarciamontoro, line 194: Super nit: we use here "Channels" and "Broadcast". Can we use the same for both? For example:
```suggestion
            hasFollowersAndChannels: Boolean(broadcastChannelCount && followersChannelCount).toString(),
            hasChannels: Boolean(broadcastChannelCount).toString(),
            hasFollowers: Boolean(followersChannelCount).toString(),
```

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -200,22 +240,37 @@ const UpdateRunStatusModal = ({
...
+            </GenericModal>
+            <RouteLeavingGuard
+                navigate={(path) => WebappUtils.browserHistory.push(path)}
+                shouldBlockNavigation={(newLoc) => {
+                    // This code will be enabled to prevent user from leave without saving
+                    // after https://github.com/mattermost/mattermost-plugin-playbooks/pull/1148 merge
+                    // In the meantime, it close the modal when navigating to run overview
+                    // if (location.pathname !== newLoc.pathname && pendingChanges) {
+                    // return true;
+                    // }
```
agarciamontoro, line 268: Not something to change, but I don't understand this comment. Why do we need that other PR merged?

webapp/src/components/modals/update_run_status_modal.tsx
trilopin: nice catch!

webapp/src/components/modals/update_run_status_modal.tsx
trilopin: From my understanding, it's an [unfinished conversation](https://community.mattermost.com/core/pl/qktui4g9yfde8xf63zyruuzxxr). If we decide to omit the overview anchor when there are broadcast items, we have to use these anchors. 

TBH, It's a bit surprising to link from these items but not open directly the modal.

webapp/src/components/modals/update_run_status_modal.tsx
trilopin: I already did a PR to show unsaved changes modal when you leave without saving changes and this is the same but with navigation (it's an issue now we do same-tab navigation) instead of accidental hide.

trilopin:
Sorry @agarciamontoro I messed up this PR with a rebase pull, I will do a fresh one

PR #1156, opened 2022-04-29
disable hover menu in checkist while dragging

trilopin:
#### Summary
Disable the checklist menu that appears when doing hover if we are currently doing drag&drop.
Menu hover is disabled for:
- the item you are currently dragging 
- any other item you may hover while dragging.

A Flag has to be managed at the top dropable level, so a new prop needs to be passed to checklist and checklist_item. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43415

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):

agarciamontoro (approved):
Elegant solution! Thanks, @trilopin :)

PR #1157, opened 2022-04-29
Add telemetry for run actions

gbochora:
#### Summary
This PR adds telemetry for run actions. 
Add tracking the following events:
- Status update broadcast to the channels
- Status update broadcast to the webhooks
- Run actions settings update

Also added broadcast to channels/webhooks flags to the `Run create` event.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43175

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


server/telemetry/rudder.go
```diff
@@ -77,6 +77,10 @@ const (
 
 	eventChannelAction     = "channel_action"
 	actionRunChannelAction = "run_channel_action"
+
+	eventRunAction         = "run_action_event"
+	actionRunAction        = "run_action"
+	actionRunActionsUpdate = "run_actions_update"
```
agarciamontoro, line 83: Nit: it's true that it's a bit of a mouthful, but if we want to be consistent with the other events and actions, this should be something like
```suggestion
	eventRunAction         = "run_action"
	actionRunAction        = "run_run_action"
	actionRunActionsUpdate = "update_run_actions"
```

agarciamontoro: Maybe instead of `run_action` it could be `playbookrun_action`, so then it would be
```suggestion
	eventRunAction         = "playbookrun_action"
	actionRunAction        = "run_playbookrun_action"
	actionRunActionsUpdate = "update_playbookrun_actions"
```

agarciamontoro: In any case, it's not that important, so I'll leave it up to you.

server/telemetry/rudder.go
```diff
@@ -598,3 +604,25 @@ func (t *RudderTelemetry) RunChannelAction(action app.GenericChannelAction, user
...
 	t.track(eventChannelAction, properties)
 }
+
+func runActionProperties(triggerType, actionType, userID string) map[string]interface{} {
+	return map[string]interface{}{
+		"UserActualID": userID,
+		"ActionType":   actionType,
+		"TriggerType":  triggerType,
+	}
+}
```
agarciamontoro, line 614: Maybe we can refactor the `channelActionProperties` into `actionProperties` to receive just a string and use the same for both channel and run actions? The reason being that product-wise, channel actions and run actions are just actions, the only thing that changes is the type of the trigger and the type of the action.

server/telemetry/rudder.go
```diff
@@ -598,3 +604,25 @@ func (t *RudderTelemetry) RunChannelAction(action app.GenericChannelAction, user
...
+
+// RunAction tracks the run actions, i.e., status broadcast action
+func (t *RudderTelemetry) RunAction(triggerType, actionType, userID string) {
+	properties := runActionProperties(triggerType, actionType, userID)
+	properties["Action"] = actionRunAction
+	t.track(eventRunAction, properties)
+}
```
agarciamontoro, line 621: The same suggestion with `RunChannelAction` to be just this `RunAction`

agarciamontoro (approved):
Looks good! Behaviour-wise it's perfect, but I suggested a slight change on the implementation to use the same telemetry functions both for channel and run actions. Thoughts?
In any case, non-blocking, so approving :)

server/telemetry/rudder.go
gbochora: I like this one, thanks.

server/telemetry/rudder.go
gbochora: I like your suggestion.

server/telemetry/rudder.go
gbochora: This one is a little trickier. `RunChannelAction` and `RunAction` have different `properties["Action"]` values . It seems we should remain them different, it's an easy way to define action parent object(channel/run). 
Passing the `Action` property as an argument doesn't seem like a good solution either. I would leave it as is. Thoughts?

agarciamontoro: Oh, that's true. Leaving it as is sounds good, thanks for the context!

server/app/playbook_run.go
```diff
@@ -810,6 +817,12 @@ type PlaybookRunTelemetry interface {
 
 	// Unfollow tracks userID following a playbook run.
 	Unfollow(playbookRun *PlaybookRun, userID string)
+
+	// RunAction tracks the run actions, i.e., status broadcast action
+	RunAction(triggerType, actionType, userID string)
```
trilopin, line 822: triggerType could have any string value but is actually a non implemented enum

No need to action here at all, but I'll send a proposal to discuss an improvement.

trilopin: same with ActionType

server/app/playbook_run_service.go
```diff
@@ -960,6 +964,7 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
 
 	if playbookRunToModify.StatusUpdateBroadcastWebhooksEnabled {
 		s.sendWebhooksOnUpdateStatus(playbookRunID)
+		s.telemetry.RunAction(TriggerTypeStatusUpdatePosted, ActionTypeBroadcastWebhooks, userID)
```
trilopin, line 967: If I have enabled channels and webhooks two tracks will be executed through telemetry:
1. TriggerTypeStatusUpdatePosted / ActionTypeBroadcastChannels / USER1
2. TriggerTypeStatusUpdatePosted / ActionTypeBroadcastWebhooks / USER1

Don't we want to know many post-update the user does no matter if they carry 0,1,2 actions?

trilopin: would it make sense to add to runActionProperties the runID and the playbookID?

That could answer later:
- avg run actions per run (across all playbooks)
- avg run actions per playbook run
- total run actions per run / playbook

trilopin: I know this could be completely out of scope, just sharing some thoughts.

trilopin (approved):
Thank you @gbochora, 
 
I shared some non-blocking thoughts, the code is more than ready.

server/app/playbook_run_service.go
gbochora: Great suggestions @trilopin, thanks! I've added additional properties.

gbochora:
@agarciamontoro @trilopin could you please review the last commit.

server/telemetry/rudder.go
```diff
@@ -605,17 +605,20 @@ func (t *RudderTelemetry) RunChannelAction(action app.GenericChannelAction, user
...
-func runActionProperties(triggerType, actionType, userID string) map[string]interface{} {
+func runActionProperties(playbookRun *app.PlaybookRun, userID, triggerType, actionType string, numActions int) map[string]interface{} {
 	return map[string]interface{}{
-		"UserActualID": userID,
-		"ActionType":   actionType,
-		"TriggerType":  triggerType,
+		"UserActualID":  userID,
+		"ActionType":    actionType,
+		"TriggerType":   triggerType,
+		"NumActions":    numActions,
```
agarciamontoro, line 613: I think I'd name this in a different way. Maybe `NumBroadcasts` or something similar. Otherwise it may be a bit confusing, because the action is only one.

server/telemetry/rudder.go
```diff
@@ -605,17 +605,20 @@ func (t *RudderTelemetry) RunChannelAction(action app.GenericChannelAction, user
...
+		"PlaybookRunID": playbookRun.ID,
+		"PlaybookID":    playbookRun.PlaybookID,
```
agarciamontoro, line 615: This is for grouping the actions of the same playbook/run, right? Maybe not something for this PR, but: Should we do the same with channels?

agarciamontoro (changes requested):
Just a naming change suggestion.

agarciamontoro (approved):
Thanks, @gbochora! :tada:

PR #1160, opened 2022-04-29
MM-43132: Run actions

agarciamontoro:
#### Summary
This PR adds the run actions modal to the run overview page. The modal lets the users change the channels and the webhook URLs to which the status updates are broadcasted after the run has started.

As usual, I encourage you to review by commit (except the one for E2E tests, though, which are fixed in a later commit too), the commit messages should be self-explanatory :)

https://user-images.githubusercontent.com/3924815/165997084-2e44c60d-eee0-4c1a-b5ca-eb2cc6d09b0d.mp4

**Note**: I added the `do not merge` label because I'm waiting for two PRs to be merged:
- #1153, which this branch depends on
- https://github.com/mattermost/compass-icons/pull/42, which will add the expected lightning bolt icon to the compass-icons library (although I kind of like the palette icon now, to be honest :joy:)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43132

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


abhijit-singh (commented):
Looks good @agarciamontoro! A couple of minor requests:

1. The channel selector dropdown seems to be getting hidden under the modal footer. It should ideally overlay the modal. This is also there in the channel actions modal - https://mattermost.atlassian.net/browse/MM-43277. I'm good if these are fixed together as a separate effort. 
2. Can we change the tooltip text to just `Run Actions` instead of `Open Run Actions modal`?

webapp/src/client.ts
```diff
@@ -692,6 +693,14 @@ export const saveChannelAction = async (action: ChannelAction): Promise<string>
     return action.id;
 };
 
+export const updateRunActions = async (playbookRunID: string, actions: RunActions) => {
+    try {
+        return await doPut(`${apiUrl}/runs/${playbookRunID}/actions`, JSON.stringify(actions));
```
trilopin, line 698: type here helps even if it's void
```suggestion
        return await doPut<void>(`${apiUrl}/runs/${playbookRunID}/actions`, JSON.stringify(actions));
```

webapp/src/client.ts
```diff
@@ -692,6 +693,14 @@ export const saveChannelAction = async (action: ChannelAction): Promise<string>
...
+    } catch (error) {
+        return {error};
```
trilopin, line 700: slightly more accurate errors (will be forced in typescript 4.4)
```suggestion
        return {error as ClientError};
```

webapp/src/client.ts
trilopin: I assumed it is void, please use the right type if it returns something

webapp/src/client.ts
trilopin: to give some context, we are still in 4.3, and errors from the catch are type=any, in 4.4 they become type=unknown so you can not use properties unless you cast the error.

https://dev.to/ankittanna/what-s-new-in-typescript-4-4-ai7

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
```diff
@@ -359,24 +363,36 @@ const PlaybookRunBackstage = () => {
...
+            shouldUpdatePosition={true}
+            content={formatMessage({defaultMessage: 'Open Run Actions modal'})}
+        >
+            <HeaderIcon
+                className='icon-palette-outline'
+                onClick={() => dispatch(showRunActionsModal())}
+                clicked={false}
+                aria-label={formatMessage({defaultMessage: 'Open Run Actions modal'})}
+            />
+        </Tooltip>
```
trilopin, line 379: there are several points I find surprising:
- Why the icon is palette instead of the lighting?
- The design has a number close to the icon, not sure what it means but should be included in this scope?

trilopin:
I tried to save webhooks from the run actions modal and didn't seem to work https://www.loom.com/share/2ec5942e77a14648985633e992ae99de

webapp/src/components/run_actions_modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+        updateRunActions(playbookRun.id, {
+            status_update_broadcast_channels_enabled: broadcastToChannelsEnabled,
+            broadcast_channel_ids: channelIds,
+            status_update_broadcast_webhooks_enabled: sendOutgoingWebhookEnabled,
+            webhook_on_status_update_urls: webhooks,
+        });
+    };
+
+    return (
+        <ActionsModal
```
trilopin, line 60: Philosophical question (discussion and action can be taken outside this PR).

ActionsModal seems to be a wrapper with some props fixed. But the "actions" related behavior is completely on children (outside ActionModal component). Does it feel right to you?

I would do one of the following:
- change ActionsModal name since seems a generic wrapper with style and structure (not only for actions)
- include part of the children logic in the ActionModal and fine-tune with some props

trilopin:
Thank you @agarciamontoro !, I left some optional comments about Typescript typing and design decisions that probably need no action or the action is not expected to be applied in this PR. 

The only blocking thing is the bug I experienced and reported with a video.

agarciamontoro:
Ugh, sorry, I committed here the changes corresponding to #1153. I'll force-push just to eliminate them.

webapp/src/client.ts
agarciamontoro: Good suggestion, thanks!

webapp/src/client.ts
agarciamontoro: This cannot be casted directly to `ClientError`. I propose to leave it as is and come back to this later, I think it'd be good to talk about how to handle errors in the client.

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
agarciamontoro: >     * Why the icon is palette instead of the lighting?

The palette icon is just a placeholder. The lightning bolt was not in the [compass-icons](https://github.com/mattermost/compass-icons) font, so I just picked a random one while we update compass-icons. And it is already updated (see [my last PR there](https://github.com/mattermost/compass-icons/pull/42)) but now we need to release a new version so we can actually use it. I asked how to do that (that's something I've never done before) [here](https://community-daily.mattermost.com/core/pl/fe3bozhi8jfc3gdwxkyrczd5ih) and [here](https://community-daily.mattermost.com/core/pl/r5oireyrutf4381c4r1agi59br). When that release is done, I can update the dependency here.

>     * The design has a number close to the icon, not sure what it means but should be included in this scope?

Yes, the design has the number, but we later decided to remove it, as it would be just a hardcoded 2 for now. Even if we count only the "enabled" actions, it would not add a lot of information.

webapp/src/components/run_actions_modal.tsx
agarciamontoro: Yeah, `ActionsModal` was before `ChannelActionsModal`, so it had a lot more of logic. Maybe it makes sense to change the name now, but I propose to tackle that outside of this PR, as you suggested.

agarciamontoro:
@abhijit-singh, thanks for the review!

- Yup, this is the exact same problem as with the channel actions modal. We can tackle it with the ticket you linked, which we may want to prioritize. Thank you!
- I changed the tooltip text to simply `Run Actions` as you suggested. See https://github.com/mattermost/mattermost-plugin-playbooks/pull/1160/commits/a9028febc89af3f86aede3fe64e62b026ecfebfc.

@trilopin, also thank you for the review!

I answered the individual comments in their corresponding threads. Regarding the bug you found: thank you for catching that, I don't know how I had missed that :open_mouth:

The problem was that we were not updating the playbook run with the new state, which is fixed by sending a websocket event from the server to let the client know about its new state. See https://github.com/mattermost/mattermost-plugin-playbooks/pull/1160/commits/165102e50f0bdcc6b64456800015ee7c64b1fdf1 This fixes things on my end, but I'd like you to take a second look too, if that's possible. Thanks!

webapp/src/components/backstage/playbook_runs/playbook_run_backstage/playbook_run_backstage.tsx
trilopin: Thanks for your explanations.

webapp/src/components/run_actions_modal.tsx
trilopin: Cool!

trilopin (approved):
Bug is solved and code looks good!

agarciamontoro:
After having merged #1153, the commits here became a mess, so I no longer recommend to review by commit, sorry! The changes are still correct, so no problem there.

agarciamontoro:
The new version of compass-icons has not been released yet, so I just used the mdi icon instead. See https://github.com/mattermost/mattermost-plugin-playbooks/pull/1160/commits/ef51523967f2f4dfa8b209369cd6bd88327ebdaa. We can change it to the one in compass-icons when it is released.

Phrynobatrachus (approved):
Awesome work! E2Es and manual use look solid.

gbochora (approved):
Great work @agarciamontoro 🥳 
Two non blocking issues:
- After closing the run actions modal, the `Run Action` tooltip is displayed, and it stays.

- If I refresh the run overview page, the run actions modal shows all channels as `Unknown`

https://user-images.githubusercontent.com/4368372/167241727-5624a35a-5e24-43fa-ab28-233bf19ee933.mp4

abhijit-singh (approved):
Thanks @agarciamontoro! Approving but there's one tiny non-blocking request — The sizes for the `Run actions` and the `Copy link` icon and button seem different. Can we standardize them to 16px x 16px for the inner icon, and 28px x 28px for the button (which includes hover area)?
image
image

agarciamontoro:
@abhijit-singh, thanks! Icon sizes fixed:

https://user-images.githubusercontent.com/3924815/167612008-61282847-00c4-4779-8486-da46606487c3.mp4

agarciamontoro:
Those 2 E2E tests are expected to fail. Merging now.

PR #1167, opened 2022-05-04
enum types proposal

trilopin:
THIS PR IS A PROPOSAL TO BE DISCUSSED - NOT AN ACTUAL PR TO BE MERGED

## current status
The current approach in many places is a combination of package constants, validation methods, and passing parameters as scalar types (string/int).  

## proposal
Create an enum type through [iota mechanism](https://go.dev/ref/spec#Iota) and implement key interfaces:
- [Scanner, Valuer](http://jmoiron.net/blog/built-in-interfaces): handles read/write conversions from database
- [Stringer](https://musse.dev/stringer-golang/): handles string version for errors or any readable format
- [Marshallers](https://pkg.go.dev/encoding#TextMarshaler): handle JSON serialization/deserialization  

All params used in functions must use the new type which implemented all this interfaces.

Given a type T with two options One and Two 
- It's not possible to create a HTTP response with a wrong value different from One and Two, if we forced it, it won't encode JSON
- It's not possible to read JSON body from HTTP requests if the value for type is different from One or Two
- It's not possible to store a wrong value in database
- It's not possible to read a wrong value from database 

## Benefits
**Safety**
The validation of the type is included in any of the serializations, from/to JSON and from/to database.

**Documentation**
It's clearer what are the possible values for a type with iota than just having a couple of constants that you are not sure if they are part of the same topic.

## Drawbacks
It's more boilerplate, not complicated code but indeed more.

### DIFF
--- a/server/api/actions.go
+++ b/server/api/actions.go
@@ -102,26 +102,14 @@ func isValidTrigger(trigger string) bool {
 	return false
 }
 
-func isValidAction(action string) bool {
-	if action == "" {
-		return true
-	}
-
-	for _, elem := range app.ValidActionTypes {
-		if action == string(elem) {
-			return true
-		}
-	}
-
-	return false
-}
-
 func parseGetChannelActionsOptions(query url.Values) (*app.GetChannelActionOptions, error) {
 	actionTypeStr := query.Get("action_type")
 	triggerTypeStr := query.Get("trigger_type")
 
-	if !isValidAction(actionTypeStr) {
-		return nil, fmt.Errorf("action_type %q not recognized; valid values are %v", actionTypeStr, app.ValidActionTypes)
+	var actionType app.ActionType
+	err := (&actionType).UnmarshalText([]byte(actionTypeStr))
+	if err != nil {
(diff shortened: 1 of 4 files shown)


server/api/actions.go
```diff
@@ -102,34 +102,22 @@ func isValidTrigger(trigger string) bool {
...
-}
-
 func parseGetChannelActionsOptions(query url.Values) (*app.GetChannelActionOptions, error) {
 	actionTypeStr := query.Get("action_type")
 	triggerTypeStr := query.Get("trigger_type")
 
-	if !isValidAction(actionTypeStr) {
-		return nil, fmt.Errorf("action_type %q not recognized; valid values are %v", actionTypeStr, app.ValidActionTypes)
+	var actionType app.ActionType
+	err := (&actionType).UnmarshalText([]byte(actionTypeStr))
```
lieut-data, line 110: I feel like this would "read" better as:
```suggestion
	actionType, err := NewAtionTypeFromString(actionTypeStr)
```

trilopin: Like it. If we go this way, I would do directly `actionType, errr := NewActionType(s string)`

In the end, there is no other constructor, and seems clearer.

PR #1171, opened 2022-05-09
MN-44032 avoid infinite loop in use ensure profiles

lieut-data:
#### Summary
In `useEnsureProfiles`, avoid re-querying users unless the requested set of users has actually changed. Also, only actually request unknown users, not ones already in Redux. Finally, introduce a `useEnsureProfile` helper that avoids clients passing `[userId]` and forcing re-evaluation on each render.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-44032

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/hooks/general.ts
```diff
@@ -381,24 +381,25 @@ export function useAllowMakePlaybookPrivate() {
...
+        const unknownIds = userIds.filter((userId) => !getUserFromStore(userId));
+        if (unknownIds.length > 0) {
+            dispatch(getProfilesByIds(unknownIds));
         }
-    }
-
-    if (unknownIds.length > 0) {
-        dispatch(getProfilesByIds(userIds));
-    }
+    }, [userIds]);
```
crspeller, line 402: This still doesn't quite "avoid re-querying users unless the requested set of users has actually changed" since it won't do a deep comparison here. But this is certainly better than before.

crspeller (approved):

calebroseland (approved):
LGTM, though indeed the list of IDs would need to be identity-stable to avoid re-requests.

There's also `{getMissingProfilesByIds} from 'mattermost-redux/actions/users'` which is already thoroughly used and tested, but I don't think it has false-retry negation built into it—so we'd need a hook/wrapper anyway.

PR #1172, opened 2022-05-10
Upgrade server dep to have safe server e2e plugin initialisation

trilopin:
#### Summary
We create two processes per e2e server test (at each `Setup()` call) because the plugin was started at regular initialization and additionally after the call to UpdateConfig in the `OnActivate` hook. [More info](https://community-daily.mattermost.com/core/pl/57pmdgm6ji867rwctqzu5wzb9a)

The mattermost-server has included a change to set the plugin active **before** OnActivate call. https://github.com/mattermost/mattermost-server/pull/20165

Postgre `too many clients already` error is also fixed with this change, I monitored a full `make test` cycle and it didn't reach 30 connections (compared to 105 previously).

In the playbooks repository, we have just to update server dep and remove the temporary explicit command to increase postgre max_connections. 

This PR should be open until https://github.com/mattermost/mattermost-server/pull/20165 merge, then update the dep.

#### Deps change
* mattermost-server updated to the branch with the change (later will be master)
* squirrel has been changed from the official repository to @lieut-data's to match the same version of the server (otherwise tests don't pass)
* check additional changes in go.sum/go.mod

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43750
Related https://mattermost.atlassian.net/browse/MM-43833

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


trilopin:
Cypress still failing due to a flaky test that will be fixed outside this PR.

I'm still going to triple-check the changes in the deps before merging.

trilopin:
cc @Phrynobatrachus , Here I revert the changes with the custom command for postgres in the CI pipeline.

go.mod
```diff
@@ -8,10 +8,12 @@ replace github.com/HdrHistogram/hdrhistogram-go => github.com/codahale/hdrhistog
 
 replace github.com/golang/mock => github.com/golang/mock v1.4.4
 
-replace github.com/mattermost/mattermost-server/v6 => github.com/mattermost/mattermost-server/v6 v6.0.0-20220303161041-4ecd0712756e
+replace github.com/mattermost/mattermost-server/v6 => github.com/mattermost/mattermost-server/v6 v6.0.0-20220510062744-292149ee4170
 
 replace github.com/mattermost/mattermost-plugin-api => github.com/mattermost/mattermost-plugin-api v0.0.22-0.20211207232216-3faec618d311
 
+replace github.com/Masterminds/squirrel v1.5.2 => github.com/lieut-data/squirrel v1.5.4
```
trilopin, line 15: I don't think this is risky since it's the same version than server is currently using. 

@lieut-data 

Could this version change have unintended side effects for playbooks?

Is there any active effort to merge this multiple-update support into the main squirrel?

lieut-data:
@trilopin, if you can update server once more, we can pull in https://github.com/mattermost/mattermost-server/pull/20162 and simplify the squirrel import. (0/5, but we can also just start using https://github.com/mattermost/squirrel ourselves. Shouldn't require any immediate changes in our code, I think.)

lieut-data (approved):

crspeller (approved):

PR #1177, opened 2022-05-10
MM-42206: avoid concurrent profile fetch

lieut-data:
#### Summary
When a checklist has multiple components that list users in team or channel:
CleanShot 2022-05-10 at 15 27 28@2x

The `useProfilesInTeam`, `useProfilesInChannel`, and `useProfilesInCurrentChannel` create a thundering herd problem, all trying to fetch the requested metadata per component:
CleanShot 2022-05-10 at 15 27 36@2x

Detect and avoid this:
CleanShot 2022-05-10 at 15 29 05@2x

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-42206

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


lieut-data (commented):
In a different universe, I would attach "middleware" to a dispatched redux action using redux saga, intelligently deciding when to fetch profiles and seamlessly merging requests from all clients, not just these special hooks that are cache aware.

webapp/src/hooks/general.test.ts
```diff
@@ -215,3 +219,481 @@ describe('useEnsureProfiles', () => {
...
+        const profilesInCurrentTeam = [] as String[];
+        const currentTeamId = 'team_id';
+
+        let calls = 0;
+        useSelectorSpy.mockImplementation(() => {
+            if (calls++ % 2 === 0) {
+                return profilesInCurrentTeam;
+            }
+            return currentTeamId;
+        });
```
agarciamontoro, line 235: It took me a while to understand this call. Could we add a comment stating that there are two selectors in the implementation called one after the other, and that this mocks the implementation of those two, just alternating between them?

webapp/src/hooks/general.test.ts
```diff
@@ -215,3 +219,481 @@ describe('useEnsureProfiles', () => {
...
+
+        const useDispatchSpy = jest.spyOn(redux, 'useDispatch');
+        const mockDispatchFn = jest.fn();
+        useDispatchSpy.mockReturnValue(mockDispatchFn);
+
+        const {rerender} = renderHook(() => {
+            useProfilesInTeam();
+        });
```
agarciamontoro, line 243: Super nit (repeated in the tests below too)
```suggestion
        const {rerender} = renderHook(useProfilesInTeam);
```

agarciamontoro (approved):
Interesting problem (and I think I actually understand now the potential of redux-sagas) and good solution! Thanks :tada:

webapp/src/hooks/general.ts
```diff
@@ -191,15 +186,42 @@ export function useCanCreatePlaybooksOnAnyTeam() {
     ));
 }
 
+const gettingProfilesInTeam = new Map<string, boolean>();
+const gettingProfilesInChannel = new Map<string, boolean>();
+
+export function clearCaches() {
```
trilopin, line 192: If I understand correctly, these maps act as a lock to avoid fetching profiles into the state when it's not really needed or risky.

Is that right?

webapp/src/hooks/general.ts
```diff
@@ -191,15 +186,42 @@ export function useCanCreatePlaybooksOnAnyTeam() {
...
     const currentTeamId = useSelector(getCurrentTeamId);
+
     useEffect(() => {
         if (profilesInTeam.length > 0) {
+            gettingProfilesInTeam.delete(currentTeamId);
             return;
         }
 
+        // Avoid issuing multiple concurrent fetches for this team.
+        if (gettingProfilesInTeam.has(currentTeamId)) {
```
trilopin, line 220: optional detail: I would use `get` instead of `has`. It will return `undefined | false | true` (and your if clause will be still valid)

It's a bit more defensive, in case someone fills values with false in a future pr

trilopin:
What would happen if the users we would want to retrieve are not available on the first page of the team?

webapp/src/hooks/general.ts
```diff
@@ -191,15 +186,42 @@ export function useCanCreatePlaybooksOnAnyTeam() {
...
+// wanting the same metadata.
 export function useProfilesInTeam() {
-    const dispatch = useDispatch() as DispatchFunc;
+    const dispatch = useDispatch();
     const profilesInTeam = useSelector(getProfilesInCurrentTeam);
     const currentTeamId = useSelector(getCurrentTeamId);
+
     useEffect(() => {
         if (profilesInTeam.length > 0) {
+            gettingProfilesInTeam.delete(currentTeamId);
```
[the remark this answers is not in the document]
lieut-data: Not exactly -- so long as we have some profiles in the team, we'll never try fetching again. And our current Redux semantics have us never discarding those profiles. But in case something ever does change that causes our Redux state to reset without also reloading the application, I wanted to ensure this "mutex" doesn't get in the way.

webapp/src/hooks/general.test.ts
lieut-data: Yes, good call!

webapp/src/hooks/general.test.ts
lieut-data: I think this works for `useProfilesInTeam`, but not `useEnsureProfiles`, since we have to pass a parameter, and then the testing blocks are "different" in ways that might seem more confusing when understanding the whole file.

webapp/src/hooks/general.ts
lieut-data: Ooh, thank you for raising this. I used a map in the Go-style of `map[string]bool` to actually represent a set, but I can just use a Set natively instead here. Fixed!

trilopin: Nice 🎉

webapp/src/hooks/general.ts
trilopin: Since your motivation is to avoid concurrent calls more than store&cache the results I would like to rename `clearCaches` to `clearLocks` and  `gettingProfilesInTeam` to `lockProfilesInTeamFetch`, You don't have to use exactly my suggestions but at least, remove cache reference.

trilopin (approved):

trilopin:
My (optional) suggestions are more related to naming and semantics than the flow or the approach. 

Thanks!

webapp/src/hooks/general.ts
lieut-data: Love your suggestions as-is! Done :)

PR #1179, opened 2022-05-11
Mm 43423 inconsistent hover icons

trilopin:
#### Summary
Make icon hover style consistent with https://zeroheight.com/29be2c109/p/00b099-icon-buttons .
dotmenu now works the following way:
- grey when no active, no hover
- grey darker background when hover
- blue when active (dropdown opened, no matter if hover)
- blue with darker background when active & hover

In the case of RHS filter, I considered active when:
- the dropdown is openend
- the dropdown is closed but we have selected an option different than default (all choices selected by clicking in "all" or by clicking any choice except "all")

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-43423

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/dot_menu.tsx
```diff
@@ -114,6 +117,7 @@ const DotMenu = (props: DotMenuProps) => {
         <MenuButton
             title={props.title}
             ref={rootRef}
+            isActive={(props.isActive ?? false) || isOpen}
```
agarciamontoro, line 120: Isn't this the same as
```suggestion
            isActive={props.isActive ?? isOpen}
```

agarciamontoro (approved):
Looks good, @trilopin! Thank you :tada:

gbochora (approved):
Cool!

webapp/src/components/dot_menu.tsx
gbochora: It seems these conditions are not the same. i.e. if `props.isActive` is `false` and `isOpen` is `true`, first condition result is `true`, and second is `false`.

agarciamontoro: Aaaah, that's true, because the ?? only checks for real `null`ity, right?

gbochora: Yep

abhijit-singh (approved):
Awesome!

PR #1181, opened 2022-05-11
Fix checklist item render issue

gbochora:
#### Summary
This PR fixes issue with RHS Checklist item showing the due date of an item with the same name from another channel. It happened because items with the same name had the same `key` attribute values, and switching between channels did not render the list component.
Updated `key` generation logic, now its value is unique.

This recording shows the problem:

https://user-images.githubusercontent.com/4368372/167859688-d401720a-c88f-40d1-8ff8-8df0809f2797.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43936

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/checklist/generic_checklist.tsx
+++ b/webapp/src/components/checklist/generic_checklist.tsx
@@ -114,7 +114,7 @@ const GenericChecklist = (props: Props) => {
         props.onUpdateChecklist(newChecklist);
     };
 
-    const keys = generateKeys(props.checklist.items.map((item) => item.title));
+    const keys = generateKeys(props.checklist.items.map((item) => props.playbookRun?.id + item.title));
 
     return (
         <Droppable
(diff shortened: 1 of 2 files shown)


webapp/src/components/checklist/generic_checklist.tsx
```diff
@@ -114,7 +114,7 @@ const GenericChecklist = (props: Props) => {
         props.onUpdateChecklist(newChecklist);
     };
 
-    const keys = generateKeys(props.checklist.items.map((item) => item.title));
+    const keys = generateKeys(props.checklist.items.map((item) => props.playbookRun?.id + item.title));
```
lieut-data, line 117: I imagine we'll need to support task ids in the very near future, especially as we introduce threads anywhere.

lieut-data (approved):

lieut-data:
@gbochora, is there an opportunity to test this via e2e?

crspeller (approved):

gbochora:
> @gbochora, is there an opportunity to test this via e2e?

I've added an e2e test.

PR #1184, opened 2022-05-13
handle optimistic updates to checkbox in checklist item

trilopin:
### Summary
Optimistic approach to checkbox state management.
The main reasons why I pushed for this change are:
- offer a more responsive UI in a task that I'd expect to be quick as a user
- be more defensive to websocket failures where users don't see the UI update and even could click several times keeping a trash value

Approach:

**client.ts**
`doPut`and `doFetchWithResponse` already throw a ClientError if status is not OK
Catch that error and return it as data.

**checklist_item_draggable and checklist_item**
type the return value and add the possibility of `undefined`

**inputs**
do the following:
1. handle current value and setter through useState
2. onChange: calculate the new value and set it immediately
3. onChange: await prop passed with `setChecklistItemState`
4. if the result is not undefined and has a property error, set the previous value 

0/5 in types mess, `Promise<void | {error: ClientError}> | undefined` is not elegant.
I kept `{error: ClientError}` instead of `ClientError` to be consistent with other client functions.

### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-43834

### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/client.ts
+++ b/webapp/src/client.ts
@@ -312,11 +312,12 @@ export async function setDueDate(playbookRunId: string, checklistNum: number, it
 }
 
 export async function setChecklistItemState(playbookRunID: string, checklistNum: number, itemNum: number, newState: ChecklistItemState) {
-    return doPut(`${apiUrl}/runs/${playbookRunID}/checklists/${checklistNum}/item/${itemNum}/state`,
-        JSON.stringify({
-            new_state: newState,
-        }),
-    );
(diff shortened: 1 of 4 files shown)


lieut-data (approved):
Love the simplicity!

calebroseland (approved):
Right to the point 😀

PR #1186, opened 2022-05-16
MM-41875 Introducing graphql with playbooks

crspeller:
This PR adds graphql fetching for a single playbook and updates for a single playbook. This allows the better playbook editor to make incremental updates to playbooks, avoiding conflicts. 

The backend is based on https://github.com/mattermost/mattermost-server/pull/19486 the mattermost-server implementation. Eventually we would be able to reuse that portion when we move over and the multi product architecture allows for combining our schemas. In turn this is based on https://github.com/graph-gophers/graphql-go

I will highlight important parts of the PR in comments. 

On the client side we have no current implementation to copy so I decided to use the simplest to use graphql client Apollo. I debated between Apollo and Relay. They both solve the caching problem for you. Relay is more complex but it allows for fragmented queries to allow sub-components to request exactly what data they need and be forced to only use that. It's a really nice architecture however it very much encompasses your code with Relay code which I didn't want to commit to. It also uses unreleased React features which would force us to upgrade to unstable React versions. (Relay is developed by FB) Apollo is simple and straightforward to understand.

Client types are quite the thing with graphql. I decided to try and generate them using `graphql-codegen`. 
In addition it is able to generate hooks compatible with Apollo. It does this by reading [queries from files. that we write. Then you can easily use that in your own hooks.


client/client.go
```diff
@@ -162,6 +162,27 @@ func (c *Client) do(ctx context.Context, req *http.Request, v interface{}) (*htt
 	return resp, err
 }
 
+type GraphQLInput struct {
+	Query         string                 `json:"query"`
+	OperationName string                 `json:"operationName"`
+	Variables     map[string]interface{} `json:"variables"`
+}
+
+func (c *Client) DoGraphql(ctx context.Context, input *GraphQLInput, v interface{}) error {
```
crspeller, line 171: Very basic for tests.

server/api/graphql.go
```diff
@@ -0,0 +1,199 @@
...
+
+	return c, nil
+}
+
+func graphiQL(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "text/html")
+	_, _ = w.Write(graphiqlPage)
+}
+
+var graphiqlPage = []byte(`
```
crspeller, line 159: This is for the query sandbox. Just visit http://localhost:8065/plugins/playbooks/api/v0/query and this page will be served to you so you can try out queries.

server/api/graphql_playbook.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"github.com/mattermost/mattermost-plugin-playbooks/server/app"
+)
+
+type PlaybookResolver struct {
+	app.Playbook
+}
+
+func (r *PlaybookResolver) DeleteAt() float64 {
```
crspeller, line 11: There is some strangeness here as graphql can not support int64 which a lot of our fields are.

server/api/graphql_root.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	if err := c.permissions.PlaybookManageProperties(userID, currentPlaybook); err != nil {
+		return "", err
+	}
+
+	setmap := map[string]interface{}{}
+	addToSetmap(setmap, "Title", args.Updates.Title)
+	addToSetmap(setmap, "Description", args.Updates.Description)
+	if args.Updates.Public != nil {
```
crspeller, line 115: Validation is handled separate for each field if it's included.

server/api/schema.graphqls
```diff
@@ -0,0 +1,139 @@
...
+	webhookOnStatusUpdateURLs: [String!]
+	webhookOnStatusUpdateEnabled: Boolean
+	signalAnyKeywords: [String!]
+	signalAnyKeywordsEnabled: Boolean
+	categorizeChannelEnabled: Boolean
+	categoryName: String
+	runSummaryTemplateEnabled: Boolean
+	runSummaryTemplate: String
+	channelNameTemplate: String
+	checklists: [ChecklistUpdates!]
```
crspeller, line 40: For now this is here for simplicity, but it should really be a separate mutation with it's own ID.

server/main.go
```diff
@@ -1,9 +1,42 @@
...
+	"fmt"
+	"os"
+
+	"github.com/graph-gophers/graphql-go"
+	"github.com/mattermost/mattermost-plugin-playbooks/server/api"
 	"github.com/mattermost/mattermost-server/v6/plugin"
 )
 
 func main() {
+	if len(os.Args) > 1 {
```
crspeller, line 13: This allows an easy way to check the graphql schema for validity automatically. Otherwise it would fail at runtime and be really annoying.

webapp/src/graphql/generated_types.ts
```diff
@@ -0,0 +1,314 @@
+import {gql} from '@apollo/client';
```
crspeller, line 1: This file is generated. Lots of stuff to make using graphql on the client easier.

webapp/src/graphql/hooks.ts
```diff
@@ -0,0 +1,33 @@
+import {PlaybookDocument, PlaybookQuery, PlaybookQueryHookResult, PlaybookUpdates, usePlaybookQuery, useUpdatePlaybookMutation} from 'src/graphql/generated_types';
+
+export type FullPlaybook = PlaybookQuery['playbook']
+
+export type Loaded<T> = Exclude<T, undefined | null>
+
+export const usePlaybook = (id: string): [FullPlaybook, PlaybookQueryHookResult] => {
```
crspeller, line 7: Wrappers for the standard generated hooks to make them easier to use.

webapp/src/graphql/hooks.ts
```diff
@@ -0,0 +1,33 @@
+import {PlaybookDocument, PlaybookQuery, PlaybookQueryHookResult, PlaybookUpdates, usePlaybookQuery, useUpdatePlaybookMutation} from 'src/graphql/generated_types';
+
+export type FullPlaybook = PlaybookQuery['playbook']
```
crspeller, line 3: Ideally we don't use this type and we refactor to use partials and individual queries. But this is quicker. Something for a performance effort to look at.

webapp/src/graphql/playbook.graphql
```diff
@@ -0,0 +1,63 @@
+query Playbook($id: String!) {
```
crspeller, line 1: Simple query that fetches everything for the editor.

crspeller:
Attempting to fix circleci at the moment.

server/api/graphql_root.go
```diff
@@ -0,0 +1,225 @@
...
+			if err := c.permissions.PlaybookMakePublic(userID, currentPlaybook); err != nil {
+				return "", errors.Wrap(err, "attempted to make playbook public without permissions")
+			}
+		}
+		if c.licenceChecker.PlaybookAllowed(*args.Updates.Public) {
+			return "", errors.Wrapf(app.ErrLicensedFeature, "the playbook is not valid with the current license")
+		}
+		addToSetmap(setmap, "Public", args.Updates.Public)
+	}
+	addToSetmap(setmap, "CreatePublicPlaybookRun", args.Updates.CreatePublicPlaybookRun)
```
calebroseland, line 130: DB: column mapping
```suggestion
	addToSetmap(setmap, "CreatePublicIncident", args.Updates.CreatePublicPlaybookRun)
```

calebroseland, line 130: DB: column mapping
```suggestion
	addToSetmap(setmap, "CreatePublicIncident", args.Updates.CreatePublicPlaybookRun)
```

server/api/graphql_root.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	addToSetmap(setmap, "InviteUsersEnabled", args.Updates.InviteUsersEnabled)
+	if args.Updates.DefaultOwnerID != nil {
+		if c.pluginAPI.User.HasPermissionToTeam(*args.Updates.DefaultOwnerID, currentPlaybook.TeamID, model.PermissionViewTeam) {
+			return "", errors.Wrap(app.ErrNoPermissions, "default owner can't view team")
+		}
+		addToSetmap(setmap, "DefaultOwnerID", args.Updates.DefaultOwnerID)
+	}
+	addToSetmap(setmap, "DefaultOwnerEnabled", args.Updates.DefaultOwnerEnabled)
```
calebroseland, line 152: DB: column mapping
```suggestion
	addToSetmap(setmap, "DefaultCommanderEnabled", args.Updates.DefaultOwnerEnabled)
```

calebroseland, line 152: DB: column mapping
```suggestion
	addToSetmap(setmap, "DefaultCommanderEnabled", args.Updates.DefaultOwnerEnabled)
```

server/api/graphql_root.go
```diff
@@ -0,0 +1,225 @@
...
+	Command          string  `json:"command"`
+	CommandLastRun   float64 `json:"command_last_run"`
+	Description      string  `json:"description"`
+	LastSkipped      float64 `json:"delete_at"`
+	DueDate          float64 `json:"due_date"`
+}
+
+func (r *RootResolver) UpdatePlaybook(ctx context.Context, args struct {
+	ID      string
+	Updates struct {
```
calebroseland, line 59: `Metrics` appears to be missing from the update pipeline/infrastructure.

webapp/src/graphql/generated_types.ts
```diff
@@ -0,0 +1,314 @@
...
+        command
+        command_last_run: commandLastRun
+        due_date: dueDate
+      }
+    }
+    members {
+      user_id: userID
+      roles
+    }
+    metrics {
```
calebroseland, line 235: If `metrics` has a non-zero length, there's an error on read:

```
{
    "errors": [
        {
            "message": "Invalid value metric_integer.\nExpected type MetricType, found metric_integer.",
            "path": [
                "playbook",
                "metrics",
                0,
                "type"
            ]
        }
    ],
    "data": {
        "playbook": null
    }
}
```

calebroseland (changes requested):
While I haven't gotten through every piece, this is really nice so far. I like the write/update hook usage. Found a couple of bugs, but otherwise first-skim looks pretty good!

webapp/src/components/backstage/playbook_edit/metrics/shared.ts
```diff
@@ -6,8 +6,8 @@ import {Duration} from 'luxon';
 import {MetricType} from 'src/types/playbook';
 import {formatDuration} from 'src/components/formatted_duration';
 
-export const metricToString = (target: number | null, type: MetricType, naturalDuration = false) => {
-    if (target === null) {
+export const metricToString = (target: number | null | undefined, type: MetricType, naturalDuration = false) => {
+    if (!target) {
```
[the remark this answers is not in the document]
crspeller: Fixed.

server/sqlstore/playbook.go
```diff
@@ -586,6 +586,23 @@ func (p *playbookStore) GetPlaybookIDsForUser(userID string, teamID string) ([]s
...
 }
 
+func (p *playbookStore) GraphqlUpdate(id string, setmap map[string]interface{}) error {
+	if id == "" {
+		return errors.New("id should not be empty")
+	}
+
+	_, err := p.store.execBuilder(p.store.db, sq.
+		Update("IR_Playbook").
+		SetMap(setmap).
```
lieut-data, line 596: I'm not a huge fan of mapping a setmap directly into the SQL query. Commented on this elsewhere, but can we avoid this?

server/main.go
lieut-data: What's the rationale for building it into the plugin main, vs. a separate binary?

server/app/playbook.go
```diff
@@ -449,3 +454,49 @@ func (o PlaybookFilterOptions) Validate() (PlaybookFilterOptions, error) {
...
+
+func ValidateWebhookURLs(urls []string) error {
+	if len(urls) > 64 {
+		return errors.New("too many registered urls, limit to less than 64")
+	}
+
+	for _, webhook := range urls {
+		url, err := url.ParseRequestURI(webhook)
+		if err != nil {
+			return err
```
lieut-data, line 466: Should we include which url is invalid when this happens?

calebroseland (approved):

client/client.go
```diff
@@ -162,6 +162,27 @@ func (c *Client) do(ctx context.Context, req *http.Request, v interface{}) (*htt
...
+	url := "query"
+	req, err := c.newRequest(http.MethodPost, url, input)
```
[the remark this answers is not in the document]
crspeller: Yes they are.

server/api/graphql.go
```diff
@@ -0,0 +1,199 @@
...
+		playbookStore:   playbookStore,
+		licenceChecker:  licenceChecker,
+	}
+
+	opts := []graphql.SchemaOpt{
+		graphql.UseFieldResolvers(),
+		graphql.MaxParallelism(5),
+	}
+
+	if !configService.IsConfiguredForDevelopmentAndTesting() {
```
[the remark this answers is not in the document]
crspeller: Yes, these settings are for security and turn off some useful debugging tools.

server/api/graphql_playbook.go
crspeller: Yes.

server/api/graphql_root.go
```diff
@@ -0,0 +1,225 @@
...
+	userID := c.r.Header.Get("Mattermost-User-ID")
+
+	if err := c.permissions.PlaybookView(userID, playbookID); err != nil {
+		c.log.Warnf("public error message: %v; internal details: %v", "Not authorized", err)
+		return nil, errors.New("Not authorized")
+	}
+
+	playbook, err := c.playbookService.Get(playbookID)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
crspeller: Only with custom errors see: https://github.com/graph-gophers/graphql-go#custom-errors

server/api/graphql_root.go
```diff
@@ -0,0 +1,225 @@
...
+	if args.Updates.DefaultOwnerID != nil {
+		if c.pluginAPI.User.HasPermissionToTeam(*args.Updates.DefaultOwnerID, currentPlaybook.TeamID, model.PermissionViewTeam) {
+			return "", errors.Wrap(app.ErrNoPermissions, "default owner can't view team")
+		}
+		addToSetmap(setmap, "DefaultCommanderID", args.Updates.DefaultOwnerID)
+	}
+	addToSetmap(setmap, "DefaultCommanderEnabled", args.Updates.DefaultOwnerEnabled)
+
+	if args.Updates.BroadcastChannelIDs != nil {
+		fmt.Println(*args.Updates.BroadcastChannelIDs)
```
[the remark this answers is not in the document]
crspeller: 🤦

server/api/graphql_root.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// Not optimal graphql. Stopgap measure. Should be updated seperately.
+	checklistsJSON, err := json.Marshal(args.Updates.Checklists)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to marshal checklist in graphql json for playbook id: '%s'", args.ID)
+	}
+	setmap["ChecklistsJSON"] = checklistsJSON
+
+	if len(setmap) > 0 {
+		if err := c.playbookStore.GraphqlUpdate(args.ID, setmap); err != nil {
```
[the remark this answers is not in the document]
crspeller: I guess it's just a design decision on if you do the validation and modification here or in the store layer. I chose to do it here. Not 100% sure if that is the right decision or if there is a better split.

server/api_graphql_playbooks_test.go
```diff
@@ -0,0 +1,144 @@
...
+	query Playbook($id: String!) {
+		playbook(id: $id) {
+			id
+			title
+		}
+	}
+	`
+		err := e.PlaybooksAdminClient.DoGraphql(context.Background(), &client.GraphQLInput{
+			Query:         testPlaybookQuery,
+			OperationName: "Playbook",
```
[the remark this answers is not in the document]
crspeller: This is a query not an update. This appears to be idiomatic query naming. You just name the query after the object you get.

webapp/src/types/playbook.ts
```diff
@@ -68,7 +68,7 @@ export interface Metric {
     type: MetricType;
     title: string;
     description: string;
-    target: number | null;
+    target?: number | null;
```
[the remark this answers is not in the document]
crspeller: Yes, that is still allowed.

webapp/src/graphql/playbook.graphql
crspeller: Probably for every query.

webapp/src/graphql/hooks.ts
```diff
@@ -0,0 +1,33 @@
...
+        nextFetchPolicy: 'cache-first',
+    });
+
+    let playbook = result.data?.playbook;
+    playbook = playbook === null ? undefined : playbook; //eslint-disable-line no-undefined
+
+    return [playbook, result];
+};
+
+export const useUpdatePlaybook = (id?: string) => {
```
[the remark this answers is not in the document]
crspeller: It returns a function that you use to update the playbook.

server/sqlstore/playbook.go
crspeller: This is the existing pattern we use just with a more systematic way of constructing it. It's no different then doing this: https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/sqlstore/playbook.go#L608

server/main.go
crspeller: Separate was my first plan. Ran into issues with the go embed stuff and getting a separate binary to import and read that embedded file. This ended up being simpler.

server/sqlstore/playbook.go
lieut-data: The significant difference is that the `sqlstore` owns the mapping from struct fields to column names, keeping the boundaries of ownership crisp. This approach feels much more like exposing the `p.store.db` and letting the app layer run "arbitrary" queries.

PR #1190, opened 2022-05-17
MM-44187: s/surveybot/feedbackbot

lieut-data:
#### Summary
The `surveybot` was recently renamed to `feedbackbot`. Fix our handling of the `Give Feedback` button and add a test to verify the functionality going forward.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-44187

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/server/bot/bot.go
+++ b/server/bot/bot.go
@@ -72,7 +72,7 @@ type Poster interface {
 	// NotifyAdmins sends a DM with the message to each admins
 	NotifyAdmins(message, authorUserID string, isTeamEdition bool) error
 
-	// PromptForFeedback sends a DM as the surveybot to the given user, prompting for product feedback.
+	// PromptForFeedback sends a DM as the feedbackbot to the given user, prompting for product feedback.
 	PromptForFeedback(userID string) error
 
 	// IsFromPoster returns whether the provided post was sent by this poster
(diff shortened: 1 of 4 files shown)


Phrynobatrachus:
Tests likely require some additional setup, looking into it.

Phrynobatrachus:
Two ways to handle the tests:

1) set `PluginSettings.PluginStates."com.mattermost.nps".Enable: true` in `on_prem_default_config` and `cloud_default_config`

2) call the plugin enable/disable api in this spec

No preference which.

lieut-data:
Thank you for the quick investigation, @Phrynobatrachus! I've pushed a change to enable NPS by default.

Phrynobatrachus:
Looks like setting the env var `MM_PLUGINSETTINGS_AUTOMATICPREPACKAGEDPLUGINS: true` for the `e2e-cypress-tests-...` jobs in the Circle config gets the tests to pass.

agarciamontoro (approved):
Looks good, thank you!

lieut-data:
> Looks like setting the env var MM_PLUGINSETTINGS_AUTOMATICPREPACKAGEDPLUGINS: true for the e2e-cypress-tests-... jobs in the Circle config gets the tests to pass.

Thanks again @Phrynobatrachus! I wonder if we explicitly disabled this to avoid auto-installing playbooks only to then replace it. I'll give it a whirl either way :)

PR #1192, opened 2022-05-17
MM-44323: disable duplicate if disallowed

lieut-data:
#### Summary
If a team disallows public playbook creation, disable the duplicate option for playbooks on that team:

CleanShot 2022-05-17 at 16 38 55@2x

I investigated adding a unit test, but found the effort to customize the system scheme or team scheme out of scope for this small fix.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-44323

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/dot_menu.tsx
```diff
@@ -164,15 +164,43 @@ export const DropdownMenuItemStyled = styled.a`
...
+    cursor: default;
+    font-family: 'Open Sans';
+    font-style: normal;
+    font-weight: normal;
+    font-size: 14px;
+    color: var(--center-channel-color-40);
+    padding: 10px 20px;
+    text-decoration: unset;
+}
+`;
```
agarciamontoro, line 190: I'm not sure I understand this, aren't the styles duplicated between `DisabledDropdownMenuItemStyled` and `DropdownMenuItemStyled`? We could simply pass the `props.disabled` to `DropdownMenuItemStyled` and add the different `color` and `cursor` if that prop is true.
We can also just avoid the different return in `DropdownMenuItem` below if we pass `props.disabledAltText` conditionally to `title`, right?

agarciamontoro (changes requested):
The logic is good, but I have a doubt about the CSS below.

webapp/src/components/dot_menu.tsx
agarciamontoro: Ah, I didn't catch that. It makes sense to have two returns, then. Thanks for the context :)

lieut-data:
Thanks again for the help, @calebroseland! Now looks like:

image

calebroseland (approved):
Looks good! For expediency, I pushed a brief set of changes:
- Intl
- Added to playbook editor
- (while there) Added a small UX tweak from feedback: default to outline tab as duplicated playbooks will have zero usage on first load

CleanShot 2022-05-31 at 17 01 30@2x

webapp/i18n/en.json
```diff
@@ -250,6 +250,7 @@
   "Sx3lHL": "Integer",
   "T5rX+W": "How often should an update be posted?",
   "TBez4r": "There are no playbooks to view. You don't have permission to create playbooks in this workspace.",
+  "TD8WrM": "Duplicate is disabled for this team.",
```
lieut-data, line 253: Sigh, I had done this correctly in an earlier cherry-pick, but forgot after cleaning up. Thank you!

agarciamontoro (approved):
:tada:

PR #1193, opened 2022-05-18
MM-44160: Fix empty channel selector

agarciamontoro:
#### Summary
**The problem**: When visiting the run overview page directly, the Redux store is not pre-populated by Channels, so the Run Actions modal did not have the information on the channels the user is a member of, thus showing an empty dropdown menu in the channel selector (and showing any channel already added as `Unknown channel`, even if the user had access to them).

**The solution**: When first mounting the channel selector, if the store is empty, just populate the Redux store with the channels the user is a member of.

@lieut-data, I was considering adding a cache similar to #1177 to prevent a problem of the same type in the future in this component. I decided not to do it, though, as it seems rare that we will have more than one of these components mounted at the same time, and much more rare to have a lot of them. But 0/5 on adding that preemptively. Thoughts?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44160

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/runs/actions_spec.js
+++ b/tests-e2e/cypress/integration/runs/actions_spec.js
@@ -33,7 +33,7 @@ describe('runs > actions', () => {
     });
 
     beforeEach(() => {
-        const now = Date.now()
+        const now = Date.now();
         const runName = 'Playbook Run ' + now;
         cy.apiRunPlaybook({
             teamId: testTeam.id,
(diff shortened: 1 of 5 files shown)


lieut-data (approved):
Yes, that makes sense to me @agarciamontoro, given only one of these components is expected.

Phrynobatrachus:
I think this might not solve the problem. My test steps:

1) `make clean-docker` + `make test-data`
2) login as sysadmin, create a playbook with broadcast enabled to a channel, start a run
3) in a different browser session, login as user-1, navigate to the run details + run actions modal
4) I see
Screenshot from 2022-05-19 11-56-35

Will dig in some more.

agarciamontoro:
@Phrynobatrachus, thank you for the test. I identified two things here:
- If the user does not have access to the run channel, then we cannot get its team and thus the component does not know where to pull the channels from. This causes the selector to be empty
-  If the user does not have access to the broadcast channel, then it cannot retrieve its name. This causes the badge to show `Unknown channel`.

Let me see what I can do.

agarciamontoro:
https://github.com/mattermost/mattermost-plugin-playbooks/pull/1193/commits/9d52211a7da5285e0725ed378a85dab3958d3910 should have fixed the first problem. The second one is a bit trickier and it's already too late in this part of the world, so I'll continue on Monday.

gbochora (approved):
Looks nice!

agarciamontoro:
@Phrynobatrachus, I think both problems are now fixed. Could you please take a second look? Thanks!

Phrynobatrachus (approved):
Seems to be fixed now, thanks! :+1:

Phrynobatrachus:
The test failures do appear to be real. 

1. create a private channel on a team
2. create or edit playbook > open the broadcast channel dropdown
3. the private channel from #1 is listed, but when on selection the badge is 'Unknown channel`

Also doubling checking the original scenario some more, now I'm unsure.

Phrynobatrachus:
Ok I think I did miss something, here's a more detailed example of what I originally looked at:

1) as sysadmin, create a playbook on eligendi
  - public
  - user-1 is not a member
  - configured 3 broadcast channels, 1 public, 1 private where is user-1 also a member, 1 private where user-1 is not a member

2) start the run

3) in a different browser session, login with a redirect straight to the run details, open the run actions modal

Screenshot from 2022-05-23 14-07-40

 `Town square` is correct, but I'm expecting to see the name of the private channel user-1 IS a member of, plus 1 `Unknown channel` for the private user-1 is not a member of.

Phrynobatrachus:
Nice, I think we're good now. :crossed_fingers:

Phrynobatrachus (approved):

PR #1194, opened 2022-05-19
do not log error if 0 actions in channel when a user joins

trilopin:
#### Summary
Avoid logging errors if we don't have actions in a channel and a new member joins.

Additionally, I changed the order in `getPlaybookRunByChannel` to have a more relevant error (before it reported permissions failure if the run does not exist).  Personally, I would omit this kind of error from the logs since it seems like an expected flow to me (to reproduce leave and rejoin a channel without a run).

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-43636

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


agarciamontoro (approved):
This makes sense, thank you!

PR #1197, opened 2022-05-20
modal manually centered

trilopin:
#### Summary
Vertically center *more* the modals for run&channel actions.

I tried the centered property in react-modal but there were lots of conflicting injected styles. The channel actions modal has a fixed height, but the run actions modal has a variable height, so in some situations could not be perfectly aligned.
 
#### Ticket Link

Fixes: https://mattermost.atlassian.net/browse/MM-44196

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/actions_modal.tsx
+++ b/webapp/src/components/actions_modal.tsx
@@ -22,6 +22,7 @@ interface Props {
     editable: boolean;
     onSave: () => void;
     children: React.ReactNode;
+    adjustTop?: number;
 }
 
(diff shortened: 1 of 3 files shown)


agarciamontoro (approved):
> I tried the centered property in react-modal but there were lots of conflicting injected styles.

This is indeed a mess and we should solve it at some point, because it's creating more work than it helps us.

But this is a good solution for now, thank you!

gbochora (approved):

PR #1198, opened 2022-05-20
RHS: keep full assignee pill when due date and not state-open

trilopin:
#### Summary
Do not shrink assigned user pill (hide name) when due date but state closed or skip. Since due date is hidden, the user name should be shown.  

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-44194

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/channels/rhs/checklist_spec.js
+++ b/tests-e2e/cypress/integration/channels/rhs/checklist_spec.js
@@ -369,7 +369,7 @@ describe('channels > rhs > checklist', () => {
             cy.findAllByTestId('checkbox-item-container').should('have.length', 12);
         });
 
-        it.only('switching between runs with the same checklist', () => {
+        it('switching between runs with the same checklist', () => {
             // # Create another run using the same playbook
             const playbookRunName2 = 'RunWithSameChecklist';
             cy.apiRunPlaybook({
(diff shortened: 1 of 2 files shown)


tests-e2e/cypress/integration/channels/rhs/checklist_spec.js
```diff
@@ -369,7 +369,7 @@ describe('channels > rhs > checklist', () => {
             cy.findAllByTestId('checkbox-item-container').should('have.length', 12);
         });
 
-        it.only('switching between runs with the same checklist', () => {
```
trilopin, line 372: I assumed this `only` wasn't intended, but please confirm that @gbochora

gbochora: Ups, I added it for testing on the local machine. Nice catch!

agarciamontoro (approved):
:tada:

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -129,11 +129,23 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
         }
 
+        const shouldHideName = () => {
+            if (isEditing) {
+                return false;
+            }
+            if (command !== '') {
+                return true;
+            }
+            if (dueDate > 0 && props.checklistItem.state === ChecklistItemState.Open) {
```
gbochora, line 139: Should we add `ChecklistItemState.InProgress` state check too? 
I'm not sure about it because it seems we never assign `InProgress` state to the task.

gbochora (approved):
🎉

PR #1199, opened 2022-05-20
force empty title in menuItem to prevent trigger title from propagating

trilopin:
#### Summary
Force the` DropdownMenuItem` title to an empty string. It's needed to prevent the trigger icon from appearing for all his children.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-44193

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/dot_menu.tsx
+++ b/webapp/src/components/dot_menu.tsx
@@ -182,6 +182,10 @@ export const DropdownMenuItem = (props: { children: React.ReactNode, onClick: ()
             onClick={props.onClick}
             className={props.className}
             role={'button'}
+
+            // Prevent trigger icon (parent) from propagating title prop to options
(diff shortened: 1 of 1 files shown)


agarciamontoro (approved):
Code looks good, but I don't understand why the prop was propagated to the child. We are not passing it and that's not inherited, right?

gbochora (approved):
Nice!

trilopin:
> Code looks good, but I don't understand why the prop was propagated to the child. We are not passing it and that's not inherited, right?

I have to admit that is not very elegant, just effective. Title prop propagates to children by default. Check this codepen https://codepen.io/trilopin/pen/PoQKBeK.

PR #1203, opened 2022-05-20
MM-44410 Add subresource integrity to graphql html

crspeller:
#### Summary
Adding subresource integrity to the graphqli html.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44410


agnivade (approved):
Oo nice, mind sending this for the server as well? :)

crspeller:
@agarciamontoro You could download the files and run the hash yourself. (it's sha256) But the point of this is that the browser does it automatically so it will fail if the hashes don't match.

PR #1204, opened 2022-05-22
MM-43951 Adding additional KPIs causes 500 error

gbochora:
Update store tests

Add api tests

#### Summary
Metrics fields on the retrospective form are mandatory, so a user must set all values in order to publish the retrospective. Based on this, I assumed that if the Playbook has several metrics, there will be no case when we have values for one of the metrics and don't have for the other. In other words, either all metrics have values in the database, or we don't have data for any of them.
My assumption was wrong because a user can add a metric later, i.e., we have a playbook with a single metric, then we start a run for this Playbook and publish a retrospective. After that, if we add another metric to the Playbook, we have a case when one metric has value in the database and another doesn't. This was causing a `slice bounds out of range` error on the backend.

I have updated store methods, so they will return a `nil` value if no data is available for the specific metric:

- `MetricOverallAverage` gets the Playbook metrics configuration and then maps it with average values. If the average value of the metric is not available, the corresponding element is set to `nil`

- `MetricValueRange` - same as in the `MetricOverallAverage` function.

- `MetricRollingAverageAndChange` - If the first period values of the metric are not available, then the `period average` and `change` are set to `nil`. If the second period values are not available, the `period change` value is set to `nil`.

I replaced metrics tests for the store layer with new ones. Old ones contained data manipulation logic similar to store layer functions, which added complexity. 
Added server `E2E` tests, basically the same as store layer tests, but I think they are still helpful.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43951

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


server/sqlstore/stats.go
```diff
@@ -333,42 +333,61 @@ func (s *StatsStore) ActiveParticipantsPerDayLastXDays(x int, filters *StatsFilt
...
-func (s *StatsStore) MetricOverallAverage(filters *StatsFilters) []int64 {
+// Returns empty list when Playbook doesn't have configured metrics
+// If for some metrics there are no published values, the corresponding element will be nil in the resulting slice
+func (s *StatsStore) MetricOverallAverage(filters *StatsFilters) []null.Int {
 	query := s.store.builder.
-		Select("FLOOR(AVG(m.Value))").
+		Select("mc.ID as ID, FLOOR(AVG(m.Value)) as Value").
 		From("IR_Metric as m").
 		InnerJoin("IR_MetricConfig as mc ON m.MetricConfigID = mc.ID").
 		GroupBy("mc.ID").
```
trilopin, line 345: 200% optional

Wow, seeing the groupby clause before the where clause is shocking to me, I guess it doesn't matter to squirrel but it's not very readable in my SQL-brain

server/sqlstore/stats.go
```diff
@@ -333,42 +333,61 @@ func (s *StatsStore) ActiveParticipantsPerDayLastXDays(x int, filters *StatsFilt
...
 	type MinMax struct {
-		Min null.Int
-		Max null.Int
+		ID  string
+		Min int64
+		Max int64
 	}
 	q := s.store.builder.
-		Select("MIN(Value) as Min, MAX(Value) as Max").
+		Select("mc.ID as ID, MIN(Value) as Min, MAX(Value) as Max").
```
trilopin, line 390: why are not we filtering by playbookID here? I guess it works because later we cross data with playbook configs.

trilopin: I'm a bit confused about two things about the approach:
- why do we query separately avg from min/max?
- why do we need an additional query to metricconfigs and cross data if we already joined the table in the first query?

server/sqlstore/stats.go
```diff
@@ -333,42 +333,61 @@ func (s *StatsStore) ActiveParticipantsPerDayLastXDays(x int, filters *StatsFilt
...
+// MetricOverallAverage for a specific playbook returns a list that contains an average value for each metric.
 // Only published metrics values are included.
-// Returns empty list when Playbook doesn't have metrics or there are no any published metrics values
-func (s *StatsStore) MetricOverallAverage(filters *StatsFilters) []int64 {
+// Returns empty list when Playbook doesn't have configured metrics
+// If for some metrics there are no published values, the corresponding element will be nil in the resulting slice
+func (s *StatsStore) MetricOverallAverage(filters *StatsFilters) []null.Int {
 	query := s.store.builder.
-		Select("FLOOR(AVG(m.Value))").
+		Select("mc.ID as ID, FLOOR(AVG(m.Value)) as Value").
```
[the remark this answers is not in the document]
gbochora: There is a check in the API layer(`parsePlaybookStatsFilters` function). If PlaybookID is missing, we return an error. Do you think we still need validation in this function?

server/sqlstore/stats.go
gbochora: Very good point! I'll move it.

server/sqlstore/stats.go
gbochora: 1. We have separate sections in the UI for `Range values`(Min, Max) and for `Average` values. Because of that, I've decided to create separate functions. But we never need only average or only range, so It will be better to retrieve them in a single query.
2. The first query will not include information about metrics that don't have any values in the database. We need configurations to get all existing metrics and build a result array, where `valueRange[i]` will be min-max values for the `metricConfigs[i]` or `nil` if there is no data in the database for this specific metric.

trilopin: 1) completely optional 
2) got it

trilopin: I would add a short comment explaining why we need the second query and recomputation. Is not that intuitive from the code and someone could be tempted to "simplify".

server/sqlstore/stats.go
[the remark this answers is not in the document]
trilopin: I would check as soon as func starts to have an early return, even if the API handles it.
It's cheap and we'll become a bit more defensive.

server/sqlstore/stats.go
```diff
@@ -333,42 +333,61 @@ func (s *StatsStore) ActiveParticipantsPerDayLastXDays(x int, filters *StatsFilt
...
+	var averages []Average
 	if err := s.store.selectBuilder(s.store.db, &averages, query); err != nil {
 		s.log.Warnf("Error retrieving stat total active participants %w", err)
-		return []int64{}
+		return []null.Int{}
+	}
+
+	configs, err := s.retrieveMetricConfigs(filters.PlaybookID)
+	if err != nil {
+		s.log.Warnf("Error retrieving metrics configs ids for playbook %w", err)
```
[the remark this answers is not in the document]
iomodo: `s.retrieveMetricConfigs` logs the same thing, so this logging is redundant.
Either `s.retrieveMetricConfigs` should not return error at all, or it should not log

iomodo (changes requested):
Looks great @gbochora, couple of nits below

server/sqlstore/stats.go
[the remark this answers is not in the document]
gbochora: What do you think if we change the argument type from `filters *StatsFilters` to `filters StatsFilters`. In this case, there is no need for `nil` check.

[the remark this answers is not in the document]
trilopin: it works for me

client/playbook.go
```diff
@@ -36,6 +36,12 @@ type PlaybookMember struct {
 	SchemeRoles []string `json:"scheme_roles"`
 }
 
+const (
+	MetricTypeDuration = "metric_duration"
```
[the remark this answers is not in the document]
gbochora: It would be better, but we will also need to change the `app/playbook` code. So, I would rather not do it in this PR. It already has quite large changes for a bug fix.

trilopin (approved):
Cool @gbochora , nice job

iomodo (approved):
LGTM!

PR #1208, opened 2022-05-23
usePortal cleanup side effects in useEffect

trilopin:
#### Summary
We are currently leaving tons of empty div nodes in the DOM.
 
Screenshot 2022-05-23 at 18 55 41

This is done by usePortal hook which adds a new div node at the end of the body to make drag&drop possible. We are currently adding one portal (div node) per draggable item, but we don't remove these nodes when we unmount the component which needed them.

This could degrade perf if opened and closed several times in the same session the Run RHS since it's creating a div node per draggable item (and one run could have dozens).

I added a call to the cleanup part of useEffect to remove the nodes once the component is unmounted. After this change, all div are removed once we close the RHS.

Further discussion: 
- do we need a portal per element? 
- can we share one portal among all checklist items?  

#### Ticket Link
No ticket

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/hooks/general.ts
+++ b/webapp/src/hooks/general.ts
@@ -636,9 +636,9 @@ export const usePrevious = (value: any) => {
 // Create a portal to render while dragging
 export const usePortal = (parent: HTMLElement) => {
     const [portal] = useState(document.createElement('div'));
-
     useEffect(() => {
         parent.appendChild(portal);
+        return () => portal.remove();
     }, [parent, portal]);
 
     return portal;


calebroseland (commented):
> do we need a portal per element?

I would say not, [webapp uses a shared `root-portal`](https://github.com/mattermost/mattermost-webapp/blob/master/root.html#L81)

Focalboard uses a plugin-level [`focalboard-root-portal`](https://github.com/mattermost/focalboard/blob/main/mattermost-plugin/webapp/src/index.tsx#L146)


> can we share one portal among all checklist items?

I would say yes


Further, I'd suggest we adopt webapp and focalboard's strategy of creating a wrapper component that handles portal creation/mounting for us (see: [focalboard](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/rootPortal.tsx#L11) or [webapp](https://github.com/mattermost/mattermost-webapp/blob/master/components/root_portal.jsx#L8))

iomodo (approved):
Wonderful @trilopin. I also like the idea about shared portal. We could track that with a different ticket, if you wish, 0/5.

trilopin:
I refactored this a bit. 

The main points are:
- better container: the portal is created inside playbooks-root-portal (registered from as root component by the plugin), not body element
- cleanup after unmounting: hook creates element inside playbooks-root-portal and cleanup when unmounted
- only one portal: a component that wraps the hook is provided to ease createPortal call and to enable conditional rendering (portal is only created when we drag and removed after drop)
- hook still can be used directly if you don't need to do conditional rendering

trilopin:
Just a note, a one-only portal is achieved not by having one dedicated container (body vs xxx-root-portal) but by providing an unmount-and-cleanup mechanism. That's the reason why I wrapped the hook, otherwise, we would end up creating as many divs as draggable items inside xxx-root-portal.

webapp/src/components/portal.tsx
```diff
@@ -0,0 +1,15 @@
...
+type Props = {
+    children: React.ReactNode
+}
+
+const Portal = ({children}: Props) => {
+    const el = usePortal();
+    return ReactDOM.createPortal(children, el);
+};
+
+export default React.memo(Portal);
```
iomodo, line 15: missing enter

iomodo (approved):
Looks great @trilopin !

webapp/src/hooks/general.ts
```diff
@@ -633,15 +636,20 @@ export const usePrevious = (value: any) => {
...
-
-    useEffect(() => {
-        parent.appendChild(portal);
-    }, [parent, portal]);
-
-    return portal;
+export const usePortal = () => {
+    const [el] = useState(document.createElement('div'));
+    useLayoutEffect(() => {
+        const rootPortal = document.getElementById('playbooks-root-portal');
```
calebroseland, line 642:
```suggestion
        const rootPortal = document.getElementById('root-portal');
```

webapp/src/index.tsx
```diff
@@ -148,6 +148,9 @@ export default class Plugin {
         registry.registerRootComponent(PostMenuModal);
         registry.registerRootComponent(ChannelActionsModal);
 
+        // Playbooks shared portal
+        registry.registerRootComponent(() => <div id='playbooks-root-portal'/>);
+
```
calebroseland, line 153:
```suggestion
```

calebroseland (changes requested):
I'd like to propose that we just use `#root-portal`, it's static/constant and I don't see any harm in sharing it.

Main reason being, in Playbooks ("backstage"), `<Pluggable name='Root'/>` isn't available, so it's currently broken for the new editor.

CleanShot 2022-05-25 at 10 29 29

trilopin:
> I'd like to propose that we just use `#root-portal`, it's static/constant and I don't see any harm in sharing it.
> 
> Main reason being, in Playbooks ("backstage"), `<Pluggable name='Root'/>` isn't available, so it's currently broken for the new editor.


I'm fine with that, but since it depends on webapp repository markup I will add a cypress test that ensures that di exists.

calebroseland (approved):
Thanks, @trilopin! Good idea on the E2E test—that'll let us know immediately if that ever changes in the future.

PR #1214, opened 2022-05-26
add computed schemeRoles to playbook query response -  fix start run

trilopin:
#### Summary
A non-admin user could not start a run from the new playbook editor.
That happened because the permissions check was expecting scheme_roles but they were not provided by graphql call.

Scheme roles are correctly computed in server and now, additionally included in the response.
Server and client schemas have been modified.

I'm 3/5 about this: Is there any solid reason to prevent schemeRoles from being sent to the webapp?

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-44550

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated


trilopin:
I was about to add an e2e test for this but I discovered that e2e is temporarily ignored. @calebroseland let me know if you want help recovering those e2e.

calebroseland (approved):

PR #1218, opened 2022-05-27
MM-44498: Fix typing UX on channel name input

calebroseland:
#### Summary

- Stabilizes `updatePlaybook` identity 
- Adds `useProxyState` hook to debounce side effects or upstream state-setting
- Debounces channel name input with `useProxyState`

CleanShot 2022-05-27 at 10 32 22

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44498

#### Checklist

Item | Y / N / NA [^1]
---|---
Telemetry updated? |
Gated by experimental feature flag? |
Unit/E2E tests updated? |

### DIFF
--- a/webapp/src/hooks/general.ts
+++ b/webapp/src/hooks/general.ts
@@ -39,7 +39,9 @@ import {useHistory, useLocation} from 'react-router-dom';
 import qs from 'qs';
 import {haveITeamPermission} from 'mattermost-webapp/packages/mattermost-redux/src/selectors/entities/roles';
 
-import {Argument} from 'classnames';
+import {useUpdateEffect} from 'react-use';
+
+import {debounce, isEqual} from 'lodash';
 
 import {FetchPlaybookRunsParams, PlaybookRun} from 'src/types/playbook_run';
 import {EmptyPlaybookStats} from 'src/types/stats';
@@ -58,7 +60,7 @@ import {
     isCurrentUserAdmin,
     myPlaybookRunsByTeam,
 } from '../selectors';
-import {formatText, messageHtmlToComponent} from 'src/webapp_globals';
+import {resolve} from 'src/utils';
 
 /**
  * Hook that calls handler when targetKey is pressed.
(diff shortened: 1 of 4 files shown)


webapp/src/hooks/general.ts
```diff
@@ -662,3 +664,47 @@ export const useScrollListener = (el: HTMLElement | null, listener: EventListene
...
+    const [value, setValue] = useState(prop);
+
+    useUpdateEffect(() => {
+        if (!isEqual(value, check.current)) {
+            // check failed; don't destroy pending changes (values set mid-cycle between send/sync)
+            return;
+        }
+        check.current = prop; // sync check
+        setValue(prop);
+    }, [prop]);
```
calebroseland, line 694: This update effect synchronizes the inner value when the prop changes—multi-tab editing, multiple people editing different things on the same playbook, etc.—but only if the check passes. This prevents lost chars and typing glitches without cutting off `value` from outer-updates.

crspeller (approved):
This is super complicated. I am not sure if I totally understand it. But I also understand why we need it. Hopefully we can remove this with some of the GraphQL performance work we might be doing soon.

calebroseland:
> This is super complicated. I am not sure if I totally understand it. But I also understand why we need it. Hopefully we can remove this with some of the GraphQL performance work we might be doing soon.


> This is super complicated. I am not sure if I totally understand it. But I also understand why we need it. Hopefully we can remove this with some of the GraphQL performance work we might be doing soon.

Very much agreed that this should be temporary—it'd be better to fix the root of the problem in the GraphQL layer, that solution would also scale better to fix other things not behind an edit+save gate like drag&drop, toggles, etc.

calebroseland:
> This is super complicated. I am not sure if I totally understand it. But I also understand why we need it. Hopefully we can remove this with some of the GraphQL performance work we might be doing soon.


> This is super complicated. I am not sure if I totally understand it. But I also understand why we need it. Hopefully we can remove this with some of the GraphQL performance work we might be doing soon.

Very much agreed that this should be temporary—it'd be better to fix the root of the problem in the GraphQL layer, that solution would also scale better to fix other things not behind an edit+save gate like drag&drop, toggles, etc.

iomodo (approved):

PR #1219, opened 2022-05-27
MM-44508 Add ESC and Enter handing to text inputs.

crspeller:
#### Summary
Adding ESC and Enter handing for text inputs. 
Shift-Enter saves for multiline inputs while unmodified enter saves for single line.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44508


calebroseland (approved):
LGTM. For Save I think the designs call for cmd + enter, and I've seen that used elsewhere in mattermost. (And maybe ctrl + enter on non-mac?). @abhijit-singh thoughts?

PR #1230, opened 2022-06-01
new run details page skeleton

trilopin:
#### Summary
Skeleton to start handling the next work on the full-featured run detail page

No RHS yet, not all logic copied from old component (actually I would want to refactor a bit the logic). The goal is to make parallel work on this page possible.

To see it, visit a run and the change URL to be /playbooks/run_details/<runid>
To develop, go to folder: components/backstage/playbook_runs/playbook_run

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44712

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


gbochora (approved):
Cool! Just a minor thing, some files are missing a newline at the end:
`checklists.tsx, finish_run.tsx, index.ts, retrospective.tsx, status_update.tsx, summary.tsx`

webapp/src/components/backstage/playbook_runs/playbook_run/index.ts
```diff
@@ -0,0 +1,4 @@
+import PlaybookRun from './playbook_run';
+
+export default PlaybookRun;
+export {};
```
agarciamontoro, line 4: Why do we need this index file?

webapp/src/hooks/routing.ts
```diff
@@ -1,10 +1,9 @@
 import {useEffect, useMemo} from 'react';
+import {useSelector} from 'react-redux';
```
trilopin, line 2: This is because of the merge conflict and npm-run fix

if you want I can revert it

agarciamontoro (approved):
Looks good! Just two things: the new lines at the end of the files, as @gbochora said, and the comment I left below. But all non-blocking, so approving already! :tada:

webapp/src/components/backstage/playbook_runs/playbook_run/index.ts
trilopin: we don't need it but makes easier the import because the parent component only has to know about this file, the rest of nested files are "more" private to external components.

PR #1236, opened 2022-06-03
MM-44763: allow direct dropdown editing again

lieut-data:
#### Summary
We unintentionally required users to enter edit mode for checklist items to be able to change the owner or due dates, but these should have stayed accessible directly:

image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-44763

### DIFF
--- a/webapp/src/components/checklist_item/checklist_item.tsx
+++ b/webapp/src/components/checklist_item/checklist_item.tsx
@@ -141,10 +141,11 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement => 
             }
             return false;
         };
+
         return (
             <AssignTo
                 assignee_id={assigneeID || ''}
-                editable={isEditing}
+                editable={!props.disabled}
                 withoutName={shouldHideName()}
                 onSelectedChange={onAssigneeChange}
             />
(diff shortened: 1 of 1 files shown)


webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -141,10 +141,11 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
             }
             return false;
         };
+
         return (
             <AssignTo
                 assignee_id={assigneeID || ''}
-                editable={isEditing}
+                editable={!props.disabled}
```
lieut-data, line 148: By constraining to `props.disabled`, we disallow changing these fields when the run has completed.

crspeller (approved):

lieut-data:
(Going to merge this as I believe it's as required, and cut an updated v1.28. We still have time to pivot as needed.)

abhijit-singh:
Sorry for the delay here @lieut-data. A couple of minor comments for the collapsed assignee chip:
1. There should be a background-colour (--center-channel-color-08) even in the default(non-hover) state. 
image
2. There's no margin to the right of the arrow icon. Ideally, there should be about 8px space. 

Ideal:
image
Current: 
image

lieut-data:
Thank you for following up, @abhijit-singh! I'm going to take a look at this in https://mattermost.atlassian.net/browse/MM-44899 to make sure we follow up.

PR #1239, opened 2022-06-06
RHS support rundetails

trilopin:
#### Summary
Naive support for RHS inside playbook run

This will allow us to work on the content of the RHS. It's not completely finished, and some improvements and changes will be necessary to support more sizes/browsers.

RHS will need to be moved to layout or even backstage, but if we start from here we allow ourselves to avoid blocking each other.

I also made "finish run" optional in the checklist, since a new component is needed for the run details page.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44711

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -0,0 +1,109 @@
...
+interface Props {
+
+    // playbookRun: PlaybookRun;
+    isOpen: boolean;
+    onClose: () => void;
+    section: RHSContent;
+}
+
+const RightHandSidebar = (props: Props) => {
+    const sidebarRight = React.useRef(null);
```
agarciamontoro, line 22: Maybe that's just me, but I like to name the ref variables as `whateverRef`. Maybe `sidebarRightRef`? Or `sidebarRef` so it is shorter?

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -0,0 +1,109 @@
...
+    case RHSContent.RunTimeline:
+        title = formatMessage({defaultMessage: 'Timeline'});
+        break;
+    case RHSContent.RunParticipants:
+        title = formatMessage({defaultMessage: 'Participants'});
+        break;
+    case RHSContent.RunStatusUpdates:
+        title = formatMessage({defaultMessage: 'Status updates'});
+        break;
+    }
```
agarciamontoro, line 44: Should the title be provided by the caller? I think that's a bit more flexible.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -0,0 +1,109 @@
...
+        </Container>);
+};
+
+export default RightHandSidebar;
+
+const Container = styled.div<{isOpen: boolean}>`
+    display: ${({isOpen}) => (isOpen ? 'flex' : 'hidden')};
+    width: 400px;
+    height: 100%;
+    z-index: 11;
```
agarciamontoro, line 72: Isn't there a way to avoid the z-index here? Maybe using a different `position`?

agarciamontoro (changes requested):
Looks good! :tada: I left some comments below

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
trilopin: I removed it for this naive iteration. As soon as we need better and deeper integration more hacks will be needed (positioning or z-index or other mechanisms)

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -0,0 +1,108 @@
...
+    justify-content: flex-end;
+    margin-right: 20px;
+`;
+
+const HeaderTitle = styled.div`
+    margin: auto 20px;
+    line-height: 32px;
+    font-size: 16px;
+    font-weight: 600;
+    color: var(--center-channel-color-rgb);
```
agarciamontoro, line 101: I think you don't need the `-rgb` here:
```suggestion
    color: var(--center-channel-color);
```

agarciamontoro (approved):
Thanks, Jose! I left a non-blocking comment below.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -92,16 +103,21 @@ const ColumnContainer = styled.div`
...
     flex-direction: row;
 `;
-const Header = styled.header`
-    min-height: 56px;
-    width: 100%;
-`;
 
-const Body = styled.main`
+const Main = styled.main`
     max-width: 780px;
```
gbochora, line 108: Nit: `max-width` property is duplicated.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -0,0 +1,108 @@
...
+                </HeaderIcon>
+            </Header>
+            <Body>{content}</Body>
+        </Container>);
+};
+
+export default RightHandSidebar;
+
+const Container = styled.div<{isOpen: boolean}>`
+    display: ${({isOpen}) => (isOpen ? 'flex' : 'hidden')};
```
gbochora, line 69: Do we need `isOpen` parameter? It seems its value will always be `true`

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -0,0 +1,108 @@
...
+    width: 400px;
+    height: 100%;
+    width: 400px;
```
gbochora, line 72: Nit: `width` property is duplicated.

gbochora (approved):
Looks great! Thanks @trilopin 🥳  I left a few non-blocking comments.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
trilopin: the open state is managed in the parent component, the main idea is to avoid unmounting rhs when closing it. 

```ts
            <RightHandSidebar
                isOpen={isRHSOpen}
                section={RHSContent.RunInfo}
                onClose={() => setIsRHSOpen(false)}
            />
```

PR #1248, opened 2022-06-13
Mm 44717 add retrospective

gbochora:
#### Summary
This PR adds the `Retrospective` section with existing UI to the run details main form. 
Mainly I've copied the `Retrospective` section code from the `backstage/playbook_runs/playbook_run_backstage/retrospective/retrospective.tsx` and made a few modification to fit it in the new run details form:
- Add `Copy link` to the title
- Fix `Upgrade` banner UI
- Remove timeline section
- Refactor `PlaybookRun` update functionality. Use `setPlaybookRun` instead of several small functions to update the run object.

https://user-images.githubusercontent.com/4368372/173279394-0173ea58-de5a-44fe-bfc5-d020cac20d6f.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44717

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,258 @@
...
+            <TextContainer>{publishButtonText}</TextContainer>
+        </PrimaryButtonSmaller>
+    );
+
+    const isPublished = playbookRun.retrospective_published_at > 0 && !playbookRun.retrospective_was_canceled;
+    if (isPublished) {
+        const publishedAt = (
+            <Timestamp
+                value={playbookRun.retrospective_published_at}
+                {...ELAPSED_TIME}
```
trilopin, line 102: I also did this, at some point we will need to converge.

https://github.com/mattermost/mattermost-plugin-playbooks/pull/1244/files#diff-0fba6ad2fcf358ae67462cb7347844a32561a72755d6d0b5c00c5f1a6c9e4b27R2

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,258 @@
...
+import Report from '../playbook_run_backstage/retrospective/report';
+import ConfirmModalLight from 'src/components/widgets/confirmation_modal_light';
+import {PrimaryButton} from 'src/components/assets/buttons';
+
+interface Props {
+    playbookRun: PlaybookRun;
+    setPlaybookRun: (value: React.SetStateAction<PlaybookRun | null>) => void
+}
+
+const editDebounceDelayMilliseconds = 2000;
```
trilopin, line 25: optional: constant style

```suggestion
const DEBOUNCE_2_SECS = 2000;
```

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,258 @@
...
+        publishRetrospective(playbookRun.id, playbookRun.retrospective, playbookRun.metrics_data);
+        setPlaybookRun((run) => ({
+            ...run,
+            retrospective_published_at: DateTime.now().valueOf(),
+            retrospective_was_canceled: false,
+        } as PlaybookRun));
+        setShowConfirmation(false);
+    };
+
+    const publishButtonText: React.ReactNode = formatMessage({defaultMessage: 'Publish'});
```
trilopin, line 80: isn't the type implicit?

```suggestion
    const publishButtonText = formatMessage({defaultMessage: 'Publish'});
```

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,258 @@
...
+    let publishComponent = (
```
trilopin, line 81: optional: I'd separate publishComponent to a local function renderPublishComponent with both cases

trilopin (changes requested):
Nice!

My only important point is if fetchPlaybook is really needed, I would do the useeffect-fetch in the parent component to avoid several sub-components doing their own fetch (potentially they fetching the same resource).

several optional comments

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,258 @@
...
+    const {formatMessage} = useIntl();
+    const [showConfirmation, setShowConfirmation] = useState(false);
+    const childRef = useRef<any>();
+    const metricsAvailable = useAllowPlaybookAndRunMetrics();
+
+    useEffect(() => {
+        const fetchData = async () => {
+            if (playbookRun?.playbook_id) {
+                const fetchedPlaybook = await clientFetchPlaybook(playbookRun.playbook_id);
+                setPlaybook(fetchedPlaybook ?? null);
```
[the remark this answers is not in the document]
gbochora: We need the playbook to get metrics configuration info: metric name, target value, metric type, and order of metrics. This info is used to build the retrospective form.

gbochora:
@trilopin I've added viewer mode for the retrospective. `Viewer` user will not be able to edit or publish retrospective.

webapp/src/components/backstage/playbook_runs/shared.tsx
```diff
@@ -181,6 +182,18 @@ const LinkTitle = styled.h3`
     }
 `;
 
+export const ELAPSED_TIME = {
+    useTime: false,
```
trilopin, line 186: Since this is not specific code to playbook_runs and already was extracted to time_spec. would make sense to use from there? Your comment regarding useTime was already fixed for the old retrospective.

gbochora: Yep, that's correct, I missed this part during the merge.

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,237 @@
...
+                        notificationType={AdminNotificationType.RETROSPECTIVE}
+                        verticalAdjustment={0}
+                        vertical={true}
+                    />
+                </BannerWrapper>
+            </Container>
+        );
+    }
+
+    const confirmedPublish = () => {
```
trilopin, line 63: optional, more consistent naming

```suggestion
    const onConfirmPublish = () => {
```

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,237 @@
...
+            }
+        }
+        setShowConfirmation(true);
+    };
+
+    const isPublished = playbookRun.retrospective_published_at > 0 && !playbookRun.retrospective_was_canceled;
+    const notEditable = isPublished || role === Role.Viewer;
+
+    const renderPublishComponent = () => {
+        const publishButtonText = formatMessage({defaultMessage: 'Publish'});
```
trilopin, line 87: if only is used once, would make sense that we simply put the formatmessage inside Publishbutton?

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,237 @@
...
+                    onClick={onPublishClick}
+                    disabled={notEditable}
+                >
+                    {publishButtonText}
+                </PublishButton>
+            </>
+        );
+    };
+
+    const persistMetricEditEvent = (metrics_data: RunMetricData[]) => {
```
trilopin, line 116: 100% optional and personal preference

I would love to have different naming here, EditEvent suffix does not say the important message to me. Examples like: saveMetrics storeMetrics persistMetics or onMetricsChange sound better to me. 

Additionally I'd create only one var already wrapped in debounce like 

```ts
const onMetricsChange = debounce((metrics_data: RunMetricData[]) => {
    onChange({...playbookRun, metrics_data});
    updateRetrospective(playbookRun.id, playbookRun.retrospective, metrics_data);
}, DEBOUNCE_2_SECS);
```

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -1,13 +1,237 @@
...
+                            playbookRun={playbookRun}
+                            onEdit={debouncedPersistReportEditEvent}
+                            flushChanges={() => debouncedPersistReportEditEvent.flush()}
+                            notEditable={notEditable}
+                        />
+                    </StyledContent>
+                </div>
+            ) : (
+                <RetrospectiveDisabledText id={'retrospective-disabled-msg'}>
+                    {formatMessage({defaultMessage: 'Retrospectives were disabled for this playbook run.'})}
```
trilopin, line 167: Shouldn't we hide the entire block in this case? https://community.mattermost.com/core/pl/tmd11zyxbjdpxn4r11umpp3nwh

gbochora: Nice catch! I forgot to remove this part, thanks.

trilopin (approved):
Thank you @gbochora !

calebroseland (approved):
I did run into a few issues, but I'm not sure if they would fit in this PR or if they're existing issues. Other than these, LGTM!


1) Metrics added after a retro is published show with a disabled Add Value prompt:
CleanShot 2022-06-16 at 23 56 37@2x

2) When you load a run with a retro, then navigate to another that doesn't have a run, and then you come back to the first, it doesn't show the metric values:
CleanShot 2022-06-16 at 23 52 25

3) When the Retrospective section UI isn't present, the content goes oddly narrow
CleanShot 2022-06-17 at 00 05 30@2x
(also shown in prior gif)

trilopin:
Point 2 is solved in another pr https://github.com/mattermost/mattermost-plugin-playbooks/pull/1256/files#diff-213f2b939a33ffd023fed43b598a2ea37dd385f32971886649f38b8ea67344f2R192

point 3 appeared after LHS integration, some changes must be done at state management.

PR #1252, opened 2022-06-14
MM-44962: Playbooks List and Playbook Editor UI changes to match new LHS experience

calebroseland:
#### Summary
Changes overall layout to match new designs but without additional functionality; new UX/functionality will come in subsequent PRs.

- Update Playbooks List design
  - Make full-width (constrained to `max-width: 1160px` on old page; unconstrained on LHS-enabled screen)
  - Minor page hierarchy changes
- Update Playbook Editor design
  - Make full-width 
- (https://github.com/mattermost/mattermost-plugin-playbooks/commit/a1c1496f46b81a778af9f6239cc0661bc5ed5d7a) Fix: Global Header menus disappearing on Playbooks List page in Safari
(I couldn't repro this one locally, but could repro on community—the fix applied in devtools fixes it)

New list screen
CleanShot 2022-06-13 at 23 29 31@2x

Current list screen (`width` constrained if `experimental == false`, all other changes ungated)
CleanShot 2022-06-13 at 23 48 58@2x

Global Header menus disappearing in Safari (community.mattermost.com)
image

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44962

#### Checklist

Item | Y / N / NA [^1]
---|---
Telemetry updated? | N
Gated by experimental feature flag? | Y&N
Unit/E2E tests updated? | N

### DIFF
--- a/webapp/src/components/backstage/playbook_list.tsx
+++ b/webapp/src/components/backstage/playbook_list.tsx
@@ -47,9 +47,11 @@ import useConfirmPlaybookArchiveModal from './archive_playbook_modal';
 import NoContentPage from './playbook_list_getting_started';
 import useConfirmPlaybookRestoreModal from './restore_playbook_modal';
 
-const ContainerMedium = styled.article`
-    margin: 0 auto;
-    max-width: 1160px;
+const ContainerMedium = styled.article<{$newLHSEnabled: boolean}>`
+    ${({$newLHSEnabled}) => !$newLHSEnabled && css`
+        margin: 0 auto;
+        max-width: 1160px;
+    `}
     padding: 0 20px;
     scroll-margin-top: 20px;
 `;
@@ -333,7 +335,10 @@ const PlaybookList = (props: {firstTimeUserExperience?: boolean}) => {
             {content.current}
             {canCreatePlaybooks && (
                 <>
-                    <ContainerMedium ref={selectorRef}>
+                    <ContainerMedium
+                        ref={selectorRef}
+                        $newLHSEnabled={newLHSEnabled}
+                    >
                         {props.firstTimeUserExperience || (!hasPlaybooks && !isFiltering) ? (
                             <AltCreatePlaybookHeader>
                                 <AltHeading>
--- a/webapp/src/components/backstage/playbook_list.tsx
+++ b/webapp/src/components/backstage/playbook_list.tsx
@@ -58,6 +58,10 @@ const PlaybookListContainer = styled.div`
     color: rgba(var(--center-channel-color-rgb), 0.9);
 `;
 
+const TableContainer = styled.div`
+    overflow: clip;
+`;
+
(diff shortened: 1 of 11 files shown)


webapp/src/components/backstage/playbook_list.tsx
```diff
@@ -7,19 +7,14 @@ import {Team} from 'mattermost-redux/types/teams';
...
 import {FormattedMessage, useIntl} from 'react-intl';
 import {useDispatch, useSelector} from 'react-redux';
-import styled from 'styled-components';
+import styled, {css} from 'styled-components';
 
 import {Redirect} from 'react-router-dom';
 
 import {displayPlaybookCreateModal} from 'src/actions';
 import {PrimaryButton, TertiaryButton} from 'src/components/assets/buttons';
-import LeftDots from 'src/components/assets/left_dots';
```
lieut-data, line 16: ❤️

webapp/src/components/backstage/playbook_list.tsx
```diff
@@ -63,15 +55,26 @@ const ContainerMedium = styled.article`
...
 const PlaybookListContainer = styled.div`
-    font-family: $font-family;
-    color: rgba(var(--center-channel-color-rgb), 0.90);
+    flex: 1 1 auto;
+    color: rgba(var(--center-channel-color-rgb), 0.9);
+`;
 
+const TableContainer = styled.div<{$newLHSEnabled: boolean;}>`
+    overflow: hidden;
+    overflow: clip;
```
lieut-data, line 64: nit: Two `overflow` directives?

lieut-data (approved):
So awesome to see the cleanup alongside these incremental changes!

abhijit-singh (changes requested):
This is great @calebroseland! A couple of minor things, (not necessary to be done in this PR though):
1. I think it'll make sense to also make the templates list span the full width. Open to feedback though. Something like this - 
image
I removed the max-width and it did the trick. Note that I also reduced the space between the cards to 4rem and it feels better. Also, all cards in a row should ideally be of the same height, even if that'll mean some extra white space in some of them. I added a min-height property to achieve that but not sure if it's the right way.
2. It seems the previous and next buttons are moving horizontally based on the width of the screen. Ideally, they should stay at a fixed distance(~16px) from the left and right edges respectively.

iomodo (approved):
Looks great Caleb! Thanks ^_^

calebroseland:
@abhijit-singh I updated for 1 & 2, but the consistent height will have to be a part of a later PR. There are some complexities to that to keep the upper color block to be of the same size while still being responsive.

CleanShot 2022-06-15 at 16 56 03@2x

calebroseland:
@abhijit-singh, done!

CleanShot 2022-06-15 at 17 07 23@2x

abhijit-singh:
Perfect 💯

PR #1253, opened 2022-06-15
live update for playbook runs member

trilopin:
#### Summary
Add real-time capabilities to Playbook run
Demo: https://www.loom.com/share/aab9892858b7426c91d491790ae79fc6
It covers:
- post an update with RHS opened -> shows immediately
- other members in another device post a new update while I have RHS opened -> shows immediately
- checkbox, command, assignee and due date are also synced live between two members in different browsers
- Not only affects the Run details page but also channels RHS 

Regarding the RDP RHS, a small refactor was needed to properly propagate changes.

We were already broadcasting (via WebSocket) for members but the state management didn't update the values (useState without useEffect with right deps).

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45129

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


trilopin:
cc @lieut-data regarding the topic we discussed on real-time updates.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -134,12 +123,18 @@ const PlaybookRunDetails = () => {
                         />
                         {role === Role.Participant ? (
                             <ParticipantStatusUpdate
-                                onViewAllUpdates={() => openRHS(RHSContent.RunStatusUpdates)}
+                                onViewAllUpdates={() => {
+                                    setIsRHSOpen(true);
+                                    setRHSSection(RHSContent.RunStatusUpdates);
+                                }}
```
gbochora, line 129: Could we assign this function to the variable? It would avoid duplicating code.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -42,47 +42,36 @@ const PlaybookRunDetails = () => {
...
-                <RHSStatusUpdates
-                    playbookRun={playbookRun}
-                    statusUpdates={statusUpdates}
-                />
-            );
-            break;
+            return formatMessage({defaultMessage: 'Status updates'});
+        default:
+            return '';
         }
```
gbochora, line 62: Would it be better if we could set the `RHS` title the same way as the section? Maybe we could change this after we start using the `RHS` outside of the run to avoid having playbooks section titles in the  `run_details`.

gbochora (approved):
Nice! I left a couple of optional comments.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
trilopin: I stopped managing the title&section as an state because I had issues to propagate properly the changes. But I don't know if you are more concerned about that or about keeping the title inside the RHS components.

If it's about the last one, I could handle the title inside the components and export as a secondary component.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -155,11 +147,16 @@ const PlaybookRunDetails = () => {
...
+                title={RHS.title}
+                onClose={RHS.close}
             >
-                {RHSData?.content}
+                {RHSContent.RunStatusUpdates === RHS.section ? (
+                    <RHSStatusUpdates
+                        playbookRun={playbookRun}
+                        statusUpdates={statusUpdates}
+                    />
+                ) : null}
```
calebroseland, line 159: Just a suggestion, but perhaps this content could be routable? Either from a query parameter in the URL or a sub route? In the future, permalinks to this 'All status updates' would be a lot easier if routing did this for you instead of having to programmatically check. Also for browser back/fwd compatibility.

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -61,6 +61,12 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
 
     const toggleDescription = () => setShowDescription(!showDescription);
 
+    useEffect(() => {
```
calebroseland, line 64: Even though this doesn't cause double-renders, `useUpdateEffect` might be a bit more in line with the usage here. (0/5)

```suggestion
    useUpdateEffect(() => {
```


Also just a quick plug for `useProxyState` that was written with this pattern in mind, but also 0/5 on this!

calebroseland (approved):
LGTM! Just a couple thoughts potentially for later.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
trilopin: Nice idea, I will explore it in a separate PR.

webapp/src/components/checklist_item/checklist_item.tsx
trilopin: Thanks! useUpdateEffect is what I was looking for.

PR #1255, opened 2022-06-16
Configure playbooks for Notice.txt file auto-generation

pfltdv:
#### Summary
We created a tool to automatically generate NOTICE.txt from project folders. And there is a scheduled pipeline configured to run at every Monday which creates a PR if there is a change.

Some sections need to be managed by developers at Notice.txt. And pipeline needs locations/files to search the dependencies. That configuration need to be stored in this repo, so developers will manage those and configure tool without leaving this repo.

Tool repo: https://github.com/mattermost/notice-file-generator

We selected `.config/notice-file` to store the configuration files. And it is configurable, feel free to change it if another location is more convenient.

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


spirosoik (approved):
🚀

PR #1256, opened 2022-06-16
RDP header sticky when doing scroll

trilopin:
#### Summary
Demo https://www.loom.com/share/c211de187c91420d83db118778684aad
Keep sticky header if we scroll in RDP.
Fix width when space is lower than 780px

#### Ticket Link
No ticket

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
@@ -180,7 +180,7 @@ const Container = styled(ColumnContainer)`
     flex: 1;
 `;
 
-const MainWrapper = styled.main<{isRHSOpen: boolean}>`
+const MainWrapper = styled.div<{isRHSOpen: boolean}>`
     flex: 1;
     display: flex;
     flex-direction: column;
@@ -189,6 +189,7 @@ const MainWrapper = styled.main<{isRHSOpen: boolean}>`
 
 const Main = styled.main<{isRHSOpen: boolean}>`
     max-width: 780px;
+    width: min(780px, 100%);
     padding: 20px;
     flex: 1;
     margin: 40px auto;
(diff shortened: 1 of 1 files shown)


gbochora (approved):
Cool!

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -189,18 +189,23 @@ const MainWrapper = styled.main<{isRHSOpen: boolean}>`
 
 const Main = styled.main<{isRHSOpen: boolean}>`
     max-width: 780px;
-    min-width: 500px;
+    width: min(780px, 100%);
```
agarciamontoro, line 192: Oh I didn't know CSS had a `min` function, cool!

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -180,7 +180,7 @@ const Container = styled(ColumnContainer)`
     flex: 1;
 `;
 
-const MainWrapper = styled.main<{isRHSOpen: boolean}>`
+const MainWrapper = styled.div<{isRHSOpen: boolean}>`
```
agarciamontoro, line 183: Why do we need this change?

agarciamontoro (approved):
Looks good!

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
trilopin: only one main per page like tags as body, header etc..

agarciamontoro: Oh, I see. Do we have more?

trilopin: there is a main: real main with content (max-width:780px) and a mainwrapper (now a div) with the full width to manage the centered main

PR #1257, opened 2022-06-16
Add create playbook dropdown

iomodo:
#### Summary
This PR adds create playbook dropdown to the LHS. 
`Invite people` is out of scope
`Start a run` is not implemented, cause when we create runs we use the interactive dialogue and it will not work without the channel. Either we have to implement our own dialogue or fix the current implementation of the interactive dialogue. 

https://user-images.githubusercontent.com/2340127/174095357-41c3b6c3-784e-4914-8303-9d09b11c1d66.mp4

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


webapp/src/components/sidebar/create_playbook_dropdown.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+        const importPlaybook = (
+            <>
+                <MenuItem
+                    id='importPlaybook'
+                    show={true}
+                    onClick={() => {
+                        if (fileInputRef && fileInputRef.current) {
+                            fileInputRef.current.click();
+                        }
```
agarciamontoro, line 84: Optional chaining again?
```suggestion
                    onClick={fileInputRef?.current?.click}
```

webapp/src/components/sidebar/create_playbook_dropdown.tsx
```diff
@@ -0,0 +1,202 @@
...
+                    }}
+                    icon={
+                        <StyledMDIIcon
+                            path={mdiImport}
+                            size={1.25}
```
agarciamontoro, line 89: Something I learned a couple of weeks ago: It turns out you can pass a string here as `12px`, which is way more  meaningful than this. Could we try it here and in the following mdi icons?

webapp/src/components/sidebar/create_playbook_dropdown.tsx
```diff
@@ -0,0 +1,202 @@
...
+        color: var(--sidebar-text);
+    }
+
+    min-width: 28px;
+    height: 28px;
+    font-size: 18px;
+    vertical-align: middle;
+
+    &.disabled {
+        background: rgba(255, 255, 255, 0.08);
```
agarciamontoro, line 188: Should this be themed?

agarciamontoro (approved):
Neat! :tada: I left some comments below, but none of them blocking.

webapp/src/components/sidebar/create_playbook_dropdown.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+interface CreatePlaybookDropdownProps {
+    team_id: string;
+}
+
+const CreatePlaybookDropdown = (props: CreatePlaybookDropdownProps) => {
+    const {formatMessage} = useIntl();
+    const dispatch = useDispatch();
+    const fileInputRef = useRef<HTMLInputElement | null>(null);
+    const teams = useSelector<GlobalState, Team[]>(getMyTeams);
```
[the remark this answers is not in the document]
calebroseland: +1, fewer type imports needed

webapp/src/components/sidebar/create_playbook_dropdown.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+        const createPlaybook = (
+            <MenuItem
+                id='createPlaybook'
+                show={true}
+                onClick={() => dispatch(displayPlaybookCreateModal({}))}
+                icon={<StyledIcon className='icon-plus'/>}
+                text={formatMessage({defaultMessage: 'Create New Playbook'})}
+            />
+        );
```
calebroseland, line 74: Shouldn't be behind a permissions gate of some sort? The one in `playbook_list` is behind `useCanCreatePlaybooksOnAnyTeam`.  This/that may require refactoring once we rescope to single-team, but I think we should at least match for now.

calebroseland (changes requested):
Just one UI issue (icon alignment):
CleanShot 2022-06-17 at 16 23 20@2x

Otherwise looks great and will be nice to have this in!

calebroseland (approved):

PR #1258, opened 2022-06-16
MM-43947: Enable App Bar in e2e tests

Phrynobatrachus:
#### Summary
Enables the App Bar for both cloud & on-prem e2e runs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43947

### DIFF
--- a/tests-e2e/cypress/integration/channels/rhs/home_spec.js
+++ b/tests-e2e/cypress/integration/channels/rhs/home_spec.js
@@ -9,6 +9,7 @@
 describe('channels > rhs > home', () => {
     let testTeam;
     let testUser;
+    let siteURL;
 
     before(() => {
         cy.apiInitSetup().then(({team, user}) => {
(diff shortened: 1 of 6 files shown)


tests-e2e/cypress/integration/channels/rhs/home_spec.js
```diff
@@ -29,9 +30,15 @@ describe('channels > rhs > home', () => {
...
                 channelNameTemplate: 'templated name',
             });
+
+            cy.apiGetConfig(true).then(({config}) => {
+                siteURL = config.SiteURL;
+            });
         });
     });
 
+    const getAppBarImageSelector = () => `.app-bar .app-bar__icon-inner img[src="${siteURL}/plugins/playbooks/public/app-bar-icon.png"]`;
```
agarciamontoro, line 40: This is in two files, should we add it as a cypress command so we can simply do `cy.getAppBarImage()`?

agarciamontoro (approved):
Thanks! :tada:

lieut-data (approved):

PR #1259, opened 2022-06-17
Mm 45108 viewer mode

gbochora:
#### Summary
This PR adds the `Viewer` mode to the `Header` and `Checklist`. Retrospective `Viewer` mode implemented in this PR: https://github.com/mattermost/mattermost-plugin-playbooks/pull/1248#pullrequestreview-1009998180
Header in `Viewer` mode:
- no `Finish run` menu item.
- no `Go to channel` button. We will need to add `Get Involved` button instead in v2

Checklist in `Viewer` mode is the same as for the finished run.
- no add checklist item
- no add checklist
- no hover menu
- no checklist drag

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45108

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_runs/playbook_run/checklists.tsx
```diff
@@ -7,15 +7,18 @@ import styled from 'styled-components';
 import {PlaybookRun} from 'src/types/playbook_run';
 
 import RHSChecklistList, {ChecklistParent} from 'src/components/rhs/rhs_checklist_list';
+import {Role} from '../shared';
 interface Props {
```
agarciamontoro, line 11: This is a personal preference, but I'd rather have absolute pahts in the imports. Also, new line missing:
```suggestion
import {Role} from '../shared';

interface Props {
```

agarciamontoro (approved):
Looks good! I'm loving these short, iterative PRs :)

webapp/src/components/backstage/playbook_editor/outline/outline.tsx
```diff
@@ -89,6 +89,7 @@ const Outline = ({playbook, refetch}: Props) => {
                 <ChecklistList
                     playbook={playbook}
                     enableFinishRun={false}
+                    viewerMode={false}
```
trilopin, line 92: 1/5 on this comment

Where we are aware of the role, it seems ok to use viewer vs participant and names like isViewer or viewerMode.

Once we are deeper in the component tree and we don't have necessarily the context of what's viewer and participant, I would use names like isReadOnly or isChangeAllowed. it's decoupled of the origin of that decision and explains more about what is going to happen.

trilopin (commented):
Nice job @gbochora 

I noticed that the collapse checklist icon doesn't work when you are a viewer. is that intended?

webapp/src/components/backstage/playbook_editor/outline/outline.tsx
gbochora: Good point, I also had doubts regarding this.

gbochora:
> I noticed that the collapse checklist icon doesn't work when you are a viewer. is that intended?

@trilopin  No, it's not intended. I tried and can't reproduce it. Do you make some steps before, or it happens every time?

trilopin:
Big move here @gbochora! 

> * `RD` stores checklist info in the same redux state under the `key=run_details_[runId]`. So the same run will not share `RHS` and `RD` states.

I understand that collapsed state is designed to behave separately between RDP and Channels RHS, but I feel them as connected. Video attached.


https://user-images.githubusercontent.com/4096774/175897039-a2548784-9748-40bf-a4af-52ab2a4d8f29.mp4

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -58,13 +51,25 @@ interface Props {
     playbookRun?: PlaybookRun;
     playbook?: Loaded<FullPlaybook>;
     enableFinishRun: boolean;
+    isReadOnly: boolean;
+    checklistsCollapseState: Record<number, boolean>;
+    onChecklistCollapsedStateChange: (checklistIndex: number, state: boolean) => void;
+    onEachChecklistCollapsedStateChange: (state: Record<number, boolean>) => void;
```
agarciamontoro, line 57: This function is called when the collapsed state of every checklist is changed at once, right? If that's correct, I proposeto use `onEveryChecklist` or `onAllChecklsits` instead of `Each`. At least for me the `Each` word threw me off.
```suggestion
    onEveryChecklistCollapsedStateChange: (state: Record<number, boolean>) => void;
```

agarciamontoro (changes requested):
Code looks good to me! There's only a naming suggestion.

Also, a more general question: why does the Better Playbook Editor now store the state locally instead of using yet another key for the store? Does that play well with different users viewing the playbook at the same time and one of them checking and unchecking tasks?

webapp/src/components/checklist/checklist_list.tsx
gbochora: > Also, a more general question: why does the Better Playbook Editor now store the state locally instead of using yet another key for the store? Does that play well with different users viewing the playbook at the same time and one of them checking and unchecking tasks?

@agarciamontoro  `BPE` never supported caching checklist state in the redux, so I used the local store. If storing the state in the redux makes sense, I can add it, it's just a few lines of code. I don't have a strong opinion regarding this. For `RHS` and `RD`, it's more clear for me because users switch between runs very often and it's important to save the state.

agarciamontoro: @gbochora Oh, wait, I completely mixed up two things: whether the checklist is collapsed or not and the actual state of the tasks (which is not modified in the playbook editor, sorry :facepalm:). If they didn't add that to redux then let's skip it for now. Thanks for the explanation and sorry for the misunderstanding!

webapp/src/components/backstage/playbook_editor/outline/outline.tsx
```diff
@@ -36,6 +36,17 @@ const Outline = ({playbook, refetch}: Props) => {
     const updatePlaybook = useUpdatePlaybook(playbook.id);
     const retrospectiveAccess = useAllowRetrospectiveAccess();
     const archived = playbook.delete_at !== 0;
+    const [checklistCollapseState, setChecklistCollapseState] = useState({} as Record<number, boolean>);
```
trilopin, line 39: nit: use generic instead of type cast
```suggestion
    const [checklistCollapseState, setChecklistCollapseState] = useState<Record<number, boolean>>({});
```

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -214,32 +219,33 @@ const ChecklistList = (props: Props) => {
...
-            // right, depending on whether srcIdx < dstIdx) one position
-            const newState = {...checklistsState};
-            if (srcIdx < dstIdx) {
-                for (let i = srcIdx; i < dstIdx; i++) {
-                    newState[i] = checklistsState[i + 1];
-                }
-            } else {
-                for (let i = dstIdx + 1; i <= srcIdx; i++) {
-                    newState[i] = checklistsState[i - 1];
+            if (playbookRun) {
```
trilopin, line 222: nice move

webapp/src/components/checklist/generic_checklist.tsx
```diff
@@ -170,7 +124,7 @@ const GenericChecklist = (props: Props) => {
                         }
                         {droppableProvided.placeholder}
                     </div>
-                    {props.playbookRun?.current_status !== PlaybookRunStatus.Finished &&
+                    {!props.disabled &&
```
trilopin, line 127: personal preference: use ternary operator over logical and

https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx
not always applies, but it's commonly safer

trilopin (approved):
What a refactor :)

I left some small and optional comments.

agarciamontoro (approved):
:tada:

PR #1261, opened 2022-06-20
Mm 45182 mm 45183 UI improvements

gbochora:
#### Summary
This PR: 
- adds run name to the RHS header
- adds placeholder/link when user doesn't have permissions to see the run page
- adds run state badge in the header 
- fixes distances between main form sections

**RHS header name**
image

**Placeholder/link when user doesn't have permissions to see the run page**
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45182
https://mattermost.atlassian.net/browse/MM-45183

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -92,11 +94,26 @@ const HeaderIcon = styled.div`
...
+
+const HeaderSubtitle = styled.div`
+    overflow: hidden;
+    height: 2.4rem;
+    padding: 0 8px;
+    border-left: 1px solid rgba(var(--center-channel-color-rgb), 0.16);
+    margin: 0 0 0 8px;
+    color: rgba(var(--center-channel-color-rgb), 0.56);
+    font-size: 12px;
+    line-height: 2.4rem;
```
trilopin, line 113: not sure about vertically aligning this with line-height, is there another better way?

gbochora: What if we divide the subtitle into two parts? The first one is `vertical divider line`, and the second `title text`.
In this case, we don't need the line-height to center the text.

webapp/src/components/backstage/status_badge.tsx
```diff
@@ -65,13 +65,13 @@ const StatusBadge = (props: BadgeProps) => {
...
         break;
     case BadgeType.Finished:
-        message = <FormattedMessage defaultMessage='Finished'/>;
+        message = <FormattedMessage defaultMessage='FINISHED'/>;
         break;
     case BadgeType.Archived:
-        message = <FormattedMessage defaultMessage='Archived'/>;
+        message = <FormattedMessage defaultMessage='ARCHIVED'/>;
         break;
     }
```
agarciamontoro, line 76: Nit: instead of relying on the string being uppercase, which can be forgotten when modifying existing statuses or adding new ones, we can add the CSS property `text-transform: uppercase` to the Badge styled component.

agarciamontoro (approved):
Looks fancy :tada: I left a minor, very non-blocking, nit :) Thanks, Giorgi!

PR #1262, opened 2022-06-20
MM-44718 finish run extacted to hook 

trilopin:
#### Summary
Extract finish run behavior to a hook to share among context menu and finish run section.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-44718

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/backstage/playbook_runs/playbook_run/hoc_finish_run.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {finishRun} from 'src/client';
+import {modals} from 'src/webapp_globals';
+import {makeUncontrolledConfirmModalDefinition} from 'src/components/widgets/confirmation_modal';
+
+interface WithFinishRun {
+    playbookRun: PlaybookRun;
+    children: React.ReactNode;
+}
+
+export default function withFinishRun<T extends WithFinishRun>(ChildComponent: React.ComponentType<T & {onClick: any}>) {
```
trilopin, line 16: not happy with that `any` but I spend too much time already trying to make it perfect.

Suggestions & TS wisdom are more than welcome

trilopin:
3/5 on this approach, if you feel this is too clever I'd fine changing it to a component composition approach.

agarciamontoro (commented):
This is really cool! But, as you said, I'm concerned this is too clever, it took me a while to understand how everything fits together. Another solution is to have a custom hook, something similar to this:
```js
const useOnFinishRunClick = (playbookRun: PlaybookRun) => {
    
    const onFinishRunClick = () => {
        // ... the whole onFinishRunClick logic
    }
    
    return onFinishRunClick;
}
```
Then you could call it in the components you need as in
```
const onFinishRunClick = useOnFinishRunClick(props.playbookRun)
```

Thoughts?

agarciamontoro (approved):
Yay! This is easier to understand for my old tired brain. Thanks :tada:

PR #1263, opened 2022-06-20
fix state management when chaging run in lhs

trilopin:
#### Summary
Prevent user from carrying rhs state between different runs.

We were doing a mix of redux check + fetch (useRun which relies on useThing) and parallel promise for extended data. I refactored a bit the code and replaced useRun with a plain selector (to avoid double fetch).

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45200

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -52,50 +52,52 @@ const useRHS = () => {
...
     const [playbookRun, setPlaybookRun] = useState<PlaybookRun | null>(null);
     const playbook = usePlaybook(playbookRun?.playbook_id);
     const [following, setFollowing] = useState<string[]>([]);
     const [fetchingState, setFetchingState] = useState(FetchingStateType.loading);
     const [playbookRunMetadata, setPlaybookRunMetadata] = useState<PlaybookRunMetadata | null>(null);
     const [statusUpdates, setStatusUpdates] = useState<StatusPostComplete[]>([]);
+    const myUser = useSelector(getCurrentUser);
 
     const RHS = useRHS();
+    const runFromStore = useSelector<GlobalState, PlaybookRun|undefined|null>(getRun(playbookRunId));
```
agarciamontoro, line 68: Do we need all the explicit typing here?
```suggestion
    const runFromStore = useSelector(getRun(playbookRunId));
```

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -52,50 +52,52 @@ const useRHS = () => {
...
 
-    const myUser = useSelector(getCurrentUser);
+    // Trigger multiple parallel fetch if the URL param changes
+    useEffect(() => {
+        if (playbookRunId !== playbookRun?.id && RHS.isOpen) {
+            RHS.close();
+        }
```
agarciamontoro, line 74: In Channels, switching channels does not close the RHS, should we follow that same pattern here?

agarciamontoro (changes requested):
I'm not sure I understand this change. Wouldn't it be better to still rely on `useRun` and remove the explicit `fetchPlaybookRun` from the parallel `useEffect`?

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
trilopin: https://community.mattermost.com/core/pl/o693b8n4sj8qxb8he8s4wbtc1h

trilopin:
> I'm not sure I understand this change. Wouldn't it be better to still rely on `useRun` and remove the explicit `fetchPlaybookRun` from the parallel `useEffect`?

My biggest concern (and it affected negatively) was to have two sources of truth regarding playbookRun (useRun and fetchPlaybookRun). It could be solved by removing fetchPlaybookRun from the parallel promise or by removing useRun. I chose to keep the promise with all fetch but I could recover the useRun if you see more value and clarity.

agarciamontoro:
I completely agree with the concern, having both is quite error-prone.

Having to choose, I like `useRun` better simply because it encapsulates all the fetch/store logic in a single generic (`useThing`) hook. I feel like that hook is very well tested throughout the codebase because we use it in many places, and that it's easier to miss bugs when repeating that same logic in the individual components. I also find it easier to read, but that's probably secondary (it's easier once you know how the `useThing` hook works, but it's probably a bit more obscure for newcomers).

trilopin:
Additionally, I added the RHS behavior that @abhijit-singh explained:

- run-info is opened by default
- change run does not close rhs
- if we navigate to a run with rhs-status-udpates open but there aren't (disabled or not posted yet), rhs-run-info will be shown otherwise

agarciamontoro (approved):
Looks good! Just a minor question: why do we need `useUpdateEffect` and not the regular `useEffect`?

webapp/src/hooks/general.ts
```diff
@@ -331,7 +332,7 @@ function useThing<T extends NonNullable<any>>(
         }
 
         if (id) {
-            fetch(id).then(setThing);
+            fetch(id).then(setThing).catch(() => setThing(null));
```
trilopin, line 335: two tiny changes here that are not minor, state hidden in return value: undefined == loading; null == not found

Ihis pattern is used in other hooks and thought that it was better than returning fetchingState in another value. However, checking everything is not easy and could bring some problems.

@agarciamontoro @gbochora

gbochora (commented):
Nice work @trilopin! Code looks much cleaner now. I've tested retrospective, and everything seems to work perfectly. 
Few issues I've noticed:
- `RHS` doesn't show the title. I think this happened after merging with the master.
- 'RHS' doesn't show the run title.
- Header icons are missing

trilopin:
> Nice work @trilopin! Code looks much cleaner now. I've tested retrospective, and everything seems to work perfectly. 
> Few issues I've noticed:
> - `RHS` doesn't show the title. I think this happened after merging with the master.
> - 'RHS' doesn't show the run title.
> - Header icons are missing
> 

Oh that must be the last master merge. Let me check those points.

webapp/src/hooks/general.ts
agarciamontoro: What if we have a global `FetchingState` type that we can rely everywhere? It's something common to all fetches, so it may be useful.

trilopin: I would love to have a simple API with the data and the state of the fetch with idle/pending/done/error and maybe error code.

But to do that we need to:
- make another different useThing for useRun and test extensively all useRun
- change useThing directly and test extensively all hooks that use useThing

Both points are a bit out of this scope. I would want to do it outside this PR.

webapp/src/hooks/general.ts
```diff
@@ -345,14 +347,75 @@ export function usePost(postId: string) {
...
+ */
+export function useFetch<T>(
+    id: string,
+    fetch: (id: string) => Promise<T>,
+    deps: Array<any> = [],
+) {
+    const [error, setError] = useState<ClientError|null>(null);
+    const [fetchState, setFetchState] = useState<FetchState>(FetchState.idle);
+    const [data, setData] = useState<T | null>(null);
+    useEffect(() => {
```
agarciamontoro, line 380: Super nit: new line before useEffect:

```suggestion
    const [data, setData] = useState<T | null>(null);

    useEffect(() => {
```

webapp/src/hooks/general.ts
```diff
@@ -345,14 +347,75 @@ export function usePost(postId: string) {
...
+/**
+ *
+ * @param id The id of the resource to be fetched
+ * @param fetch The function used to make the fetch
+ * @param deps Additional deps that might be needed to trigger again the fetch func
+ * @returns array tuple with the data in the first position and the fetchState in the second
+ */
+export function useFetch<T>(
+    id: string,
+    fetch: (id: string) => Promise<T>,
```
agarciamontoro, line 374: Could we rename this so we don't shadow the global `fetch` function? The suggested name is just a proposal, any other thing is good for me:
```suggestion
    fetchFunction: (id: string) => Promise<T>,
```

webapp/src/hooks/general.ts
```diff
@@ -345,14 +347,75 @@ export function usePost(postId: string) {
...
+            .then((res) => {
+                setFetchState(FetchState.done);
+                setData(res);
+            })
+            .catch((err) => {
+                if (err instanceof ClientError) {
+                    setError(err);
+                }
+                setFetchState(FetchState.error);
+                setData(null);
```
agarciamontoro, line 395: Really cool :pray:

agarciamontoro (changes requested):
Love the `useFetch` proposal, thank you! :tada: I left mostly nits, the only blocking comment is the one about fetch.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -56,50 +53,24 @@ const useRHS = () => {
...
+    const playbookRun = useRun(playbookRunId);
     const playbook = usePlaybook(playbookRun?.playbook_id);
-    const [following, setFollowing] = useState<string[]>([]);
-    const [fetchingState, setFetchingState] = useState(FetchingStateType.loading);
-    const [playbookRunMetadata, setPlaybookRunMetadata] = useState<PlaybookRunMetadata | null>(null);
-    const [statusUpdates, setStatusUpdates] = useState<StatusPostComplete[]>([]);
-
-    const RHS = useRHS();
-
+    const [metadata, metadataRes] = useRunMetadata(playbookRunId);
```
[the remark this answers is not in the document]
trilopin: 😂

webapp/src/hooks/general.ts
trilopin: good catch

agarciamontoro (approved):
:tada:

PR #1264, opened 2022-06-21
Remove position correction when dropdown is closed

gbochora:
#### Summary
The unexpected scroll problem happens in the checklist on both pages `RHS` and `PB Editor`. It's related to the `due date` and `profile` selectors. 

**Steps to reproduce it:**
- open `PB  editor` or `RHS` with lots of checklist items
- scroll down two pages
- click on the `due date` or `assign` hover menu button
Result: On button click page automatically scrolls to the top(the video is attached).

From my understanding, during the initial render, `setMoveUp` function is called for every checklist item, and all select components(`rect`) are moved to the top of the page. When we click on the `due date` button, at first dropdown popup is located at the top of the page, and that causes an unexpected scroll. Then the popup is moved to the correct location, but the page remains scrolled at the top. I've attached the debugger recording.

Fix: I've added a condition, so now `setMoveUp` function is called only when the dropdown is visible.

Unexpected scroll:
https://user-images.githubusercontent.com/4368372/174768669-d92da82f-166e-4774-9c19-abe1463e61dc.mp4

Debugging:
https://user-images.githubusercontent.com/4368372/174768700-fc3b666e-b627-4bd6-b507-00537ec78d9e.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45148

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/channels/rhs/checklist_spec.js
+++ b/tests-e2e/cypress/integration/channels/rhs/checklist_spec.js
@@ -57,23 +57,76 @@ describe('channels > rhs > checklist', () => {
             cy.apiCreatePlaybook({
                 teamId: team.id,
                 title: 'Playbook',
-                checklists: [{
-                    title: 'Stage 1',
-                    items: [
-                        {title: 'Step 1', command: '/invalid'},
-                        {title: 'Step 2', command: '/echo VALID'},
-                        {title: 'Step 3'},
-                        {title: 'Step 4'},
-                        {title: 'Step 5'},
-                        {title: 'Step 6'},
-                        {title: 'Step 7'},
-                        {title: 'Step 8'},
-                        {title: 'Step 9'},
-                        {title: 'Step 10'},
-                        {title: 'Step 11'},
-                        {title: 'Step 12'},
-                    ],
-                }],
+                checklists: [
+                    {
+                        title: 'Stage 1',
+                        items: [
+                            {title: 'Step 1', command: '/invalid'},
+                            {title: 'Step 2', command: '/echo VALID'},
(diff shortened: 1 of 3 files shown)


amyblais:
@gbochora There are some failing tests here.

lieut-data:
@calebroseland, can I assign my review your way to help take a look at this? We'd like to include it in the upcoming v7.0.1 as a patch to Playbooks.

iomodo (approved):
Wow, great catch!. 
We might have similar errors in other drop downs as well(team selector, filters, etc.). Probably not be a part of this PR though

calebroseland (approved):
Nice debugging! And straightforward fix for the current implementation; no objections for the sake of expediency.

That said, I would like to propose we move this logic over to what `DotMenu` is using— [Floating UI](https://floating-ui.com/)'s `useFloating`, which is trivial to use IMHO. It also automatically handles `placement`, `shift`, and `flip`

use-floating

This is just a part of floating ui. Just as a reference to how good this lib is, [here's an example of a full-featured dropdown with the kitchen sink](https://codesandbox.io/s/admiring-lamport-5wt3yg?file=/src/DropdownMenu.tsx)!

gbochora:
@calebroseland Looks very good, even supports submenus. I'll add a task as tech debt to switch to [Floating UI](https://floating-ui.com/)

amyblais:
@gbochora Would this be cherry-picked to ``release-1.28`` branch? cc @lieut-data

lieut-data:
Thanks, yes, @amyblais, cherry-picking and preparing a release now.

PR #1266, opened 2022-06-21
Add endpoints for categories

iomodo:
#### Summary
This PR adds endpoint for categories
It also performs a migration and adds `Favorite` category for every user in every team.

No UI changes

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


server/api/categories.go
```diff
@@ -0,0 +1,131 @@
...
+		return
+	}
+
+	// user can only create category for themselves
+	if category.UserID != userID {
+		h.HandleErrorWithCode(w, http.StatusBadRequest, fmt.Sprintf("userID %s and category userID %s mismatch", userID, category.UserID), nil)
+		return
+	}
+
+	if category.TeamID != teamID {
```
crspeller, line 73: Why not just use the category team ID?

server/app/category_service.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// category is already deleted. This avoids
+	// overriding the original deleted at timestamp
+	if existingCategory.DeleteAt != 0 {
+		return nil
+	}
+
+	// verify if category belongs to the user
+	if existingCategory.UserID != userID {
```
crspeller, line 98: The permissions validations here seem to be spread over api and app layer. Can we move them all to one or the other? Generally the API layer does permissions.

crspeller (changes requested):

crspeller (approved):

server/sqlstore/migrations.go
```diff
@@ -2024,4 +2024,106 @@ var migrations = []Migration{
...
+			}
+
+			idString := "REPLACE(uuid_in(md5(random()::text || clock_timestamp()::text)::cstring)::varchar, '-', '')"
+			timeNowString := "(extract(epoch from now())*1000)::bigint"
+			if e.DriverName() == model.DatabaseDriverMysql {
+				idString = "LEFT(MD5(RAND()), 26)"
+				timeNowString = "UNIX_TIMESTAMP() * 1000"
+			}
+
+			// Create `Favorite` category for every user in every team
```
calebroseland, line 2108: What about users that are created after this migration? How will they get a static `'Favorite'` category?

calebroseland (approved):
LGTM, just one question/concern about newly-created users after the migration.

server/sqlstore/migrations.go
iomodo: @calebroseland good catch!
I removed the migration entirely, since we anyway have to check/create for new users.
We will create `Favorite` category when user first favorites an item.

Alternative approach would be creating category in `UserHasBeenCreated` hook, but it will increase complexity.

PR #1270, opened 2022-06-23
hackaton task inbox - prepare task list

trilopin:
#### Summary
Show real and relevant tasks inside the task inbox.

Parts:
- to work with real data, I fetch all active runs and filter all relevant tasks. This should be replaced later with sth better.
- Component task rendering the info and being able to: change state (checkbox), change assignee, due date, command, description, collapse/expand desc, etc..
- Hover menu shows inside the Task
- Naive sorting mechanism

Not covered:
- Make disappear if you check/skip/assign-to-another-user a task (animation?) --> will be live if we connect them to redux
- Filters and overdue counters

This could be merged after the branch is.

#### Ticket Link
No ticket

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data (approved):
Sweet.

PR #1275, opened 2022-06-28
1st pass replacing imports to types package

trilopin:
#### Summary
Replace all imports related to types with the desired package. Even with this change, it's still working through module mappers against node_modules/mattermost-webapp/packages/ etc...

This is the first change where we change all types imports. After that, we'll change the [dependency](https://www.npmjs.com/package/@mattermost/types) and the modules mapper aliases in jest, webpack and tsconfig.

#### Ticket Link
no ticket

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):
Yay!

calebroseland (approved):

PR #1276, opened 2022-06-29
Mm 44766 participants

gbochora:
#### Summary
This PR adds `RHS > Participants` view:
- Participants list
- `RHS` back navigation
- `Send message` button on hover 

NOTE: Search is not implemented yet. We decided to do it in v2.

https://user-images.githubusercontent.com/4368372/176377807-fddbdfe7-9cb6-442b-97eb-87493094bbe5.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44766

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_runs/playbook_run/header.tsx
```diff
@@ -75,13 +75,13 @@ export const RunHeader = ({playbookRun, playbookRunMetadata, openRHS, role}: Pro
                 tooltipId={'timeline-button-tooltip'}
                 tooltipMessage={formatMessage({defaultMessage: 'View Timeline'})}
                 className={'icon-update'}
-                onClick={() => openRHS(RHSContent.RunTimeline, formatMessage({defaultMessage: 'Timeline'}), playbookRun.name)}
+                onClick={() => onViewTimeline()}
```
trilopin, line 78:
```suggestion
                onClick={onViewTimeline}
```

webapp/src/components/backstage/playbook_runs/playbook_run/header.tsx
```diff
@@ -75,13 +75,13 @@ export const RunHeader = ({playbookRun, playbookRunMetadata, openRHS, role}: Pro
...
             />
             <HeaderButton
                 tooltipId={'info-button-tooltip'}
                 tooltipMessage={formatMessage({defaultMessage: 'View Info'})}
                 className={'icon-information-outline'}
-                onClick={() => openRHS(RHSContent.RunInfo, formatMessage({defaultMessage: 'Run info'}), playbookRun.name)}
+                onClick={() => onViewInfo()}
```
trilopin, line 84:
```suggestion
                onClick={onViewInfo}
```

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -40,6 +41,14 @@ const RightHandSidebar = ({isOpen, onClose, title, children, subtitle}: Props) =
             isOpen={isOpen}
         >
             <Header>
+                {
+                    onBack ? <BackIcon>
```
trilopin, line 45: this is a personal preference, but you could use a better indentation when doing ternary operator.

Example.

```ts 
            {props.parent === 'retro' ? (
                <TimeContainer parent={props.parent}>
                    {timeSincePrevEvent}
                </TimeContainer>
            ) : null}
```

trilopin (changes requested):
Thank you @gbochora 

I left some details and doubts

webapp/src/components/backstage/playbook_runs/playbook_run/send_message_button.tsx
```diff
@@ -0,0 +1,41 @@
...
+    teamName: string | null;
+}
+
+export const SendMessageButton = (props: Props) => {
+    const dispatch = useDispatch();
+    const user = useSelector<GlobalState, UserProfile>((state) => getUser(state, props.userId));
+
+    useEffect(() => {
+        if (!user) {
+            dispatch(fetchUser(props.userId));
```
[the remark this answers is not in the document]
gbochora: Good point. I moved fetching inside the `onClick` function.

webapp/src/components/backstage/playbook_runs/playbook_run/send_message_button.tsx
```diff
@@ -0,0 +1,51 @@
...
+        if (user) {
+            navigateToUrl(`/${props.teamName}/messages/@${user.username}`);
+            return;
+        }
+
+        // if user profile is missing fetch it
+        const actionFunc = dispatch(fetchUser(props.userId));
+
+        //@ts-ignore
+        actionFunc.then((data) => {
```
trilopin, line 36: I tested this and works fine. But as far as I know, waiting for a dispatched action does not guarantee that the reducer has been executed. 

what happens if you navigate without doing the fetch first?

trilopin (commented):
I left a comment about promises and actions. Additionally, I'm wondering if we should avoid navigating to our own user.

webapp/src/components/backstage/playbook_runs/playbook_run/send_message_button.tsx
trilopin: not the selector but the action.

I think with redux, the fact that the promise of the action has been solved doesn't mean that the reducer has been executed. 

I'm still not sure if we expected that or if we simply do a best-effort strategy (that would be enough most of the cases).

agarciamontoro (changes requested):
Impressive work! The only blocking comment is the one about the plural, because if there's one participant, it still reads "1 participants". All the other comments are nits. Thanks! :tada:

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_participants.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <Container>
+            <SearchSection>
+                {/* <StyledSearchInput
+                    testId={'search-filter'}
+                    default={''}
+                    onSearch={setSearchTerm}
+                    placeholder={formatMessage({defaultMessage: 'Search'})}
+                /> */}
+                <ParticipantsNumber>
+                    {formatMessage({defaultMessage: '{num} Participants'}, {num: participantsIds.length})}
```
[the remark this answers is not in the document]
gbochora: Nice catch!

trilopin (approved):
🎉

PR #1278, opened 2022-06-29
e2e: Fix RHS failures

Phrynobatrachus:
To be honest, I still don't know why those failures are occurring. I'm unable to reproduce locally, and nothing seems out of the ordinary in the CI video artifacts. These changes are passing repeatedly though.


trilopin (approved):
I also couldn't reproduce locally

Thanks for taking the time @Phrynobatrachus

PR #1280, opened 2022-06-30
Reset page to 0 after update the params with the search param

Afsoon:
#### Summary
I'm not using directly the function `setPage` to update the state because it can generate a flashy change between UI states if the user clear the input. The flashy situation doesn't happen always but enough to be detected. 

I created a new spec for E2E because I have to populate the database with elements to force a pagination situation and it was generating conflict with the rest of tests. I thought to declare the test as the last one but I think I'm moving forward the problem to the future, this the reason I created a new spec only for pagination use cases. 

#### Ticket Link
Fixes: https://github.com/mattermost/mattermost-server/issues/20422

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data:
Wow, thank you @Afsoon! @calebroseland & @iomodo, let's see if we can include this in v7.1 and before code freeze on Friday, July 8th.

iomodo (approved):
Thanks for the contribution @Afsoon
Looks great!

calebroseland (approved):
LGTM!

PR #1287, opened 2022-07-04
MM-45412: Fix categorization of users in owner dropdown list

agarciamontoro:
#### Summary
This fixes the bug in the Run Details Page where the members in the owner selector appeared in the wrong category. We can now pass an explicit channel ID to the `AssignTo` component instead of always relying on the ID of the current channel, which may not exist.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45412

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


iomodo (approved):
Looks great @agarciamontoro !

PR #1288, opened 2022-07-04
MM-45500: Add owners not in channel

agarciamontoro:
#### Summary
This follows the same pattern we use in the channel RHS when selecting an owner that is not in the channel: we show a modal to confirm that we want to add them to the channel. I slightly modified the wording here to match the participant/viewer terms, but apart from that it's pretty much the same.

https://user-images.githubusercontent.com/3924815/177178109-9ce2facb-5bf4-4da5-ab99-0daa3b6685de.mp4

Note that this PR depends on #1287.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45500

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
```diff
@@ -110,12 +123,70 @@ const RHSInfoOverview = ({run, runMetadata, editable, onViewParticipants}: Props
...
+                }}
+            />
         </Section>
     );
 };
 
 export default RHSInfoOverview;
 
+interface AddToChannelModalProps {
+    editable: boolean;
```
trilopin, line 144: I see that this is useful, but IMHO not very semantic. editable make sense in the context of the parent component where things can be edited, here is only an exit condition.

That could be applied in `<AddToChannelModal` with a ternary condition or even on the triggers for that modal.

trilopin: Actually, my feeling is that this is already handled and we can remove that prop.

trilopin (approved):
It's really nice to have this flow in mind.

I'd check some extra props that seem unnecessary to me. As a personal preference, I wouldn't link user names in the modal, but I can live with that anyway.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
agarciamontoro: You're right, this doesn't add any value. Fixed in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1288/commits/c8d21d35f0750ef34a8c3f1a85a68e7405d0dc5a.

webapp/src/hooks/general.ts
```diff
@@ -775,3 +778,22 @@ export const useExportLogAvailable = () => {
...
     return useSelector<GlobalState, boolean>((state) => Boolean(state.plugins?.plugins?.['com.mattermost.plugin-channel-export']));
 };
+
+interface MarkdownOptions {
+    singleline: boolean;
+    mentionHighlight: boolean;
+    atMentions: boolean;
+}
+
+export const useMarkdownFormatter = (options: MarkdownOptions) => {
```
crspeller, line 788: There is already a useDefaultMarkdownOptions in https://github.com/mattermost/mattermost-plugin-playbooks/blob/3ba005f300cba07de41929da853814877d67e833/webapp/src/components/formatted_markdown.tsx#L14 maybe integrate that here as well?

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
```diff
@@ -109,12 +123,68 @@ const RHSInfoOverview = ({run, runMetadata, editable, onViewParticipants}: Props
...
+    const mdText = useMarkdownFormatter(markdownOptions);
+
+    if (!user) {
+        return null;
+    }
+
+    return (
+        <ConfirmModal
+            show={show}
+            title={mdText(formatMessage(
```
crspeller, line 169: I thought we where not supposed to use markdown with formatMessage? Or was that just for links?

crspeller (approved):
Looks good. Just a couple non-blocking things.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
agarciamontoro: Oh, I had completely forgotten that, thanks for the catch! I just removed all markdown text, which effectively removed the clickable username in the modal. Not a big deal, in my opinion.

PR #1289, opened 2022-07-04
MM-45474: Add active state to header buttons

agarciamontoro:
#### Summary

Add an active state to the header buttons so that they're highlighted when the corresponding RHS section is open.

https://user-images.githubusercontent.com/3924815/177185461-219797ff-2f56-47d6-9095-0c19549069c5.mp4

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45474

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


trilopin (approved):
Thanks @agarciamontoro 

Code is fine. 

I'm wondering which state prevails when we are active and hover. have you dealt with that in the past?

agarciamontoro:
@trilopin: right now, it's the hover styles that prevail. I've checked in the webapp and I found that there are examples of both strategies in different places. I'm 0/5 on which to apply (maybe 1/5 on applying the current one just because it's already done)
active-hover

trilopin:
@agarciamontoro thanks for the context. I'd merge this and raise the same question to @abhijit-singh. Consistency is obviously important, but I would like to know if that effect is desired or organically adopted.

PR #1291, opened 2022-07-06
MM-44548, MM-44816: Fix dnd jumps/glitches in playbook editor

calebroseland:
#### Summary
- Fix: drag and drop jumpiness in playbook editor.
This is done by intercepting the incoming prop with `useProxyState`, giving it a local `setPlaybook` with a side effect/callback to do the real update as changes are made. This should be removed after graphql performance improvements are done.

CleanShot 2022-07-05 at 22 24 08

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44548
- https://mattermost.atlassian.net/browse/MM-44816

#### Checklist

Item | Y / N / NA [^1]
---|---
Telemetry updated? |
Gated by experimental feature flag? |
Unit/E2E tests updated? |


crspeller (approved):
Agree with Jesse that we should at least centralize that mapping.

calebroseland:
I had tried some changes to centralize this mapping operation, but I'm getting a number of uncertainties and bugs that would need more time to iron out. I was tentatively thinking this could go into 7.1 and would like to propose leaving the mapping as it was/is for now until we can address it more holistically.

Let's discuss this and form a plan for how we want to address the casing differences between graphql read/write, particularly how it might affect further graphql usage and development.

lieut-data:
I'm :+1: for keeping the duplication and simplifying later.

PR #1296, opened 2022-07-06
create a new telemetry event for run details page view

trilopin:
#### Summary
Implement telemetry event for tracking views of the run details page.

New event instead of reusing previous ones. Additionally, a hack has been included in the telemetry hook to prevent retrospective events to be triggered when we re-use the component inside run-details page.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45546

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/hooks/telemetry.ts
+++ b/webapp/src/hooks/telemetry.ts
@@ -1,4 +1,5 @@
 import {useEffect} from 'react';
+import {useLocation} from 'react-use';
 
 import {telemetryEventForPlaybook, telemetryEventForPlaybookRun} from 'src/client';
 
@@ -9,8 +10,13 @@ export enum PlaybookTarget {
 }
 
 export enum PlaybookRunTarget {
+
+    // @deprecated triggered at old run details
     Overview = 'view_run_overview',
+
+    // @deprecated triggered at old run details
     Retrospective = 'view_run_retrospective',
+    Details = 'view_run_details',
     ChannelsRHSDetails = 'view_run_channels_rhs_details',
 }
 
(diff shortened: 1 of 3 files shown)


webapp/src/hooks/telemetry.ts
```diff
@@ -23,7 +29,12 @@ export const usePlaybookViewTelemetry = (target: PlaybookTarget, playbookID?: st
 };
 
 export const usePlaybookRunViewTelemetry = (target: PlaybookRunTarget, playbookRunID?: string) => {
+    const {pathname} = useLocation();
     useEffect(() => {
+        // Needed until we remove the old run details
+        if (pathname?.includes('/playbooks/run_details/') && target === PlaybookRunTarget.Retrospective) {
+            return;
+        }
```
agarciamontoro, line 37: I'm not 100% sure I'm understanding this correctly, but: what if we just remove the usage of this hook when calling it with `target` set to `PlaybookRunTarget.Retrospetive`? Or will we do that once we change the old url to point to the new component?

agarciamontoro (commented):
A question below

lieut-data (approved):
I'm fine with this, provided we remember to remove it before shipping :)

webapp/src/hooks/telemetry.ts
trilopin: Yes, When we remove the old URL the shared component will be used only from the new RUn Details Page and the inner track could be removed. This is only needed while we have both URLs live.

trilopin: > I'm fine with this, provided we remember to remove it before shipping :)

There's a task with some cleanup here https://mattermost.atlassian.net/browse/MM-45451

trilopin:
> I'm fine with this, provided we remember to remove it before shipping :)

There's a task with some cleanup here https://mattermost.atlassian.net/browse/MM-45451

webapp/src/hooks/telemetry.ts
agarciamontoro: Sounds good!

agarciamontoro:
@trilopin I wanted to approve and hit merge :facepalm: but I guess it's the same result. Too early in the morning, I guess.

trilopin:
> I'm fine with this, provided we remember to remove it before shipping :)

hack removed at https://github.com/mattermost/mattermost-plugin-playbooks/pull/1322

PR #1299, opened 2022-07-07
Check .only mark in e2e test as part of check-style

trilopin:
#### Summary
Additional linter check in case we forget some `it.only` or `describe.only` in our e2e tests.

Screenshot 2022-07-07 at 18 20 45

#### Ticket Link
no ticket

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Gated by experimental feature flag~~
- [ ] ~~Unit tests updated~~

PR #1301, opened 2022-07-08
run eslint over e2e tests on make check-style + add only check

trilopin:
#### Summary

I often forget  to remove `.only` modifiers in cypress e2e tests. I added https://github.com/mattermost/mattermost-plugin-playbooks/pull/1299 to tackle that with old-style grep/bash.

After @saturninoabril [comment](https://community-daily.mattermost.com/core/pl/fatrdcyqfpyy38po97io1uhioh), I suggest this improvement:
- check it in a more JS-native way
- check eslint rules, not only for `only` but also for the already existing rules as part of `make check-style` Makefile command.

I had to clean some existing failures, most of them unused-var. 1/5 with some of them.

I could ignore the rules for the `_ignore_.js` specs but I'm wondering if it's maybe better to clean them up (by removing them or fixing them).

#### Ticket Link
Nope

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,10 +53,7 @@ ifneq ($(HAS_WEBAPP),)
 	cd webapp && npm run check-types
 endif
 
-	@if grep -rin -e 'it\.only' -e 'describe\.only' tests-e2e/cypress/integration; then \
-		echo "\nThere are at least one e2e test marked as only, only should be removed.\n"; \
-		exit 1; \
-	fi
+	cd tests-e2e && npm run check
 
 ifneq ($(HAS_SERVER),)
 	@if ! [ -x "$$(command -v golangci-lint)" ]; then \
--- a/tests-e2e/.eslintrc
+++ b/tests-e2e/.eslintrc
@@ -4,7 +4,7 @@
     "plugin:cypress/recommended",
     "plugin:react/recommended"
   ],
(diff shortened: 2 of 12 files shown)


Phrynobatrachus (approved):
Nice, thanks for the extra lint cleanup as well.

agarciamontoro (changes requested):
Much much better, thank you!

About the changes in the tests themselves:
- There are several cases where an assignment is removed after an API call (create channel/playbook). If we don't use that resource, can we remove the API call altogether?
- There are some `eslint-disable-next-line no-unused-vars` comments. Why do we need those?

trilopin:
> Much much better, thank you!
> 
> About the changes in the tests themselves:
> 
> * There are several cases where an assignment is removed after an API call (create channel/playbook). If we don't use that resource, can we remove the API call altogether?

The fact that you don't need the var to do an assertion, does not mean that the API call is not necessary. does it? Are you concerned about some specifically?

> * There are some `eslint-disable-next-line no-unused-vars` comments. Why do we need those?

I was 0/5 and pretty conservative on some ignored specs. Also, there's a ticket to check ignored specs at https://mattermost.atlassian.net/browse/MM-45611

PR #1221, opened 2022-05-28
Mm 20021

lapaz17:
#### Summary
Opens a new modal whenever Use button is pressed inside RHS template instead of navigating to playbooks/new url. 

#### Ticket Link

  Fixes: https://github.com/mattermost/mattermost-server/issues/20021

#### Checklist

~~[ ] Telemetry updated~~
~~[ ] Gated by experimental feature flag~~
~~[ ] Unit tests updated~~

PR #1304, opened 2022-07-08
Mm 45608 create public playbooks from rhs when no e20

trilopin:
#### Summary
Environment:
- branch 7.1 (without new modal with team selector)
- below E20 (no private playbook permission)

Steps to reproduce:
- open RHS in a channel without RUNs
- click on "use" template to create a playbook

Result: 403 forbidden since we try to create a public playbook

3/5 on solution, I'm not very familiar with permissions. I've been inspired from create_playbook_modal but I'd appreciate some review.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45608

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/rhs/rhs_home.tsx
+++ b/webapp/src/components/rhs/rhs_home.tsx
@@ -33,7 +33,7 @@ import {
     usePlaybooksCrud,
     getPlaybookOrFetch,
     usePlaybooksRouting,
-    useCanCreatePlaybooksOnAnyTeam,
+    useHasTeamPermission,
 } from 'src/hooks';
 import {navigateToUrl} from 'src/browser_routing';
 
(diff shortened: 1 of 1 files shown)


trilopin:
@lieut-data any idea why CI is complaining about golang-ci?

lieut-data:
@trilopin, I added one small tweak to use your code to also gate the screen that's supposed to show when a user can't create a playbook at all. This appears to only work previously when all teams were configured in the same way, but since we're scoped to a team in this context, we can just use the current team's permissions.

lieut-data:
(I'm also investigating what's wrong with CI.)

lieut-data:
@trilopin, I think the issue here is the attempt to merge directly into the release branch, vs. `master`. I'm going to rebase and pull in your changes, then cherry-pick back to the release branch after.

lieut-data:
@trilopin, it looks like the bulk of this was fixed with https://github.com/mattermost/mattermost-plugin-playbooks/pull/1221, where we just used the proper dialog instead of auto-creating it. So I opted to pull /that/ change into our release branch, and just kept the part I contributed here to scope the permission checks to team more correctly.

lieut-data (approved):

PR #1308, opened 2022-07-11
recover kebab and request update as participant

trilopin:
#### Summary
Allow Participants to also request updates

#### Ticket Link
FIxes https://mattermost.atlassian.net/browse/MM-45598

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
```diff
@@ -72,12 +75,39 @@ interface ViewerProps {
...
+            message={formatMessage({defaultMessage: 'A message will be sent to the run channel, requesting them to post an update.'})}
+            confirmButtonText={formatMessage({defaultMessage: 'Request update'})}
+            onConfirm={() => {
+                requestStatusUpdate();
+                setShowRequestUpdateConfirm(false);
+            }}
+            onCancel={() => setShowRequestUpdateConfirm(false)}
+        />
+    );
+    return {RequestUpdateConfirmModal, setShowRequestUpdateConfirm};
```
agarciamontoro, line 103: Nit: I'm perfectly ok with this, but if I'm not mistaken, the convention is to return an array instead of an object.

agarciamontoro: By the way, great work encapsulating this, I love seeing small hooks uncluttering the main component code.

agarciamontoro: Another nit: you could even return a function only to show the modal, insteasd of the general `setShowRequestUpdateConfirm`, since `setShowRequestUpdateConfirm(false)` is only called inside the hook, if I read the code correctly.

agarciamontoro (approved):
Code looks perfect!

webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
trilopin: I like the point for exposing only the modal open. Regarding array vs object return on hooks, I find the object more suited in two cases:
- more than 2 items were returned, hard to remember the positions
- avoid explicit typing (this case). If I want to return in an array of two different types, I have to type it explicitly and the whole thing ends up being more complex. 

When you say convention, do you mean react or mattermost? I would love to read about the reasons why that convention.

webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
```diff
@@ -75,12 +77,42 @@ interface ViewerProps {
     openRHS: (section: RHSContent, title: React.ReactNode, subtitle?: React.ReactNode) => void;
 }
 
-export const ViewerStatusUpdate = ({id, playbookRun, openRHS, lastStatusUpdate}: ViewerProps) => {
+const useRequestUpdate = (playbookRunId: string) => {
```
iomodo, line 80: maybe moving this in hooks folder?

iomodo (approved):
Looks Great!

PR #1309, opened 2022-07-11
track role data for playbookRun + new events preparation

trilopin:
#### Summary
Any event sent to rudder that is related to a playbook run will include a field to reflect if the user who triggered the event is a participant or a viewer (server&frontend events).

Starts work on get-involved / request an update tracking.

Telemetry types moved outside hooks and created the structure for events. Eventually, I'd like to migrate all events to these types and remove string compatibility. 
 
#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45414

#### Checklist
- [x] Telemetry updated
- [ ] ~~Unit tests updated~~


webapp/src/types/telemetry.ts
```diff
@@ -0,0 +1,26 @@
+export enum PlaybookViewTarget {
+    Usage = 'view_playbook_usage',
+    Outline = 'view_playbook_outline',
+    Reports = 'view_playbook_reports'
+}
+
+export enum PlaybookEventTarget {
```
lieut-data, line 7: Is this a placeholder?

lieut-data (approved):

webapp/src/types/telemetry.ts
trilopin: yes, I started migrating all current event right away. I found things that are complex and I need a wider view of all telemetry. That's why I stopped that approach and just did for the new events.

I can remove it until we have the first event in the new way.

agarciamontoro (approved):
Thanks! :tada:

PR #1311, opened 2022-07-11
MM-44722: Checklist improvements

agarciamontoro:
#### Summary

This PR slightly changes the behaviour of the buttons in the checklist item: the ones for modifying the assignee, the slash command and the due date. In particular:
- The buttons that have no value now have a transparent background and a border. This applies to the Run Details page and the RHS.
- The rest of the behaviour in the RHS is unchanged.
- In the Run Details page, the viewers can only see buttons that do have a value, as before, but now the owner button always shows the full name. There's a bug right now through which viewers can change the assignee, but that's not a regression and will be tackled in a different PR. I created a ticket:
- In the Run Details page, the participants see all buttons all the time, regardless of whether they have a value or not.

Check the screenshots below to see the new style.

##### Viewer mode
image

##### Participant mode
image

##### Checklist in the RHS is still the same
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44722

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


agarciamontoro:
I'll take a look at those failing tests as soon as possible!

trilopin (approved):
Cool!

webapp/src/components/checklist_item/command.tsx
```diff
@@ -27,6 +27,8 @@ interface CommandProps {
     isEditing: boolean;
 
     onChangeCommand: (newCommand: string) => void;
+
+    showAlways?: boolean;
```
iomodo, line 31: are we using this prop?

iomodo (approved):
Looks Great @agarciamontoro !

PR #1312, opened 2022-07-11
[MM-44755] Favoriting Playbooks

iomodo:
#### Summary
This PR adds means to favorite Playbook.
Feature is implemented with GraphQL.
There are REST API endpoints `favorite`, `unfavorite` and `isfavorite` they are not being used in this PR, but we will need them for favoriting runs.

Currently list of runs is not being updated, so after favoriting the playbook you have to refresh to see it in the LHS.

https://user-images.githubusercontent.com/2340127/178334235-66ff4734-0298-4ed5-a9c4-a7681a318c5c.mp4

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-44755

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


server/sqlstore/category.go
```diff
@@ -180,3 +180,66 @@ func (c *categoryStore) Delete(categoryID string) error {
...
+			ID:        model.NewId(),
+			Name:      "Favorite",
+			TeamID:    teamID,
+			UserID:    userID,
+			Collapsed: false,
+			CreateAt:  now,
+			UpdateAt:  now,
+			Items:     []app.CategoryItem{},
+		}
+		if err := c.Create(category); err != nil {
```
crspeller, line 205: Why bother actually writing it back? Shouldn't it be created on write not read? It seems like as written if I where to try and add a favorite before reading (using API directly) it would fail.

server/api/categories.go
```diff
@@ -35,6 +35,9 @@ func NewCategoryHandler(router *mux.Router, api *pluginapi.Client, logger bot.Lo
 	categoriesRouter := router.PathPrefix("/my_categories").Subrouter()
 	categoriesRouter.HandleFunc("", handler.getMyCategories).Methods(http.MethodGet)
 	categoriesRouter.HandleFunc("", handler.createMyCategory).Methods(http.MethodPost)
+	categoriesRouter.HandleFunc("/favorites", handler.addFavorite).Methods(http.MethodPost)
```
crspeller, line 38: I thought we where going to do this through GraphQL?

webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
```diff
@@ -61,16 +62,21 @@ const PlaybookEditor = () => {
...
 
+    const [isFavoritePlaybook, setIsFavoritePlaybook] = useState(false);
+
     useEffect(() => {
         const teamId = playbook?.team_id;
         if (!teamId) {
             return;
         }
 
+        isFavoriteItem(teamId, playbookId, CategoryItemType.PlaybookItemType)
```
crspeller, line 73: We already have the playbook and it should include the isFavourite you added right?

crspeller (changes requested):
Nice! Love the GraphQL usage!
Not sure why there is also a REST way to do this? Seems like we should be able to use GraphQL all the way here.

server/sqlstore/category.go
iomodo: GetFavoriteCategory is called on write actually so no fail would occur, but you are right, readability is bad here, will refactor

crspeller (changes requested):

crspeller (approved):

calebroseland (approved):
This is great! ⭐️

iomodo:
Removing @abhijit-singh's review, will address his suggestions in the next PR.

calebroseland:
UX review deferred cc @abhijit-singh

PR #1315, opened 2022-07-12
fine tune run title control + toggle open in dotmenu

trilopin:
#### Summary
Feedback on the run title:
- adjusted padding 
- hover/active/open styles as the channel context menu
- dot menu works like a toggle now **globally** (click again if open to close it). This affect lots of other dotmenus, but it seems globally desired 

```
2. The hover state on the run name is mis-aligned. Recommended change: padding: 2px 2px 2px 6px;
3. When the run name dropdown is open, the run name should stay in the highlighted state. Please check channel name dropdown for reference.
``` 

Demo:
https://user-images.githubusercontent.com/4096774/178491441-c1236ac1-63b3-40e1-b0c5-5febff991e02.mp4

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-45674

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/dot_menu.tsx
+++ b/webapp/src/components/dot_menu.tsx
@@ -80,7 +80,7 @@ const DotMenu = (props: DotMenuProps) => {
 
     const [isOpen, setOpen] = useState(false);
     const toggleOpen = () => {
-        setOpen(true);
+        setOpen(!isOpen);
     };
 
     useClickOutsideRef(refs.reference, () => {
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
@@ -107,7 +107,6 @@ const Title = styled.h1`
     `;
 
 export const TitleButton = styled.div<{isActive: boolean}>`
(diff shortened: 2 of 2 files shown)


webapp/src/components/dot_menu.tsx
```diff
@@ -80,7 +80,7 @@ const DotMenu = (props: DotMenuProps) => {
 
     const [isOpen, setOpen] = useState(false);
     const toggleOpen = () => {
-        setOpen(true);
+        setOpen(!isOpen);
     };
```
agarciamontoro, line 84: A toggle that always set open to true :joy:

agarciamontoro (approved):
I think this makes sense, although I'm concerned about toggle that always set open to true. Was that intentional for some component or a mistake we made when we added that code?

trilopin:
> I think this makes sense, although I'm concerned about toggle that always set open to true. Was that intentional for some component or a mistake we made when we added that code?

I have a similar feeling, feels ok but it's widely used. We can pay special attention to qa here or we can make it optional.

abhijit-singh (approved):
💯

calebroseland (approved):

PR #1317, opened 2022-07-12
MM-45575 Playbook member migration.

crspeller:
#### Summary
Add all users who are participating in runs to playbooks to they will appear in the LHS when we switch over. 

Any user who is a participant in any of a public playbook's runs should be added to that playbook as a member. 

@Phrynobatrachus could I have your help testing this?

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45575

### DIFF
--- a/server/sqlstore/migrations.go
+++ b/server/sqlstore/migrations.go
@@ -2108,24 +2108,24 @@ var migrations = []Migration{
 			// Find all users who are members of channels where runs have been created.
 			// Add them as members of the playbook but only if it's a public playbook.
 			if _, err := e.Exec(`
-			INSERT INTO IR_PlaybookMember
-				SELECT
-					pb.ID as PlaybookID,
-					cm.UserID as MemberID,
-					'playbook_member' as Roles
-				FROM IR_Playbook as pb
-				JOIN IR_Incident as run on run.PlaybookID = pb.ID
-				LEFT JOIN ChannelMembers as cm on cm.ChannelID = run.ChannelID
-				LEFT JOIN IR_PlaybookMember as pm on pm.PlaybookID = pb.ID AND pm.MemberID = cm.UserID
-				LEFT JOIN Bots as b ON b.UserID = cm.UserID
-				WHERE
-					pb.Public = true AND
-					pb.DeleteAt=0 AND
-					pm.PlaybookID is NULL AND
-					b.UserId IS NULL
-				GROUP BY pb.ID, cm.UserID
+				INSERT INTO IR_PlaybookMember
+					SELECT DISTINCT
+						pb.ID as PlaybookID,
+						cm.UserID as MemberID,
+						'playbook_member' as Roles
+					FROM IR_Playbook as pb
+					JOIN IR_Incident as run on run.PlaybookID = pb.ID
+					JOIN ChannelMembers as cm on cm.ChannelID = run.ChannelID
+					LEFT JOIN IR_PlaybookMember as pm on pm.PlaybookID = pb.ID AND pm.MemberID = cm.UserID
+					LEFT JOIN Bots as b ON b.UserID = cm.UserID
+					WHERE
+						pb.Public = true AND
+						pb.DeleteAt = 0 AND
+						pm.PlaybookID IS NULL AND
+						b.UserId IS NULL
 			`); err != nil {
-				return errors.Wrapf(err, "failed to add existing users as playbook members")
+				// Migration is optional so no failure just logging. (it will not try again)
+				sqlStore.log.Debugf("%w", errors.Wrapf(err, "failed to add existing users as playbook members"))
 			}
 
 			return nil
--- a/server/sqlstore/migrations.go
+++ b/server/sqlstore/migrations.go
@@ -2101,4 +2101,34 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.55.0"),
+		toVersion:   semver.MustParse("0.56.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			// Find all users who are members of channels where runs have been created.
+			// Add them as members of the playbook but only if it's a public playbook.
+			if _, err := e.Exec(`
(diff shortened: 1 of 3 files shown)


server/sqlstore/migrations.go
```diff
@@ -2098,6 +2098,36 @@ var migrations = []Migration{
...
+				LEFT JOIN IR_PlaybookMember as pm on pm.PlaybookID = pb.ID AND pm.MemberID = cm.UserID
+				LEFT JOIN Bots as b ON b.UserID = cm.UserID
+				WHERE
+					pb.Public = true AND
+					pb.DeleteAt=0 AND
+					pm.PlaybookID is NULL AND
+					b.UserId IS NULL
+				GROUP BY pb.ID, cm.UserID
+			`); err != nil {
+				return errors.Wrapf(err, "failed to add existing users as playbook members")
```
lieut-data, line 2128: Should we consider /not/ failing the migration in this case, but just log an error? (So that Playbooks at least keeps going.)

server/sqlstore/migrations.go
```diff
@@ -2098,6 +2098,36 @@ var migrations = []Migration{
...
+		},
+	},
+	{
+		fromVersion: semver.MustParse("0.55.0"),
+		toVersion:   semver.MustParse("0.56.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			// Find all users who are members of channels where runs have been created.
+			// Add them as members of the playbook but only if it's a public playbook.
+			if _, err := e.Exec(`
+			INSERT INTO IR_PlaybookMember
```
lieut-data, line 2111: nit: worth indenting the block once more?

server/sqlstore/migrations.go
```diff
@@ -2098,6 +2098,36 @@ var migrations = []Migration{
...
+					cm.UserID as MemberID,
+					'playbook_member' as Roles
+				FROM IR_Playbook as pb
+				JOIN IR_Incident as run on run.PlaybookID = pb.ID
+				LEFT JOIN ChannelMembers as cm on cm.ChannelID = run.ChannelID
+				LEFT JOIN IR_PlaybookMember as pm on pm.PlaybookID = pb.ID AND pm.MemberID = cm.UserID
+				LEFT JOIN Bots as b ON b.UserID = cm.UserID
+				WHERE
+					pb.Public = true AND
+					pb.DeleteAt=0 AND
```
lieut-data, line 2123: nit:
```suggestion
					pb.DeleteAt = 0 AND
```

server/sqlstore/migrations.go
```diff
@@ -2098,6 +2098,36 @@ var migrations = []Migration{
...
+					pm.PlaybookID is NULL AND
```
lieut-data, line 2124: nit:
```suggestion
					pm.PlaybookID IS NULL AND
```

server/sqlstore/migrations.go
```diff
@@ -2098,6 +2098,36 @@ var migrations = []Migration{
...
+	},
+	{
+		fromVersion: semver.MustParse("0.55.0"),
+		toVersion:   semver.MustParse("0.56.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			// Find all users who are members of channels where runs have been created.
+			// Add them as members of the playbook but only if it's a public playbook.
+			if _, err := e.Exec(`
+			INSERT INTO IR_PlaybookMember
+				SELECT
```
lieut-data, line 2112: Do we need a `SELECT DISTINCT` here in case the query yields a user as being part of the playbook multiple times?

server/sqlstore/migrations.go
```diff
@@ -2098,6 +2098,36 @@ var migrations = []Migration{
...
+					pb.ID as PlaybookID,
+					cm.UserID as MemberID,
+					'playbook_member' as Roles
+				FROM IR_Playbook as pb
+				JOIN IR_Incident as run on run.PlaybookID = pb.ID
+				LEFT JOIN ChannelMembers as cm on cm.ChannelID = run.ChannelID
```
lieut-data, line 2118: Should this just be a `JOIN` since we don't want to yield results without a channel membership?

lieut-data (changes requested):
Awesome! A few comments inline on edge cases.

server/sqlstore/migrations.go
crspeller: That's probably for the best.

server/sqlstore/migrations.go
crspeller: I think the group by should cover it, but I guess it doesn't hurt.

lieut-data: Ah, you're right! I might actually suggest writing this as a `DISTINCT` vs. `GROUP BY`, the latter of which I usually associate with aggregate queries only.

server/sqlstore/migrations.go
```diff
@@ -2108,24 +2108,24 @@ var migrations = []Migration{
...
+					LEFT JOIN Bots as b ON b.UserID = cm.UserID
+					WHERE
+						pb.Public = true AND
+						pb.DeleteAt = 0 AND
+						pm.PlaybookID IS NULL AND
+						b.UserId IS NULL
 			`); err != nil {
-				return errors.Wrapf(err, "failed to add existing users as playbook members")
+				// Migration is optional so no failure just logging. (it will not try again)
+				sqlStore.log.Debugf("%w", errors.Wrapf(err, "failed to add existing users as playbook members"))
```
lieut-data, line 2128: nit: Propose still keeping this as `Errorf`, since I do think it's actionable (we probably have to fix something).

lieut-data (approved):

server/sqlstore/migrations.go
iomodo: Since `GROUP BY` is used mainly for aggregates I'd suggest to use `DISTINCT` and get rid of `GROUP BY`

crspeller: ^ already done.

iomodo: Did not refresh :D

iomodo (approved):
Looks great 👍

PR #1327, opened 2022-07-14
MM-44754, MM-45087: Playbooks LHS Team Sidebar

calebroseland:
#### Summary
- Add: Team Sidebar 
  - Change: re-scope much of Playbooks to be team-specific again
  - No routing changes, deep-linking/permalinks still supported
- Fix: minor responsive layout issue that cause LHS and CenterPane to overlap. Note that mobile breakpoints cause team sidebar and global header to disappear—these are out of our control atm
- Add/change: LHS item active+hover styling
- Fix/preemptive: LHS item `key` could contain duplicates, changed to playbook/run `id`, falls back to `display_name`
- Remove: experimental flag for LHS (also see #1322)

NS:
- NPS/Feedback needs a new home (subsequent PR) https://mattermost.atlassian.net/browse/MM-45738
- ~~E2E needs fixing (14 failing, probably related to old navbar)~~ remaining failing in `runs/list_spec` fixed by #1322 
- Remove dead/removed codepaths (subsequent PR(s))

| before | after |
|---|---|
| CleanShot 2022-07-14 at 02 09 04@2x | image |
|  image  |  image |
| CleanShot 2022-07-14 at 02 04 38 | CleanShot 2022-07-14 at 02 02 31 |

#### Remarks

Loading playbooks and changing teams will keep in sync with Channels' localstorage item `user_prev_team:<user_id>` for better product transitions and refresh-safe UX. 

Re: Team Sidebar routing changes: This is a "Plan B" implementation, in that it doesn't change the routing structure or add url parameters.

Plan A (possibly to come back and do eventually) is to add a teamName parameter to the beginning of all routes in playbooks e.g. `mylocalinstance.local/teamNameHere/playbooks/*`. This was too complicated to change for this first go-round, but I have a stash that's mostly working with some redirect/history bugs to iron out.

This Plan B has some tradeoffs compared with A (see complicated side-effects in `main_body`).

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44754
- https://mattermost.atlassian.net/browse/MM-45087

#### Checklist

Item | Y / N / NA [^1]
---|---
Telemetry updated? | N
Gated by experimental feature flag? | Removed
Unit/E2E tests updated? | Y


calebroseland:
UX review deferred cc @abhijit-singh

webapp/src/components/backstage/main_body.tsx
```diff
@@ -11,12 +21,70 @@ import PlaybookEdit from 'src/components/backstage/playbook_edit/playbook_edit';
...
+    const dispatch = useDispatch();
+    const location = useLocation();
+    const {url} = useRouteMatch();
+    const teams = useSelector(getMyTeams);
+    const currentTeamId = useSelector(getCurrentTeamId);
+    const currentUserId = useSelector(getCurrentUserId);
+
+    // ? consider moving to multi-product or plugin infrastructure
+    // see https://github.com/mattermost/mattermost-webapp/blob/25043262dbab1fc2f9ac6972b1f1b0b1f9c20ae0/stores/local_storage_store.jsx#L9
+    const [prevTeamId, setPrevTeamId] = useLocalStorage(`user_prev_team:${currentUserId}`, teams[0].id, {raw: true});
```
trilopin, line 38: silly question: Is this because we want to share the team selector from channels tab?

webapp/src/components/backstage/main_body.tsx
```diff
@@ -11,12 +21,70 @@ import PlaybookEdit from 'src/components/backstage/playbook_edit/playbook_edit';
...
+
+    /**
+     * * These routes will select the proper team they belong too.
+     * ! Don't restore prev team on these routes or those routes will redirect back to default route.
+     * @see {useDefaultRedirectOnTeamChange}
+     */
+    const negateTeamRestore = matchPath<{playbookRunId?: string; playbookId?: string;}>(location.pathname, {
+        path: [
+            `${url}/run_details/:playbookRunId`,
```
trilopin, line 47: important we clean this url as part of https://github.com/mattermost/mattermost-plugin-playbooks/pull/1322

webapp/src/components/backstage/main_body.tsx
```diff
@@ -11,12 +21,70 @@ import PlaybookEdit from 'src/components/backstage/playbook_edit/playbook_edit';
 import PlaybookEditor from 'src/components/backstage/playbook_editor/playbook_editor';
 import {NewPlaybook} from 'src/components/backstage/new_playbook';
 import {ErrorPageTypes} from 'src/constants';
-import {pluginErrorUrl} from 'src/browser_routing';
+import {pluginErrorUrl, pluginUrl} from 'src/browser_routing';
 import ErrorPage from 'src/components/error_page';
 import RunsPage from 'src/components/backstage/runs_page';
 
+const useInitTeamRoutingLogic = () => {
```
crspeller, line 28: This is for cold start correct? Because there is no team selected from channels.

crspeller (approved):
Looks great! Just one check my understanding question.

webapp/src/components/backstage/main_body.tsx
calebroseland: Cold start team selection, yes, but also updating to keep in sync as you switch between products

calebroseland:
> Looks great! one comment should not we remove teamSelects from [here](https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/8ac3bd2874de025baf36eaa4366a57d22b3183bc/webapp/src/components/backstage/playbook_editor/playbook_editor.tsx#L71) and [here](https://github.com/mattermost/mattermost-plugin-incident-collaboration/blob/4f9094cf6394c0aa4959c3cb17e59f057f1e05ac/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx#L99) ?

No, we actually need those for now for permalink purposes. We could place it behind a conditional to not fire if the currentTeamId is the same, but indeed we'll remove these selects once we get a url parameter 👍

webapp/src/components/backstage/main_body.tsx
calebroseland: Which one should we merge first?

crspeller: @calebroseland Looks like https://github.com/mattermost/mattermost-plugin-playbooks/pull/1322 is going to merge first so could you modify this URL here?

calebroseland: It merges cleanly locally, so I'll go ahead and merge

PR #1328, opened 2022-07-14
[MM-45728] Favoriting runs

iomodo:
#### Summary
This PR adds support of favoriting runs.  Currently you'll have to refresh to see changes in the LHS

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-45728

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
+++ b/webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
@@ -117,14 +117,14 @@ const PlaybookEditor = () => {
             <NavBackdrop/>
             <TitleBar>
                 <div>
-                    <Button
+                    <StarButton
                         onClick={toggleFavorite}
                         className={playbook.is_favorite ? 'active' : ''}
                     >
                         <div>
                             <i className={'icon ' + favoriteIcon}/>
                         </div>
-                    </Button>
+                    </StarButton>
                     <TextEdit
                         disabled={archived}
                         placeholder={formatMessage({defaultMessage: 'Playbook name'})}
@@ -580,12 +580,11 @@ const Editor = styled.main<{$headingVisible: boolean}>`
     }
 `;
 
-const Button = styled.button`
+export const StarButton = styled.button`
     border-radius: 4px;
     border: 0;
     padding: 12px 0 10px 0;
     background: none;
-    flex: 1;
     margin: 0 6px;
 
     &:hover {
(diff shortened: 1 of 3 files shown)


webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
```diff
@@ -580,12 +580,11 @@ const Editor = styled.main<{$headingVisible: boolean}>`
     }
 `;
 
-const Button = styled.button`
+export const StarButton = styled.button`
```
trilopin, line 583: from the figma files I have for run details page. this icon should be 18px18 size and color sidebar-active-border instead of button-bg

webapp/src/components/backstage/playbook_runs/playbook_run/header.tsx
```diff
@@ -76,8 +93,29 @@ export const RunHeader = ({playbookRun, playbookRunMetadata, role, onInfoClick,
...
+        }
+        favoriteItem(playbookRun.team_id, playbookRun.id, CategoryItemType.RunItemType);
+        setIsFavoriteRun(true);
+    };
+
     return (
         <Container data-testid={'run-header-section'}>
+            <StarButton
+                onClick={toggleFavorite}
+                className={isFavoriteRun ? 'active' : ''}
```
trilopin, line 113: optional:

both icons are available in compass, could we use the compass icons instead of mdi?  (as the other icons in this file)

webapp/src/components/backstage/playbook_runs/playbook_run/header.tsx
```diff
@@ -76,8 +93,29 @@ export const RunHeader = ({playbookRun, playbookRunMetadata, role, onInfoClick,
         addToast(formatMessage({defaultMessage: 'You\'ve joined this run.'}), ToastType.Success);
     };
 
+    // Favorite Button State
+    const favoriteIcon = isFavoriteRun ? 'icon-star' : 'icon-star-outline';
+
+    const toggleFavorite = () => {
```
trilopin, line 99: optional: I'd love to have a hook useFavorite holding useEffect+useState and returning isFavoriteRun and the toggle func.

trilopin (changes requested):
Thanks!

The only blocking point is the size&color. The rest is optional.

crspeller (approved):
Looks good! Agree with the comment that I would love to see a custom hook to clean up the component.

webapp/src/hooks/general.ts
```diff
@@ -723,3 +734,24 @@ export const useExportLogAvailable = () => {
...
+    const toggleFavorite = () => {
+        if (isFavoriteRun) {
+            unfavoriteItem(teamID, runID, CategoryItemType.RunItemType);
+            setIsFavoriteRun(false);
+            return;
+        }
+        favoriteItem(teamID, runID, CategoryItemType.RunItemType);
+        setIsFavoriteRun(true);
+    };
+    return [isFavoriteRun, toggleFavorite];
```
trilopin, line 756: wonderful!

PR #1329, opened 2022-07-14
add 72px to avoid teambar in max-width calculations

trilopin:
#### Summary
Now we have team bar, we have to add 72px to our calculatios for max-width

#### Ticket Link
no ticket

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/runs/list_spec.js
+++ b/tests-e2e/cypress/integration/runs/list_spec.js
@@ -127,21 +127,23 @@ describe('runs > list', () => {
             // # Open the product
             cy.visit('/playbooks/runs');
 
-            // # Filter to all runs
-            cy.findByTestId('my-runs-only').click();
+            cy.get('#playbookRunList').within(() => {
+                // # Filter to all runs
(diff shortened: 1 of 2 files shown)


crspeller (approved):

calebroseland (approved):
This is an improvement, but I think we should do a bit more here overall (not now, but followup PRs). I have some doubts on using calcs in this way. A layout like this using grid or flex should be able to resize to it's container without having to use max-widths and such.

CleanShot 2022-07-14 at 11 55 23@2x
CleanShot 2022-07-14 at 11 55 09@2x

trilopin:
> This is an improvement, but I think we should do a bit more here overall (not now, but followup PRs). I have some doubts on using calcs in this way. A layout like this using grid or flex should be able to resize to it's container without having to use max-widths and such.
> 
> CleanShot 2022-07-14 at 11 55 23@2x
> 
> CleanShot 2022-07-14 at 11 55 09@2x

Will check it and you'll be the lucky reviewer

PR #1331, opened 2022-07-15
fetchstate in useThing

trilopin:
### Problem

We currently have a great hook `useThing` which extracts data from the redux store and, if there's no data does a fetch call to get the data from the API. This is currently used by `useChannel`, `usePost` and `useRun`.

This useThing derived functions usually work with the following return type:
- undefined: loading
- null: not found 
- not null and not undefined: the real data 

The `null == not found` is not necessarily true, it covers actually any flow that makes the fetch promise enter the catch. So it could be 500, 403, 401, 400, or even a timeout.

Current code at playbook_run/header.tsx
```ts
const channel = useChannel(playbookRun.channel_id);

...

// Channel null value comes from error response (and we assume that is mostly 403)
// If we don't have access to channel we'll send a request to be added,
// otherwise we directly join it
if (channel === null) {
    const response = await requestGetInvolved(playbookRun.id);
    if (response?.error) {
        addToast(formatMessage({defaultMessage: 'Your request wasn\'t successful.'}), ToastType.Failure);
    } else {
        addToast(formatMessage({defaultMessage: 'Your request has been sent to the run channel.'}), ToastType.Success);
    }
    return;
}
```

In this code, I want to check if the user could have access to the channel to decide if they would join directly or send a request to be added. I'm assuming that any error means no access, which is not very accurate. Additionally, I need a comment explaining what null means in this case.

Even if you don't need to know which kind of error, it is not very readable to know undefined=loading, null=error.

### Existing approach for useFetch

`useFetch` hook was recently added. It does only the API part (compared to useThing) but also adds some metadata to track the status of the call. Since some JSX components were doing their own fetch state management I included it in `useFetch` like this:

hooks/general.ts
```ts
export function useFetch<T>(
    id: string,
    fetchFunction: (id: string) => Promise<T>,
    deps: Array<any> = [],
) {
    const [error, setError] = useState<ClientError | null>(null);
    const [fetchState, setFetchState] = useState<FetchState>(FetchState.idle);
    const [data, setData] = useState<T | null>(null);

    useEffect(() => {
        if (!id) {
            return;
        }
        setFetchState(FetchState.loading);
        fetchFunction(id)
            .then((res) => {
                setFetchState(FetchState.done);
                setData(res);
            })
            .catch((err) => {
                if (err instanceof ClientError) {
                    setError(err);
                }
                setFetchState(FetchState.error);
                setData(null);
            });
    }, [id, ...deps]);

    return [data, {state: fetchState, error}] as [T | null, FetchMetadata];
}
```

### Proposal

The proposal is to bring the same strategy for `useThing`, where the same previous example would be:

```ts
const [channel, channelFetchMetadata] = useChannel(playbookRun.channel_id);

...

if (channelFetchMetadata.isErrorCode(403)) {
    const response = await requestGetInvolved(playbookRun.id);
    if (response?.error) {
        addToast(formatMessage({defaultMessage: 'Your request wasn\'t successful.'}), ToastType.Failure);
    } else {
        addToast(formatMessage({defaultMessage: 'Your request has been sent to the run channel.'}), ToastType.Success);
    }
    return;
}  else if (channelFetchMetadata.state === FetchState.error) {
    addToast(formatMessage({defaultMessage: 'Your request wasn\'t successful.'}), ToastType.Failure);
    return;
}

```

If you are not interested on the fetch metadata, you can simply ignore second param:

```ts
const [channel] = useChannel(playbookRun.channel_id);
```

The actual implementation of the proposal is in this PR.

### Benefits

- More explicit code 
- Be able to be more accurate while handling errors.
- Be more consistent fetch state management, since it's provided by the hook and there's no need to do it in the component (with different names in each one).
- If the error is CLientError, we'll be able to handle/expose it. 

### Drawbacks

- Add [ ] to any return values of a useThing derived func 

### Risks

We have to check carefully where `usePost`, `useChannel` and `useRun` were used and review flows looking for null/undefined checks.  This PR covers the `tsc` compilation and the e2e test running green. A deeper check would be needed if we want to go ahead.


trilopin:
The proposal has evolved with the received feedback: simplify states and check if useFetch is really necessary.

**Simplify states**
The 4-states have been replaced with just one boolean `isFetching`. It will start as true and change to false when the fetch ends (success or error).

**Removed useFetch**
A no-op selector has been created (just a ()=>undefined) to be passed as a selector to useThing if we want the same behavior as the removed useFetch.

**Dep list as optional parameter in useThing**
There is a new 4th optional param with additional dependencies we could want to pass to useThing. That deps will be injected in the inner useEffect and will re-trigger the fetch if they change.

One example of this need is at runMetadata case. Sometimes we want to refetch metadata when the participant list (outside metadata) changes.

webapp/src/hooks/general.ts
```diff
@@ -320,91 +326,67 @@ export function useProfilesInChannel(channelId: string) {
...
+ * @returns Array with data in first parameter and metadata in the second.
  */
-function useThing<T extends NonNullable<any>>(
-    id: string,
-    fetch: (id: string) => Promise<T>,
-    select?: (state: GlobalState, id: string) => T,
+export function useThing<T extends NonNullable<any>>(
+    id: string| undefined,
+    fetchFunc: (id: string) => Promise<T>,
+    select: (state: GlobalState, id: string) => T,
```
crspeller, line 339: Maybe we can make this optional too with an `= noopSelector`? Or do you think we should make this explicit?

crspeller (approved):
Awesome! Love it!

webapp/src/hooks/general.ts
trilopin: yes, as long as we are aware that we'll have to pass manually the noopSelector for those cases we want to pass additional deps

trilopin: but indeed that are few cases, not the main flow

gbochora (approved):
Great work @trilopin, thanks! 🥳

PR #1332, opened 2022-07-15
Recover golangci staticcheck, gosimple and unused linters

trilopin:
#### Summary
Recover golangci linters: gosimple, unused and staticcheck.

At CI:
- Since cicleci image is forcing 1.45.2 version, a new step is added to force the tool to be udpated
- PR [sent to circleci](https://github.com/CircleCI-Public/cimg-go/pull/183) . If they accept this change, we can remove the additional step to replace the binary.

At Local dev env:
- force version check to 1.46.2 in `make check-style` to force everyone to update :) (Sorry if this makes some noise, but it's important to have the linters locally)

Changes in the codebase:
//nolint tag in the deprecated lines. VScode will still warn about it but we won't break CI.

#### Ticket Link
NOpe

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,7 @@ ifneq ($(HAS_SERVER),)
 	golangci-lint run ./...
 endif
 
+.PHONY: check-golangci
 check-golangci:
 ifneq ($(HAS_SERVER),)
 	@echo Ckecking golangci-lint
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,12 @@ DLV_DEBUG_PORT := 2346
 DEFAULT_GOOS := $(shell go env GOOS)
 DEFAULT_GOARCH := $(shell go env GOARCH)
 
+GOLANGCI_CURRENT_MAJOR := $(shell golangci-lint --version | cut -f 4 -d ' ' | cut -f 1 -d '.')
+GOLANGCI_CURRENT_MINOR := $(shell golangci-lint --version | cut -f 4 -d ' ' | cut -f 2 -d '.')
+GOLANGCI_MINIMUM_MINOR := 46
+GOLANGCI_MINIMUM_MAJOR := 1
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -56,16 +56,27 @@ endif
...
+
 	@if ! [ -x "$$(command -v golangci-lint)" ]; then \
 		echo "golangci-lint is not installed. Please see https://github.com/golangci/golangci-lint#install for installation instructions."; \
 		exit 1; \
 	fi; \
 
-	@echo Running golangci-lint
-	golangci-lint --version
-	golangci-lint run ./...
+	@if [ $(shell golangci-lint --version | cut -f 4 -d ' ') != "1.46.2" ]; then \
```
crspeller, line 72: This is going to be annoying when someone upgrades to `1.46.3` or something. Can we do a greater than check? You can see how the server does it for golang version here: https://github.com/mattermost/mattermost-server/blob/049e67b863e5dcf1eb81b53a63f89664d517a19c/Makefile#L553

crspeller (changes requested):

trilopin:
Oh wait, I made a mistake with the version check 🤦🏽

Makefile
```diff
@@ -56,16 +62,33 @@ endif
...
 
 ifneq ($(HAS_SERVER),)
+	@echo Running golangci-lint
+	golangci-lint run ./...
+endif
+
+.PHONY: check-golangci
+check-golangci:
+ifneq ($(HAS_SERVER),)
+	@echo Ckecking golangci-lint
```
Phrynobatrachus, line 72:
```suggestion
	@echo Checking golangci-lint
```

Phrynobatrachus (approved):
Thanks! Looks like the CircleCI PR is merged already as well.

crspeller (approved):

PR #1336, opened 2022-07-18
migrate RHS UI management to  cssgrid

trilopin:
#### Summary
Use cssgrid to be more resilient to changes in other external points (i.e team bar, app bar).

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45855

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
@@ -253,6 +253,7 @@ const Container = styled(ColumnContainer)`
     display: grid;
     grid-auto-flow: column;
     grid-auto-columns: 2fr minmax(400px, 1fr);
+    overflow-y: hidden;
 
     @media screen and (min-width: 1600px) {
         grid-auto-columns: 2.5fr 500px;
@@ -263,17 +264,16 @@ const MainWrapper = styled.div`
     display: grid;
     grid-auto-flow: row;
     height: 100vh;
-    `;
+    overflow-y: hidden;
+`;
 
 const Main = styled.main`
+    min-height: 0;
     padding: 0 20px 60px;
     display: grid;
     overflow-y: auto;
     place-content: start center;
     grid-auto-columns: min(780px, 100%);
-    ::-webkit-scrollbar {
-        display: none;
-    }
 `;
 const Body = styled(RowContainer)`
 `;
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
@@ -175,8 +175,8 @@ const PlaybookRunDetails = () => {
     const onTimelineClick = RHS.isOpen && RHS.section === RHSContent.RunTimeline ? RHS.close : onViewTimeline;
 
     return (
-        <Container>
-            <MainWrapper isRHSOpen={RHS.isOpen}>
+        <Container isRHSOpen={RHS.isOpen}>
+            <MainWrapper>
                 <Header isRHSOpen={RHS.isOpen}>
                     <RunHeader
                         playbookRunMetadata={metadata ?? null}
@@ -187,7 +187,7 @@ const PlaybookRunDetails = () => {
                         rhsSection={RHS.isOpen ? RHS.section : null}
                     />
                 </Header>
-                <Main isRHSOpen={RHS.isOpen}>
+                <Main>
                     <Body>
                         <Summary
                             id={PlaybookRunIDs.SectionSummary}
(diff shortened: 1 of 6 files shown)


calebroseland (commented):
I do think this fixes much of the column-calc stuff, but I think it may have regressed the 2-column scrolling contexts (now the whole page scrolls and the headers aren't sticky)

trilopin:
Hey @calebroseland, I appreciate a lot your comments in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1338.

I applied your feedback and fixed the sticky header. 

From here I have a couple of doubts:
- make sense to keep using react-custom-scrollbars for the RHS? I feel we could do the same as the main container.
- do you have any idea of how to solve the annoying scroll issue introduced by team sidebar? video attached

https://user-images.githubusercontent.com/4096774/179721893-cf85d60d-d74b-4449-b073-75f33e2a1f5a.mp4

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -250,42 +250,36 @@ const ColumnContainer = styled.div`
...
-    width: min(780px, 100%);
-    padding: 20px;
-    flex: 1;
-    margin: 40px auto;
-    display: flex;
-    flex-direction: column;
+const MainWrapper = styled.div`
+    display: grid;
+    grid-auto-flow: row;
+    height: 100vh;
```
calebroseland, line 265: Constraining to 100vh won't take the global header height into account.

```suggestion
    overflow-y: hidden;
```
`min-height: 0` would also work

the frustratingly simple reason that took me a long time to learn: https://datacadamia.com/web/css/grid/overflow#solution_set_an_explicit_minimum_not_to_auto

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -250,42 +250,36 @@ const ColumnContainer = styled.div`
...
-    max-width: ${({isRHSOpen}) => (isRHSOpen ? 'calc(100% - 400px)' : '100%')};
+    display: grid;
+    grid-auto-flow: column;
+    grid-auto-columns: 2fr minmax(400px, 1fr);
 
     @media screen and (min-width: 1600px) {
-        max-width: ${({isRHSOpen}) => (isRHSOpen ? 'calc(100% - 500px)' : '100%')};
+        grid-auto-columns: 2.5fr 500px;
     }
 `;
```
calebroseland, line 260: Same is needed here

```suggestion
    overflow-y: hidden;
`;
```
min-height: 0 would also work

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -250,42 +250,36 @@ const ColumnContainer = styled.div`
...
+
+const Main = styled.main`
+    padding: 0 20px 60px;
+    display: grid;
+    overflow-y: auto;
+    place-content: start center;
+    grid-auto-columns: min(780px, 100%);
+    ::-webkit-scrollbar {
+        display: none;
+    }
```
calebroseland, line 276: Is there a need to hide the scrollbar? I just haven't seen this used much elsewhere.

trilopin: it felt weird, but I've just checked channels and scrollbar is there. I will keep it to be consistent with channels RHS

calebroseland:
> make sense to keep using react-custom-scrollbars for the RHS? I feel we could do the same as the main container.

I don't think it hurts to use it, and either way could work—it's being used in LHS and quite a number of other places, too. I've never felt the need to use it over native scrollbars, but I do like the fade-in stuff it does.

If adding it to the main container doesn't break anything, you could for sure do that. Or you could drop it from RHS. Up to you.

trilopin:
> Just two overflow-y: hidden;s needed! (Container and MainWrapper in playbook_run)

This fixed the bug!

> Is there a need to hide the scrollbar? I just haven't seen this used much elsewhere.

I've added it again


> Constraining to 100vh won't take the global header height into account.

If I remove the 100vh I lose the ability to scroll in the main container.

calebroseland:
> If I remove the 100vh I lose the ability to scroll in the main container.

Isn't that what you want?

With `height: 100vh;`, MainWrapper is clipped/overflows by `40px`
CleanShot 2022-07-19 at 14 25 38@2x


Without `height: 100vh;` cleanly comes to an end
CleanShot 2022-07-19 at 14 25 17@2x

trilopin:
> > If I remove the 100vh I lose the ability to scroll in the main container.
> 
> Isn't that what you want?


No, I want to scroll in the main, but not the main wrapper (which includes the header). I'm happy with the current behavior. 

The extra 40px height problem (coming from  team sidebar) has been solved with your overflow-y recommendation.

trilopin:
> > > If I remove the 100vh I lose the ability to scroll in the main container.
> > 
> > 
> > Isn't that what you want?
> 
> No, I want to scroll in the main, but not the main wrapper (which includes the header). I'm happy with the current behavior.
> 
> The extra 40px height problem (coming from team sidebar) has been solved with your overflow-y recommendation.

You were right. 100vh has been removed

calebroseland (approved):
🚀

gbochora (approved):
Cool stuff! Works perfectly!

PR #1338, opened 2022-07-19
rdp grid suggestions

calebroseland:
Suggestions/thoughts for #1336

CleanShot 2022-07-19 at 00 57 16

### DIFF
--- a/webapp/src/components/backstage/backstage.tsx
+++ b/webapp/src/components/backstage/backstage.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useEffect} from 'react';
-import {Switch, Route, NavLink, useLocation, useRouteMatch} from 'react-router-dom';
+import {Switch, Route, NavLink, useRouteMatch, matchPath, useLocation} from 'react-router-dom';
 import {useSelector} from 'react-redux';
 import {useIntl} from 'react-intl';
 import styled, {css} from 'styled-components';
@@ -26,20 +26,10 @@ import {ToastProvider} from './toast_banner';
 import LHSNavigation from './lhs_navigation';
 import MainBody from './main_body';
 
-const BackstageContainer = styled.div<{isRunDetailsPage: boolean}>`
+const BackstageContainer = styled.div`
     background: var(--center-channel-bg);
-    display: flex;
-    flex-direction: column;
     overflow-y: auto;
     height: 100%;
-
-    ${({isRunDetailsPage}) => isRunDetailsPage && css`
-        // Accounts for the run header, so when scrolling with
-        // anchor links, the sections are not hidden under it.
-        scroll-padding-top: 64px;
-
-        scroll-behavior: smooth;
-    `}
 `;
 
 const BackstageTitlebarItem = styled(NavLink)`
@@ -76,17 +66,15 @@ const BackstageTitlebarItem = styled(NavLink)`
     }
 `;
 
-const BackstageBody = styled.div`
-    flex-grow: 1;
-    display: flex;
-    flex-direction: column;
-`;
-
 const Backstage = () => {
     const {pathname} = useLocation();
+    const {url} = useRouteMatch();
 
-    // TODO: Change  to /playbooks/runs when we get the new run details page to production
-    const isRunDetailsPage = pathname.startsWith('/playbooks/run_details');
+    const noContainerScroll = matchPath<{playbookRunId?: string; playbookId?: string;}>(pathname, {
+        path: [
+            `${url}/runs/:playbookRunId`,
+        ],
+    });
 
     const currentTheme = useSelector<GlobalState, Theme>(getTheme);
     useEffect(() => {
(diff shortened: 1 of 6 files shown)


webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -244,40 +244,33 @@ const RowContainer = styled.div`
...
-    padding: 20px;
-    flex: 1;
-    margin: 40px auto;
-    display: flex;
-    flex-direction: column;
+    padding: 0 20px 60px;
+    display: grid;
+    overflow-y: auto;
+    place-content: start center;
+    grid-auto-columns: min(780px, 100%);
```
calebroseland, line 268: maybe `scroll-behavior: smooth;` would go back in here?

webapp/src/components/backstage/backstage.tsx
```diff
@@ -26,20 +26,10 @@ import {ToastProvider} from './toast_banner';
 import LHSNavigation from './lhs_navigation';
 import MainBody from './main_body';
 
-const BackstageContainer = styled.div<{isRunDetailsPage: boolean}>`
+const BackstageContainer = styled.div`
     background: var(--center-channel-bg);
-    display: flex;
-    flex-direction: column;
```
calebroseland, line 32: These wouldn't have to be removed, but they don't do anything in my testing

webapp/src/components/backstage/backstage.tsx
```diff
@@ -122,10 +107,15 @@ const Backstage = () => {
...
+const MainContainer = styled.div<{noContainerScroll: boolean}>`
     display: grid;
     grid-auto-flow: column;
-    grid-template-columns: max-content auto;
+    grid-auto-columns: max-content auto;
+    ${({noContainerScroll}) => (noContainerScroll ? css`
+        height: 100%;
+    ` : css`
+        min-height: 100%;
+    `)}
```
calebroseland, line 118: This is crucial:
- `min-height: 100%` to prevent some page-height issues in various routes/children
- `height: 100%` to constrain the page and negate `BackstageContainer`'s scrolling and give scrolling control to the rdp

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -244,40 +244,33 @@ const RowContainer = styled.div`
...
-`;
-
-const Container = styled(ColumnContainer)<{isRHSOpen: boolean}>`
+const Container = styled.div`
     display: grid;
     grid-auto-flow: column;
-    grid-template-columns: ${({isRHSOpen}) => (isRHSOpen ? 'auto 400px' : 'auto')};
-
-
+    grid-auto-columns: 2fr minmax(300px, 1fr);
```
calebroseland, line 250: `grid-auto-columns` makes the grid implicit, so when RDP RHS is hidden the second column doesn't apply; it doesn't leave an empty column taking up space—the main content column goes full-width without having to redefine an explicit grid 
(just learned this via skimming https://bitsofco.de/understanding-the-difference-between-grid-template-and-grid-auto/)

calebroseland: Under 1600px, the `2fr`:`1fr` ratio gives more of the extra space to center content area (rhs expands at half the rate that center content does)

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -244,40 +244,33 @@ const RowContainer = styled.div`
...
     @media screen and (min-width: 1600px) {
-        grid-template-columns: ${({isRHSOpen}) => (isRHSOpen ? 'auto 500px' : 'auto')};
+        grid-auto-columns: 2.5fr 500px;
     }
```
calebroseland, line 253: 1600px and over cuts rhs off at 500px like before

webapp/src/components/backstage/playbook_runs/playbook_run/rhs.tsx
```diff
@@ -87,7 +87,6 @@ const Container = styled.div<{isOpen: boolean}>`
     flex-direction: column;
     border-left: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
     right: 0;
-    z-index: 2;
```
calebroseland, line 90: I don't think this is needed anymore?

webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
```diff
@@ -240,7 +240,10 @@ export const ParticipantStatusUpdate = ({id, playbookRun, openRHS}: ParticipantP
                         </PostUpdateButton>
                     ) : null}
                     <Kebab>
-                        <DotMenu icon={<ThreeDotsIcon/>}>
+                        <DotMenu
+                            icon={<ThreeDotsIcon/>}
+                            placement='bottom-end'
+                        >
```
calebroseland, line 246: Just to fix a scrolling glitch now that this would open past the new scrolling boundary

PR #1139, opened 2022-04-17
[MM-42660] Create a timeline event when an update is snoozed.

jeffreyssmith2nd:
#### Summary

Adds a timeline event to record when a status update reminder is snoozed.

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/19908

#### Checklist

~~- Telemetry updated~~
~~- Gated by experimental feature flag~~
- [X] Unit tests updated

PR #1350, opened 2022-07-20
MM-42660 - Timeline event for status update snooze

Phrynobatrachus:
Finishing up #1139, it was only missing the case in `timeline_event_item.tsx`.  I also did not include the generated mocks from the parent as it seems like those are no longer used. I pushed up a separate branch since I had to fix some master conflicts and didn't want to potentially screw up the original branch.

Apologies for the long delay in getting this sorted @jeffreyssmith2nd, thank you for the contribution!

Screenshot from 2022-07-20 14-50-40


lieut-data (approved):
Nice work, @jeffreyssmith2nd and @Phrynobatrachus!

iomodo (approved):
Perfect!
Thanks @Phrynobatrachus  and @jeffreyssmith2nd ^_^

PR #1351, opened 2022-07-20
Collapse categories

iomodo:
#### Summary
This PR will support collapsing categories. 
Currently we do not save `RUNS` and `Playbooks` categories in the DB(we generate them), so this PR does not add the possibility to save collapsed state for those 2 categories. For other categories everything works fine.

If you collapse `RUNS` or `PLAYBOOKS` and refresh they will appear expanded. If this is not usable, I'll add the collapsed state for those too.

https://user-images.githubusercontent.com/2340127/180065696-ac0f76d2-7a46-450e-8d91-dfc39b893c41.mp4

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/client.ts
+++ b/webapp/src/client.ts
@@ -778,6 +778,14 @@ export const fetchMyCategories = async (teamID: string): Promise<Category[]> => 
     return data;
 };
 
+export const setCategoryCollapsed = async (categoryID: string, collapsed: boolean) => {
+    try {
+        return await doPost(`${apiUrl}/my_categories/${categoryID}/collapse`, collapsed);
+    } catch (error) {
+        return {error};
+    }
+};
+
 export const doGet = async <TData = any>(url: string) => {
     const {data} = await doFetchWithResponse<TData>(url, {method: 'get'});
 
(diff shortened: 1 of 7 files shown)


webapp/src/client.ts
```diff
@@ -778,6 +778,14 @@ export const fetchMyCategories = async (teamID: string): Promise<Category[]> =>
     return data;
 };
 
+export const setCategoryCollapsed = async (categoryID: string, collapsed: boolean) => {
+    try {
+        return await doPost(`${apiUrl}/my_categories/${categoryID}/collapse`, collapsed);
```
lieut-data, line 783: For REST semantics (if that even matters given our eventual move to GraphQL), should this be a PUT?

lieut-data (approved):
Nice! A handful of comments: will let you drive on next steps since I won't be back to re-review tomorrow.

stephenvanhemmen (approved):
I don't see this in the demo, but I'm assuming if one category is collapsed I can still interact with things in other categories without changing that state. For example, if I have runs collapsed I want to switch between different playbooks while the runs category stays collapsed. If that's the case, then looks great!

crspeller (approved):

PR #1352, opened 2022-07-21
Mm 45834 leave run

gbochora:
#### Summary
This PR adds `Leave run` functionality to the RDP. `Leave run` removes the user from the run's channel, so the user becomes a `Viewer`. We show the `Leave run` menu item only to participants.

https://user-images.githubusercontent.com/4368372/180148025-a946f2c6-f99f-4a57-94f7-67e4f9d94428.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45834

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated


trilopin:
I'm here in two minds... 

On one side I would like to do the same as we do for get-involved: just dispatch(leaveChannel(channelId)) and avoid creating more code to replicate webapp/server code.

On the other side, this could be more the way we work if we separate channel members from run participants.

server/app/playbook_run_service.go
```diff
@@ -2745,6 +2745,26 @@ func (s *PlaybookRunServiceImpl) RequestGetInvolved(playbookRunID, requesterID s
...
 
+// Leave removes user from the run's participants list
+func (s *PlaybookRunServiceImpl) Leave(playbookRunID, requesterID string) error {
+	playbookRun, err := s.store.GetPlaybookRun(playbookRunID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve playbook run")
+	}
+
+	// Check if user is not a member of the channel
+	member, _ := s.pluginAPI.Channel.GetMember(playbookRun.ChannelID, requesterID)
```
trilopin, line 2756: should the owner be able to leave the run? it's a weird result. isn't it?

webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
```diff
@@ -80,22 +82,81 @@ export const ContextMenu = ({playbookRun, role}: Props) => {
...
+                            <Separator/>
+                            <DropdownMenuItem
+                                onClick={onFinishRun}
+                            >
+                                <FormattedMessage defaultMessage='Finish run'/>
+                            </DropdownMenuItem>
+                        </>
+                }
+                {
+                    role === Role.Participant &&
```
trilopin, line 95: I would also prevent the owner from leaving the run

trilopin:
I'm not 100% sure, but is there any scenario where you don't have permission to see the run because you left? In that case, we should additionally redirect to a safe place.

gbochora:
> I'm not 100% sure, but is there any scenario where you don't have permission to see the run because you left? In that case, we should additionally redirect to a safe place.

Very good point. I can think of only one scenario. If we add a user directly to the private channel, they will become a participant. For such a user, leaving the run means losing access to it.
I've added a redirect to the plugin page for such situations.

crspeller (approved):

server/app/playbook_run_service.go
gbochora: Added check, if the requester user is the owner, an error is returned.

justinegeffen (commented):
Thank you! Some minor suggestions inline. Happy to re-review when needed. :)

server/api/playbook_runs.go
```diff
@@ -76,6 +76,7 @@ func NewPlaybookRunHandler(
 	playbookRunRouter.HandleFunc("/status-updates", handler.getStatusUpdates).Methods(http.MethodGet)
 	playbookRunRouter.HandleFunc("/request-update", handler.requestUpdate).Methods(http.MethodPost)
 	playbookRunRouter.HandleFunc("/request-get-involved", handler.requestGetInvolved).Methods(http.MethodPost)
+	playbookRunRouter.HandleFunc("/leave", handler.leave).Methods(http.MethodDelete)
```
trilopin, line 79: I'd use the delete method if the resource to be deleted is reflected on the URL, or the post method if we use the `/leave` route. But I wouldn't use a mix of both.

webapp/src/client.ts
```diff
@@ -752,6 +752,14 @@ export const requestGetInvolved = async (playbookRunId: string) => {
     }
 };
 
+export const leaveRun = async (playbookRunId: string) => {
+    try {
+        return await doDelete(`${apiUrl}/runs/${playbookRunId}/leave`);
```
trilopin, line 757: same comment, the way I read this is "The user wants to delete the leave"

webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
```diff
@@ -80,22 +84,89 @@ export const ContextMenu = ({playbookRun, role}: Props) => {
...
+const useLeaveRun = (playbookRun: PlaybookRun) => {
+    const {formatMessage} = useIntl();
+    const currentUserId = useSelector(getCurrentUserId);
+    const addToast = useToaster().add;
+    const [showLeaveRunConfirm, setLeaveRunConfirm] = useState(false);
+
+    const onLeaveRun = async () => {
+        const response = await leaveRun(playbookRun.id);
+        if (response?.error) {
+            addToast(formatMessage({defaultMessage: "It wasn't possible to leave the run."}), ToastType.Failure);
```
trilopin, line 125: this is not only related to your PR, but I had a bad time reading the toasts. Don't you think they disappear too fast?

trilopin (commented):
Cool @gbochora!

added some comments about the HTTP method used for the endpoint.

webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
gbochora: I Agree, it's too fast for me too. I'll add 1 sec, it should make a difference.

trilopin (approved):
🎉

PR #1354, opened 2022-07-21
extract from redux event filters

trilopin:
#### Summary
The main goal of this task is to be coherent when navigating from recent activity to RHS timeline. We were navigating from partial to complete information, having actually less events (due to filters)

To fix that, several actions have been addressed:
- extract feature from redux, the main component handles the filter state. if you reload the page, you lose the state and have the default.
- when navigating from recent activity "all" filter is automatically enabled 
- tasks events are visible by default
- if you change run in LHS, filters are reset

2/5 on the behavior. With this change, we keep the "all" filter after navigating through recent activity. We could, in addition to that, revert to previous filters if wanted (with some internal double state).

Note that a big part of this PR is fixing the code that is not being used anymore. Don't be scared about the removed code in the playbook_runs_backstage folder. I had to migrate it to be compliant with the redux removal.

CleanShot 2022-07-21 at 15 56 29

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45896

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/actions.ts
+++ b/webapp/src/actions.ts
@@ -12,7 +12,7 @@ import {getCurrentChannelId} from 'mattermost-webapp/packages/mattermost-redux/s
 
 import {PlaybookRun} from 'src/types/playbook_run';
 import {selectToggleRHS, canIPostUpdateForRun} from 'src/selectors';
-import {RHSState, TimelineEventsFilter} from 'src/types/rhs';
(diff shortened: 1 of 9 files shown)


crspeller (approved):
Looks good!

gbochora (approved):
Nice!

PR #1355, opened 2022-07-21
[MM-45729] Update playbook editor header UI

iomodo:
#### Summary
This PR updates playbook editor header UI

https://user-images.githubusercontent.com/2340127/180268465-66a77c83-a3e0-4871-8f50-077f8343f15e.mp4

Bit confused about the `globe` icon in the [Figma](https://www.figma.com/file/f02HsYE1ftPoB44iWYoyIH/⬅%EF%B8%8F-LHS-Navigation?node-id=513%3A168245). Not sure what should happen when playbook is private. Another thing is that it looks kinda awkward next to the star, Star is clickable and I assume globe is clickable too and it does nothing... So I'm removing it for now. Let me know what you think

#### Ticket Link
  Fixes: https://mattermost.atlassian.net/browse/MM-45729

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
+++ b/webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
@@ -2,12 +2,11 @@
 // See LICENSE.txt for license information.
 
 import styled, {css} from 'styled-components';
-import React, {useRef, useEffect} from 'react';
+import React, {useEffect} from 'react';
 import {Switch, Route, Redirect, NavLink, useRouteMatch} from 'react-router-dom';
 
 import {useIntl} from 'react-intl';
 
-import {useIntersection} from 'react-use';
 import {selectTeam} from 'mattermost-redux/actions/teams';
 import {fetchMyChannelsAndMembers} from 'mattermost-redux/actions/channels';
 import {fetchMyCategories} from 'mattermost-redux/actions/channel_categories';
@@ -59,10 +58,6 @@ const PlaybookEditor = () => {
     const stats = useStats(playbookId);
 
     useForceDocumentTitle(playbook?.title ? (playbook.title + ' - Playbooks') : 'Playbooks');
-
-    const headingRef = useRef<HTMLDivElement>(null);
-    const headingIntersection = useIntersection(headingRef, {threshold: 1});
-    const headingVisible = headingIntersection?.isIntersecting ?? true;
 
     useEffect(() => {
         const teamId = playbook?.team_id;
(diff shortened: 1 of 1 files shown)


webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
```diff
@@ -174,48 +169,15 @@ const PlaybookEditor = () => {
...
-                                <span data-testid={'playbook-editor-title'}>
-                                    {playbook.title}
-                                </span>
-                            </Controls.TitleMenu>
-                            {privateTooltip}
-                            {archivedTooltip}
-                        </Heading>
-                    )}
-                </TextEdit>
+            <Header>
```
crspeller, line 172: I am a bit confused. It appears we have removed the ability to edit the header here?

crspeller (changes requested):
Also noticed that the new version looks a little squished and removes the fancy scroll behavior. Is that intentional?

webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
iomodo: Yes, it's the new designs [here](https://www.figma.com/file/f02HsYE1ftPoB44iWYoyIH/⬅%EF%B8%8F-LHS-Navigation?node-id=422%3A150446). You can still edit it from the top left corner

calebroseland (approved):
Approving to not block, but I agree removing the ability to rename the playbook name/title in the larger one is a bit of a step back IMHO. 

If we're removing the header-title menu, I'd like to maybe see a pencil icon next to it?
CleanShot 2022-07-22 at 16 37 20@2x
CleanShot 2022-07-22 at 16 40 01@2x

calebroseland (commented):
Also: 

Definitely in the scope of polish, but at some point this should probably be made more square.

CleanShot 2022-07-22 at 16 44 35@2x

calebroseland (commented):
> Bit confused about the globe icon in the [Figma](https://www.figma.com/file/f02HsYE1ftPoB44iWYoyIH/%E2%AC%85%EF%B8%8F-LHS-Navigation?node-id=513%3A168245). Not sure what should happen when playbook is private. Another thing is that it looks kinda awkward next to the star, Star is clickable and I assume globe is clickable too and it does nothing... So I'm removing it for now. Let me know what you think



My $0.02: Channels relies on the LHS globe icon. Perhaps we could add a globe to the playbook icon to make it more obvious without adding extra UI in the playbook editor title area?


CleanShot 2022-07-22 at 16 49 47@2x
CleanShot 2022-07-22 at 16 50 15@2x

PR #1357, opened 2022-07-22
MM-44756: playbook join/leave, list view and editor view

calebroseland:
#### Summary
CleanShot 2022-07-22 at 04 46 28

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44756

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_editor/controls.tsx
```diff
@@ -288,6 +302,30 @@ export const RunPlaybook = ({playbook}: ControlProps) => {
...
+export const JoinPlaybook = ({playbook: {id: playbookId}, refetch}: ControlProps & {refetch: () => void;}) => {
+    const {formatMessage} = useIntl();
+    const currentUser = useSelector(getCurrentUser);
+    const {join} = usePlaybookMembership(playbookId, currentUser.id);
+    const {setFollowing} = useEditorFollowersMeta(playbookId);
+
+    return (
+        <PrimaryButtonLarger
+            onClick={async () => {
+                join();
```
gbochora, line 314: Do we need to wait for the `join` to complete?
```suggestion
                await join();
```

webapp/src/components/backstage/playbook_editor/controls.tsx
```diff
@@ -346,30 +402,65 @@ const TitleMenuImpl = ({playbook, children, className, editTitle, refetch}: Titl
...
-                        <RedText>
-                            <FormattedMessage defaultMessage='Archive playbook'/>
-                        </RedText>
-                    </DropdownMenuItem>
+                {currentUserMember && (
+                    <>
+                        <div className='MenuGroup menu-divider'/>
+                        <DropdownMenuItem
+                            onClick={() => {
+                                leave();
```
gbochora, line 429: Same is here:
```suggestion
                                await leave();
```

gbochora (approved):
Great work! I left a few comments, but nothing blocking. 
Also, there is an issue with leaving an archived Playbook, it fails silently.

https://user-images.githubusercontent.com/4368372/182576081-1b0fc8fe-0aa4-4b7f-ab9c-a5baa9eb8fbe.mp4

webapp/src/graphql/generated_types.ts
```diff
@@ -66,7 +66,9 @@ export {MetricType};
 export type Mutation = {
     __typename?: 'Mutation';
     addMetric: Scalars['String'];
+    addPlaybookMember: Scalars['String'];
```
lieut-data, line 69: Based on the file name / content, I'm assuming this is generated -- should track this in a .gitattributes as per https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github?

server/api/graphql_root.go
```diff
@@ -382,6 +382,66 @@ func (r *RootResolver) UpdateRun(ctx context.Context, args struct {
...
 
+func (r *RootResolver) AddPlaybookMember(ctx context.Context, args struct {
+	PlaybookID string
+	UserID     string
+}) (string, error) {
+	c, err := getContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	userID := c.r.Header.Get("Mattermost-User-ID")
```
lieut-data, line 393: nit: given we unwrap via `getContext`, I wonder if we can add a helper method to get the user id vs. having to duplicate this string everywhere and know to look on the request header.

server/api/graphql_root.go
```diff
@@ -382,6 +382,66 @@ func (r *RootResolver) UpdateRun(ctx context.Context, args struct {
...
+	currentPlaybook, err := c.playbookService.Get(args.PlaybookID)
+	if err != nil {
+		return "", err
+	}
+
+	if currentPlaybook.DeleteAt != 0 {
+		return "", errors.New("archived playbooks can not be modified")
+	}
+
+	if err := c.permissions.PlaybookManageMembers(userID, currentPlaybook); err != nil {
```
lieut-data, line 404: nit: Should we check permissions before any other constraints?

server/api/graphql_root.go
```diff
@@ -382,6 +382,66 @@ func (r *RootResolver) UpdateRun(ctx context.Context, args struct {
...
+	}
+	userID := c.r.Header.Get("Mattermost-User-ID")
+
+	currentPlaybook, err := c.playbookService.Get(args.PlaybookID)
+	if err != nil {
+		return "", err
+	}
+
+	if currentPlaybook.DeleteAt != 0 {
+		return "", errors.New("archived playbooks can not be modified")
```
lieut-data, line 431: Will this mean I can't leave an archived playbook?

lieut-data (approved):
A handful of comments, but nothing blocking I think.

webapp/src/graphql/generated_types.ts
crspeller: It is generated bu I actually would like to see it in diffs by default.

server/api/graphql_root.go
crspeller: That makes sense. Will do in a separate PR.

server/api/graphql_root.go
crspeller: I fell like expensive checks should go last, but 0/5.

server/api/graphql_root.go
crspeller: Correct. This is the existing behavior.

crspeller:
I believe cypress test failures are unrelated: https://community.mattermost.com/core/pl/t1zkk7h77bnddk476r3gaacfpo

PR #1360, opened 2022-07-22
MM-45933: Playbook List Columns: Last used, Active runs, Runs in 30 days

calebroseland:
CleanShot 2022-08-29 at 11 40 29@2x

#### Summary

- [x] New column: `Last used`
- [x] New column: `Active runs`
~~- [ ] New column: `Runs in 30 days` (TODO)~~

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-45933

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


iomodo:
Added the active runs column.
regarding the Runs in 30 days not quite sure what that means. Runs that ended 30 days ago or less + active runs? cc @stephenvanhemmen. Or runs that started 30 days ago?

stephenvanhemmen:
> Added the active runs column. regarding the Runs in 30 days not quite sure what that means. Runs that ended 30 days ago or less + active runs? cc @stephenvanhemmen. Or runs that started 30 days ago?

I'm not sure, I'd want to leave this question to @abhijit-singh. For this release these changes aren't critical, the only thing that is absolutely necessary is the join/run buttons. Can the categories (checklists, tasks, runs) be kept the same for this release? I think we should focus on a list upgrade (giving dashboard functionality) as a different feature.

server/sqlstore/playbook.go
```diff
@@ -60,7 +61,9 @@ func applyPlaybookFilterOptionsSort(builder sq.SelectBuilder, options app.Playbo
 		sort = "NumRuns"
 	case app.SortByCreateAt:
 		sort = "CreateAt"
-	case "":
+	case app.SortByLastRunAt:
+		sort = "LastRunAt"
+	case "", app.SortByActiveRuns:
```
lieut-data, line 66: This doesn't seem right -- should we be setting `sort` to something else when we match on `app.SortByActiveRuns`?

server/sqlstore/playbook.go
```diff
@@ -494,6 +535,21 @@ func (p *playbookStore) GetPlaybooksForTeam(requesterInfo app.RequesterInfo, tea
 	addMembersToPlaybooks(members, playbooks)
 	addMetricsToPlaybooks(metrics, playbooks)
 
+	if err := p.populateActiveRuns(playbooks); err != nil {
+		return app.GetPlaybooksResults{}, err
+	}
+	if opts.Sort == app.SortByActiveRuns {
```
lieut-data, line 541: If I'm understanding this correctly, this won't work past page 1 of the results since the server will always return the same set for the first page?

lieut-data (changes requested):

lieut-data (approved):
Sweet!

crspeller:
Need rebasing after https://github.com/mattermost/mattermost-plugin-playbooks/pull/1357

calebroseland:
@crspeller this is ready now

crspeller (approved):

PR #1363, opened 2022-07-25
e2e: Update Cypress

Phrynobatrachus:
#### Summary
Updates Cypress and RTL dependencies. Also contains a couple changes for tests that continue to be flaky. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41527


tests-e2e/cypress.config.js
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+const {defineConfig} = require('cypress');
+
+const cypressPlugins = require('./cypress/plugins/index');
+
+module.exports = defineConfig({
```
trilopin, line 7: has any config changed or just taken the new filename convention?

tests-e2e/package.json
```diff
@@ -1,11 +1,11 @@
...
     "@cypress/skip-test": "^2.6.1",
-    "@testing-library/cypress": "7.0.6",
+    "@testing-library/cypress": "^8.0.3",
     "authenticator": "^1.1.5",
     "axios": "^0.21.1",
     "axios-retry": "^3.1.9",
-    "cypress": "7.7.0",
+    "cypress": "^10.3.0",
     "cypress-file-upload": "5.0.8",
     "cypress-multi-reporters": "^1.5.0",
```
trilopin, line 10: I'm surprised that cypress plugins are kept in old versions.

trilopin (commented):
🎉 This is great @Phrynobatrachus!

I understood that all specs and the folder would need to be renamed to follow the new convention. Why is this not needed in your PR?

tests-e2e/cypress.config.js
Phrynobatrachus: Actually that's not quite right, some of the config options moved to an `e2e` property. No functional changes though.

Phrynobatrachus:
> I understood that all specs and the folder would need to be renamed to follow the new convention. Why is this not needed in your PR?

My read of the [migration guide](https://docs.cypress.io/guides/references/migration-guide#testFiles-specPattern) is that while there's new defaults, there's no hard requirement to follow a particular structure. All that matters is that the config `specPattern` globs the tests you want.

tests-e2e/package.json
Phrynobatrachus: All are on the latest major versions. I will update them though, as I did mean to remove `cypress-file-upload`.

trilopin (approved):
Thanks @Phrynobatrachus for taking the time for this.

it's really cool!

lieut-data (approved):

PR #1364, opened 2022-07-25
MM-45575 GraphQL updates for LHS list

crspeller:
#### Summary
Move playbooks LHS listing to GraphQL to allow for basic automatic updates of the LHS.
Note that this is the first iteration of this and only works locally. 

#### Demos
Favorites:

https://user-images.githubusercontent.com/3191642/180880689-5504327f-ef50-40a2-9d6c-a83bce7e57fe.mp4

New Playbooks:

https://user-images.githubusercontent.com/3191642/180880781-b1a69d34-d3a5-4f16-b8b6-6e6ad9bb4784.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45575

### DIFF
--- a/server/api/graphql_playbook.go
+++ b/server/api/graphql_playbook.go
@@ -1,12 +1,36 @@
 package api
 
 import (
+	"context"
+
 	"github.com/mattermost/mattermost-plugin-playbooks/server/app"
+	"github.com/pkg/errors"
 )
 
 type PlaybookResolver struct {
 	app.Playbook
-	IsFavorite bool
+}
+
+func (r *PlaybookResolver) IsFavorite(ctx context.Context) (bool, error) {
+	c, err := getContext(ctx)
+	if err != nil {
+		return false, err
+	}
+	userID := c.r.Header.Get("Mattermost-User-ID")
+
+	isFavorite, err := c.categoryService.IsItemFavorite(
+		app.CategoryItem{
+			ItemID: r.ID,
+			Type:   app.PlaybookItemType,
+		},
+		r.TeamID,
+		userID,
+	)
+	if err != nil {
+		return false, errors.Wrap(err, "can't determine if item is favorite or not")
+	}
+
+	return isFavorite, nil
 }
 
 func (r *PlaybookResolver) DeleteAt() float64 {
--- a/server/api/graphql_root.go
+++ b/server/api/graphql_root.go
@@ -32,19 +32,56 @@ func (r *RootResolver) Playbook(ctx context.Context, args struct {
 	if err != nil {
 		return nil, err
 	}
-	isFavorite, err := c.categoryService.IsItemFavorite(
-		app.CategoryItem{
-			ItemID: playbookID,
-			Type:   app.PlaybookItemType,
-		},
-		playbook.TeamID,
-		userID,
-	)
+
+	return &PlaybookResolver{playbook}, nil
+}
+
+func (r *RootResolver) Playbooks(ctx context.Context, args struct {
+	TeamID       string
+	Sort         string
+	Direction    string
+	SearchTerm   string
+	WithArchived bool
+}) ([]*PlaybookResolver, error) {
+	c, err := getContext(ctx)
 	if err != nil {
-		return nil, errors.Wrap(err, "can't determine if item is favorite or not")
+		return nil, err
+	}
+	userID := c.r.Header.Get("Mattermost-User-ID")
+
+	if args.TeamID != "" {
+		if err := c.permissions.PlaybookList(userID, args.TeamID); err != nil {
+			c.log.Warnf("public error message: %v; internal details: %v", "Not authorized", err)
+			return nil, errors.New("Not authorized")
+		}
 	}
 
-	return &PlaybookResolver{playbook, isFavorite}, nil
+	requesterInfo := app.RequesterInfo{
+		UserID:  userID,
+		TeamID:  args.TeamID,
+		IsAdmin: app.IsSystemAdmin(userID, c.pluginAPI),
+	}
+
+	opts := app.PlaybookFilterOptions{
+		Sort:         app.SortField(args.Sort),
+		Direction:    app.SortDirection(args.Direction),
+		SearchTerm:   args.SearchTerm,
+		WithArchived: args.WithArchived,
+		Page:         0,
+		PerPage:      10000,
+	}
+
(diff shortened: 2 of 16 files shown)


webapp/src/graphql_client.ts
```diff
@@ -0,0 +1,13 @@
+import {ApolloClient, NormalizedCacheObject} from '@apollo/client';
+
+// Global for use in gloablly mounted modals that won't work with react context.
+// Don't use this. If you need the client itself use `useApolloClient`.
+let playbooksGraphQLClient: ApolloClient<NormalizedCacheObject>;
```
trilopin, line 5: nice warning in the comment

server/api/graphql_root.go
```diff
@@ -250,6 +332,56 @@ func (r *RootResolver) UpdatePlaybook(ctx context.Context, args struct {
...
+	playbookRun, err := c.playbookRunService.GetPlaybookRun(args.ID)
+	if err != nil {
+		return "", err
+	}
+
+	if err := c.permissions.RunManageProperties(userID, playbookRun.ID); err != nil {
+		return "", err
+	}
+
+	if args.Updates.IsFavorite != nil {
```
iomodo, line 356: Maybe less nesting for readability like:
```go
if args.Updates.IsFavorite == nil {
    return playbookRun.ID, nil
}
```

server/api/graphql_root.go
```diff
@@ -250,6 +332,56 @@ func (r *RootResolver) UpdatePlaybook(ctx context.Context, args struct {
...
+		if *args.Updates.IsFavorite {
+			if err := c.categoryService.AddFavorite(
```
iomodo, line 358: It's kinda strange that get and update for `IsFavorite` are in different files? Is there a way to consolidate somehow?

webapp/src/components/sidebar/playbooks_sidebar.tsx
```diff
@@ -20,51 +22,73 @@ export const PlaybooksCategoryName = 'playbooksCategory';
...
+            id: pb.id,
+            icon,
+            link,
+            isCollapsed: false,
+            itemMenu: null,
+            isFavorite: pb.isFavorite,
+            className: '',
+        };
+    });
+    const playbookFavorites = playbookItems.filter((group) => group.isFavorite);
```
iomodo, line 61: In previous implementation server returned the categories and frontend just displayed them. Current implementation server returns playbooks and runs and frontend has to figure out what's where. 
Would not it be better for the server to stick with returning categories in GraphQL rather than playbooks and runs?

iomodo (commented):
Looks really great @crspeller. 
My only concern is GraphQL returning categories vs playbooks and runs. What do you think?

server/api/graphql_playbook.go
```diff
@@ -1,12 +1,36 @@
...
+}
+
+func (r *PlaybookResolver) IsFavorite(ctx context.Context) (bool, error) {
+	c, err := getContext(ctx)
+	if err != nil {
+		return false, err
+	}
+	userID := c.r.Header.Get("Mattermost-User-ID")
+
+	isFavorite, err := c.categoryService.IsItemFavorite(
```
trilopin, line 21: question: when this is checked for a list, do we still run a query for each item?

trilopin: btw not asking for optimization, just wanted to know how it works

webapp/src/graphql/playbook.graphql
```diff
@@ -64,3 +64,17 @@ query Playbook($id: String!) {
 mutation UpdatePlaybook($id: String!, $updates: PlaybookUpdates!) {
   updatePlaybook(id: $id, updates: $updates)
 }
+
+query PlaybookLHS($userID: String!, $teamID: String!) {
```
trilopin, line 68: aside from the latency(1 call less), what's the benefit of doing a lhsquery instead of using runs/playbooks queries?

webapp/src/components/sidebar/playbooks_sidebar.tsx
trilopin: I don't have a strong opinion on whether the backend or the frontend should handle the categorization. 

But if we go ahead with the frontend, What do you think about extracting the logic to manipulate all the returned data in a separate function. Would that improve a bit the readability? 

Could be a local hook like useLHSData which contains usePlaybookLhsQuery.

server/api/graphql_root.go
crspeller: Except that in the future it's going to look like UpdatePlaybook above where there are more values below. This is more obviously expandable.

server/api/graphql_root.go
crspeller: Really the get and update are in the same file since the resolvers for Runs and UpdateRun are here.  It's just that the run resolver is in a separate file. We could of course cram them both into the same file, but I am not sure why we would do that. 

I do think there is a better way to organize this. Not sure how but let's think of that as a separate PR.

webapp/src/components/sidebar/playbooks_sidebar.tsx
crspeller: I do think that we will eventually need an endpoint for categories when we implement custom categories. But until then we can get away with this. To take advantage of the graphql client cache we need to return the objects with matching IDs so this was the easiest way to do it. 

Agree with the local hook will do.

server/api/graphql_playbook.go
crspeller: Yes it will! A lovely pitfall of GraphQL that needs to be fixed using the dataloader pattern which is not implemented yet. I believe we can get away with this for now until we introduce data loaders that the server implmented: https://github.com/mattermost/mattermost-server/blob/master/api4/graphql.go#L110

webapp/src/graphql/playbook.graphql
crspeller: You can retrieve exactly the data you need. If you need to re-fetch (which we do in this case) you can do it independently of other queries. Its more clear what the component depends on.

iomodo (approved):

webapp/src/graphql/playbook.graphql
trilopin: got it, more accurate calls (with only the data7refetch you need) that internally reuse the same core resources (run, pb). I have to get used to this paradigm

trilopin (approved):
This review has been more useful for me than for you! 

Please keep throwing graphql prs to my plate 🙏🏽.

PR #1365, opened 2022-07-26
Add hiding due date if task is done

gbochora:
#### Summary
This PR hides due date info for finished/skipped tasks on the RDP.

https://user-images.githubusercontent.com/4368372/180964956-57ba5896-8852-4310-92c8-1cb3b9f7c3f4.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45961

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/checklist_item/checklist_item.tsx
+++ b/webapp/src/components/checklist_item/checklist_item.tsx
@@ -200,7 +200,12 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement => 
 
     const renderDueDate = (): null | React.ReactNode => {
         const isTaskOpenOrInProgress = props.checklistItem.state === ChecklistItemState.Open || props.checklistItem.state === ChecklistItemState.InProgress;
-        if (buttonsFormat !== ButtonsFormat.Long && (!dueDate || !isTaskOpenOrInProgress) && !isEditing) {
+
+        // if task is done hide due date info
+        if (!isTaskOpenOrInProgress) {
+            return null;
+        }
+        if (buttonsFormat !== ButtonsFormat.Long && (!dueDate && !isEditing)) {
             return null;
         }
 


webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -200,7 +200,12 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
 
     const renderDueDate = (): null | React.ReactNode => {
         const isTaskOpenOrInProgress = props.checklistItem.state === ChecklistItemState.Open || props.checklistItem.state === ChecklistItemState.InProgress;
-        if (buttonsFormat !== ButtonsFormat.Long && (!dueDate || !isTaskOpenOrInProgress) && !isEditing) {
+
+        // if task is done hide due date info
+        if (!isTaskOpenOrInProgress) {
```
trilopin, line 205: shouldn't it be shown if you are editing?

gbochora: I think there should not be a need to modify the due date of the finished/skipped task. We have similar behavior in the Channel-RHS.

trilopin (approved):
Thanks!

iomodo (approved):

webapp/src/components/checklist_item/checklist_item.tsx
lieut-data: The one nuance I see with the run details is that we currently show all the widets all the time, even if empty. I wonder if instead of hiding, we might consider just removing the "late" warning and rendering it neutrally?

(I do see a case for adjusting the due date as a "matter of record", just like we can change the assignee retroactively as well.)

gbochora: Removing the "late" warning is a good idea.

PR #1369, opened 2022-07-27
Add retrospective broadcast

gbochora:
#### Summary
This PR adds retrospective broadcast to the channels. When the run's actions are configured to broadcast status updates to channels, we broadcast retrospective to the same thread.

Updated the `Run actions` dialog copies, needs review.
Screenshot 2022-07-27 at 16 31 02

https://user-images.githubusercontent.com/4368372/181244657-d7148c57-fdee-40de-baf7-5985abf9cd16.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45175

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated


server/app/playbook_run_service.go
```diff
@@ -2568,6 +2568,19 @@ func (s *PlaybookRunServiceImpl) PublishRetrospective(playbookRunID, publisherID
 		return errors.Wrap(err, "failed to post to channel")
 	}
 
+	if playbookRunToPublish.StatusUpdateBroadcastChannelsEnabled {
+		T := i18n.GetUserTranslations(publisherUser.Locale)
+		data := map[string]interface{}{
+			"Name": publisherUser.Username,
+			"URL":  retrospectiveURL + "#playbook-run-retrospective",
```
trilopin, line 2575: Let's replace `getRunRetrospectiveURL`method since it's still linking the old retrospective tab. That could be an excellent place to include the hashtag.

What do you think?

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
```diff
@@ -210,7 +212,7 @@ describe('runs > run details page > header', () => {
                             openRunActionsModal();
 
                             // * Verify that the broadcast-to-channels toggle is checked
-                            cy.findByText('Broadcast update to selected channels').parent().within(() => {
+                            cy.findByText(CHANNEL_BROADCAST_TEXT).parent().within(() => {
```
trilopin, line 215: Optional suggestions: personal preferences

```ts
const getBroadcastSection = () => cy.FindByText('Broadcast to selected channels');
...
getBroadcastSection().click();
getBroadcastSection().parent().get('input').should('be.checked');
```
1. You can define and reuse the command locally instead of using a constant
2. I usually chain commands directly when within only has one call


thoughts?

tests-e2e/cypress/support/plugin/api_commands.js
```diff
@@ -133,6 +133,25 @@ Cypress.Commands.add('apiUpdateStatus', (
...
+        url: `${playbookRunsEndpoint}/${playbookRunId}/retrospective/publish`,
+        method: 'POST',
+        body: {
+            retrospective,
+        },
+    }).then((response) => {
+        expect(response.status).to.equal(StatusOK);
+        cy.wrap(response.body);
+    });
+});
```
trilopin, line 153: Thanks, very useful

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
```diff
@@ -331,6 +333,87 @@ describe('runs > run details page > header', () => {
...
+                            playbookRunId: playbookRun.id,
+                            retrospective,
+                        });
+
+                        // # Navigate to the town square channel
+                        cy.visit(`/${testTeam.name}/channels/town-square`);
+
+                        // * Verify that the last post contains the retrospective post
+                        cy.getLastPost().then((post) => {
+                            cy.get(post).contains(retrospective);
```
trilopin, line 364: [question/nofeedback]: what is injected into that promise? 

Is this function which is called?

```ts
Cypress.Commands.add('getLastPost', () => {
    waitUntilPermanentPost();

    cy.findAllByTestId('postView').last();
});
```

trilopin (changes requested):
I left some comments, suggestions and questions.

The only blocker is the URL thing that I feel we must fix.

server/app/playbook_run_service.go
gbochora: I agree 💯.

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
gbochora: Again, an excellent suggestion 👍

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
gbochora: It's unclear to me too. We use `getLastPost` in many places, and I did the same.

trilopin (approved):
Cool, thanks @gbochora !

justinegeffen (approved):

justinegeffen:
Thanks, @gbochora, this looks good to me! :)

crspeller (approved):

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
Phrynobatrachus: This is something I've wondered about before as well, the official docs don't explain the semantics very well IMO. This post by Gleb (one of the maintainers) notes:

https://glebbahmutov.com/blog/writing-custom-cypress-command/#simple-custom-command

`Notice that we did not have to return the command chain from the custom command, this is done automatically for every command.`

PR #1370, opened 2022-07-27
[MM-45741] Add following button to runs list

iomodo:
#### Summary
This PR adds following button to runs list. It uses already existed `useRunMetadata` hook. To avoid creating additional api endpoint or hook. It will change when we move to GraphQL for runs.

https://user-images.githubusercontent.com/2340127/181278512-722add60-f9ba-4bf1-9a3a-66437db99686.mp4

#### Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-45741

#### Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


webapp/src/components/backstage/runs_list/row.tsx
```diff
@@ -184,3 +165,84 @@ const formatDate = (millis: number) => {
...
+     display: flex;
+     align-items: center;
+     gap: 8px;
+     height: 32px;
+     padding: 0 16px;
+`;
+
+const FollowButton = styled(SecondaryButton)`
+    border: 1px solid var(--center-channel-color-08);
+    color: var(--center-channel-color-64);
```
trilopin, line 243: buttons are completely different here and on the run details page (run info overview). is that intended?

webapp/src/components/backstage/runs_list/row.tsx
```diff
@@ -184,3 +165,84 @@ const formatDate = (millis: number) => {
 };
 
 export default Row;
+
+const FollowPlaybookRun = ({id}: {id: string}) => {
+    const {formatMessage} = useIntl();
+    const currentUser = useSelector(getCurrentUser);
+    const [metadata] = useRunMetadata(id);
+    const [followers, setFollowers] = useState(metadata?.followers || []);
+    const [isFollowing, setIsFollowing] = useState(followers.includes(currentUser.id));
```
trilopin, line 174:
```suggestion
    const isFollowing = followers.includes(currentUser.id);
```

I may be wrong, but this seems to be a derived property that don't need to be state

webapp/src/components/backstage/runs_list/row.tsx
```diff
@@ -184,3 +165,84 @@ const formatDate = (millis: number) => {
...
+    const addToast = useToaster().add;
+
+    useEffect(() => {
```
trilopin, line 177: we could use useUpdateEffect from react-use to make more clear that we only need it for future changes (not the initial one)

trilopin (changes requested):
Thanks @iomodo 👍🏽 

I left some questions and comments

webapp/src/components/backstage/runs_list/row.tsx
```diff
@@ -143,34 +146,12 @@ const Row = (props: Props) => {
                 </SmallText>
             </div>
             <div className='col-sm-2'>
-                {/* eslint-disable-next-line formatjs/no-literal-string-in-jsx */}
-                <NormalText>{completedTasks + ' / ' + totalTasks}</NormalText>
-                <ProgressBar
-                    completed={completedTasks}
-                    total={totalTasks}
-                />
```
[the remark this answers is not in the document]
iomodo: These are no longer in the design

trilopin (approved):
Thanks @iomodo

crspeller (approved):

PR #1372, opened 2022-07-28
leave & unfollow run

trilopin:
#### Summary
Leave and unfollow run

#### Ticket Link
Not yet

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
@@ -10,7 +10,7 @@ import {getCurrentUserId} from 'mattermost-webapp/packages/mattermost-redux/src/
 import {StarIcon, StarOutlineIcon, LightningBoltOutlineIcon, LinkVariantIcon, ArrowDownIcon, FlagOutlineIcon, CloseIcon} from '@mattermost/compass-icons/components';
 
 import {showRunActionsModal} from 'src/actions';
-import {exportChannelUrl, getSiteUrl, leaveRun} from 'src/client';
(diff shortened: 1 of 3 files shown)


webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
```diff
@@ -140,18 +140,20 @@ const useLeaveRun = (playbookRun: PlaybookRun) => {
         if (response?.error) {
             addToast(formatMessage({defaultMessage: "It wasn't possible to leave the run."}), ToastType.Failure);
         } else {
-            addToast(formatMessage({defaultMessage: "You've left the run."}), ToastType.Success);
-            if (!response.has_view_permission) {
-                navigateToUrl(pluginUrl(''));
-            }
+            unfollowPlaybookRun(playbookRun.id).then(() => {
```
gbochora, line 143: From my understanding, we changed the meaning of `Leave run`, which now means leave + unfollow. Would it be better to add the `unfollow` part to the  `PlaybookRun.Leave` function in the backend?

trilopin: Good point @gbochora , even if later we make this conditional it could be a parameter of leave run. I'm going to move it to server and the webapp code will be simplified.

trilopin: updated, moved to back, and added a test

gbochora (approved):
Great! Thanks @trilopin

iomodo (approved):
Looks great! Thanks @trilopin

PR #1374, opened 2022-08-01
share follower state between rhs and main + leave-run copies accuracy

trilopin:
#### Summary
Make leave-run aware of the current follower state to set the copies to leave-run or leave-and-unfollow. 

Follower state is not handled at redux, it had to be moved to parent component so we can share it between RHS-Info (where the follow/unfollow button is) and header/context-menu

Telemetry for page view moved from RHS to actual page (no effect in metrics but I wasn't the right palce).

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-46085

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
@@ -35,7 +35,6 @@ import RHSTimeline from './rhs_timeline';
 const RHSRunInfoTitle = <FormattedMessage defaultMessage={'Run info'}/>;
 
 const useRHS = (playbookRun?: PlaybookRun|null) => {
-    usePlaybookRunViewTelemetry(PlaybookRunViewTarget.Details, playbookRun?.id);
     const [isOpen, setIsOpen] = useState(true);
     const [scrollable, setScrollable] = useState(true);
     const [section, setSection] = useState<RHSContent>(RHSContent.RunInfo);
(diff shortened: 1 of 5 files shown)


webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -62,6 +61,19 @@ const useRHS = (playbookRun?: PlaybookRun|null) => {
...
+    const currentUser = useSelector(getCurrentUser);
+    const [followers, setFollowers] = useState(metadataFollowers);
+    const [isFollowing, setIsFollowing] = useState(followers.includes(currentUser.id));
+
+    useEffect(() => {
+        setFollowers(metadataFollowers);
+        setIsFollowing(metadataFollowers.includes(currentUser.id));
+    }, [currentUser.id, JSON.stringify(metadataFollowers)]);
+
+    return {isFollowing, followers, setIsFollowing, setFollowers};
```
gbochora, line 74: According to the current implementation, we should always call both setters(`setFollowers` and `setIsFollowing`). Otherwise, one of the states(`isFollowing` or `followers`) might be incorrect.
To avoid this, could we return only `setFollowers` and move `setIsFollowing` logic inside this hook?

trilopin: good suggestion! that makes completely sense 💯

gbochora (approved):
Looks great! Thanks @trilopin 🥳

crspeller (approved):

PR #1375, opened 2022-08-01
prevent request-update from execute when run is finished

trilopin:
#### Summary
Prevent participants from requesting updates once the run is finished

CleanShot 2022-08-01 at 11 42 47

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-46073

#### Checklist

- [ ] ~~Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/runs/rdp_main_statusupdate_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_main_statusupdate_spec.js
@@ -114,6 +114,22 @@ describe('runs > run details page > status update', () => {
         });
 
         describe('request an update', () => {
+            it('is disabled if the run is finished', () => {
+                cy.apiFinishRun(testRun.id).then(() => {
+                    // # reload url
+                    cy.visit(`/playbooks/runs/${testRun.id}`);
+
+                    // # Click on kebab menu
+                    cy.findByTestId('run-statusupdate-section').getStyledComponent('Kebab').click();
+
+                    // # Click on request update
+                    cy.findByText('Request update...').click();
+
+                    // # Click on modal confirmation
+                    cy.get('#confirmModalButton').should('not.exist');
+                });
+            });
+
             it('requests and confirm', () => {
                 // # Click on kebab menu
                 cy.findByTestId('run-statusupdate-section').getStyledComponent('Kebab').click();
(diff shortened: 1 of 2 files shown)


tests-e2e/cypress/integration/runs/rdp_main_statusupdate_spec.js
```diff
@@ -114,6 +114,22 @@ describe('runs > run details page > status update', () => {
...
+                    cy.visit(`/playbooks/runs/${testRun.id}`);
+
+                    // # Click on kebab menu
+                    cy.findByTestId('run-statusupdate-section').getStyledComponent('Kebab').click();
+
+                    // # Click on request update
+                    cy.findByText('Request update...').click();
+
+                    // # Click on modal confirmation
+                    cy.get('#confirmModalButton').should('not.exist');
```
gbochora, line 129: Nit: 
```suggestion
                    // * Verify there is no confirmation modal
```

gbochora (approved):
Nice!

webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
```diff
@@ -251,8 +251,8 @@ export const ParticipantStatusUpdate = ({id, playbookRun, openRHS}: ParticipantP
                                 {openRHSText}
                             </DropdownItem>
                             <DropdownItem
-                                onClick={showRequestUpdateConfirm}
-                                disabled={false}
+                                onClick={playbookRun.current_status === PlaybookRunStatus.Finished ? undefined : showRequestUpdateConfirm}
```
crspeller, line 254: Is this necessary? Isn't it rendered unclickable by the disabled property? Actually I guess they are `a` tags could we change  them to buttons or use `pointer-events: none;`?

crspeller (approved):
Looks good. One minor suggestion.

webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
trilopin: you made me doubt and double-check 😹 

Apparently, disabled has the effect that you described for buttons, inputs, and textarea. In this case, the underlying component is <a> which has no native disabled property.

The disabled prop is added in a StyledComponent
```ts
const DropdownItem = styled(DropdownMenuItemStyled)<{disabled: boolean}>`
    opacity: ${({disabled}) => (disabled ? '0.50' : '1')};
    cursor: ${({disabled}) => (disabled ? 'not-allowed' : 'pointer')};
`;
```

trilopin: Since this DotMenu component is widely used, I'll play and maybe change them in a follow-up PR.

PR #1376, opened 2022-08-01
Mm 46070 lhs not updated

gbochora:
#### Summary
This PR adds LHS refresh for some actions:
- Finish run
- Become a participant
- Leave a run
- Archive playbook
- Restore playbook
- Create playbook

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46070

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


tests-e2e/cypress/integration/playbooks/creation_button_spec.js
```diff
@@ -61,6 +61,11 @@ describe('playbooks > creation button', () => {
 
         // * Verify playbook outline page opened
         verifyPlaybookOutlineOpened(playbookName);
+
+        // * Verify playbook was added to the LHS
+        cy.findByTestId('lhs-navigation').within(() => {
+            cy.findByText(playbookName).should('exist');
+        });
```
trilopin, line 68: optional: would make sense to make this a one-liner?
```suggestion
        cy.findByTestId('lhs-navigation').findByText(playbookName).should('exist');
```

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
```diff
@@ -395,6 +395,31 @@ describe('runs > run details page > header', () => {
...
+                    cy.wait(500);
+
+                    // # Click on leave run
+                    getDropdownItemByText('Leave and unfollow run').click();
+
+                    // # confirm modal
+                    cy.get('#confirmModal').get('#confirmModalButton').click();
+
+                    // * Assert that the Participate button is shown
+                    getHeader().findByText('Participate').should('exist');
```
trilopin, line 415: would you want to additionally check that is visible in the screen?
```suggestion
                    getHeader().findByText('Participate').should('be.visible');
```

webapp/src/components/backstage/lhs_navigation.tsx
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
 
+import {PlaybookLhsDocument} from 'src/graphql/generated_types';
+import {getPlaybooksGraphQLClient} from 'src/graphql_client';
```
trilopin, line 5: we have the following disclaimer at this singleton
```ts
// Global for use in gloablly mounted modals that won't work with react context.
// Don't use this. If you need the client itself use `useApolloClient`.
```

should we use it for this goal? can't we call useApolloClient? 0/5 on this (on both approaches)

trilopin (changes requested):
I'm unsure about the GraphQL client thing.  I'll let @crspeller spread his wisdom on that 🧙🏽‍♂️.

Regarding e2e within calls, I personally chain calls when there's only one call inside the within. From my point of view, in that case, it's better to have a longer line than nesting contexts.

webapp/src/components/backstage/lhs_navigation.tsx
crspeller: @trilopin is right here we should use `useApolloClient` for this. This get was just a hack for global modals that don't have the context wrapped. 

Overall this is a pattern to be avoided but it's the best one I have found for interfacing between non-graphql and graphql for updates (until we can get push updates from server side)

crspeller (changes requested):

webapp/src/components/backstage/lhs_navigation.tsx
```diff
@@ -19,4 +21,16 @@ const LHSNavigation = () => {
...
+export const useLHS = () => {
+    const apolloClient = useApolloClient();
+
+    const refreshLists = () => {
+        apolloClient.refetchQueries({
+            include: [PlaybookLhsDocument],
+        });
+    };
+
+    return refreshLists;
```
trilopin, line 33: if you only expose the refresh op, should the hook be called `useLHSRefresh` ?

trilopin (approved):

crspeller (approved):

PR #1379, opened 2022-08-01
Improved LHS styling & use of icons

lieut-data:
#### Summary
Fix a number of styling issues with the LHS, bringing the font, spacing and background colours more into alignment with the Channels LHS.

 Before:
CleanShot 2022-08-01 at 15 12 51

After:
CleanShot 2022-08-01 at 15 36 27

Update a few icons, including the "View Timeline" icon:
image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-46076


trilopin (approved):
Amazing!

I'm worried about having so many ways of doign icons but we'll discuss it as a separate conversation.

crspeller (approved):

abhijit-singh (approved):
Looks so much better! Thank you @lieut-data! 

I spotted a few other random things. Not necessary to be done in this PR, and not blocking at all. Approving the PR and listing them down here:

1. The `+` icon button in the top right should have a background by default like in the designs and like it is for channels. The interaction (hover state, active state) should be identical to the channels LHS.
2. In the `+` menu:
  - the colour of the icons for each of the menu items should be 0.56% opacity of the center-channel-text color variable.
  - can we increase the space between the icon and text by ~3px? There's a margin-right on the icons of 3px right now which can be changed to 6px.
  
image

3. The spacing in the LHS changes slightly based on the height of the window. That shouldn't be the case.

4. The height of the top row in the LHS - the one with the team name and the `+` icon - should be 52 px. I can see the CSS where it is set to 52 px but it actually seems less than that. Might be because of the issue mentioned in the above point.
5. There seem to be tooltips appearing for all LHS items. They shouldn't be there unless the  name is getting truncated.
image

lieut-data:
Thank you so much for the review, @abhijit-singh! I'm going to merge this set of changes and then take a closer look at the additional improvements you've helped identify in a separate PR.

PR #1380, opened 2022-08-02
prevent users from saving invalid webhooks

trilopin:
#### Summary
Prevent webhooks from being saved at "Run actions modal" when they are invalid. The save button handler is removed temporarily (until content is valid) and the style changed (less opacity). I considered that approach better than just hiding the save button (which is a bit unexpected).

CleanShot 2022-08-02 at 10 56 50

Disabled text color for webhooks (when they are non-editable).

CleanShot 2022-08-02 at 10 58 13

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45959
Fixes https://mattermost.atlassian.net/browse/MM-45958

#### Checklist
- [ ] ~~Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
@@ -175,6 +175,31 @@ describe('runs > run details page > header', () => {
                         saveRunActionsModal();
                     });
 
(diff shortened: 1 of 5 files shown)


crspeller (approved):

gbochora (approved):
Cool!

PR #1383, opened 2022-08-02
recover channel and all timeline by default

trilopin:
#### Summary
Don't remove the channel from RHS when the run is finished (only when you don't have permission to join)

Keep all events at timeline enabled by default (no matter the navigation). Filters are kept on the playbook_run state (instead of rhs_timeline) to avoid resetting them if close/reopen rhs.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-46119
Fixes https://mattermost.atlassian.net/browse/MM-46123

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated


gbochora (approved):
Nice!

crspeller (approved):

PR #1384, opened 2022-08-02
Add sort by name

gbochora:
#### Summary
This PR adds sorting LHS run list by name. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46116

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated


server/api/schema.graphqls
```diff
@@ -10,9 +10,10 @@ type Query {
 	): [Playbook!]!
 
 	runs(
-		teamID: String = ""
-		statuses: [String!] = []
-		participantOrFollowerID: String = ""
+		teamID: String = "",
+		sort: String = "name",
```
crspeller, line 14: I don't think we want to set this as the default in the schema. We probably just want to modify the LHS query in https://github.com/mattermost/mattermost-plugin-playbooks/blob/f529e4a00faf188b4d6628e22fc9643bb215e2c1/webapp/src/graphql/playbook.graphql#L69

crspeller (changes requested):

crspeller (approved):

tests-e2e/cypress/integration/runs/list_spec.js
```diff
@@ -216,4 +216,51 @@ describe('runs > list', () => {
             });
         });
     });
+
+    describe('LHS run list', () => {
+        before(() => {
+            // # Login as testUser
+            cy.apiLogin(testUser);
+
+            cy.apiRunPlaybook({
```
trilopin, line 225: optional nit:

I've never tried in cypress but what if you do a Promise.all to avoid 4 level nesting.

trilopin (approved):
Thanks @gbochora !

PR #1385, opened 2022-08-03
fix several tiny ui issues

trilopin:
#### Summary

**RDP: RHS overview, Owner has no clickable feeling when the user is viewer**

|t|image|
|--|--|
|Before|CleanShot 2022-07-29 at 12 30 58|
|After|CleanShot 2022-08-03 at 16 57 19|

**Added padding to the placeholder when adding a new checklist**

|t|image|
|--|--|
|Before | Image Pasted at 2022-7-29 12-35|
|After|Screenshot 2022-08-03 at 16 41 07|

**Status post misaligned in channel RHS when the summary is collapsed**
|t|image|
|--|--|
|Before |Screenshot 2022-08-03 at 16 26 26|
|After |Screenshot 2022-08-03 at 16 36 01|

**Run actions enabled by default even without content (webhooks or broadcast channels)**

0/5 on the new magic there. We currently rely on the flags copied from the playbook to show enabled the toggles, no matter if they have content.

The new behavior is the following:
- if the prop is enabled but the content is empty, it shows the toggle off in the UI
- when saving (to server), if the user didn't add any content we send the property as off to the server side save

CleanShot 2022-08-03 at 18 01 20

#### Ticket Link
Fixes:
- https://mattermost.atlassian.net/browse/MM-46074
- https://mattermost.atlassian.net/browse/MM-46080
- https://mattermost.atlassian.net/browse/MM-45660
- https://mattermost.atlassian.net/browse/MM-46075

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data:
> when saving (to server), if the user didn't add any content we send the property as off to the server side save

@trilopin, would it make sense to have the server send down `false` if the corresponding value is empty, saving the client from having to "know" about this, but also making the API consistent for REST users?

lieut-data (approved):
Nice! Thanks for knocking these out :)

crspeller (approved):

trilopin:
> > when saving (to server), if the user didn't add any content we send the property as off to the server side save
> 
> @trilopin, would it make sense to have the server send down `false` if the corresponding value is empty, saving the client from having to "know" about this, but also making the API consistent for REST users?

yeah, to be fair, I was not very confident on the solution. What about saving as false (if related field is empty) when we copy the data from playbook to run?

lieut-data:
> yeah, to be fair, I was not very confident on the solution. What about saving as false (if related field is empty) when we copy the data from playbook to run?

One downside is that this only works for new runs, vs. fixing it consistently for all active ones. I'm 0/5 on fixing any of this now, just jotting ideas down :)

trilopin:
> > yeah, to be fair, I was not very confident on the solution. What about saving as false (if related field is empty) when we copy the data from playbook to run?
> 
> One downside is that this only works for new runs, vs. fixing it consistently for all active ones. I'm 0/5 on fixing any of this now, just jotting ideas down :)

looking at both proposals, I'm more inclined to solve for all new runs in an explicit way in createRun procedure (much better than the surprising magic I did on the webapp which is more error-prone). All the existing runs will finish naturally, and even if they don't, the worst-case scenario is not that bad.

trilopin:
merging this now.  

However, there are a few changes I did after the approval, let me know if you would want sth differently and I'll do a follow-up PR.

PR #1387, opened 2022-08-03
RDP RHS info: remove onclick events and keep regular a elements

trilopin:
#### Summary
Remove onclick handlers in the items that are pure links in RHS overview at Run Details Page.
That make the cmd+click works natively due to the inner `<a>`. We lose the click handler for the whole row, but it's not critical since the links at the right part catch naturally all the clicks.

Done for playbook and channel links.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-45960

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/runs/rdp_main_retrospective_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_main_retrospective_spec.js
@@ -172,7 +172,7 @@ describe('runs > run details page', () => {
                 editAndPublishRetro();
 
                 // # Switch to the run channel
-                cy.findByTestId('runinfo-channel').click();
+                cy.findByTestId('runinfo-channel-link').click();
 
                 // * Verify the modified retro text is posted
                 cy.getStyledComponent('CustomPostContent').should('exist').contains('Edited retrospective.');
--- a/tests-e2e/cypress/integration/runs/rdp_rhs_runinfo_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_rhs_runinfo_spec.js
@@ -68,7 +68,7 @@ describe('runs > run details page > run info', () => {
         const commonTests = () => {
             it('Playbook entry links to the playbook', () => {
                 // # Click on the Playbook entry
-                getOverviewEntry('playbook').click();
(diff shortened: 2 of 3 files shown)


gbochora (approved):

lieut-data (approved):
Nice! One possible suggestion: would it make sense to have `ItemLink` itself be an `Item` (vs. being contained within one), preserving the "whole row" clickability? 

Right now, the `Participants` row will still be fully clickable, but it might be odd to have more rows that highlight on hover but are only clickable in small places?

trilopin:
> Nice! One possible suggestion: would it make sense to have `ItemLink` itself be an `Item` (vs. being contained within one), preserving the "whole row" clickability?
> 
> Right now, the `Participants` row will still be fully clickable, but it might be odd to have more rows that highlight on hover but are only clickable in small places?

I tried this first. I made the row clickable with <a> and some ugly hacks to remove <a> formatting (text decoration, focus, color, etc..). In the end, it seemed more natural to just use links in the links.

That said, it's true that we have 3 different interactions in a very short space:
- channel and playbook -> pure link on the right
- participants -> whole row clickable and navigate to other RHS
- followers -> button clickable (not the row, or even the right part)

cc @abhijit-singh to get some insights

abhijit-singh (approved):
The clickable area is different for different rows but that's okay in my opinion because it is different kinds of objects. The downside of making the entire row clickable in case of `playbook` and `channel` is that accidental clicks can happen a lot more and it takes you to a completely different page/product. 

Listing a couple of random things I noticed around `run info` but might not be related to this PR at all so feel free to go ahead and merge this and I can create separate tickets for these. 
1. Noticed there's no arrow at the end of the `Participants` row which was there in the designs. 
image
2. The `1 participant` label is misaligned. It should be left aligned with the user avatars.
image

lieut-data:
Thanks, @abhijit-singh! Filed https://mattermost.atlassian.net/browse/MM-46256 and https://mattermost.atlassian.net/browse/MM-46257 to track these two observations and queued them up for this sprint.

PR #1388, opened 2022-08-03
MM-45975 Update default landing page to playbooks list.

crspeller:
#### Summary
Update the default landing page to be playbooks list instead of runs list.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45975


lieut-data:
@crspeller, is this one still valid?

crspeller:
@lieut-data I keep meaning to ask about this one. I think the runs list makes more sense then the playbooks list to drop on. So going to close this.

PR #1390, opened 2022-08-03
MM-45989 Fix overview navigation spacing.

crspeller:
#### Summary
Update the playbook navigation spacing to have some more spacing at low screen sizes. 

#### Screenshot
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45989


lieut-data (approved):

gbochora (approved):

abhijit-singh (changes requested):
Thanks, @crspeller! A few requests:
1. Can we double the space between the navigation and the white box containing the content? (Adding a margin-right: 1.5rem on the `playbook-preview-navbar` did it for me)
2. Can we ensure the text in the navigation never has to wrap like in the screenshot below? (I couldn't figure out why that was happening but setting a min-width of anything more than 140px solved the issue, but that's probably not the correct way)
image

abhijit-singh (approved):
Pefect 💯 

Thanks, @crspeller!

PR #1396, opened 2022-08-09
Add LHS polling.

crspeller:
#### Summary
Adds a poll every 60 seconds to update the LHS with any new content. Only active if you have the LHS in view.

### DIFF
--- a/webapp/src/components/sidebar/playbooks_sidebar.tsx
+++ b/webapp/src/components/sidebar/playbooks_sidebar.tsx
@@ -25,6 +25,7 @@ const useLHSData = (teamID: string) => {
             teamID,
         },
         fetchPolicy: 'cache-and-network',
+        pollInterval: 60000, // Poll every minute for updates
     });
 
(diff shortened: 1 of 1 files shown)


webapp/src/components/sidebar/playbooks_sidebar.tsx
```diff
@@ -25,6 +25,7 @@ const useLHSData = (teamID: string) => {
             teamID,
         },
         fetchPolicy: 'cache-and-network',
+        pollInterval: 60000, // Poll every minute for updates
```
lieut-data, line 28: What's the mechanism to stop doing this after the component is no longer in view?

crspeller: It's a hook. So somewhere there is probably a `useEffect` keeping track. I did check and make sure the functionality of only polling when visible was working.

lieut-data (approved):

gbochora (approved):

PR #1397, opened 2022-08-10
Morph adoption

gbochora:
#### Summary
Adds morph migrations from `v0.1.0` to `v0.3.0`. We have separate files for MySQL and PostgreSQL, located in
`server/sqlstore/migrations/mysql` and `server/sqlstore/migrations/postgres`.
Files with prefix `00001 - 00004` correspond to the migration `v0.1.0` -> `v0.2.0`
Files with prefix `00005` correspond to the migration `v0.2.0` -> `v0.3.0`
The migration filenames attempt to describe changes to the database.

Changes in the `migrate.go`:
- Commented morph migration temporary until properly tested. 
- Split `runMigrationsWithMorph` into several functions to make them reusable

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44243
https://mattermost.atlassian.net/browse/MM-44244
https://mattermost.atlassian.net/browse/MM-44245

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


server/sqlstore/migrations/mysql/000005_add_active_stage_title_to_ir_incident.down.sql
```diff
@@ -0,0 +1,14 @@
+SET @preparedStatement = (SELECT IF(
+    EXISTS(
+        SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS
+        WHERE table_name = 'IR_Incident'
+        AND table_schema = DATABASE()
+        AND column_name = 'ActiveStageTitle'
+    ) > 0,
```
lieut-data, line 7: Is `> 0` necessary here?

server/sqlstore/migrations/mysql/000005_add_active_stage_title_to_ir_incident.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    'ALTER TABLE IR_Incident ADD COLUMN ActiveStageTitle VARCHAR(1024) DEFAULT "";',
+    'SELECT 1;'
+));
+
+PREPARE addColumnIfNotExists FROM @preparedStatement;
+EXECUTE addColumnIfNotExists;
+DEALLOCATE PREPARE addColumnIfNotExists;
+
+
+UPDATE IR_Incident
```
lieut-data, line 17: Ooh, this goes back to early days! 0/5 for this PR, but I'm wondering if we should consider deprecating this more aggressively, since I believe we don't use this field at all anymore, and it would be a shame to permanently keep that complexity. Thoughts?

server/sqlstore/migrations/postgres/000003_create_ir_playbook.up.sql
```diff
@@ -0,0 +1,28 @@
...
+    TeamID TEXT NOT NULL,
+    CreatePublicIncident BOOLEAN NOT NULL,
+    CreateAt BIGINT NOT NULL,
+    DeleteAt BIGINT NOT NULL DEFAULT 0,
+    ChecklistsJSON JSON NOT NULL,
+    NumStages BIGINT NOT NULL DEFAULT 0,
+    NumSteps BIGINT NOT NULL DEFAULT 0
+);
+
+DO $$
```
lieut-data, line 14: Since we require PostgreSQL 10.0+ now, can we just rely on `CREATE INDEX IF NOT EXISTS`?

server/sqlstore/migrations/postgres/000005_add_active_stage_title_to_ir_incident.up.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE IR_Incident ADD COLUMN IF NOT EXISTS ActiveStageTitle TEXT DEFAULT '';
+
+CREATE OR REPLACE function json_array_length_safe(p_json text)
```
lieut-data, line 3: Should we drop this function afterwards?

lieut-data (changes requested):
Nice! Awesome to see us moving towards the multi-product architecture :)

A few comments below.

server/sqlstore/migrations/postgres/000003_create_ir_playbook.up.sql
gbochora: Excellent point!

server/sqlstore/migrations/mysql/000005_add_active_stage_title_to_ir_incident.up.sql
gbochora: I agree, it would be better if we removed this unnecessary complexity. I'll do it in a different PR, this one already has a lot of changes.

lieut-data (approved):

server/sqlstore/migrations/mysql/000003_create_ir_playbook.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS IR_Playbook;
```
crspeller, line 1: I feel like these down migrations are like a bomb waiting to go off and delete all our data.

crspeller (approved):

PR #1398, opened 2022-08-10
[MM-45116] Adding playbooks client stuff for insights and adding new webapp package dependencies

BenCookie95:
#### Summary
I had to upgrade the webapp commit hash so we can use the new insights handler, this also meant having to add in the client  and types package from webapp as dependencies. I had to make some lint fixes with these new additions.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42744

#### Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


plant99 (approved):

webapp/webpack.config.js
```diff
@@ -50,6 +50,7 @@ let config = {
             src: path.resolve(__dirname, './src/'),
             'mattermost-redux': path.resolve(__dirname, './node_modules/mattermost-webapp/packages/mattermost-redux/src/'),
             '@mattermost/types': path.resolve(__dirname, './node_modules/mattermost-webapp/packages/types/src/'),
```
crspeller, line 52: I don't think webpack needs these anymore right? Since we are using the ones from the package.json? Same with the lines in the `tsconfig.json`

crspeller (changes requested):

webapp/webpack.config.js
BenCookie95: Removed from here but I needed to keep it in `tsconfig.json`, there was a bunch of lint errors when it's removed

crspeller (approved):

PR #1407, opened 2022-08-15
Explain why run is disabled

lieut-data:
#### Summary
We already do this:
CleanShot 2022-08-15 at 15 19 49@2x

So now do this too:
CleanShot 2022-08-15 at 15 23 19@2x

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-40666


gbochora (approved):
Cool!

PR #1408, opened 2022-08-16
Add arrow forward icon to participants section

gbochora:
#### Summary
This PR fixes two minor UI issues in the `RDP > RHS > Run overview`:
- adds a missing arrow at the end of the `Participants`
- fixes the channel name truncating with an ellipsis

**Before**                                     vs                                      **After**
Screenshot 2022-08-16 at 10 08 22

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46256

#### Checklist

~~- [ ] Telemetry updated~~
- [ ] Unit tests updated

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
@@ -8,7 +8,7 @@ import {useIntl} from 'react-intl';
 import styled, {css} from 'styled-components';
 import {Channel} from '@mattermost/types/channels';
 
-import {AccountOutlineIcon, AccountMultipleOutlineIcon, BookOutlineIcon, BullhornOutlineIcon, ProductChannelsIcon, OpenInNewIcon} from '@mattermost/compass-icons/components';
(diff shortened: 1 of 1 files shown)


iomodo (approved):

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
```diff
@@ -273,21 +283,21 @@ const Item = (props: ItemProps) => {
...
     align-items: center;
 
     svg {
         margin-left: 3px;
-    }
+    }    
+`;
+
+const ItemContent = styled.div`
+    max-width: 230px;
```
trilopin, line 296: should we take into account that RHS could be 400px or 500px?

trilopin (approved):
Thanks @gbochora 

It's an already existing issue but I would want to double-check that the 230px max-width is valid for all possible RHS widths.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
gbochora: I've checked, and it looks well for 400px too.

PR #1412, opened 2022-08-16
Update run participants table and related queries.

### LINKED ISSUE #1399: Update table for run participants
There exists a table that tracks followers. For tracking participants we will update it. 

The `IR_Run_Participants` table currently has the following fields:
| Column | Description |
| -- | -- |
| IncidentID | ID of the run the user may be a participant or follower in. Refrences `ir_incident` |
| UserID | ID of the user. References `users` |
| IsFollower | true if the user is a follower false if not |

Despite the name it doesn't track participants, only followers. We must add the following fields:
| Column | Description |
| -- | -- |
| IsParticipant | true of the user is actually a participant |

Existing queries such as the one in https://github.com/mattermost/mattermost-plugin-playbooks/blob/0cd966b186b0c1d805620f30d4cc8e244f8fdd95/server/sqlstore/playbook_run.go#L159 should be updated to match.

### LINKED ISSUE #1401: Actions for syncronizing channel members with run participants (no UI)
In an effort to restore existing functionality as soon as possible create hidden actions on run channels that are enabled by default but hidden. **No UI is necessary for this ticket.** 

After this is complete this will restore the existing functionality so we may continue iterating on UI for participant separation. 

#### Run Actions
- [ ] When a participant joins a run --> Add them to linked channel
- [ ] When a participant leaves a run --> Remove the from the linked channel

In addition we will need a migration for existing runs. All existing runs should have all the above actions enabled by default.

#### Channel Actions (fake)
As we currently don't intend to actually support these channel actions beyond the scope of compatibility, these can be fake non-actions.

- [ ] When a member joins the run channel --> Add them to the linked run
- [ ] When a member leaves the run channel --> Remove them from the linked run

crspeller:
#### Summary
Adding participants to the `IR_Run_Participants` table. Modifying the queries to pull from the updated list.
Note that there is currently no way to actually become a participant from the UI.

Also adding automatic adding and removing of participants/members as this was required to fix tests.

Commits are logical if you want to review incrementally.  

#### Ticket Link
Fixes #1399
Fixes #1401

### DIFF
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -743,10 +743,10 @@ type PlaybookRunService interface {
 	// RequestGetInvolved posts a join request message in the run's channel
 	RequestGetInvolved(playbookRunID, requesterID string) error
 
-	// RemoveParticipants removes user from the run's participants
+	// RemoveParticipants removes users from the run's participants
 	RemoveParticipants(playbookRunID string, userIDs []string) error
 
-	// AddParticipants adds a user to the participants list
+	// AddParticipants adds users to the participants list
 	AddParticipants(playbookRunID string, userIDs []string) error
 }
 
@@ -843,10 +843,10 @@ type PlaybookRunStore interface {
 	// if a user is member of more than one channel, it will be counted multiple times
 	GetParticipantsActiveTotal() (int64, error)
 
-	// AddParticipants adds a particpant to the run
+	// AddParticipants adds particpants to the run
 	AddParticipants(playbookRunID string, userIDs []string) error
 
-	// RemoveParticipants removes a participant from the run
+	// RemoveParticipants removes participants from the run
 	RemoveParticipants(playbookRunID string, userIDs []string) error
 }
 
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -743,8 +743,11 @@ type PlaybookRunService interface {
 	// RequestGetInvolved posts a join request message in the run's channel
 	RequestGetInvolved(playbookRunID, requesterID string) error
 
-	// Leave removes user from the run's participants&followers list
+	// Leave removes user from the run's participants
 	Leave(playbookRunID, requesterID string) error
+
+	// Participate adds a user to the participants list
+	Participate(playbookRunID, userID string) error
 }
 
 // PlaybookRunStore defines the methods the PlaybookRunServiceImpl needs from the interfaceStore.
@@ -839,6 +842,12 @@ type PlaybookRunStore interface {
 	// (i.e. members of the playbook run channel when the run is active)
 	// if a user is member of more than one channel, it will be counted multiple times
 	GetParticipantsActiveTotal() (int64, error)
+
+	// AddParticipant adds a particpant to the run
+	AddParticipant(playbookRunID, userID string) error
+
+	// RemoveParticipant removes a participant from the run
+	RemoveParticipant(playbookRunID, userID string) error
 }
 
 // PlaybookRunTelemetry defines the methods that the PlaybookRunServiceImpl needs from the RudderTelemetry.
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -743,11 +743,11 @@ type PlaybookRunService interface {
 	// RequestGetInvolved posts a join request message in the run's channel
 	RequestGetInvolved(playbookRunID, requesterID string) error
 
-	// Leave removes user from the run's participants
-	Leave(playbookRunID, requesterID string) error
+	// RemoveParticipants removes user from the run's participants
+	RemoveParticipants(playbookRunID, userID string) error
 
-	// Participate adds a user to the participants list
-	Participate(playbookRunID, userID string) error
+	// AddParticipants adds a user to the participants list
+	AddParticipants(playbookRunID, userID string) error
 }
 
 // PlaybookRunStore defines the methods the PlaybookRunServiceImpl needs from the interfaceStore.
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -744,10 +744,10 @@ type PlaybookRunService interface {
 	RequestGetInvolved(playbookRunID, requesterID string) error
 
 	// RemoveParticipants removes user from the run's participants
-	RemoveParticipants(playbookRunID, userID string) error
+	RemoveParticipants(playbookRunID string, userIDs []string) error
 
 	// AddParticipants adds a user to the participants list
-	AddParticipants(playbookRunID, userID string) error
+	AddParticipants(playbookRunID string, userIDs []string) error
 }
 
 // PlaybookRunStore defines the methods the PlaybookRunServiceImpl needs from the interfaceStore.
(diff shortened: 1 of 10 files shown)


server/app/playbook_run.go
```diff
@@ -743,8 +743,11 @@ type PlaybookRunService interface {
 	// RequestGetInvolved posts a join request message in the run's channel
 	RequestGetInvolved(playbookRunID, requesterID string) error
 
-	// Leave removes user from the run's participants&followers list
+	// Leave removes user from the run's participants
 	Leave(playbookRunID, requesterID string) error
+
+	// Participate adds a user to the participants list
+	Participate(playbookRunID, userID string) error
```
trilopin, line 750: I'd use the same naming as mutations and store `AddParticipants`, `RemoveParticipants`. Otherwise, it could be misleading when we use leave on other users' behalf.

server/app/playbook_run.go
```diff
@@ -839,6 +842,12 @@ type PlaybookRunStore interface {
 	// (i.e. members of the playbook run channel when the run is active)
 	// if a user is member of more than one channel, it will be counted multiple times
 	GetParticipantsActiveTotal() (int64, error)
+
+	// AddParticipant adds a particpant to the run
+	AddParticipant(playbookRunID, userID string) error
```
trilopin, line 847: If it is not hard, I'd add support to pass multiple userID in one shot to avoid multiple queries. Also for RemoveParticipant.

server/app/playbook_run.go
trilopin: additionally, I'd add support to multiple userIDs in both funcions

server/app/playbook_run_service.go
```diff
@@ -2098,14 +2098,32 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
...
+	// Automaticly participate if you join the channel
+	// To be removed when separating members and participants is complete.
+	if err := s.Participate(playbookRunID, user.Id); err != nil {
+		s.logger.Errorf("faied to add participant that joined channel for run '%s', user '%s'; error: %s", playbookRunID, user.Id, err.Error())
+	}
+
+	_ = s.sendPlaybookRunToClient(playbookRunID)
+}
+
+func (s *PlaybookRunServiceImpl) addChannelJoinTimelineEvent(user *model.User, channel *model.Channel, actorID string, playbookRunID string, userID string) error {
```
trilopin, line 2118: nice refactor 👍🏽

server/app/playbook_run_service.go
```diff
@@ -2748,31 +2763,68 @@ func (s *PlaybookRunServiceImpl) RequestGetInvolved(playbookRunID, requesterID s
...
+	}
+
+	s.leaveActions(playbookRun, userID)
+
+	return nil
+}
+
+func (s *PlaybookRunServiceImpl) leaveActions(playbookRun *PlaybookRun, userID string) {
+	// Don't do anything if the user not a channel member
+	member, _ := s.pluginAPI.Channel.GetMember(playbookRun.ChannelID, userID)
```
trilopin, line 2789: nice! almost done waiting for just a couple of flags :)

server/sqlstore/playbook_run.go
```diff
@@ -929,26 +932,30 @@ func (s *playbookRunStore) buildPermissionsExpr(info app.RequesterInfo) sq.Sqliz
 		return nil
 	}
 
-	// Guests must be channel members
+	// Guests must be participants
```
trilopin, line 935: why do we need an special block for guests?

trilopin (changes requested):
Cool move!

I left some comments about naming and multiple-users support

server/sqlstore/playbook_run.go
crspeller: Because they don't have permissions to view all the playbooks on a team.

server/app/playbook_run_service.go
```diff
@@ -2098,14 +2098,32 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
...
 	summary := fmt.Sprintf("@%s joined ~%s", user.Username, channel.Name)
 	if actorID != "" {
-		actor, err2 := s.pluginAPI.User.Get(actorID)
-		if err2 != nil {
-			s.logger.Errorf("failed to resolve user for userID '%s'; error: %s", actorID, err2.Error())
-			return
+		actor, err := s.pluginAPI.User.Get(actorID)
+		if err != nil {
+			s.logger.Errorf("failed to resolve user for userID '%s'; error: %s", actorID, err.Error())
+			return err
```
trilopin, line 2126: we are currently both loogging the error and returning it. After that, we ignore it in the upper level. 

I see two options here:
- return err in this scope and log it in the upper level (remove logging here)
- log err in this level and don't return it to the upper level (remove return type)

Personally, I don't have any special preference for any of those.

Thoughts?

server/app/playbook_run_service.go
```diff
@@ -2181,14 +2185,26 @@ func (s *PlaybookRunServiceImpl) UserHasLeftChannel(userID, channelID, actorID s
 		return
 	}
 
+	_ = s.addChannelLeaveTimelineEvent(user, channel, actorID, playbookRunID, userID)
+
+	// Automaticly leave if you leave the channel
```
trilopin, line 2190:
```suggestion
	// Automatically leave run if you leave the channel
```

server/app/playbook_run_service.go
```diff
@@ -2205,11 +2221,10 @@ func (s *PlaybookRunServiceImpl) UserHasLeftChannel(userID, channelID, actorID s
 		CreatorUserID: actorID,
 	}
 
-	if _, err = s.store.CreateTimelineEvent(event); err != nil {
+	if _, err := s.store.CreateTimelineEvent(event); err != nil {
 		s.logger.Errorf("failed to create timeline event; error: %s", err.Error())
 	}
-
-	_ = s.sendPlaybookRunToClient(playbookRunID)
+	return nil
```
trilopin, line 2227: same comment than the one for joinTimelineEvent

trilopin (approved):
Thanks @crspeller: looks good to merge

I left a style comment about error returned+logged in the timeline functions

server/sqlstore/playbook_run.go
trilopin: yeah, makes sense when I look at the whole piece of code 👍🏽

server/app/playbook_run_service.go
```diff
@@ -2098,14 +2098,32 @@ func (s *PlaybookRunServiceImpl) UserHasJoinedChannel(userID, channelID, actorID
...
+
+	if !user.IsBot {
+		if err := s.Follow(playbookRunID, user.Id); err != nil {
+			s.logger.Errorf("user `%s` was not able to follow the run `%s`; error: %s", user.Id, playbookRunID, err.Error())
+		}
+	}
+
+	// Automaticly participate if you join the channel
+	// To be removed when separating members and participants is complete.
+	if err := s.AddParticipants(playbookRunID, []string{user.Id}); err != nil {
```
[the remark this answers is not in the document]
crspeller: Good question! They where allowed to be participants before, but that might change now. I suspect we don't.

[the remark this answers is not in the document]
crspeller: Decision was to allow it: https://community.mattermost.com/core/pl/jus93dbx1bn5urbxqz4qe4s35a

gbochora (approved):
Great stuff! 🎉

PR #1420, opened 2022-08-17
e2e: Cypress 10.3 -> 10.6

Phrynobatrachus:
The new Cypress 10 runner has been locking up on me pretty frequently while working on tests. Saturn noticed the same but also tried `10.4` and said that it seemed much better, so :crossed_fingers:.


calebroseland (approved):
LGTM. I was getting consistent lockups when saving/updating a spec file, which is not happening anymore with this version. https://github.com/cypress-io/cypress/issues/22026 perhaps?

PR #1431, opened 2022-08-23
Add channel scheme role support

gbochora:
#### Summary
Before: we passed default role values to update a user channel role, which didn't work in cases when the channel had schema-managed roles. 
After: If the channel has scheme-managed roles, we retrieve them from the database and pass them to the `UpdateChannelMemberRoles`. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39387

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


crspeller (approved):

trilopin (approved):
Thanks @gbochora 

I wonder if the team overrides should have preference over the system one. I really don't know that.

PR #1437, opened 2022-08-24
Mm 46379 fav run in LHS queries

trilopin:
#### Summary
Make fav runs to come in LHSDocument graphql query.
The reported Jira ticket wanted to make "fav" invoke the "follow" action, but this approach seems a better solution to me.

CleanShot 2022-08-24 at 11 41 28

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-46379

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


server/sqlstore/playbook_run.go
```diff
@@ -301,6 +301,17 @@ func (s *playbookRunStore) GetPlaybookRuns(requesterInfo app.RequesterInfo, opti
 			AND cm.UserId = ?)`, userIDFilter)
 		myRunsClause := sq.Or{followerFilterExpr, participantFilterExpr}
 
+		if options.IncludeFavorites {
```
trilopin, line 304: I wonder if this should be also used when participantID is not empty

server/api/graphql_root.go
```diff
@@ -351,7 +352,8 @@ func (r *RootResolver) UpdateRun(ctx context.Context, args struct {
 		return "", err
 	}
 
-	if err := c.permissions.RunManageProperties(userID, playbookRun.ID); err != nil {
+	// Enough permissions to do a fav/unfav, check if future ops need RunManageProperties
+	if err := c.permissions.RunView(userID, playbookRun.ID); err != nil {
```
gbochora, line 356: Should we have a separate function for the favorite/unfavorite? 
From my understanding purpose of the `UpdateRun` function is to modify the run and in this case we don't make changes to the run object.

trilopin: from my point of view, it would make sense because this is the relation user<->run, not the run itself. 

@crspeller thoughts?

if we go ahead, I'd prefer to do it in a different pr

server/sqlstore/playbook_run.go
gbochora: Not sure, it seems that yes, we should, but I've tested different combinations(participant, viewer, participant&follower, viewer&follower), and everything works perfectly.

server/sqlstore/playbook_run.go
```diff
@@ -301,6 +301,17 @@ func (s *playbookRunStore) GetPlaybookRuns(requesterInfo app.RequesterInfo, opti
...
+			favoriteFilterExpr := sq.Expr(`EXISTS(SELECT 1
+				FROM IR_Category AS cat
+				INNER JOIN IR_Category_Item it ON cat.ID = it.CategoryID
+				WHERE cat.Name = 'Favorite'
+				AND	it.Type = 'r'
+				AND	it.ItemID = i.ID
```
gbochora, line 310: Why do we need this condition?

trilopin: there are 4, which one exactly? :)

gbochora (approved):

server/sqlstore/playbook_run.go
gbochora: This one: `AND it.ItemID = i.ID`

trilopin: we want the runs that are marked as favorite, it.ItemID is the runID

the same way we match followers with WHERE rp.IncidentID = i.ID, and participants with  cm.ChannelId = i.ChannelID

webapp/package-lock.json
```diff
@@ -34795,7 +34794,6 @@
     },
     "mattermost-webapp": {
       "version": "git+ssh://git@github.com/mattermost/mattermost-webapp.git#c00fde021889aaddfcd3afa5ce18254756fc4185",
-      "integrity": "sha512-uBL3cO278tHi4lQV2fO1ytDKbRuFJ50o9OV3Z8CoCE3xd6CK1cLcF5XXylsYllK4aqmk7p/1GY9NZ3gOo5B5Lg==",
```
iomodo, line 34798: Why remove integrity checks?

iomodo (approved):
Looks perfect! Thanks @trilopin

webapp/package-lock.json
trilopin: without this, it's impossible to install with `npm i` because apparently integrity checks are arch oriented. that's why @lieut-data added the `node skip-integrity-check.ts` step.

somehow this lines came again to the repo

Issue #1434, opened 2022-08-23
complete add/join runs with private channels


trilopin:
Given the following resources:

- Run (Public or Private) "R1" linked to a private Channel "C1"
- A user "U1" who is a participant of "R1" and member of "C1"
- A user "U2" who is a participant of "R1" but NOT member of "C1"
- A user "U3" who is NOT a participant of "R1" and NOT member of "C1". Has view acess to "R1"

When:
- "U3" clicks on RDP's Participate button -> "U3" is added as participant of "R1", A message is sent to "C1" requesting to be added to channel.
- "U1" adds "U3" as participant of "R1" -> "U3" is added as participant of "R1" and as a member "C1"
- "U2" adds "U3" as participant of "R1" -> "U3" is added as participant of "R1" and a message is sent to "C1" requesting to be added to the channel.
Messages sent to the channel can slighly vary to reflect the different situations.

https://community.mattermost.com/core/pl/o3zhgbs1b3rmf89grhjqzf7zxa

Tasks:
- [ ] participate: when requester has no access to private channel -> plug message again
- [ ] add to run: when requester has no access to private channel -> create message (could be multiple users)

trilopin:
edge case that can be treated as independent

PR #1440, opened 2022-08-24
Separate run participants base

crspeller:
#### Summary
This is a combination of #1412 #1423 #1424 #1446 which all need to go into master at once. 

#### Ticket Link
Fixes #1399 
Fixes #1401 
Fixes #1400 
Fixes #1402
Fixes #1434

PR #1446, opened 2022-08-25
send message to channel when joined run has private channel

trilopin:
#### Summary
When some user joins the run through participate but the channel is private, user is added to the run but not to the channel. Same happens when the user who add another new user to the run doesn't have access to the private channel.

For those cases, a message will be sent to the channel as playbooks bot.

cleanup:
- `leave` and `requestGetInvolved` REST run endpoints removed, also removed from client.ts
- `requestGetInvolved` service removed, part of it done in `participateActions`
- `leave` and `requestGetInvolved` removed from client

**Graphql & client**
I was tempted to add the functions addParticipants/removeParticipants to the client package, but I ended up thinking that graphql/schema is already the API (and we should not restrict/hide to API clients). I've just added a shortcut to handle the mutations more easily in our server tests (and not write them each time).

0/5 on the messages, they need more work

#### Ticket Link
Fixes: https://github.com/mattermost/mattermost-plugin-playbooks/issues/1434

#### Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


assets/i18n/en.json
```diff
@@ -86,8 +86,12 @@
...
   },
   {
-    "id": "app.user.run.request_get_involved",
-    "translation": "@here — @{{.Name}} wants to participate in this run. To include them as a participant, add them to this channel.\n"
+    "id": "app.user.run.joined_run_channel_private_add_participant",
+    "translation": "@here — @{{.Name}} has been added to the run by @{{.RequesterName}}. They haven't been automatically added to the channel because @{{.RequesterName}} is not a channel member and the channel is private.\n"
+  },
+  {
+    "id": "app.user.run.joined_run_channel_private_participate",
+    "translation": "@here — @{{.Name}} has joined the run. They haven't been automatically added to the channel because it's private.\n"
```
trilopin, line 94: context: https://community.mattermost.com/core/pl/o3zhgbs1b3rmf89grhjqzf7zxa

suggestions welcome, these messages are not ready @itao @justinegeffen

gbochora (approved):
Looks good! Thanks @trilopin

assets/i18n/en.json
```diff
@@ -86,8 +86,12 @@
     "translation": "You have 0 assigned tasks."
   },
   {
-    "id": "app.user.run.request_get_involved",
-    "translation": "@here — @{{.Name}} wants to participate in this run. To include them as a participant, add them to this channel.\n"
+    "id": "app.user.run.joined_run_channel_private_add_participant",
+    "translation": "@here — @{{.Name}} has been added to the run by @{{.RequesterName}}. They haven't been automatically added to the channel because @{{.RequesterName}} is not a channel member and the channel is private.\n"
```
crspeller, line 90: This feels like it's a bit noisy. Are we sure we want `@here`?

crspeller (approved):

assets/i18n/en.json
trilopin: there was already here, but 1/5 on that one

justinegeffen:
Sorry I missed this, @trilopin! I was out of the office. Let me know if you still need my review. :)

trilopin:
> Sorry I missed this, @trilopin! I was out of the office. Let me know if you still need my review. :)

no problem @justinegeffen, your help would be very appreciated on https://github.com/mattermost/mattermost-plugin-playbooks/pull/1440

PR #1452, opened 2022-08-29
RHS: do not trigger inline edit if user is doing a text-selection

trilopin:
#### Summary
Prevent users from accidentally enter into inline edit when they want to select the title or description in the channel RHS.

CleanShot 2022-08-29 at 13 31 54

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-38717

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):

calebroseland (approved):
Nice 😃

PR #1458, opened 2022-08-30
Mm 44752 add dot menu to lhs

gbochora:
#### Summary
This PR adds a dot menu to the LHS. 
Playbook menu items:
- Favorite / Unfavorite
- Copy link
- Leave

Run menu items:
- Favorite / Unfavorite
- Copy link
- Follow / Unfollow
- Leave (Only if participants)

Changelog highlights:
- DotMenu had issues with z-index and content overflow, thanks to @calebroseland switched to using `dropdown` instead.
- Moved menu items from `contex_menu` to `controls` to reuse them in the LHS 
- Moved shared components to `shared.tsx`

https://user-images.githubusercontent.com/4368372/187399285-3803723d-10e4-421f-a7fe-b9e021a38a0b.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44752

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/backstage/lhs_run_dot_menu.tsx
```diff
@@ -0,0 +1,85 @@
...
+export const LHSRunDotMenu = ({playbookRunId, teamId}: Props) => {
+    const {formatMessage} = useIntl();
+    const {add: addToast} = useToaster();
+    const [isFavoriteRun, toggleFavorite] = useFavoriteRun(teamId, playbookRunId);
+    const [playbookRun] = useRun(playbookRunId);
+    const currentUser = useSelector(getCurrentUser);
+    const [metadata] = useRunMetadata(playbookRun?.id, [JSON.stringify(playbookRun?.participant_ids)]);
+    const followState = useFollowers(metadata?.followers || []);
+    const {isFollowing, followers, setFollowers} = followState;
+    const {leaveRunConfirmModal, showLeaveRunConfirm} = useLeaveRun(playbookRunId, isFollowing);
```
trilopin, line 35: This is getting from fetch every run and metadata individually. 

I'd rather add to LHSQuery the data you need to handle dotmenu and pass to this component as a prop instead of fetching.

trilopin: I guess you need participantIDs and isFollower. 

Would that be enough to remove the useRunMetadata, useRun and useFavoriteRun?

trilopin (changes requested):
Huge effort in this PR, thanks @gbochora 

I'm concerned on the data flow approach. The right approach to me would be to get the data from LHSQuery and avoid all the additional fetch.

We could delay it a bit for useRun and rely on useThing getting that from redux for the following useRun calls. Unfortunately, we still have useRunMetadata with relies on useFetch and it's not stored in redux.

Run resolver is a bit more finished on [this branch](https://github.com/mattermost/mattermost-plugin-playbooks/tree/graphql-run-resolver). We may want to pick some of that code to solve this problem.

Phrynobatrachus:
Something I noticed - for runs, favorite/unfavorite using the dot menu updates the LHS as expected, however the star icon in the run details does not change until a refresh.

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
```diff
@@ -76,7 +76,7 @@ describe('runs > run details page > header', () => {
 
     const getDropdownItemByText = (text) => {
         cy.findByTestId('run-header-section').find('h1').click();
-        return cy.findByTestId('run-header-section').findByTestId('dropdownmenu').findByText(text);
+        return cy.findByTestId('dropdownmenu').findByText(text);
```
trilopin, line 79: why did you change this func? it was designed to find dropdownmenu inside run-header-section, not globally.

If it failed, I would have tried to change the chaining to a within statement

webapp/src/components/backstage/playbook_editor/controls.tsx
```diff
@@ -318,6 +324,58 @@ export const JoinPlaybook = ({playbook: {id: playbookId}, refetch}: ControlProps
     );
 };
 
+export const FavoritePlaybookMenuItem = (props: {isFavorite: boolean, toggleFavorite: () => void}) => {
+    const {formatMessage} = useIntl();
```
trilopin, line 328: Once you made a separated component for this, would make sense to import and use the useUpdateRun here instead of passing from parent? It may hide some complexity. You already did that for leave op.

webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
```diff
@@ -66,63 +59,31 @@ export const ContextMenu = ({playbookRun, role, isFavoriteRun, isFollowing, togg
...
+                <CopyRunLinkMenuItem
+                    playbookRunId={playbookRun.id}
+                />
+                <Separator/>
+                <RunActionsMenuItem
+                    showRunActionsModal={showRunActionsModal}
+                />
+                <ExportLogsMenuItem
+                    exportAvailable={exportAvailable}
+                    onExportClick={onExportClick}
```
trilopin, line 76: would make sense to hide complexity and move that flow onExportClick to the controls file?

webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
```diff
@@ -4,27 +4,25 @@
...
-import {PlaybookRun, playbookRunIsActive} from 'src/types/playbook_run';
-import DotMenu, {DropdownMenuItem} from 'src/components/dot_menu';
+import {exportChannelUrl, leaveRun} from 'src/client';
+import {PlaybookRun} from 'src/types/playbook_run';
+import DotMenu from 'src/components/dot_menu';
 import {SemiBoldHeading} from 'src/styles/headings';
-import {copyToClipboard} from 'src/utils';
 import {ToastType, useToaster} from 'src/components/backstage/toast_banner';
-import {useAllowChannelExport, useExportLogAvailable} from 'src/hooks';
+import {useAllowChannelExport, useExportLogAvailable, useRun} from 'src/hooks';
```
trilopin, line 17: you can avoid calling useRun if you retrieve owneruserid from GQL

webapp/src/components/dot_menu.tsx
```diff
@@ -30,16 +31,12 @@ export const DotMenuButton = styled.div<{isActive: boolean}>`
...
-    position: relative;
-`;
-
 export const DropdownMenu = styled.div`
     display: flex;
     flex-direction: column;
 
     width: max-content;
-    min-width: 160px;
+    min-width: 16rem;
```
trilopin, line 39: why this change? just curiosity

trilopin (changes requested):
great work on graphql move


I still see tons of run requests, probably because uiseRun inside useLeaveRun. Could you please add owneruserid to gql so we can avoid that one?

the comments on moving logic to controls is optional and I'd do it in future prs.

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
gbochora: DotMenu now works differently; The drop-down is not added inside the header element. Because of that, we can't find it inside the header.

webapp/src/components/backstage/playbook_editor/controls.tsx
gbochora: Good point; this way, we will encapsulate logic inside menu item.

webapp/src/components/dot_menu.tsx
gbochora: It's Caleb's change. I guess now the dropdown is more flexible; if someone zooms, the page dropdown will increase proportionally to font size.

Phrynobatrachus:
Favorite/unfavorite is working better in that the run details star now updates when clicking away and back to the run without a full page refresh, but I still would like that to happen immediately in the same view if possible.

gbochora:
> Favorite/unfavorite is working better in that the run details star now updates when clicking away and back to the run without a full page refresh, but I still would like that to happen immediately in the same view if possible.

Unfortunately, there is no easy way to make the live update work for Follow/Unfollow and Favorite/Unfavorite until we move entirely run API to GraphQL. I'll add Jira bugs and will fix them after API is migrated.

Phrynobatrachus (approved):
LGTM, thanks!

webapp/src/graphql/runs.graphql
```diff
@@ -1,3 +1,13 @@
+query Run($id: String!) {
```
trilopin, line 1: this is not really needed, but is not harmful too

trilopin (changes requested):
Just a detail, you must force a refreshLHS when setting a new owner to force LHS to know about the change, and allow you to leave the run.

gbochora:
> Just a detail, you must force a refreshLHS when setting a new owner to force LHS to know about the change, and allow you to leave the run.

Nice catch! thanks.

calebroseland (approved):
Everything LGTM! Thank you for all the work on this. Really great stuff.

PR #1463, opened 2022-09-01
limit size for all incoming requests and migrate postgres text column 

trilopin:
## Summary

**HTTP size**
Limit incoming HTTP requests to be 5MB large. Any requests larger than that will fail. mattermost-server limit request with more size than max_file_size, playbooks plugin adds a lower limit: 5MB.

**Database**
Additionally, PostgreSQL text fields have been migrated to varchar with the limits we already apply to MySQL (some of them need revision, but it's not part of this PR). The following strategy has been used:
- if text in postgre and varchar in MySQL -> change to varchar with MySQL limit
- if text in postgre and text in MySQL -> change to varchar with implicit MySQL limit (65535)
- if JSON, checked explicitly in code

Migrations have been done inside one huge migration but with individual queries. Key points:
- no data loss, if current data is longer than the target type, the individual change will fail
- if one field change fails, the rest will be applied

Migration was also added in morph-style to ease close-future work. 

**Ability to override config for server e2e tests**
Sometimes we need a specific config for the temporary server we start with server e2e tests. 

The old fake logger has been removed and logging has been directly disabled. Additionally, a mechanism has been added in order to support any config we may want for the e2e server (`example at server/e2etest.config.json.sample`). 

If `server/e2etest.config.json` exists, it will be read and the values will take preference in server config (same structure as mattermost-server's config.json but only with the keys we want to override)

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-46829

#### Checklist
- [ ] ~~Telemetry updated~~
- [X] Unit tests updated
- [X] Manually tested


crspeller (approved):

gbochora (approved):

srkgupta (changes requested):
Thanks @trilopin. The issue is partially fixed now. I still see the client-side JS issue as shown in this screenshot. Can you please fix this as well.

server/sqlstore/migrations_utils.go
```diff
@@ -57,6 +57,22 @@ var addColumnToPGTable = func(e sqlx.Ext, tableName, columnName, columnType stri
...
 
+var changeColumnTypeToPGTable = func(e sqlx.Ext, tableName, columnName, columnType string) error {
+	_, err := e.Exec(fmt.Sprintf(`
+		DO
+		$$
+		BEGIN
+			ALTER TABLE %s ALTER COLUMN %s TYPE %s;
+		EXCEPTION
+			WHEN others THEN
+				RAISE NOTICE 'Ignoring ALTER TABLE statement. Column "%s" can not be changed to type %s in table "%s".';
```
trilopin, line 68: @gbochora @lieut-data I followed the same pattern but I couldn't make this notice appear when I forced the failure (manually I could, programmatically I couldn't). 

should it be an error instead?

server/e2etest.config.json.sample
```diff
@@ -0,0 +1,14 @@
+// Rename this file (and remove comments) to e2etest.config.json to force some server settings to e2e engine
```
lieut-data, line 1: 0/5, but wondering if we should just force these on all the time? The way I /think/ I did it in my PR, the logs only get spit out if the test fails, and it's likely you want these in that case all the time.

server/sqlstore/migrations_utils.go
lieut-data: If an error, playbooks as a product can't be upgraded: is it worth leaving things as-is in such a case instead?

server/sqlstore/store.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+// MaxJSONLength holds the limit we set for JSON data in postgres
+// Since JSON data type is unboounded, we need to set a limit
+// that we'll control manually.
+const MaxJSONLength = 65535
```
lieut-data, line 18: Do we need to export this, or could it be internal only? (Also, propose we raise to something like 256kb to avoid artificially low limits.)

lieut-data (approved):
A handful of comments, but nothing strictly blocking. Nice work :)

server/sqlstore/migrations/future/postgres/0.57.up.sql
```diff
@@ -0,0 +1,92 @@
+ALTER TABLE ir_incident ALTER COLUMN id TYPE varchar(26);
```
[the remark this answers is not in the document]
trilopin: good idea 👍🏽

gbochora (commented):
Great work @trilopin! I left a few comments. Also, the table `IR_ViewedChannel` is missing in migration; was it intentional?

server/sqlstore/migrations.go
```diff
@@ -2128,6 +2130,145 @@ var migrations = []Migration{
...
+				"ir_category_item": {
+					{"type", 1},
+					{"categoryid", 26},
+					{"itemid", 26},
+				},
+				"ir_channelactions": {
+					{"id", 26},
+					{"actiontype", 65535},
+					{"triggertype", 65535},
+				},
```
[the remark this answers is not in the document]
trilopin: yep, it's a typo. channelid was already a varchar(26)

trilopin:
> the table `IR_ViewedChannel` is missing in migration; was it intentional?
not intentional, I added it and will be pushed soon. Nice catch @gbochora

server/sqlstore/migrations_utils.go
trilopin: > If an error, playbooks as a product can't be upgraded: is it worth leaving things as-is in such a case instead?

unless we print and return nil as error inside the migration

trilopin:
Feedback addressed:

@gbochora 
- added missing field migration to ir_incident
- added missing table migration
- fixed typo at ir_channelaction

@lieut-data 
- created placeholder MySQL migrations to be more explicit (future morph)
- increased JSON limit to 256KB and graphql request size to 300KB (It was at 100KB) 

@srkgupta 
- added test & error handling for playbook update

gbochora:
Looks great! 🎉 thanks @trilopin

srkgupta (approved):
Thanks @trilopin. Changes LGTM

PR #1464, opened 2022-09-05
[GH-19785, MM-39579] - Added timezone settings selected by user to playbook

rolwin100:
* added a selected for getting the timezone (getCurrentUserTimezone) similar to the one used in the web-app
* passed timezone parameter to parse function

resolves https://github.com/mattermost/mattermost-server/issues/19785

#### Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


calebroseland (changes requested):
Hi @rolwin100, thank you for the PR. I tested it in a couple scenarios and it doesn't seem to apply the timezone on the parsing side. `Luxon.DateTime.setZone` will not affect the inner UTC time produced by `chrono-node`, but rather just specify which timezone it would be displayed in when formatting later on.

I think the issue for this ticket is getting `chrono-node`'s inner `Date` creation to respect the user-selected timezone for fixed-time based queries. (It's currently only respecting the browser/system timezone.)

This might be possible by using Chrono's [Parsing Reference Date/Timezone](https://github.com/wanasit/chrono#parsing-reference-date--timezone) or [Parser and Refiner](https://github.com/wanasit/chrono#customize-chrono) customization methods. If those don't work, another option might be calculating the timezone offset difference between the _Mattermost User-selected timezone_ and the _browser's system timezone_ and applying that offset after getting the resulting `Date`.


Re: `getCurrentUserTimezone`,  I'm not sure it's necessary to copy that over to Playbooks. `Luxon.Settings.defaultZone` is already provided here: https://github.com/mattermost/mattermost-webapp/blob/master/components/root/effects.ts#L26. That tells Luxon to format according to the mattermost user timezone. It returns a [`Zone`](https://moment.github.io/luxon/api-docs/index.html#zone), so you could potentially use that inside of `parseDateTime`/`parseDateTimes`.

`master` and this branch both behave the same:
(Browser/system set to: `America/Chicago`, MM User set to: `America/Anchorage`)
CleanShot 2022-09-12 at 06 44 41@2x
CleanShot 2022-09-12 at 07 44 39@2x
CleanShot 2022-09-12 at 07 49 11@2x

rolwin100:
got it @calebroseland will work on it

calebroseland:
Hi @rolwin100, just a gentle check in: is there anything further you wanted to do with this? Either way is totally fine! Just checking if there's anything we can help with; else we might want to re-open it if someone else might want to work on it.

rolwin100:
Hey sure let me know wht needs to be done @trilopin was telling some tasks for this. Please create a ticket and assign it to me. Thanks

rolwin100:
@calebroseland you can assign this to someone

PR #1467, opened 2022-09-06
[MM-45662] Fix context dropdown closing when clicked

MusabShakeel576:
#### Summary
Resolved task filter context dropdown closing when checked and unchecked.

Now context dropdown is closed only when clicking on parent element, clicking outside the menu, or if dropdown parent appears on hover then hovering outside of the element.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/20821

#### Checklist
- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
@@ -124,6 +124,9 @@ describe('runs > run details page > header', () => {
                 // * Verify clipboard content
                 cy.get('@clipboard').its('contents').should('contain', `/playbooks/runs/${playbookRun.id}`);
             });
+
+            // * Verify context menu is opened
+            cy.findByTestId('run-header-section').findByTestId('dropdownmenu').should('be.visible');
         });
     };
 
--- a/tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
@@ -124,9 +124,6 @@ describe('runs > run details page > header', () => {
                 // * Verify clipboard content
                 cy.get('@clipboard').its('contents').should('contain', `/playbooks/runs/${playbookRun.id}`);
             });
-
-            // * Verify context menu is opened
-            cy.findByTestId('run-header-section').findByTestId('dropdownmenu').should('be.visible');
         });
     };
 
--- a/webapp/src/components/dot_menu.tsx
+++ b/webapp/src/components/dot_menu.tsx
@@ -66,6 +66,7 @@ type DotMenuProps = {
     className?: string;
     isActive?: boolean;
     onOpenChange?: (isOpen: boolean) => void;
+    isMultiCheckbox?: boolean;
 };
 
 type DropdownProps = Omit<ComponentProps<typeof Dropdown>, 'target' | 'children'>;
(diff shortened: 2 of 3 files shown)


tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
```diff
@@ -124,6 +124,9 @@ describe('runs > run details page > header', () => {
                 // * Verify clipboard content
                 cy.get('@clipboard').its('contents').should('contain', `/playbooks/runs/${playbookRun.id}`);
             });
+
+            // * Verify context menu is opened
+            cy.findByTestId('run-header-section').findByTestId('dropdownmenu').should('be.visible');
```
Phrynobatrachus, line 129: This should not be necessary, for example.

Phrynobatrachus (changes requested):
Copying my comment from Dev: Playbooks:

I think your solution is too general. The failures are from usages of the Dot Menu that your PR affects, but should not. The plainest way I see is to add a new prop to `Dot_Menu` that you pass to its usage in `Multi_Checkbox`, and then make your original change conditional on that.

tests-e2e/cypress/integration/runs/rdp_main_header_spec.js
MusabShakeel576: alright, removed [056946c](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1467/commits/056946c2e9d193da79f560ada6f07ae14e01e8fa#diff-60962e12ef06c5823b22865dfa586ffbf548aeb7e12548efd361c0b48a8c9678)

webapp/src/components/dot_menu.tsx
```diff
@@ -129,7 +131,7 @@ const DotMenu = ({
                 data-testid='dropdownmenu'
                 onClick={(e) => {
                     e.stopPropagation();
-                    setOpen(false);
+                    return isMultiCheckbox ? null : setOpen(false);
```
calebroseland, line 134: This return ternary feels a little odd to me. Could we just do an if block?

```suggestion
                   if (closeOnClick) {
                       setOpen(false);
                   }
```

calebroseland (changes requested):
I think the prop approach is great solution, but maybe it could be named according to the behavior?

webapp/src/components/dot_menu.tsx
MusabShakeel576: resolved [f5bf468](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1467/commits/f5bf468b7f99baa539993767d8dc49f7436ab0a5#diff-cd9f22e13bba5ded76a48757c8918b7ca3e7dd1dc01cd055ba7543d9a64b522bR134-R136)

webapp/src/components/dot_menu.tsx
```diff
@@ -77,6 +78,7 @@ const DotMenu = ({
     className,
     disabled,
     isActive,
+    isMultiCheckbox,
```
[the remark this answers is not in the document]
MusabShakeel576: resolved [f5bf468](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1467/commits/f5bf468b7f99baa539993767d8dc49f7436ab0a5#diff-cd9f22e13bba5ded76a48757c8918b7ca3e7dd1dc01cd055ba7543d9a64b522bR81)

Phrynobatrachus (approved):
Nice. The failing E2E is unrelated, it's something I need to look at separately. Thank you for the contribution!

calebroseland (approved):
Awesome, thanks @MusabShakeel576!

PR #1470, opened 2022-09-09
Telemetry: new approach draft + PostUpdate viewed

trilopin:
## Summary
This ticket covers three goals:
- add missing tracking for StatusUpdate impressions
- migrate run_detail page view tracking to the new approach
- draft the new telemetry approach

### New telemetry approach
Key points:
- use different types for pageviews "page" and event tracking "track", this is natively supported by rudder/segment. Before it was always "track" and we included the "view_" prefix in the event names.
- server acts more like a proxy (+ some necessary validations)
- event names should be part of enums at server/client and, will be validated (to avoid typos). 
- we accept arbitrary extra data from the webapp to enrich the event. 
- generic hook `useViewTelemetry(target: TelemetryViewTarget, trigger?: string, data = {})` to ease adoption 
- e2e testing for telemetry

Side note, we usually inject some data server-side every time we report an event linked to a playbook/run/template. It can be technically done in the new approach, but I chose to avoid it until each additional attribute is necessary.

### status update tracking
We want to know how many people use playbooks outside playbooks, and identify all those "readers" that are aware through broadcasts to channels or DM.
As a use case for the recent approach change, some data has been added to the event to enrich it: channel_type, playbook_id, playbookrun_id and, post_id. From that info I could answer:
- How many total status impressions do we have?
- How many unique post impressions do we have?
- How many unique users have seen a status post? 
- How many unique runs have been viewed (status) by users?
- How many unique playbooks have been viewed (status) by users?

### migrate run_details_page event
- now is tracked as page, instead of event (track)
- removed view_ prefix
- added rich data to the pageview: role (participant/viewer), where we navigated from (LHS,run_list,channel_rhs, etc..)

New capabilities:
- know from which page the users navigate to RDP
- know the ratio of participant vs viewer

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-46690

## Checklist
- [X] Telemetry updated 😂 
- [X] Unit tests updated


client/client.go
```diff
@@ -48,6 +48,8 @@ type Client struct {
 	Stats *StatsService
 	// Reminders is a collection of methods used to interact with reminders.
 	Reminders *RemindersService
+	// Telemetry is a collection of methods used to interact with telemetry.
+	Telemetry *TelemetryService
```
trilopin, line 52: if we look at client lib as more than a test helper, this may not be here.

server/app/telemetry.go
```diff
@@ -0,0 +1,52 @@
...
+type TelemetryPage string
+
+const (
+	telemetryRunStatusUpdate TelemetryPage = "run_status_update"
+)
+
+// NewTelemetryPage creates an instance of TelemetryPage from a string.
+// It's useful to validate that the arbitrary string has a equivalent constant
+// for what pages we want to track (and avoid typos).
+func NewTelemetryPage(name string) (TelemetryPage, error) {
```
trilopin, line 24: Personally I'm more inclined to create a nested package app.telemetry and avoid prefixing everything with Telemetry. However, I followed the current pattern

tests-e2e/cypress/integration/runs/rdp_main_statusupdate_spec.js
```diff
@@ -110,6 +110,22 @@ describe('runs > run details page > status update', () => {
...
+
+                // * check that post has been added
+                cy.getLastPost().contains('my nice update');
+
+                // * assert  telemetry pageview
+                cy.wait('@telemetry').then((interception) => {
+                    expect(interception.request.body.name).to.eq('run_status_update');
+                    expect(interception.request.body.type).to.eq('page');
+                    expect(interception.request.body.properties.channel_type).to.eq('P');
+                });
```
trilopin, line 128: e2e check for tracking! It feels easy even without creating new commands.

I'd love to hear @Phrynobatrachus thoughts anyway

tests-e2e/cypress/integration/lhs_spec.js
```diff
@@ -56,6 +56,47 @@ describe('lhs', () => {
...
+
+        it('click run', () => {
+            // # Intercept all calls to telemetry
+            cy.intercept('/plugins/playbooks/api/v0/telemetry').as('telemetry');
+
+            // # Click on run at LHS
+            cy.findByTestId('Runs').findByTestId(playbookRun.name).click();
+
+            // * assert  telemetry pageview
+            cy.wait('@telemetry').then((interception) => {
```
lieut-data, line 89: Sweet!

lieut-data (approved):
Awesome -- just of a handful of comments, but a solid improvement!

server/telemetry/rudder.go
```diff
@@ -136,7 +137,26 @@ func (t *RudderTelemetry) track(event string, properties map[string]interface{})
...
+}
+
+// Page is the generic tracker for pageviews to rudderstack
+func (t *RudderTelemetry) Page(name app.TelemetryPage, properties map[string]interface{}) {
+	t.mutex.RLock()
+	defer t.mutex.RUnlock()
+
+	if !t.enabled {
+		return
+	}
```
gbochora, line 152: Why don't we do the same check in the `Track` function?

gbochora (approved):
Great work @trilopin!

tests-e2e/cypress/integration/runs/rdp_main_statusupdate_spec.js
Phrynobatrachus: No real comments here, `intercept` is definitely the correct way to do this. :+1:

server/telemetry/rudder.go
trilopin: it's already done, Track function already existed and had this code

PR #1472, opened 2022-09-09
MM-45225 Absolute task due date in playbook editor

m1lt0n:
#### Summary

This PR switches to using absolute due date in the playbook editor (as is the case while editing a run).

#### Ticket Link

Jira ticket: https://mattermost.atlassian.net/browse/MM-45225

#### Checklist

~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~


crspeller (changes requested):
This ticket takes more than just changing the UI. The backend will need to be updated to recognize this as not a duration but an absolute time value.

Also I believe we also want to preserve the functionality of entering a duration here (so 7 days after the run is started for example).

m1lt0n:
Thanks a lot for the context @crspeller I'll look into this again on Friday in the context of OSF.

PR #1475, opened 2022-09-12
clean unused code inside backstage

trilopin:
#### Summary
Cleanup all backstage unused code.
Quick try that would need some global QA.

#### Ticket Link
Nope

#### Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data (approved):
Given the tests pass (wow!), I'm a fan of just merging this.

calebroseland (approved):
This feels so much like cleaning out the garage... 💯

If there truly is something we need from here, we can always rewind!

PR #1478, opened 2022-09-12
Mm 45998 clear playbook before saving

gbochora:
#### Summary
If a playbook’s checklist is updated using the API, it’s possible to set bits like the checked status and assignee, even though these don’t show up in the UI as such.

When running the playbook, those bits are copied, resulting in pre-assignees (yay) and pre-checked tasks (booo). 

This PR sets empty values for noneditable fields of the playbook object before storing it in the database. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45998

#### Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


crspeller (approved):
I like the duplication as those will be removed separately. However maybe a comment would help in case someone needs to update them before that happens?

iomodo (approved):

Issue #1444, opened 2022-08-24
Add participate button to the participants list in the RHS


crspeller:
Participation icon. 

[Figma Designs](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=275%3A279713)

Issue #1443, opened 2022-08-24
Add follow button to the run RHS


crspeller:
Add a follow button to the header of the run RHS.

[Figma Designs](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=275%3A279713)

PR #1484, opened 2022-09-14
Gh 1444 1443 add participate/follow button channel rhs

trilopin:
## Summary
Add participate and follow buttons to Channels RHS.

Participate flow has been simplified, messages were different when you had no access to the channel (now you can participate anyways)

Some code has been moved to share logic properly between RDP and channels RHS:
- new run.tsx file in the hooks folder with custom hooks for folowers/follow/unfollow/participate
- RDP code migrated to use the hooks
- new Code in rhs_about or rhs_title to use the same hooks 

### Follow/Unfollow Run
It's possible to follow/unfollow run from 4 places: run_list, channels_rhs, rdp-rhs and playbooks-lhs (some of them still don't use the new hook).
They all have telemetry (and additionally include where the action was triggered).
E2E tests included for follow/unfollow (tested telemetry too)

### Participate
It's possible to participate in a run:
- from RDP if you were not a participant but had access to RDP
- from Channel RHS, if you were a channel member but not a participant
E2E tests included for participate/leave with telemetry

### Telemetry
Tiny refactor on telemetry to accommodate new events with the new approach. Enum types for telemetry have been improved.

### demo videos
Some flows are still not reproducible (without circuit breakers in the code). I added some recordings to see them in action and they will be available after the run-action is done. 

**follow/unfollow run from channels rhs**
https://user-images.githubusercontent.com/4096774/193800082-725be264-8d15-426e-92b7-de5cbd33b52b.mp4

**participate from channel when there are no participants**
https://user-images.githubusercontent.com/4096774/193797716-4112d5e2-f558-4f2e-aa3f-63e1ef1ee643.mp4

**participate from channel when there are participants**
https://user-images.githubusercontent.com/4096774/193797699-bb76a659-b380-46ba-88ce-71ef9bf40fc4.mp4

**I'm the owner and there's no any other participant**
Still needs new code at RDP RHS participants by @gbochora .
https://user-images.githubusercontent.com/4096774/193812377-1948230c-3faf-4fb9-ae86-88429db5122c.mp4

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1444
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1443

## Checklist
- [X] Telemetry updated
- [X] Unit tests updated


webapp/src/components/rhs/rhs_participants.tsx
```diff
@@ -4,49 +4,82 @@
...
-            <NoParticipants>
-                <FormattedMessage defaultMessage='Nobody yet.'/>
-                {/* eslint-disable-next-line formatjs/no-literal-string-in-jsx */}
-                {' '}
-                <AddParticipants/>
-            </NoParticipants>
+            <Container>
+                <NoParticipants>
+                    <FormattedMessage defaultMessage='Nobody yet.'/>
+                    {/* eslint-disable-next-line formatjs/no-literal-string-in-jsx */}
```
crspeller, line 47: Could also add ' ' to the list of exceptions for that rule.

webapp/src/components/rhs/rhs_title.tsx
```diff
@@ -58,6 +62,9 @@ const RHSTitle = () => {
                         </StyledButtonIcon>
                     </RHSTitleLink>
                 </OverlayTrigger>
+                <FollowingWrapper>
+                    {FollowingButton ? <FollowingButton/> : null}
```
crspeller, line 66: This is kinda ugly. Could we turn that hook into an actual component?

webapp/src/hooks/run.tsx
```diff
@@ -0,0 +1,173 @@
...
+    };
+};
+
+interface FollowState {
+    isFollowing: boolean;
+    followers: string[];
+    setFollowers: (followers: string[]) => void;
+}
+
+export const useFollowRun = (runID: string, followState: FollowState | undefined, trigger: 'run_details'|'playbooks_lhs'|'channel_rhs') => {
```
crspeller, line 87: Maybe this should be a component and not a hook?

crspeller (approved):
Nice! Just a couple non blockers.

webapp/src/hooks/run.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    const toggleFollow = () => {
+        const action = isFollowing ? unfollowPlaybookRun : followPlaybookRun;
+        const eventTarget = isFollowing ? PlaybookRunEventTarget.Unfollow : PlaybookRunEventTarget.Follow;
+        action(runID)
+            .then(() => {
+                const newFollowers = isFollowing ? followers.filter((userId: string) => userId !== currentUserId) : [...followers, currentUserId];
+                setFollowers(newFollowers);
+                refreshLHS();
+                telemetryEvent(eventTarget, {
```
gbochora, line 118: Nit: If we want to track click on the button, it might be better to move the telemetry call outside the `then` method, like on the line 58.

webapp/src/hooks/run.tsx
gbochora: Or we could do hook and component separately:
- hook will be responsible only for follow/unfollow functionality, and we can reuse it in other places like the LHS dot menu.
- button can reuse the follow/unfollow hook and add UI logic.

We can do it in a separate branch if this seems a good idea.

gbochora (approved):
Looks great! 🥳

abhijit-singh (approved):
Looks good @trilopin! Thank you for the recordings, they were super helpful because it was difficult to reproduce some of the cases! A couple of non-blocking requests:
1. For the follow button in the RHS header, the font should be`Open sans` and not `Metrolpolis`. Basically we should make the styling identical to the follow button on the threads RHS.
2. For the case when we have the `Participate` button in the RHS (that is when there are no participants), please add ~4px space on both sides inside the button — between the icon and the border on the left and the text and the border on the right.

trilopin:
> Looks good @trilopin! Thank you for the recordings, they were super helpful because it was difficult to reproduce some of the cases! A couple of non-blocking requests:
> 
> 1. For the follow button in the RHS header, the font should be`Open sans` and not `Metrolpolis`. Basically we should make the styling identical to the follow button on the threads RHS.

Could you confirm that follow/unfollow button is 'Open Sans' in all his versions (channels_rhs, run_detail_rhs, run_list, playbooks_lhs)?

webapp/src/components/rhs/rhs_participants.tsx
trilopin: [eslint plugin]()https://formatjs.io/docs/tooling/linter/#no-literal-string-in-jsx) seems to work excluding attribute/component types, not their content. So, it's not possible with this plugin

webapp/src/components/rhs/rhs_title.tsx
trilopin: makes completely sense, hook grew organically until made no sense to continue as a hook

PR #1487, opened 2022-09-15
MM-46975: Variables in summary and slash commands

ashishbhate:
#### Summary

- Add support for variables in the playbook summary, which are then interpolated in the slash commands

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46975

#### Checklist

- [ ] Telemetry updated
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -1531,19 +1531,34 @@ func (s *PlaybookRunServiceImpl) RunChecklistItemSlashCommand(playbookRunID, use
 		return "", errors.New("no slash command associated with this checklist item")
 	}
 
+	// parse playbook summary for variables and values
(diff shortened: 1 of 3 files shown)


ashishbhate:
1. The interpolated slash command is added to the timeline, so I've opted not to post an ephemeral message with the interpolated slash command
2. I'm not sure if this needs to be behind a feature flag
3. I'm not sure why the e2e cypress test are failing. It looks unrelated but I'm checking.

Phrynobatrachus (approved):
Cool! Works well and I haven't been able to break it in any weird ways.

iomodo (approved):
So cool, thanks @ashishbhate !

crspeller (approved):

PR #1490, opened 2022-09-21
Collapse checklists when dragging

safakkizkin:
## Summary
Whenever dragging starts on checklists, its automatically collapsed

## Ticket Link

  Fixes: https://github.com/mattermost/mattermost-server/issues/20822
  JIRA: https://mattermost.atlassian.net/browse/MM-45654

## Checklist

- [ ] ~Telemetry updated~
- [ ] ~Unit tests updated~


safakkizkin:
Newbie here,

Please let me know if I need to do same changes on `outline.tsx`.

safakkizkin:
Here is an example,

CollapseExample

safakkizkin:
Hi @lieut-data, I am totally aware of it! Best wishes for MatterCon.

crspeller:
@abhijit-singh @safakkizkin Wondering if we should re-expand the checklists after the drag is complete?

safakkizkin:
@crspeller sorry for the late responding, I was waiting for @abhijit-singh response.
 
I've a question, If we do it, do we need to do it for all checklists or just the drags?
Also whatever we do, we need to stick to the state of the checklists before the drag, right?

abhijit-singh:
@crspeller @safakkizkin Apologies, I completely missed this.

> Wondering if we should re-expand the checklists after the drag is complete?

Yes, ideally we should be reverting back to the original state after the drag is complete. If it was expanded originally, it should expand after drag. If it was collapsed originally and you drag it around, it should stay collapsed. Is that feasible?

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -262,6 +282,8 @@ const ChecklistList = ({
                 // Persist the new data in the server
                 clientMoveChecklist(playbookRun.id, srcIdx, dstIdx);
             }
+
+            restoreCollapseState(dstIdx);
```
crspeller, line 286: Looks like this only restores the checklist actually being dragged. Can we restore all of them?

safakkizkin: I quouted @abhijit-singh thought below. if its okay to go with all then I could do that too.

> Just the drags. It should not effect any other checklists apart from the one you drag.

crspeller: I don't think he meant that we should permanently collapse all the checklists except the one that we are dragging. Could be wrong though.

abhijit-singh: @safakkizkin Sorry for the ambiguity. @crspeller is correct, we should restore all of them to their original state.

lieut-data:
@safakkizkin, I see you've helped push some changes here. Are you still interested in pushing this one through? No pressure: we can close it out for now as well.

safakkizkin:
@lieut-data IMHO, the PR supports restoring the original state of all checklist after the drag ended. I'm so sorry not to take action.

lieut-data:
No worries, thanks for marking it as ready for review @safakkizkin!

calebroseland (changes requested):
Thanks, @safakkizkin! It looks like the expand/collapse state restoration is working now, but there seems to be some issues with re-ordering: 

- While reordering sections, sometimes the collapsing seems to cause the positioning to get out-of-sync with the cursor, ideally when grabbing the handle, the cursor and handle should stay in the same place relative to each other.
- Reordering checklist items causes a JS error and all sections collapse, making it impossible to reorder checklist items currently.

https://user-images.githubusercontent.com/11724372/216125119-5df6e97f-a975-4798-861f-8eb28465f266.mp4

safakkizkin:
Hi @calebroseland, I'll solve them as soon as possible. 

Do you have any suggestions for the first one? 

I could easily say that I wasn't aware of we are could reorder checklist items.

Thanks a lot!

safakkizkin:
Reordering checklist items have been solved, currently looking for a solution cursor sync problem.

safakkizkin:
Hi @calebroseland, As far as I tested, the issue currently be solved. Can you please check again? 🥺

calebroseland:
I'm sorry for the delay @safakkizkin. I've been reviewing your changes and experimenting with this over the past couple of days. I've tested your latest changes, and I'm still seeing the cursor get out-of-sync with the handle. Occasionally I've seen it re-sync while dragging, but mostly it's still out-of-sync for me. 

Also, all checklists will remain closed unless the drag-drop is completed correctly. If something goes wrong or the drag is aborted—esc. key, checklist released outside of the list, etc.—then the previous state doesn't get restored. Instead of modifying/restoring the source-of-truth state, IMO it would be better to have an  `or` condition on the `collapse` prop of `CollapsibleChecklist`—perhaps something like:

```diff
<CollapsibleChecklist
     //...
-    collapsed={Boolean(checklistsCollapseState[checklistIndex])}
+    collapsed={Boolean(checklistsCollapseState[checklistIndex]) || isAnyChecklistDragging}
```


https://user-images.githubusercontent.com/11724372/216723388-39e539c2-0ee1-4c08-b527-f763a8ca7b32.mp4


Before continuing, I want to thank you @safakkizkin for the work you've done on this and the change requests you've worked through!

TL;DR: But in summary, I think this ticket needs to go back to the drawing board. Some unfortunate unforeseen things that we failed to investigate beforehand. 

### UX Issues
cc @abhijit-singh @trilopin @m-ripper @iomodo
After some experimentation, I'm starting to think this needs to go back to the drawing board. While the scroll-shifting isn't too jarring with only a few checklists or for the first few, but the problem grows significantly as the number scales. Take a look at this example.


https://user-images.githubusercontent.com/11724372/216722959-8b299149-7356-44c4-9f42-7582557885de.mp4

(First examples: the checklist being dragged is outside of the viewport, and you're suddenly scrolled to the end of the page.)

At a certain point (`>= N checklists`)—even if the cursor/drag-handle sync issue were fixed—the scroll position is now wildly out of place and you lose context as to where you were. The only solution to this problem IMO would be a scroll-lock mechanism that would 1. detect where the cursor is, 2. keep the checklist item under it, and 3. smoothly scroll the new scroll position to this place. All of that would take a lot of very complex or sophisticated code IMO—I don't think it'd be worth it.


@abhijit-singh I'm wondering if we could design an explicit "rearrange" button that would enter a sort of edit mode:
1. collapse all checklists
2. make the entire checklist header area the drag handle
3. have a "done" button to exit rearrange mode and restore the previous expand/collapse state.

CleanShot 2023-02-03 at 16 56 20@2x

lieut-data:
Thanks, @safakkizkin & @calebroseland for the deep dive here. We're moving in a different direction with some of the Playbook features, and won't be needing this change right now. I appreciate all the effort that went into getting this far, and regret that we couldn't ship this enhancement together.

PR #1491, opened 2022-09-22
Fix styled component errors

lieut-data:
## Summary
Several styled components were being created on each render: avoid this (and squelch the corresponding console log).

## Ticket Link
N/A


webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
```diff
@@ -266,21 +266,25 @@ interface ItemProps {
     onClick?: () => void;
 }
 
+const StyledIcon = styled.div`
```
lieut-data, line 269: I think this creates an extra DOM element -- any suggested improvements?

trilopin: you could remove this one and set at OverviewitemName if you are concerned about it

```ts
const OverviewItemName = styled.div`
    display: flex;
    align-items: center;
    svg {
        margin-right: 11px;
    }
`;
```

trilopin (approved):
nice catch @lieut-data

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
calebroseland: The container is already `flex`, so I think `gap` would be the most straightforward:
```diff
 const OverviewItemName = styled.div`
     display: flex;
     align-items: center;
+    gap: 11px;
 `;
```

calebroseland (approved):

PR #1493, opened 2022-09-22
Fix a setState after unmount

lieut-data:
## Summary
Ensure debounce is cleaned up on unmount to avoid it running asynchronously after the tooltip component has been unmounted:

image

## Ticket Link
N/A

### DIFF
--- a/webapp/src/components/widgets/text_with_tooltip_when_ellipsis.tsx
+++ b/webapp/src/components/widgets/text_with_tooltip_when_ellipsis.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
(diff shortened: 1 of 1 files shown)


webapp/src/components/widgets/text_with_tooltip_when_ellipsis.tsx
```diff
@@ -19,7 +19,7 @@ const TextWithTooltipWhenEllipsis = (props: Props) => {
     const ref = useRef<HTMLElement|null>(null);
     const [showTooltip, setShowTooltip] = useState(false);
 
-    const resizeListener = useCallback(debounce(() => {
+    const resizeListener = useMemo(() => debounce(() => {
```
lieut-data, line 22: Switching to `useMemo` to avoid even re-creating the debounced callback.

gbochora (approved):

calebroseland (approved):

PR #1494, opened 2022-09-23
GH-20841 - Enable/Disable run status updates

rolwin100:
## Summary

- added apis to enable/disable status run
- added button for enabling and disabling in the ui

## Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/20841

## Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~


rolwin100:
@gbochora Pla have a look at the PR and suggest few improvements if necessary

webapp/src/types/playbook_run.ts
```diff
@@ -151,6 +151,10 @@ export function playbookRunIsActive(playbookRun: PlaybookRun): boolean {
     return playbookRun.current_status === PlaybookRunStatus.InProgress;
 }
 
+export function playbookStatusUpdateEnabled(playbookRun: PlaybookRun): boolean {
+    return playbookRun.status_update_enabled;
```
trilopin, line 155: is this function really needed? Why not using directly the field?

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective/timeline_event_item.tsx
```diff
@@ -277,6 +278,16 @@ const TimelineEventItem = (props: Props) => {
         summaryTitle = formatMessage({defaultMessage: 'Retrospective canceled by {name}'}, {name: props.event.subject_display_name});
         testid = TimelineEventType.CanceledRetrospective;
         break;
+    case TimelineEventType.StatusUpdateEnabled:
+        iconClass = 'icon icon-flag-outline';
+        summaryTitle = formatMessage({defaultMessage: 'Run status update enabled by {name}'}, {name: props.event.subject_display_name});
```
trilopin, line 283:
```suggestion
        summaryTitle = formatMessage({defaultMessage: 'Run status updates enabled by {name}'}, {name: props.event.subject_display_name});
```

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective/timeline_event_item.tsx
```diff
@@ -277,6 +278,16 @@ const TimelineEventItem = (props: Props) => {
...
+        testid = TimelineEventType.StatusUpdateEnabled;
+        break;
+    case TimelineEventType.StatusUpdateDisabled:
+        iconClass = 'icon icon-flag-outline';
+        summaryTitle = formatMessage({defaultMessage: 'Run status update disabled by {name}'}, {name: props.event.subject_display_name});
```
trilopin, line 288:
```suggestion
        summaryTitle = formatMessage({defaultMessage: 'Run status updates disabled by {name}'}, {name: props.event.subject_display_name});
```

webapp/src/components/backstage/playbook_runs/playbook_run/enable_disable_run_status_update.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const dispatch = useDispatch();
+    const {formatMessage} = useIntl();
+    const refreshLHS = useLHSRefresh();
+
+    return (status:'enable' | 'disable') => {
+        const confirmationMessage = formatMessage({defaultMessage: 'Are you sure you want to {status} status update for this run?'}, {status});
+
+        const onConfirm = async () => {
+            await runStatusUpdate(playbookRun.id, status);
+            refreshLHS();
```
trilopin, line 23: Why do we need to refreshLHS? enable/disable status updates should not change at all LHS, so I don't think we have to refresh it.

webapp/src/components/backstage/playbook_runs/playbook_run/enable_disable_run_status_update.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {makeUncontrolledConfirmModalDefinition} from 'src/components/widgets/confirmation_modal';
+import {useLHSRefresh} from '../../lhs_navigation';
+
+export const useEnableOrDisableRunStatusUpdate = (playbookRun: PlaybookRun) => {
+    const dispatch = useDispatch();
+    const {formatMessage} = useIntl();
+    const refreshLHS = useLHSRefresh();
+
+    return (status:'enable' | 'disable') => {
+        const confirmationMessage = formatMessage({defaultMessage: 'Are you sure you want to {status} status update for this run?'}, {status});
```
trilopin, line 19: This approach is correct for English but could be problematic for other languages. I think you should do two separate strings for enabling and disabling (with no variables interpolated) and then use the right one using a ternary operator.

webapp/src/components/backstage/playbook_runs/playbook_run/controls.tsx
```diff
@@ -174,3 +176,44 @@ export const RestoreRunMenuItem = (props: {playbookRun: PlaybookRun, role: Role}
...
+                        <ClockOutlineIcon size={18}/>
+                        <FormattedMessage defaultMessage='Enable status update'/>
+                    </StyledDropdownMenuItem>
+                </>
+            }
+        </>
+    );
+};
+
+export const DisableRunStatusUpdateMenuItem = (props: {playbookRun: PlaybookRun, role: Role}) => {
```
trilopin, line 200: These two components are almost the same. what if you merge them and make handle some strings with a ternary operator?

server/api/playbook_runs.go
```diff
@@ -905,6 +907,34 @@ func (h *PlaybookRunHandler) finishDialog(c *Context, w http.ResponseWriter, r *
...
+	playbookRunID := mux.Vars(r)["id"]
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	if err := h.playbookRunService.UpdatePlaybookRunStatusUpdate(playbookRunID, userID, true); err != nil {
+		h.HandleError(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte(`{"status":"OK"}`))
```
trilopin, line 920: I think that there's no need of this write operation. When needed it's better to use ReturnJSON to force setting the right headers.

server/api/playbook_runs.go
```diff
@@ -905,6 +907,34 @@ func (h *PlaybookRunHandler) finishDialog(c *Context, w http.ResponseWriter, r *
...
+	playbookRunID := mux.Vars(r)["id"]
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	if err := h.playbookRunService.UpdatePlaybookRunStatusUpdate(playbookRunID, userID, false); err != nil {
+		h.HandleError(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte(`{"status":"OK"}`))
```
trilopin, line 934: same, line not needed

server/api/playbook_runs.go
```diff
@@ -89,6 +89,8 @@ func NewPlaybookRunHandler(
 	playbookRunRouterAuthorized.HandleFunc("/update-description", withContext(handler.updateDescription)).Methods(http.MethodPut)
 	playbookRunRouterAuthorized.HandleFunc("/restore", withContext(handler.restore)).Methods(http.MethodPut)
 	playbookRunRouterAuthorized.HandleFunc("/actions", withContext(handler.updateRunActions)).Methods(http.MethodPut)
+	playbookRunRouterAuthorized.HandleFunc("/status-update/enable", handler.enableStatusUpdate).Methods(http.MethodPut)
+	playbookRunRouterAuthorized.HandleFunc("/status-update/disable", handler.disableStatusUpdate).Methods(http.MethodPut)
```
trilopin, line 93: what if we have only add one endpoint with the desired value?

server/app/playbook_run_service.go
```diff
@@ -1032,6 +1053,104 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
...
+
+	user, err := s.pluginAPI.User.Get(userID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to to resolve user %s", userID)
+	}
+
+	statusUpdate := "enabled"
+	eventType := StatusUpdateEnabled
+
+	if enable == false {
```
trilopin, line 1076:
```suggestion
	if !enable {
```

server/app/playbook_run_service.go
```diff
@@ -3208,6 +3327,8 @@ const (
 	restoreMessage             messageType = "restore"
 	retroMessage               messageType = "retrospective"
 	statusUpdateMessage        messageType = "status update"
+	statusUpdateEnabled        messageType = "status update enabled"
+	statusUpdateDisabled       messageType = "status update disabled"
```
trilopin, line 3331: these values are unused

server/app/playbook_run_service.go
```diff
@@ -1032,6 +1053,104 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
...
+	if playbookRunToModify.StatusUpdateBroadcastChannelsEnabled {
+		s.broadcastPlaybookRunMessageToChannels(playbookRunToModify.BroadcastChannelIDs, &model.Post{Message: message}, statusUpdateMessage, playbookRunToModify)
+		s.telemetry.RunAction(playbookRunToModify, userID, TriggerTypeStatusUpdatePosted, ActionTypeBroadcastChannels, len(playbookRunToModify.BroadcastChannelIDs))
+	}
+
+	runStatusUpdateMessage := s.buildStatusUpdateMessage(playbookRunToModify, user.Username, statusUpdate)
+	s.dmPostToRunFollowers(&model.Post{Message: runStatusUpdateMessage}, statusUpdateMessage, playbookRunToModify.ID, userID)
+
+	// Remove pending reminder (if any), even if current reminder was set to "none" (0 minutes)
+	s.RemoveReminder(playbookRunID)
```
trilopin, line 1099: nice! 👍🏽

server/app/playbook_run_service.go
```diff
@@ -1032,6 +1053,104 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
...
+
+	statusUpdate := "enabled"
+	eventType := StatusUpdateEnabled
+
+	if enable == false {
+		statusUpdate = "disabled"
+		eventType = StatusUpdateDisabled
+	}
+
+	message := fmt.Sprintf("@%s %s the status update for this run", user.Username, statusUpdate)
```
trilopin, line 1081: Not sure if we want to send this message, but if we want it must be a translatable string.

[example of how to use translations at server](https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/app/playbook_run_service.go#L2933)

server/app/playbook_run_service.go
```diff
@@ -1032,6 +1053,104 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
...
+
+	err = s.resetReminderTimer(playbookRunID)
+	if err != nil {
+		s.pluginAPI.Log.Warn("failed to reset the reminder timer when updating status to Archived", "playbook ID", playbookRunToModify.ID, "error", err)
+	}
+
+	// We are resolving the playbook run. Send the reminder to fill out the retrospective
+	// Also start the recurring reminder if enabled.
+	if s.licenseChecker.RetrospectiveAllowed() {
+		if playbookRunToModify.RetrospectiveEnabled && playbookRunToModify.RetrospectivePublishedAt == 0 {
```
trilopin, line 1109: I imagine that this piece comes from the finish-run code, but we shouldn't invoke any code related to retrospectives here. Enabling/Disabling status updates is independent of retrospectives.

Screenshot 2022-09-28 at 18 42 08

server/app/playbook_run_service.go
trilopin, line 1081:
```suggestion
	message := fmt.Sprintf("@%s %s the status updates for this run", user.Username, statusUpdate)
```

server/api/playbook_runs.go
trilopin, line 93: you will need the withContext wrapper to add context as the first parameter of the endpoint (latest changes in log management).
The you'll have to get from there the logger to inject it into the handleError func (second parameter)

```suggestion
	playbookRunRouterAuthorized.HandleFunc("/status-update/disable", withContext(handler.disableStatusUpdate)).Methods(http.MethodPut)
```

trilopin (changes requested):
Hey @rolwin100 ,

Thanks a lot for all the energy you put in this PR.

There are some points that need revision to be ready but it's basically remove code and make a couple of points simpler.

server/api/playbook_runs.go
rolwin100: @trilopin I had thought about this but again we need to add validation to take only enable and disable and then do conditional check which will increase slight complexity and not much line of code difference.

webapp/src/types/playbook_run.ts
rolwin100: not sure! The reason I did this because I noticed other place people have done this so it motivated me to write this function.

rolwin100: Should I remove it

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective/timeline_event_item.tsx
```diff
@@ -196,7 +196,8 @@ const TimelineEventItem = (props: Props) => {
             />
         </TimeBetween>
     );
-
+        // eslint-disable-next-line no-console
+    console.log(props.event.event_type);
```
gbochora, line 200: I'm assuming this was for debugging purposes, do we still need it?

gbochora (commented):
Thanks a lot, @rolwin100, for the huge amount of work. I've left a few comments.

webapp/src/types/playbook_run.ts
trilopin: IMHO if it is a function, I'd expect to have more complex behavior (and that's not the case). 

I would use directly the property.

server/app/playbook_run_service.go
```diff
@@ -1032,6 +1053,105 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
...
+func (s *PlaybookRunServiceImpl) UpdatePlaybookRunStatusUpdate(playbookRunID, userID string, enable bool) error {
+
+	playbookRunToModify, err := s.store.GetPlaybookRun(playbookRunID)
+	logger := logrus.WithField("playbook_run_id", playbookRunID)
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve playbook run")
+	}
+
+	updateAt := model.GetMillis()
+	if err = s.store.UpdatePlaybookStatusUpdateEnable(playbookRunID, enable); err != nil {
```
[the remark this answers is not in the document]
rolwin100: It's needed right since we are updating a single field I didn't find a function doing this also other functionality also it takes care. Please let me know your thoughts.

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective/timeline_event_item.tsx
rolwin100: sorry

server/app/playbook_run_service.go
```diff
@@ -1032,6 +1053,105 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
...
+	runStatusUpdateMessage := s.buildStatusUpdateMessage(playbookRunToModify, user.Username, statusUpdate)
+	s.dmPostToRunFollowers(&model.Post{Message: runStatusUpdateMessage}, statusUpdateMessage, playbookRunToModify.ID, userID)
+
+	// Remove pending reminder (if any), even if current reminder was set to "none" (0 minutes)
+	s.RemoveReminder(playbookRunID)
+
+	err = s.resetReminderTimer(playbookRunID)
+	if err != nil {
+		s.pluginAPI.Log.Warn("failed to reset the reminder timer when updating status to Archived", "playbook ID", playbookRunToModify.ID, "error", err)
+	}
```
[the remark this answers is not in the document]
rolwin100: can you add more detail on this what needs to be done

webapp/src/types/playbook_run.ts
rolwin100: okay will change it then

server/app/playbook_run_service.go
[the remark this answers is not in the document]
rolwin100: if you see `UpdatePlaybookRun` making a complete update of all the values and currently that function is not having a `StatusUpdateEnabled` field for an update I may need to modify that function as well so I didn't use it. Let me know your thoughts. I feel we can proceed because this function will be a bit faster than the `UpdatePlaybookRun`. @gbochora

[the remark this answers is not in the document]
rolwin100: if you notice `FinishPlaybookRun` even this function is the same I referred this and made the change

server/app/playbook_run_service.go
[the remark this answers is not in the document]
gbochora: When we want to disable status updates:
- we need to remove the scheduled reminder. 
- we don't need to reset the reminder timer because it only sets the `run.PreviousReminder's` value to zero.  `run.PreviousReminder's` value is never used when status updates are disabled, so no need to set it to zero.

We don't need to do anything when we want to enable status updates. 

So my suggestion is:
```suggestion
	if !enable {
		s.RemoveReminder(playbookRunID)
	}
```

server/api/playbook_runs.go
trilopin: There's no need for additional checks. You could create one endpoint that accepts a JSON with a bool parameter which is passed to the service function `UpdatePlaybookRunStatusUpdate`.

I see a lot of value in having less API surface with deeper endpoints.

rolwin100: okay you mean I need to create an endpoint `/status-update` and then pass ` {statusEnabled: boolean}` in the body right?

assets/i18n/en.json
```diff
@@ -100,5 +100,9 @@
   {
     "id": "app.user.run.request_update",
     "translation": "@here — @{{.Name}} requested a status update. \n"
+  },
+  {
+    "id": "app.user.run.status_update",
+    "translation": "@{{.Username}} %{{.statusUpdate}} the status update for this run"
```
gbochora, line 106: This text also could be problematic to translate into other languages. I think it would be better to have two strings:
1. "@{{.Username}} enabled the status update for this run"
2. "@{{.Username}} disabled the status update for this run"

server/app/playbook_run_service.go
```diff
@@ -981,6 +1002,95 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
...
+	s.dmPostToRunFollowers(&model.Post{Message: runStatusUpdateMessage}, statusUpdateMessage, playbookRunToModify.ID, userID)
+
+	// Remove pending reminder (if any), even if current reminder was set to "none" (0 minutes)
+	if !enable {
+		s.RemoveReminder(playbookRunID)
+	}
+
 	// We are resolving the playbook run. Send the reminder to fill out the retrospective
 	// Also start the recurring reminder if enabled.
 	if s.licenseChecker.RetrospectiveAllowed() {
```
gbochora, line 1096: I think you removed the license check code from the `FinishPlaybookRun` function instead of the `UpdatePlaybookRunStatusUpdate`

webapp/src/components/backstage/playbook_runs/playbook_run/context_menu.tsx
```diff
@@ -75,6 +75,16 @@ export const ContextMenu = ({playbookRun, hasPermanentViewerAccess, role, isFavo
...
+                <EnableDisableRunStatusUpdateMenuItem
+                    playbookRun={playbookRun}
+                    role={role}
+                    status='enable'
+                />
+                <EnableDisableRunStatusUpdateMenuItem
+                    playbookRun={playbookRun}
+                    role={role}
+                    status='disable'
+                />
```
gbochora, line 87: We can have a single menu item here that handles both enable and disable based on the current state. You can see `FollowRunMenuItem` as an example.

webapp/src/components/backstage/playbook_runs/playbook_run/controls.tsx
```diff
@@ -174,3 +176,28 @@ export const RestoreRunMenuItem = (props: {playbookRun: PlaybookRun, role: Role}
 
     return null;
 };
+
+export const EnableDisableRunStatusUpdateMenuItem = (props: {playbookRun: PlaybookRun, role: Role, status: 'enable' | 'disable'}) => {
```
gbochora, line 180: You don't need the `status` parameter here; you can rely on the `playbookRun.status_update_enabled`

rolwin100:
@trilopin @gbochora I have made all the changes suggested I guess we can call it a day guys.

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective/timeline_event_item.tsx
```diff
@@ -277,6 +277,16 @@ const TimelineEventItem = (props: Props) => {
         summaryTitle = formatMessage({defaultMessage: 'Retrospective canceled by {name}'}, {name: props.event.subject_display_name});
         testid = TimelineEventType.CanceledRetrospective;
         break;
+    case TimelineEventType.StatusUpdateEnabled:
+        iconClass = 'icon icon-flag-outline';
```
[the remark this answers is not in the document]
trilopin: maybe ClockOutlineIcon / icon-clock-outline?

webapp/src/components/backstage/playbook_runs/playbook_run/enable_disable_run_status_update.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {runStatusUpdate} from 'src/client';
+import {modals} from 'src/webapp_globals';
+import {makeUncontrolledConfirmModalDefinition} from 'src/components/widgets/confirmation_modal';
+
+export const useEnableOrDisableRunStatusUpdate = (playbookRun: PlaybookRun) => {
+    const dispatch = useDispatch();
+    const {formatMessage} = useIntl();
+
+    return (status: boolean) => {
+        const confirmTitle = status ? formatMessage({defaultMessage: 'Confirm enable status update'}) : formatMessage({defaultMessage: 'Confirm disable status update'});
```
trilopin, line 17:
```suggestion
        const confirmTitle = status ? formatMessage({defaultMessage: 'Confirm enable status updates'}) : formatMessage({defaultMessage: 'Confirm disable status updates'});
```

webapp/src/components/backstage/playbook_runs/playbook_run/enable_disable_run_status_update.tsx
```diff
@@ -0,0 +1,33 @@
...
+        const confirmationMessage = status ? formatMessage({defaultMessage: 'Are you sure you want to enable status update for this run?'}) : formatMessage({defaultMessage: 'Are you sure you want to disable status update for this run?'});
```
trilopin, line 18:
```suggestion
        const confirmationMessage = status ? formatMessage({defaultMessage: 'Are you sure you want to enable status updates for this run?'}) : formatMessage({defaultMessage: 'Are you sure you want to disable status updates for this run?'});
```

webapp/src/client.ts
```diff
@@ -288,6 +288,14 @@ export async function restoreRun(playbookRunId: string) {
     }
 }
 
+export async function runStatusUpdate(playbookRunId: string, status_enabled: boolean) {
```
trilopin, line 291: naming nit, what do you think of `toggleRunStatusUpdates` ?

server/sqlstore/playbook_run.go
```diff
@@ -587,6 +587,21 @@ func (s *playbookRunStore) FinishPlaybookRun(playbookRunID string, endAt int64)
 	return nil
 }
 
+func (s *playbookRunStore) UpdatePlaybookStatusUpdateEnable(playbookRunID string, enable bool) error {
```
trilopin, line 590: we could extend `UpdatePlaybookRun` instead of creating a new method

trilopin (changes requested):

gbochora (changes requested):
I think you have missed one of my comments from the previous review. In the `playboook_run_service.go` file, you have removed the license check(`s.licenseChecker.RetrospectiveAllowed()`) from the the`FinishPlaybookRun` function and copied it into the `ToggleStatusUpdates` function. 
We should revert this particular change, the `ToggleStatusUpdates` function should not make any license check.

gbochora (approved):
Looks great! Thanks a lot for your effort @rolwin100

rolwin100:
> Looks great! Thanks a lot for your effort @rolwin100

thank you so much

rolwin100:
https://github.com/mattermost/mattermost-plugin-playbooks/pull/1494#discussion_r991928052

@trilopin regarding this. I don't want to play around with this, the reason is simple. In the database, we are storing this field as StatusUpdateEnabled. changing it will introduce confusion for sure. Also this was already present and is being used by many other api's we may not be able to change this at this point of time.

trilopin (approved):
THanks a lot for all your time @rolwin100 , this is a really valuable addition to the Run Details Page.

rolwin100:
@amyblais this PR is safe to be merged can we close this

trilopin:
> @amyblais this PR is safe to be merged can we close this

@rolwin100 I labeled it for UX review, please wait a bit more for that. I'll merge it when it's ready.

abhijit-singh (changes requested):
Thanks @trilopin for looping me in and thank you @rolwin100 for the awesome work on this! I have a couple of change requests:
- For the confirmation modal, can we change the CTA of the button to `Enable updates` and `Disable updates` for the respective case instead of `Ok`? 
image

- In the dropdown, the icon for the enable/disable status update menu item needs to be changed to [`icon-update`](https://mattermost.github.io/compass-icons/) instead of the current clock icon. 
image

rolwin100:
@abhijit-singh I have made the changes.

abhijit-singh (approved):
Perfect! 💯 

Thank you @rolwin100!

PR #1502, opened 2022-10-03
Task Inbox

trilopin:
## Summary
Taskinbox rescued from the last hackathon:
- RHS is handled as it was done (before RDP)
- feature behind playbooks experimental features
- added e2e tests
- added telemetry
- change globalheader icon approach to use directly SVG icon (and avoid webapp exposing IconButton on window)

RHS is open above Run Details Page RHS, covering it. If you close tasks inbox RHS, RDP RHS will be shown.

This pr DOESN'T include some optimizations such as:
- handle data through graphql to reduce payload
- use global RHS 
- use CSS grid

## Ticket Link
Nope

## Checklist
- [X] Telemetry updated
- [X] Unit tests updated


server/app/playbook_run_service.go
```diff
@@ -1252,7 +1253,9 @@ func (s *PlaybookRunServiceImpl) ChangeOwner(playbookRunID, userID, ownerID stri
 	}
 
 	playbookRunToModify.OwnerUserID = ownerID
-	if err = s.store.UpdatePlaybookRun(playbookRunToModify); err != nil {
+
+	playbookRunToModify, err = s.store.UpdatePlaybookRun(playbookRunToModify)
```
trilopin, line 1257: I'm not sure where all the updatePlaybookRun changes come from. I'm going to do some spelunking to be sure...

trilopin: @lieut-data it seems to come from https://github.com/mattermost/mattermost-plugin-playbooks/pull/1502/commits/fe0215b6d971fb221e5986e1288784c7367ccb9e . Do you think is still relevant?

lieut-data:
@calebroseland & @iomodo, @trilopin and I worked together on this, so we thought we'd get a second review of this "feature branch merge".

server/app/playbook_run.go
```diff
@@ -759,7 +759,7 @@ type PlaybookRunStore interface {
 	CreatePlaybookRun(playbookRun *PlaybookRun) (*PlaybookRun, error)
 
 	// UpdatePlaybookRun updates a playbook run.
-	UpdatePlaybookRun(playbookRun *PlaybookRun) error
+	UpdatePlaybookRun(playbookRun *PlaybookRun) (*PlaybookRun, error)
```
lieut-data, line 762: For context here, `UpdatePlaybookRun` creates some new ids for checklist items, which the caller couldn't previously know unless it knew to re-request the `PlaybookRun` from the database. These changes effectively facilitate access to that information inline.

abhijit-singh:
This is awesome @trilopin! Love the transition when a task is checked off 😍 Would be so awesome to have that in all checklist views!

I have a few requests for the task inbox:
1. Can we make the run and checklist icons of the color `rgba(63, 67, 80, 0.56)`. Right now they are too dark and prominent. 
2. Can we make the font-weight of the run name 600 or semi-bold? The checklist name can remain as is. 
3. Can we add 4px padding above and below the header row that contains the run name and checklist name?
image
4. Can we change the padding for the task container to top:16px, bottom:12px? Currently it seems to be 15px and 5px. Just looking to add some white space. 
image
5. The filters button text changes to grey when you hover on it. It should stay blue. 
image
6. The first item in the filters dropdown seems to be blue. It should have the `center-channel-color`.
image

abhijit-singh (changes requested):
Added some comments above.

webapp/src/selectors.ts
```diff
@@ -261,3 +277,61 @@ export const selectTeamsIHavePermissionToMakePlaybooksOn = (state: GlobalState)
 };
 
 export const selectExperimentalFeatures = (state: GlobalState) => Boolean(globalSettings(state)?.enable_experimental_features);
+
+// Select tasks assigned to the current user, or unassigned but belonging to a run owned by the
+// current user.
+export const selectMyTasks = createSelector(
```
calebroseland, line 283: This selector (and the additional filtering in the render of task_inbox) was locking up the UI for about 700ms for me.  Just this part is ~O(N<sup>4</sup>). Could we consider using `reduce` (or `for ... of` directly on the object) to reduce the loop count, and maybe even include the user-filtering in this selector and also returning the count/metadata with that to avoid the additional loops in render?

webapp/src/selectors.ts
```diff
@@ -261,3 +277,61 @@ export const selectTeamsIHavePermissionToMakePlaybooksOn = (state: GlobalState)
...
+                    ...item,
+                    item_num: itemNum,
+                    playbook_run_id: playbookRun.id,
+                    playbook_run_name: playbookRun.name,
+                    playbook_run_owner_user_id: playbookRun.owner_user_id,
+                    playbook_run_create_at: playbookRun.create_at,
+                    checklist_title: checklist.title,
+                    checklist_num: checklistNum,
+                }))
+            )
```
calebroseland, line 308: Maybe this `flatMap` could be done inside `tast_inbox` to reduce the op time and trim the size a bit?

calebroseland (changes requested):
Only a partial review so far, and I really like a lot of it, but one blocker for me is the delay when opening task inbox.

webapp/src/selectors.ts
lieut-data: I think these are my part of hackathon efforts :p -- let me take a look at how to improve :)

lieut-data: I do see that slowdown with about 94 tasks in the list, but it doesn't appear to be related to any of the selector or rendering code:
CleanShot 2022-10-20 at 17 24 45@2x

I /think/ it might be an artifact of the checklist tasks rendering so many elements, including hidden elements for on hover effects:
CleanShot 2022-10-20 at 17 26 56@2x

I'll take a closer look at our checklist components to see if there's an opportunity to simplify.

lieut-data: Indeed, just skipping the render of the hover menu halves the execution time:
CleanShot 2022-10-20 at 17 33 16@2x

I'm inclined to address this as part of a larger performance issue, though perhaps leveraging a virtualized list in this context if we can't simplify things downstream. Thoughts?

calebroseland: I think a quick shortcut could even be just waiting until after the first render to displau all the tasks. (Open container empty instantly, then load tasks afterwards). There would still be a delay, but you could at least paint something within 100ms and avoid the `click but nothing happens for 1s` UX

abhijit-singh (commented):
Thanks @trilopin, looks good! Just one final thing — I noticed some odd wrapping for the run name/checklist name. We should probably just truncate it instead of wrapping it in two lines.  

Current:
image

Ideal: 
image

A similar pattern in channels for reference:
image

abhijit-singh:
Also, I noticed a significant delay while loading the task inbox, especially for the first time. Other than any optimization possible, I'm wondering if a loader or a skeleton screen (like we have for `insights`) would make sense here instead od showing `No tasks`?

trilopin:
> Also, I noticed a significant delay while loading the task inbox, especially for the first time. Other than any optimization possible, I'm wondering if a loader or a skeleton screen (like we have for `insights`) would make sense here instead od showing `No tasks`? 

Agree on this one, we have two challenges in perf (reduce payload and improve selectors/render) but the experience could improve with the loading skeleton while we are retrieving the data.

iomodo (approved):

trilopin:
New approach with [react-infinite-scroller](https://github.com/danbovey/react-infinite-scroller) to avoid rendering all at once

https://user-images.githubusercontent.com/4096774/199942309-4f20c156-b17c-4589-8807-98572bfc0dac.mp4

webapp/src/components/backstage/rhs/task_inbox/task_inbox.tsx
```diff
@@ -0,0 +1,330 @@
...
+    display: flex;
+    align-items: center;
+    padding: 0 30px;
+    text-align: center;
+    word-break: break-word;
+    font-size: 14px;
+`;
+
+const InfiniteScrollContainer = styled.div`
+    height: calc(100vh - 129px);
```
trilopin, line 328: not very proud of this one, suggestions welcome

lieut-data:
> New approach with [react-infinite-scroller](https://github.com/danbovey/react-infinite-scroller) to avoid rendering all at once

🤯

abhijit-singh (changes requested):
Thanks @trilopin! Looks so much better and feels much faster! 

Noticed a few things:
1. When I open the task inbox for the first time, it does not show some tasks, especially the overdue tasks which should be shown up top. They only come in after closing and opening again. 


2. (Non-blocking) There seems to be some truncation of the text even where it is not needed. When I inspected and removed the `max-width: 43%` property from the checklist name it seemed to work fine, but I could be missing something.
 

3. (Non-blocking) The task inbox RHS width seems to be behaving differently from the run details RHS in screen size ranging from around 1500px to 1600px. This looks slightly odd and is a little bit of an issue because 1536x864 is a common resolution. They should have the same width ideally.  
Screenshot 2022-11-05 at 2 30 09 AM

trilopin:
> 1. When I open the task inbox for the first time, it does not show some tasks, especially the overdue tasks which should be shown up top. They only come in after closing and opening again.

I left a couple of states without the right update-flow..Fixed

> 2. (Non-blocking) There seems to be some truncation of the text even where it is not needed. When I inspected and removed the `max-width: 43%` property from the checklist name it seemed to work fine, but I could be missing something.

I've adjusted to 50% which seems far better. It's not perfect tho because if the checklist name is very long and the playbook name is very short it will be ellipsified anyways (more edge case than before).



> 3. (Non-blocking) The task inbox RHS width seems to be behaving differently from the run details RHS in screen size ranging from around 1500px to 1600px. This looks slightly odd and is a little bit of an issue because 1536x864 is a common resolution. 

I noticed this. The only solution that comes to my mind is to jump to @calebroseland's global RHS since it's really hard to match a hardcoded media-queried absolute positioning 400/500px with the minmax fr grid style in the RDP RHS. If this is important enough, we should use the new stuff.

lieut-data:
@calebroseland / @abhijit-singh, thoughts on merging this in time to deploy live tomorrow? I'd love to be able to get it in the hands of community behind the experimental flag and iterate as needed.

abhijit-singh (approved):
Good to go ahead from my side 👍

calebroseland (approved):

PR #1505, opened 2022-10-04
Migrate upgrade_badge icon usage to compass icons.

hawkril:
## Summary
This PR removes the usage of upgrade_badge icon and uses the same icon that is already provided by the compass-icons package. The only difference between the icons is the naming and that the fill color is no longer hardcoded, but needs to be provided.

## Ticket Link
  Fixes: https://github.com/mattermost/mattermost-server/issues/21154

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


trilopin (approved):
@hawkril Thank you very much for your help

calebroseland (approved):
LGTM, thanks @hawkril!

PR #1509, opened 2022-10-09
[MM-42886] Introduce weekly digest

batebobo:
## Summary

Introduces a weekly digest, similar to the daily one. 
Extracted functions that decide whether a daily/weekly digest should be sent.
Currently users would receive weekly digest if the daily digest is enabled.

What's left:
- [ ] Introduce a enable/disable weekly digest flag
- [ ] The above flag should default to the current daily digest setting
- [ ] Add a feature flag, as this is a user-facing feature and includes refactoring
- [ ] Add more tests for corner cases around sending daily/weekly digests

## Ticket Link
[Weekly Digest](https://github.com/mattermost/mattermost-server/issues/20825)

## Checklist
- [ ] Telemetry updated
- [x] Unit tests updated

### DIFF
--- a/server/api/bot.go
+++ b/server/api/bot.go
@@ -13,6 +13,7 @@ import (
 	"github.com/mattermost/mattermost-plugin-playbooks/server/app"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/bot"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/config"
+	"github.com/mattermost/mattermost-plugin-playbooks/server/timeutils"
 	"github.com/mattermost/mattermost-server/v6/model"
 
 	pluginapi "github.com/mattermost/mattermost-plugin-api"
@@ -180,6 +181,10 @@ func (h *BotHandler) promptForFeedback(c *Context, w http.ResponseWriter, r *htt
 	w.WriteHeader(http.StatusOK)
 }
 
+type DigestSenderParams struct {
+	isWeekly bool
+}
+
 // connect handles the GET /bot/connect endpoint (a notification sent when the client wakes up or reconnects)
 func (h *BotHandler) connect(c *Context, w http.ResponseWriter, r *http.Request) {
 	userID := r.Header.Get("Mattermost-User-ID")
@@ -202,25 +207,39 @@ func (h *BotHandler) connect(c *Context, w http.ResponseWriter, r *http.Request)
 	offset, _ := strconv.Atoi(r.Header.Get("X-Timezone-Offset"))
 	timezone = time.FixedZone("local", offset*60*60)
 
-	// DM message if it's the next day and been more than an hour since the last post
-	// Hat tip to Github plugin for the logic.
-	now := model.GetMillis()
-	nt := time.Unix(now/1000, 0).In(timezone)
-	lt := time.Unix(info.LastDailyTodoDMAt/1000, 0).In(timezone)
-	if nt.Sub(lt).Hours() >= 1 && (nt.Day() != lt.Day() || nt.Month() != lt.Month() || nt.Year() != lt.Year()) {
+	sendRegularDigest := h.createDigestSender(c, w, userID, &info)
+
+	// we want to first try a weekly digest
+	// if we have already sent it this week, try with a daily one
+	currentTime := timeutils.GetCurrentUnixTime(timezone)
+	if app.ShouldSendWeeklyDigestMessage(info, timezone, currentTime) {
+		sendRegularDigest(DigestSenderParams{isWeekly: true})
+	} else if app.ShouldSendDailyDigestMessage(info, timezone, currentTime) {
+		sendRegularDigest(DigestSenderParams{isWeekly: false})
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+func (h *BotHandler) createDigestSender(c *Context, w http.ResponseWriter, userID string, userInfo *app.UserInfo) func(DigestSenderParams) {
+	return func(params DigestSenderParams) {
+		now := model.GetMillis()
 		// record that we're sending a DM now (this will prevent us trying over and over on every
 		// response if there's a failure later)
-		info.LastDailyTodoDMAt = now
-		if err = h.userInfoStore.Upsert(info); err != nil {
+		userInfo.LastDailyTodoDMAt = now
+		if err := h.userInfoStore.Upsert(*userInfo); err != nil {
 			h.HandleError(w, c.logger, err)
 			return
 		}
 
-		if err = h.playbookRunService.DMTodoDigestToUser(userID, false); err != nil {
-			h.HandleError(w, c.logger, errors.Wrapf(err, "failed to DMTodoDigest to userID '%s'", userID))
+		regulartity := "daily"
+		if params.isWeekly {
+			regulartity = "weekly"
+		}
+
+		if err := h.playbookRunService.DMTodoDigestToUser(userID, params.isWeekly); err != nil {
+			h.HandleError(w, c.logger, errors.Wrapf(err, "failed to send '%s' DMTodoDigest to userID '%s'", regulartity, userID))
 			return
 		}
 	}
-
-	w.WriteHeader(http.StatusOK)
 }
--- a/server/app/regular_digest_service.go
+++ b/server/app/regular_digest_service.go
@@ -1,0 +1,31 @@
+package app
+
+import (
+	"time"
+
+	"github.com/mattermost/mattermost-plugin-playbooks/server/timeutils"
+)
+
+func ShouldSendWeeklyDigestMessage(userInfo UserInfo, timezone *time.Location, currentTime time.Time) bool {
+	lastSentTime := timeutils.GetUnixTimeForTimezone(userInfo.LastDailyTodoDMAt, timezone)
+
+	currentYear, currentWeek := currentTime.ISOWeek()
+	lastSentYear, lastSentWeek := lastSentTime.ISOWeek()
+	isFirstLoginOfTheWeek := currentYear != lastSentYear || currentWeek != lastSentWeek
+
+	return ShouldSendDailyDigestMessage(userInfo, timezone, currentTime) && isFirstLoginOfTheWeek
+}
+
+func ShouldSendDailyDigestMessage(userInfo UserInfo, timezone *time.Location, currentTime time.Time) bool {
+	// DM message if it's the next day and been more than an hour since the last post
+	// Hat tip to Github plugin for the logic.
+	lastSentTime := timeutils.GetUnixTimeForTimezone(userInfo.LastDailyTodoDMAt, timezone)
+
+	isMoreThanOneHourPassed := currentTime.Sub(lastSentTime).Hours() >= 1
+
+	isDifferentDay := currentTime.Day() != lastSentTime.Day() ||
+		currentTime.Month() != lastSentTime.Month() ||
+		currentTime.Year() != lastSentTime.Year()
+
+	return isMoreThanOneHourPassed && isDifferentDay
+}
--- a/server/timeutils/timeutils.go
+++ b/server/timeutils/timeutils.go
@@ -69,3 +69,11 @@ func GetDaysDiff(start, end time.Time) int {
 	}
 	return days
 }
+
+func GetUnixTimeForTimezone(currentTime int64, timezone *time.Location) time.Time {
+	return time.Unix(currentTime/1000, 0).In(timezone)
+}
+
+func GetCurrentUnixTime(timezone *time.Location) time.Time {
+	return GetUnixTimeForTimezone(model.GetMillis(), timezone)
+}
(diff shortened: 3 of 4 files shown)


server/api/bot.go
```diff
@@ -202,25 +207,39 @@ func (h *BotHandler) connect(c *Context, w http.ResponseWriter, r *http.Request)
 	offset, _ := strconv.Atoi(r.Header.Get("X-Timezone-Offset"))
 	timezone = time.FixedZone("local", offset*60*60)
 
-	// DM message if it's the next day and been more than an hour since the last post
-	// Hat tip to Github plugin for the logic.
-	now := model.GetMillis()
-	nt := time.Unix(now/1000, 0).In(timezone)
-	lt := time.Unix(info.LastDailyTodoDMAt/1000, 0).In(timezone)
-	if nt.Sub(lt).Hours() >= 1 && (nt.Day() != lt.Day() || nt.Month() != lt.Month() || nt.Year() != lt.Year()) {
+	sendRegularDigest := h.createDigestSender(c, w, userID, &info)
```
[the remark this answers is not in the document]
batebobo: I extracted it purely for readability purposes and not repeating common arguments to method calls.
Does the alternative you suggested look like that?
```go
	if app.ShouldSendWeeklyDigestMessage(info, timezone, currentTime) {
		sendRegularDigest(c, w, userID, &info, true)
	} else if app.ShouldSendDailyDigestMessage(info, timezone, currentTime) {
		sendRegularDigest(c, w, userID, &info, false)
	}
```

server/timeutils/timeutils.go
```diff
@@ -69,3 +69,11 @@ func GetDaysDiff(start, end time.Time) int {
 	}
 	return days
 }
+
+func GetUnixTimeForTimezone(currentTime int64, timezone *time.Location) time.Time {
```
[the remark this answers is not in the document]
batebobo: Yeah, you're right. I'm thinking of something like `currentTimeMillis` for the parameter name and a godoc. Does this sound alright?

server/api/bot.go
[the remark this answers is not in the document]
NikolaVuAstea: Unrelated, but isn't `ShouldSendDailyDigest` descriptive enough?
Also, what's the reason behind passing `currentTime` as a parameter? Maybe testing convenience?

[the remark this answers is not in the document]
batebobo: I'm extra verbose with the names for my own convenience and understanding :)
The current time is indeed a testing convenience

lieut-data:
Hey @batebobo! I'd love to have @crspeller take a look at this review -- should we wait until it leaves draft status?

batebobo:
Hi @lieut-data! Yesterday I re-requested a review from @mkraft and GitHub automatically removed @crspeller :( I'd also like another person to review the task!

lieut-data:
No worries! That's weird behaviour on GitHub's part -- fixed :)

server/api/bot.go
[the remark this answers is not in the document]
mkraft: > I extracted it purely for readability purposes and not repeating common arguments to method calls

I have no strong objection to this, I just wanted to understand your thinking. My opinion is that a bit of repetition is OK to avoid having to grok the extra function, but I'm not right, it's just an opinion.

mkraft (approved):

server/api/bot.go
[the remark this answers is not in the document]
batebobo: Thanks for sharing and explaining! (I also learned what "grok" means :D) Let's also hear what @crspeller thinks :pray:

server/timeutils/timeutils.go
```diff
@@ -69,3 +69,12 @@ func GetDaysDiff(start, end time.Time) int {
 	}
 	return days
 }
+
+// GetUnixTimeForTimezone returns the unix time for a given timezone and a time in milliseconds
```
crspeller, line 73: Are these helpers really necessary?

server/sqlstore/user_info.go
```diff
@@ -97,6 +97,11 @@ func toUserInfo(rawUserInfo sqlUserInfo) (app.UserInfo, error) {
 		return userInfo, errors.Wrapf(err, "failed to unmarshal DigestNotificationSettings for userid: %s", userInfo.ID)
 	}
 
+	// if weekly digest is not yet configured, we use the daily digest setting as a default
+	if userInfo.DigestNotificationSettings.DisableWeeklyDigest == nil {
```
crspeller, line 101: Would it maybe be better to do a DB migration rather than keeping this nil check here?

server/api/bot.go
```diff
@@ -202,25 +207,39 @@ func (h *BotHandler) connect(c *Context, w http.ResponseWriter, r *http.Request)
...
 		}
 
-		if err = h.playbookRunService.DMTodoDigestToUser(userID, false); err != nil {
-			h.HandleError(w, c.logger, errors.Wrapf(err, "failed to DMTodoDigest to userID '%s'", userID))
+		regulartity := "daily"
+		if params.isWeekly {
+			regulartity = "weekly"
+		}
+
+		if err := h.playbookRunService.DMTodoDigestToUser(userID, params.isWeekly); err != nil {
```
crspeller, line 240: So the only difference between daily and weekly is that we force it here?

crspeller (changes requested):

server/timeutils/timeutils.go
batebobo: I don't have a strong opinion here. I extracted them as the first helper was used twice and makes sense as a `timeutil`.

server/sqlstore/user_info.go
batebobo: Yeah, you're right. I did this having in mind the difference between `not yet configured` (nil) and `explicitly configured` (true/false), but according to the spec and this discussion we don't need that difference and could just add a new json field that has the daily digest setting as initial value. Is this what you had in mind?

server/api/bot.go
batebobo: According to my understanding of the specifications, yes. With the `force` flag, we achieve the same as `/playbooks todo` without any filtering.
It makes sense to verify this with tests though.

server/sqlstore/user_info.go
crspeller: Yes, we don't need to know if it wasn't configured since we can set the correct value automatically.

server/api/bot.go
crspeller: Ah I see the force flag does more than it seems. Maybe we can update the name and comment to be more clear.

server/timeutils/timeutils.go
batebobo: Alright, thanks for explaining :) I'm new to the Go ecosystem and community and I'll keep that in mind for the future. Will also remove these two.

What does 1/5 mean?

server/api/bot.go
batebobo: I'd only add a comment as the `force` semantics is still the dominant one IMO. It just happens that it matches the `weekly digest` one. That said I can't figure a name that captures both `force` and `weekly digest`. The meaning of the flag in the code is such that it would send a full message with unfiltered data even if nothing is "actionable". Let me know if you have any suggestions though!

batebobo: Oof, I've forgotten that we shouldn't really `force` the weekly digest. We should still send the fullest possible information (identical to `/playbooks todo`), but if the list is empty, we should skip the message. I've added a separate flag in the DM sending functions that tells whether we should also send `runsInProgress` (I'm wondering whether I should comment this in the code as well). It should behave like this:
daily digest - force: false, includeRunsInProgress: false
weekly digest - force: false, includeRunsInProgress: true
/playbooks todo - force: true, includeRunsInProgress: true

server/sqlstore/migrations/mysql/000049_add_weekly_digest_setting.down.sql
```diff
@@ -0,0 +1,3 @@
+
+update IR_UserInfo
```
crspeller, line 2: I think you also need to add these migrations to https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/sqlstore/migrations.go as we are not done migrating to the new system.

crspeller (changes requested):
One last request.

server/sqlstore/migrations/mysql/000049_add_weekly_digest_setting.down.sql
batebobo: Thanks! I added them to `migrations.go` and to the `future` folder as well. Also included a test in `migrations_test.go`. This helped me find some issues with the scripts, which should now be resolved

batebobo:
Thanks for re-requesting @Phrynobatrachus! I didn't mean to remove you, but rather to re-request review from @crspeller :(

Phrynobatrachus:
No worries, it's a Github bug I've seen a few times recently.

crspeller (approved):
Great!

server/command/command.go
```diff
@@ -842,14 +854,18 @@ func (r *Runner) actionSettings(args []string) {
 	settingsHelpText := "###### Playbooks Personal Settings - Slash Command Help\n" +
 		"* `/playbook settings` - display current settings. \n" +
 		"* `/playbook settings digest on` - turn daily digest on. \n" +
-		"* `/playbook settings digest off` - turn daily digest off."
+		"* `/playbook settings digest off` - turn daily digest off. \n" +
+		"* `/playbook settings weekly-digest on` - turn weekly digest on. \n" +
```
Phrynobatrachus, line 858: Perhaps should explain that the weekly setting is dependent on the daily setting.

Phrynobatrachus (changes requested):
I'm having trouble triggering the weekly digest. Here's my test setup:

Screenshot from 2022-10-31 16-28-30


Screenshot from 2022-10-31 17-04-44

1 active run with an assigned task due today, where I've manually set the `lastdailytododmat: 767227425358` seems to give me a daily digest, not a weekly digest. 

Screenshot from 2022-10-31 17-08-44

Am I missing something?

server/command/command.go
batebobo: Hey @Phrynobatrachus! It's not really dependent on the daily setting when reconfiguring. Only the initial value (the default) for each user is taken from the daily digest. I should've updated the PR description, sorry. Do you still think we need to explain this in the help texts?

batebobo:
Hey @Phrynobatrachus! Thanks for the comment and the screenshots! It seems that the timestamp is from a while ago (~30 years), so it definitely doesn't make sense to receive a daily digest. I'll take a look and let you know

batebobo:
Hey again @Phrynobatrachus! There was still code left to change with regard to the difference between daily digest, weekly digest and /playbook todo. I pushsed the necessary changes and tested it with the following cases:
* Didn't send any digest because of no data to show
* /playbook todo still showed everything (even with 0s all around)
* Added 5 playbooks and 1 run with tasks assigned to me. Weekly digest was sent (all sections, even the ones with 0s)
* Daily digest was sent (not showing the 0 overdue tasks)
* /playbook todo still works with all the data

Let me know if anything else comes up :)

Phrynobatrachus:
Apologies for the delay, once we sort out 

https://github.com/mattermost/mattermost-plugin-playbooks/pull/1509#discussion_r1016979222

this should be good, everything else appears to be working after the updates.

batebobo:
Hi @Phrynobatrachus! Thanks for the feedback! I'll check the case with the daily/weekly flags later today and let you know :)

server/command/command.go
batebobo: Yup, that's true. I used to have the functions that determine whether we should send a digest separate from the configuration, but this changed at some point, but not everywhere, it seems. It should be alright now.

server/app/regular_digest_service.go
```diff
@@ -15,13 +15,12 @@ func ShouldSendWeeklyDigestMessage(userInfo UserInfo, timezone *time.Location, c
...
-	return ShouldSendDailyDigestMessage(userInfo, timezone, currentTime) && isFirstLoginOfTheWeek
+	return !userInfo.DigestNotificationSettings.DisableDailyDigest &&
+		ShouldSendDailyDigestMessage(userInfo, timezone, currentTime) &&
+		isFirstLoginOfTheWeek
 }
 
 func ShouldSendDailyDigestMessage(userInfo UserInfo, timezone *time.Location, currentTime time.Time) bool {
-	if userInfo.DigestNotificationSettings.DisableDailyDigest {
-		return false
-	}
```
Phrynobatrachus: This doesn't seem right, I think the only change needed was to remove the daily check in `ShouldSendWeekly...` and just `return isFirstLoginOfTheWeek` there.

batebobo: Yeah, makes sense, thanks!

Phrynobatrachus (approved):
Thank you for following through with the review cycle, nice contribution! This is ready to merge @crspeller

mkraft:
It looks like another pull request won the migration race for the name 0.61.

PR #1510, opened 2022-10-11
e2e: Fix recent failures

Phrynobatrachus:
Failures from these specs were due to tests interacting with certain elements while the DOM was unstable. The fixes rely on more accurately describing expected page state (`should(...)`) and/or `wait()`ing on a common request responsible for most of the data used on the page, prior to the problematic interactions.

### DIFF
--- a/tests-e2e/cypress/integration/lhs_spec.js
+++ b/tests-e2e/cypress/integration/lhs_spec.js
@@ -181,7 +181,7 @@ describe('lhs', () => {
 
             // # The assertions here guard against the click() on 194
             // # happening on a detached element.
-            cy.assertBackstageRenderComplete(testUser.username);
+            cy.assertRunDetailsPageRenderComplete(testUser.username);
             cy.findByTestId('runinfo-following').should('be.visible').within(() => {
                 // # Verify follower icon
                 cy.findAllByTestId('profile-option', {exact: false}).should('have.length', 1);
--- a/tests-e2e/cypress/integration/lhs_spec.js
+++ b/tests-e2e/cypress/integration/lhs_spec.js
@@ -126,16 +126,12 @@ describe('lhs', () => {
 
                 // # Intercept these graphQL requests for wait()'s
                 // # that help ensure rendering has finished.
-                cy.intercept('/plugins/playbooks/api/v0/query', (req) => {
-                    if (req.body?.operationName === 'PlaybookLHS') {
-                        req.alias = 'gqlUpdateLHS';
-                    }
-                });
+                cy.gqlInterceptQuery('PlaybookLHS');
             });
         });
 
         it('shows on click', () => {
-            cy.wait('@gqlUpdateLHS').wait('@gqlUpdateLHS');
+            cy.wait('@gqlPlaybookLHS').wait('@gqlPlaybookLHS');
 
             // # Click dot menu
             cy.findByTestId('Runs')
@@ -148,7 +144,7 @@ describe('lhs', () => {
         });
 
         it('can copy link', () => {
-            cy.wait('@gqlUpdateLHS').wait('@gqlUpdateLHS');
+            cy.wait('@gqlPlaybookLHS').wait('@gqlPlaybookLHS');
             stubClipboard().as('clipboard');
 
             // # Click on Copy link menu item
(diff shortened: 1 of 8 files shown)


tests-e2e/cypress/integration/lhs_spec.js
```diff
@@ -116,20 +124,31 @@ describe('lhs', () => {
                 // # Visit the playbook run
                 cy.visit(`/playbooks/runs/${playbookRun.id}`);
 
-                // # LHS render takes a few seconds, wait for it
-                cy.wait(5000);
+                // # Intercept these graphQL requests for wait()'s
+                // # that help ensure rendering has finished.
+                cy.intercept('/plugins/playbooks/api/v0/query', (req) => {
+                    if (req.body?.operationName === 'PlaybookLHS') {
+                        req.alias = 'gqlUpdateLHS';
```
trilopin, line 131: would make sense to do `req.alias = req.body.operationName` and make this a command?

just a thought, probably is too early to generalize

tests-e2e/cypress/integration/lhs_spec.js
```diff
@@ -52,8 +52,14 @@ describe('lhs', () => {
     });
 
     const getRunDropdownItemByText = (groupName, runName, itemName) => {
-        cy.findByTestId(groupName).findByTestId(runName).findByTestId('menuButton').click();
-        return cy.findByTestId('dropdownmenu').findByText(itemName);
+        cy.findByTestId(groupName).should('exist')
+            .findByTestId(runName).should('exist')
+            .findByTestId('menuButton')
```
trilopin, line 57: shouldn't this be enclosed in a within clause?

I'm confused about whether pure chaining sets the scope or within is required. I first thought that chaining-finds would take the parent's scope but then I realize that doesn't happen.

in this case, wouldn't any menubutton be found? would be found just the one behind runName?

tests-e2e/cypress/integration/lhs_spec.js
Phrynobatrachus: Yep, if this approach seems solid I'd like to go clean up other simple timeout `wait` calls, and can do that alongside. I also found https://github.com/bahmutov/cypress-network-idle which could be useful, as opposed to relying on a specific number of requests - in this case 2, with the `wait(...).wait(...)` calls - to happen on load.

tests-e2e/cypress/integration/lhs_spec.js
Phrynobatrachus: I realized I did use `within` in a few place where it isn't necessary, namely

```
cy.findByTestId('lhs-navigation').within((() => {
    cy.contains(playbookRun.name).should('be.visible');
}));
```
I'll change those, sorry for the extra confusion!

Phrynobatrachus:
Ok I think this should be good to merge now.

tests-e2e/cypress/integration/runs/rdp_main_statusupdate_spec.js
```diff
@@ -260,8 +285,10 @@ describe('runs > run details page > status update', () => {
         });
 
         it('requests an update and cancel', () => {
-            // # Click on request update
-            cy.findByTestId('run-statusupdate-section').findByText('Request update...').click();
+            // # Click post update
```
trilopin, line 288:
```suggestion
            // # Click request update
```

tests-e2e/cypress/integration/runs/rdp_main_statusupdate_spec.js
```diff
@@ -150,8 +158,12 @@ describe('runs > run details page > status update', () => {
                 // # Click on kebab menu
                 cy.findByTestId('run-statusupdate-section').getStyledComponent('Kebab').click();
 
-                // # Click on request update
-                cy.findByText('Request update...').click();
+                cy.findByTestId('dropdownmenu').within(($dropdown) => {
+                    cy.wrap($dropdown).children().should('have.length', 2);
```
trilopin, line 162: didn't know about wrap 👀

trilopin (approved):
Nice work on removing flakiness (and I'm sure that this will be faster)

I'd love to have a small recommendation list about:
- more reliable selectors 
- wait for interception vs wait for fixed time
- asserting visibility before click

I understand why you did those. However, I'd want to understand if we can convert that into general recommendations or if they are only the solutions to very specific problems.

PR #1511, opened 2022-10-11
MM-47454: Introduce a highly visible feedback button

lieut-data:
## Summary
The feedback button is an `a` that links out to a permalink, eventually landing on a Google Form. The current server version is passed as a query string parameter, though I note it's a Google Form specific parameter, meaning if we migrate to some other platform, we'll need to support this particular name.

image

CleanShot 2022-10-11 at 18 13 45

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-47454


webapp/src/components/assets/buttons.tsx
```diff
@@ -16,14 +16,24 @@ export const Button = styled.button`
...
     position: relative;
+    justify-content: center;
+
+    transition: all 0.15s ease-out;
 
     &:hover{
         background: rgba(var(--center-channel-color-rgb), 0.12);
     }
 
+    &&, &&:focus {
```
lieut-data, line 29: Many of the CSS rules here are to allow me to use the `as="a"` parameter, rendering a link instead of a button, and saving the `onClick` handling and complexities of launching a popup. Overriding the default `a` styling then becomes a juggle, such as the `not([disabled])`.

Feedback welcome!

webapp/src/components/global_header_right.tsx
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
lieut-data, line 1: This component is lifted out of @trilopin's pending changes for task inbox -- should be "minimal" merge conflicts.

crspeller (approved):
Couple non-blocking.

webapp/src/components/assets/buttons.tsx
lieut-data: Entirely because of the global `.app_body` CSS rules:
image

webapp/src/components/give_feedback_button.tsx
```diff
@@ -0,0 +1,57 @@
...
+            {formatMessage({defaultMessage: 'Give feedback about Playbooks.'})}
+        </Tooltip>
+    );
+
+    const giveFeedbackURL = new URL('https://mattermost.com/pl/playbooks-feedback');
+    const searchParams = giveFeedbackURL.searchParams;
+
+    // Set the 'Server Version' field on the linked Google Form. (If we move to a different
+    // form platform, we'll need to preserve backwards compatibility.)
+    searchParams.set('entry.796122893', serverVersion);
```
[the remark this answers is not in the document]
lieut-data: Indeed. Focalboard managed this by just routing to their custom focalboard.com domain. For expediency, I'm relying on the WordPress mattermost.com permalink system, and aiming to keep things simple there.

lieut-data:
CI failure looks to be spurious. I'm going to merge this given the overall success in passing, and may revisit the approach to CSS later. (Ideally, I'd just use the common components for these buttons eventually :))

PR #1514, opened 2022-10-12
reduce GraphQL error logs

lieut-data:
## Summary
I was seeing a rash of permission logs on community from the GraphQL endpoints, that essentially looked like these two logs:
```
{
  "timestamp": "2022-10-12 15:03:58.990 -03:00",
  "level": "warn",
  "msg": "Not authorized",
  "caller": "app/plugin_api.go:979",
  "plugin_id": "playbooks",
  "error": "does not have permissions\ngithub.com/mattermost/mattermost-plugin-playbooks/server/app.init\n\tgithub.com/mattermost/mattermost-plugin-playbooks/server/app/permissions_service.go:15\nruntime.doInit\n\truntime/proc.go:6321\nruntime.d
oInit\n\truntime/proc.go:6298\nruntime.doInit\n\truntime/proc.go:6298\nruntime.main\n\truntime/proc.go:233\nruntime.goexit\n\truntime/asm_arm64.s:1165",
  "request_id": "h4f5h3up7jfoi8saxikzpiypch",
  "plugin_caller": "github.com/mattermost/mattermost-plugin-playbooks/server/api/graphql_root_playbook.go:56"
}
{
  "timestamp": "2022-10-12 15:03:58.997 -03:00",
  "level": "error",
  "msg": "Error executing request",
  "caller": "app/plugin_api.go:976",
  "plugin_id": "playbooks",
  "request_id": "h4f5h3up7jfoi8saxikzpiypch",
  "error": "graphql: Not authorized",
  "operation": "PlaybookLHS",
  "plugin_caller": "github.com/mattermost/mattermost-plugin-playbooks/server/api/graphql.go:146"
}
```

This PR changes this into the single warning log:
```
{
  "timestamp": "2022-10-12 15:05:54.229 -03:00",
  "level": "warn",
  "msg": "Warning executing request",
  "caller": "app/plugin_api.go:979",
  "plugin_id": "playbooks",
  "request_id": "bk8gu5qa83noicwboo9yzgrmir",
  "error": "graphql: userID `ts3e4e98npggzkkj53s4t6cgrr` is not on team `abc`: does not have permissions",
  "operation": "PlaybookLHS",
  "plugin_caller": "github.com/mattermost/mattermost-plugin-playbooks/server/api/graphql.go:149"
}
```

### DIFF
--- a/server/api/graphql.go
+++ b/server/api/graphql.go
@@ -143,7 +143,13 @@ func (h *GraphQLHandler) graphQL(c *Context, w http.ResponseWriter, r *http.Requ
 	r.Header.Set("X-GQL-Operation", params.OperationName)
 
 	for _, err := range response.Errors {
-		c.logger.WithError(err).WithField("operation", params.OperationName).Error("Error executing request")
+		errLogger := c.logger.WithError(err).WithField("operation", params.OperationName)
+
+		if errors.Is(err, app.ErrNoPermissions) {
+			errLogger.Warn("Warning executing request")
+		} else {
+			errLogger.Error("Error executing request")
+		}
 	}
 
 	if err := json.NewEncoder(w).Encode(response); err != nil {
--- a/server/api/graphql_root_playbook.go
+++ b/server/api/graphql_root_playbook.go
@@ -134,7 +134,7 @@ func (r *PlaybookRootResolver) UpdatePlaybook(ctx context.Context, args struct {
 	}
 
 	if err := c.permissions.PlaybookManageProperties(userID, currentPlaybook); err != nil {
-		return "", errors.Wrapf(err, "permissions check failed for playbook `%s`", args.ID)
+		return "", err
 	}
 
 	if currentPlaybook.DeleteAt != 0 {
--- a/server/api/graphql_root_playbook.go
+++ b/server/api/graphql_root_playbook.go
@@ -133,12 +133,12 @@ func (r *PlaybookRootResolver) UpdatePlaybook(ctx context.Context, args struct {
 		return "", err
 	}
 
-	if currentPlaybook.DeleteAt != 0 {
-		return "", errors.New("archived playbooks can not be modified")
+	if err := c.permissions.PlaybookManageProperties(userID, currentPlaybook); err != nil {
+		return "", errors.Wrapf(err, "permissions check failed for playbook `%s`", args.ID)
 	}
 
-	if err := c.permissions.PlaybookManageProperties(userID, currentPlaybook); err != nil {
-		return "", err
+	if currentPlaybook.DeleteAt != 0 {
+		return "", errors.New("archived playbooks can not be modified")
 	}
 
 	setmap := map[string]interface{}{}
@@ -147,11 +147,11 @@ func (r *PlaybookRootResolver) UpdatePlaybook(ctx context.Context, args struct {
 	if args.Updates.Public != nil {
 		if *args.Updates.Public {
 			if err := c.permissions.PlaybookMakePublic(userID, currentPlaybook); err != nil {
-				return "", errors.Wrap(err, "attempted to make playbook public without permissions")
+				return "", err
 			}
 		} else {
 			if err := c.permissions.PlaybookMakePrivate(userID, currentPlaybook); err != nil {
-				return "", errors.Wrap(err, "attempted to make playbook private without permissions")
+				return "", err
 			}
 		}
 		if !c.licenceChecker.PlaybookAllowed(*args.Updates.Public) {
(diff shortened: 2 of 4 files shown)


server/api/graphql.go
```diff
@@ -143,7 +143,13 @@
 	r.Header.Set("X-GQL-Operation", params.OperationName)
 
 	for _, err := range response.Errors {
-		c.logger.WithError(err).WithField("operation", params.OperationName).Error("Error executing request")
+		errLogger := c.logger.WithError(err).WithField("operation", params.OperationName)
```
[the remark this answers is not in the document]
lieut-data: @srkgupta, can you help investigate? I'm /intending/ to log user input here, so not sure why it's complaining.

server/api/graphql_root_playbook.go
```diff
@@ -25,8 +25,7 @@ func (r *PlaybookRootResolver) Playbook(ctx context.Context, args struct {
 	userID := c.r.Header.Get("Mattermost-User-ID")
 
 	if err := c.permissions.PlaybookView(userID, playbookID); err != nil {
-		c.logger.WithError(err).Warn("Not authorized")
-		return nil, errors.New("Not authorized")
+		return nil, err
```
trilopin, line 28: Maybe https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/api/graphql_root_run.go#L25 needs to be simplified too

trilopin: Should https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/api/graphql_root_run.go#L180 and https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/api/graphql_root_run.go#L216 just return the original err?

trilopin (approved):
Nice catch @lieut-data 

I left some comments of additional points that could be also taken into account.

server/api/graphql_root_playbook.go
lieut-data: Ooh, good catch re: graphql_root_run.go! Fixed.

re: original err, I'm less inclined to strip these given we're already wrapping (so can expect sane semantics from `errors.Is`). One benefit to the wrapping is the extra context to disambiguate which of the permission steps failed, as right now I think the permission checks aren't consistently verbose. Thoughts?

crspeller (approved):

server/api/graphql_root_playbook.go
trilopin: > Ooh, good catch re: graphql_root_run.go! Fixed.
> 
> re: original err, I'm less inclined to strip these given we're already wrapping (so can expect sane semantics from `errors.Is`). One benefit to the wrapping is the extra context to disambiguate which of the permission steps failed, as right now I think the permission checks aren't consistently verbose. Thoughts?

I'm fine keeping them or simplifying them, no strong opinion here. There are different benefits in both approaches: simpler-code vs more-accurate-message but I would try to use one approach for all of them.

lieut-data: Sounds good: I've pushed another change that attempts to do this. Let me know your thoughts?

server/api/graphql_root_playbook.go
```diff
@@ -135,12 +133,12 @@ func (r *PlaybookRootResolver) UpdatePlaybook(ctx context.Context, args struct {
 		return "", err
 	}
 
-	if currentPlaybook.DeleteAt != 0 {
-		return "", errors.New("archived playbooks can not be modified")
+	if err := c.permissions.PlaybookManageProperties(userID, currentPlaybook); err != nil {
```
lieut-data, line 136: Note that I've also inverted the order here, so as to avoid leaking information about a playbook to which a user doesn't have access.

server/api/graphql_root_playbook.go
```diff
@@ -288,12 +286,12 @@ func (r *PlaybookRootResolver) AddPlaybookMember(ctx context.Context, args struc
 		return "", err
 	}
 
-	if currentPlaybook.DeleteAt != 0 {
-		return "", errors.New("archived playbooks can not be modified")
+	if err := c.permissions.PlaybookManageMembers(userID, currentPlaybook); err != nil {
```
lieut-data, line 289: Same reordering as above.

server/app/permissions_service.go
```diff
@@ -299,7 +299,7 @@ func (p *PermissionsService) PlaybookManageMembers(userID string, playbook Playb
 		return nil
 	}
 
-	return ErrNoPermissions
+	return errors.Wrapf(ErrNoPermissions, "user `%s` does not have permission to manage members for playbook `%s`", userID, playbook.ID)
```
lieut-data, line 302: My strategy has been to wrap at the permission layer with enough detail, and allow layers above to mostly just `return nil` unless they have something else to clarify.

server/api/graphql_root_playbook.go
```diff
@@ -135,12 +133,12 @@ func (r *PlaybookRootResolver) UpdatePlaybook(ctx context.Context, args struct {
 		return "", err
 	}
 
-	if currentPlaybook.DeleteAt != 0 {
-		return "", errors.New("archived playbooks can not be modified")
+	if err := c.permissions.PlaybookManageProperties(userID, currentPlaybook); err != nil {
+		return "", errors.Wrapf(err, "permissions check failed for playbook `%s`", args.ID)
```
trilopin, line 137: why in this case is different than `return "", err`?

server/app/permissions_service.go
trilopin: You took the best of both approaches, with more accuracy in the messages and more simplicity in the resolvers. Very nice!

server/api/graphql_root_playbook.go
lieut-data: Oops, just missed that! Thank you :)

PR #1515, opened 2022-10-13
MM-47558: fix a crash re: status update posts

lieut-data:
## Summary
If a status update is permalinked and posted into another channel, and a user views that permalink without being a member of the run channel, the application crashes due to an attempt to resolve channel metadata.

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-47558

PR #1516, opened 2022-10-14
add telemetry testing to post_udpate + rely on post internal props

trilopin:
## Summary

Follow up https://github.com/mattermost/mattermost-plugin-playbooks/pull/1515

- Send empty string when some extra props are undefined
- avoid relying on selectors whose triggers are not ready 
- add telemetry for 3 cases of post_udpate render (run channel + permalink scenarios)

## Ticket Link
No

## Checklist
- [X] Telemetry updated
- [X] Unit tests updated

### DIFF
--- a/webapp/src/components/update_post.tsx
+++ b/webapp/src/components/update_post.tsx
@@ -54,9 +54,8 @@ export const UpdatePost = (props: Props) => {
     const runName = props.post.props.runName ?? '';
     useViewTelemetry(PlaybookRunViewTarget.StatusUpdate, props.post.id, {
         post_id: props.post.id,
-        playbook_id: currentRun?.playbook_id, // not always available
-        channel_type: channel?.type, // not always available
-        playbook_run_id: currentRun?.id, // not always available
+        channel_type: channel?.type || '', // not always available
+        playbook_run_id: props.post.props.playbookRunId || '',
     });
 
     return (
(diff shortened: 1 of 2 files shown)


webapp/src/components/update_post.tsx
```diff
@@ -54,9 +54,8 @@ export const UpdatePost = (props: Props) => {
     const runName = props.post.props.runName ?? '';
     useViewTelemetry(PlaybookRunViewTarget.StatusUpdate, props.post.id, {
         post_id: props.post.id,
-        playbook_id: currentRun?.playbook_id, // not always available
```
lieut-data, line 57: To confirm, not interested in the playbook?

lieut-data (approved):
Nice follow up, @trilopin! Thanks :)

webapp/src/components/update_post.tsx
trilopin: I would want it but it's too complicated. 

We are currently reporting it as undefined even when you are at the Run's channel because curerntRun is not ready when telemetry is called (it'll be ready in the next render).

I can not force waiting for the currentRun to trigger the telemetry because in some cases there is not currentRun (permalink in other channels)

trilopin: The best way to have it is to inject it into the post.props and then, rely on post data to report to telemetry.

crspeller (approved):

PR #1517, opened 2022-10-17
[MM-46382] Make convert to private playbook more prominent

cannalee90:
## Summary

- Add convert to playbook private menu in the dropdown
- Create custom hook for converting playbook private 
- Apply custom hook for playbook access modal for converting private

## Ticket Link

Fixes: https://github.com/mattermost/mattermost-server/issues/21184
JIRA: https://mattermost.atlassian.net/browse/MM-46382

## Checklist
- [x] Telemetry updated
- [x] Unit tests updated(everything except formatted_duration.test.tsx because of i18n issue)

### DIFF
--- a/webapp/src/components/backstage/convert_private_playbook_modal.tsx
+++ b/webapp/src/components/backstage/convert_private_playbook_modal.tsx
@@ -4,7 +4,13 @@ import {useIntl} from 'react-intl';
 import ConfirmModal from 'src/components/widgets/confirmation_modal';
 import {useEditPlaybook} from 'src/hooks';
 
-const useConfirmPlaybookConvertPrivateModal = (playbookId: string, refetch: () => void) => {
+type ConfirmPlaybookConvertPrivateReturn = [React.ReactNode, (show: boolean) => void];
+type Props = {
+    playbookId: string,
+    refetch?: () => void | undefined,
+}
+
+const useConfirmPlaybookConvertPrivateModal = ({playbookId, refetch}: Props): ConfirmPlaybookConvertPrivateReturn => {
     const {formatMessage} = useIntl();
     const [showMakePrivateConfirm, setShowMakePrivateConfirm] = useState(false);
     const [playbook, updatePlaybook] = useEditPlaybook(playbookId, refetch);
--- a/webapp/src/components/backstage/convert_private_playbook_modal.tsx
+++ b/webapp/src/components/backstage/convert_private_playbook_modal.tsx
@@ -1,0 +1,30 @@
+import React, {useState} from 'react';
+import {useIntl} from 'react-intl';
+
+import ConfirmModal from 'src/components/widgets/confirmation_modal';
+import {useEditPlaybook} from 'src/hooks';
+
+const useConfirmPlaybookConvertPrivateModal = (playbookId: string, refetch: () => void) => {
+    const {formatMessage} = useIntl();
+    const [showMakePrivateConfirm, setShowMakePrivateConfirm] = useState(false);
+    const [playbook, updatePlaybook] = useEditPlaybook(playbookId, refetch);
+
+    const modal = (
+        <ConfirmModal
+            show={showMakePrivateConfirm}
+            title={formatMessage({defaultMessage: 'Convert to Private playbook'})}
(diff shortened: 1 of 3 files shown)


cannalee90:
gonna fix it asap

webapp/src/components/backstage/convert_private_playbook_modal.tsx
```diff
@@ -0,0 +1,36 @@
...
+import ConfirmModal from 'src/components/widgets/confirmation_modal';
+import {useEditPlaybook} from 'src/hooks';
+
+type ConfirmPlaybookConvertPrivateReturn = [React.ReactNode, (show: boolean) => void];
+type Props = {
+    playbookId: string,
+    refetch?: () => void | undefined,
+}
+
+const useConfirmPlaybookConvertPrivateModal = ({playbookId, refetch}: Props): ConfirmPlaybookConvertPrivateReturn => {
```
crspeller, line 13: I like that you pulled this out from the access modal. Could we update the usage there?

webapp/src/components/backstage/playbook_editor/controls.tsx
```diff
@@ -440,6 +445,18 @@ const TitleMenuImpl = ({playbook, children, className, editTitle, refetch}: Titl
                     <ExportVariantIcon size={18}/>
                     <FormattedMessage defaultMessage='Export'/>
                 </DropdownMenuItemStyled>
+                {isEligibleToMakePrivate && (
+                    <DropdownMenuItemStyled
+                        role={'button'}
+                        css={`${iconSplitStyling}`}
+                        onClick={() => {
+                            telemetryEventForPlaybook(playbook.id, 'playbook_makeprivate');
```
crspeller, line 453: I think we forgot to actually open the modal.

webapp/src/components/backstage/playbook_editor/controls.tsx
```diff
@@ -440,6 +445,18 @@ const TitleMenuImpl = ({playbook, children, className, editTitle, refetch}: Titl
...
+                        }}
+                    >
+                        <FormattedMessage defaultMessage='Convert to private playbook'/>
```
crspeller, line 456: We probably want an icon here. @abhijit-singh Can you provide an icon recommendation?

crspeller (changes requested):
Thanks for the contribution!

webapp/src/components/backstage/convert_private_playbook_modal.tsx
cannalee90: So are you suggesting to make change to use this hook  for access modal?

webapp/src/components/backstage/playbook_editor/controls.tsx
cannalee90: I just fix it

webapp/src/components/backstage/convert_private_playbook_modal.tsx
crspeller: Exactly, so we are using the same code in both places.

webapp/src/components/backstage/playbook_editor/controls.tsx
abhijit-singh: The lock-outline icon should work here 
image

cannalee90: I added icon-lock-outline icon!

cannalee90:
@crspeller Hi. As you suggested I just apply my custom hook for playbook access modal but there is a problem. If I convert it to private access modal, the private link I just clicked not disappeared even though lock-icon appeared in background.

The reason why it happed it that the playbook object that used in access-modal is not same as the one that I used in the custom hook. so I just passed the updater function to solve this problem. If you have any other idea please give me a hint for that

Thank you!

crspeller (approved):
@cannalee90 That's fine for now. Thanks!

lieut-data (approved):
Super slick! Thanks, @cannalee90 :)

abhijit-singh (approved):
Looks good, thanks @cannalee90!

jordanafung:
Hi @cannalee90 
You were selected as one of our Top Contributors for Hacktoberfest!
Please fill out [this form](https://docs.google.com/forms/d/11pLB3pF1CmOslYqHr7c7bxFWU6QAEAJuiaKgHD1vJtg) to select your swag as thanks!
Thank you for all you do and for making Mattermost even better!

Issue #1507, opened 2022-10-06
UI for actions syncronizing channel members with run participants


trilopin:
Actions for synchronizing should be exposed in the UI.

- [x] When a participant joins a run --> Add them to linked channel
- [x] When a participant leaves a run --> Remove the from the linked channel

[Figma](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=275%3A273275)

Not as actions for now:
- When a member joins the run channel --> Add them to the linked run
- When a member leaves the run channel --> Remove them from the linked run

PR #1518, opened 2022-10-17
Gh 1507 pb run actions joinleave

trilopin:
## Summary
Add actions to playbook and run to:
- create a channel member when a new participant joins the run
- remove a channel member when a existent participant leaves the run

Remove run participant creation/removal on channel join/leave events. This is a major change since we relied a lot on visiting public channels to join public runs (e2e).

Note that:
- runs `options.ParticipantOrFollowerID` sqlstore function has been updated to get the participant from the ir_run_participants table instead of the channel members.
- there was a problem that didn't redirect the user to the error page when trying to access a non-existent/forbidden Run (blank page instead of redir to error page).

### migrations
Two new fields have been created in playbooks and run table.
- Default values are the same as the current behavior (true=create/remove channel members)
- Migrations in morph style added to `future` folder

### UI
CleanShot 2022-10-21 at 12 43 24

At playbook editor side:
- added two new triggers
- moved channel triggers to a modal

At run detail page side:
- added two new triggers

### E2E
Many tests relied on adding users to the channel to let them join the run. Now that's changed.

New command `cy.apiAddUsersToRun(runID, userIDs)` to replace `cy.apiAddUserToChannel`.

### graphql
Updated playbook and run graphql schemas to handle the two new fields 

This task involves the migration of updateRunActions to existent graphql updateRun mutation.
Key points:
- old endpoint, service and test functions removed
- managed permissions RunMageProperties in updateRun endpoint unless the user changes only isFavorite (isFavorite should be a different mutation IMHO)
- same strategy as playbook with setmap to do the update, except that a step through the service has been forced to handle side tasks (like real time through WS). Even if it's just wrapper I would keep it and avoid resolver<->store direct calls. 
- tests done in graphql server e2e

### telemetry
- update run action event moved from server to frontend

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1507

## Checklist

- [ ] ~~Telemetry updated~~
- [X] Unit tests updated


server/main_test.go
```diff
@@ -169,7 +170,6 @@ func Setup(t *testing.T) *TestEnvironment {
 
 	options := []sapp.Option{
 		sapp.ConfigStore(configStore),
-		sapp.SetLogger(testLogger),
```
trilopin, line 172: don't know why, but no-logger was back in the code

if you want to disable the e2e logging you should do `config.LogSettings.EnableConsole = model.NewBool(false)`in the code above instead of setting a no-logger

server/app/playbook_run_service.go
```diff
@@ -344,6 +344,16 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
...
 
+	// add owner as user
+	err = s.AddParticipants(playbookRun.ID, []string{playbookRun.OwnerUserID}, playbookRun.OwnerUserID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to add owner as a participant")
+	}
+	err = s.Follow(playbookRun.ID, playbookRun.OwnerUserID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to make owner follow run")
+	}
```
trilopin, line 355: this code was done before by joining channel and being subscribed to userJoinedChannel hook. Since we removed the hook, we have to manually add the owner to the list of followers and participants

server/api/graphql_root_run.go
```diff
@@ -78,28 +78,68 @@ func (r *RunRootResolver) Runs(ctx context.Context, args struct {
...
+
 	playbookRun, err := c.playbookRunService.GetPlaybookRun(args.ID)
 	if err != nil {
 		return "", err
 	}
 
-	// Enough permissions to do a fav/unfav, check if future ops need RunManageProperties
-	if err := c.permissions.RunView(userID, playbookRun.ID); err != nil {
-		return "", err
+	// scalar updates
```
crspeller, line 110: Nice! 😄

tests-e2e/cypress/support/plugin/api_commands.js
```diff
@@ -423,6 +423,33 @@ Cypress.Commands.add('apiUnfollowPlaybookRun', (playbookRunId) => {
...
+        runID: playbookRunId,
+        userIDs: usersIds,
+    };
+    return doGraphqlQuery(query, 'AddRunParticipants', vars).then((response) => {
+        expect(response.status).to.equal(StatusOK);
+        cy.wrap(response.body);
+    });
+});
+
+const doGraphqlQuery = (query, operationName, variables) => {
```
crspeller, line 443: Helpful!

webapp/src/components/actions_modal_action.tsx
```diff
@@ -28,14 +29,14 @@ const Action = (props: Props) => {
...
+                    disabled={!props.editable}
                     isChecked={props.enabled}
                     onChange={() => {/* do nothing, clicking logic lives in Container's onClick */}}
-                    disabled={!props.editable}
                 />
             </Container>
             {props.enabled &&
             <ChildrenContainer>
-                {props.children}
+                {props.children ?? null}
```
crspeller, line 39: Does it barf on undefined? Why do we need the `?? null`?

webapp/src/components/playbook_actions_modal.tsx
```diff
@@ -0,0 +1,103 @@
...
+            setWelcomeMessageEnabled(false);
+        }
+        if (categoryName === '' && categorizeChannelEnabled) {
+            setCategorizeChannelEnabled(false);
+        }
+        updatePlaybook({
+            categoryName: categoryName ?? '',
+            categorizeChannelEnabled,
+            messageOnJoin: welcomeMessage,
+            messageOnJoinEnabled: welcomeMessageEnabled,
```
crspeller, line 50: Do these values update immediately? I think you need to have local updated copies of these.

server/plugin.go
```diff
@@ -310,18 +310,9 @@ func (p *Plugin) UserHasJoinedChannel(c *plugin.Context, channelMember *model.Ch
 	if actor != nil && actor.Id != channelMember.UserId {
 		actorID = actor.Id
 	}
-	p.playbookRunService.UserHasJoinedChannel(channelMember.UserId, channelMember.ChannelId, actorID)
```
crspeller, line 313: We should remove these functions from the service as well.

server/app/playbook_run_service.go
```diff
@@ -344,6 +344,16 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
 		return nil, errors.Wrap(err, "failed to add users to playbook run channel")
 	}
 
+	// add owner as user
+	err = s.AddParticipants(playbookRun.ID, []string{playbookRun.OwnerUserID}, playbookRun.OwnerUserID)
```
crspeller, line 348: These actually get added above in `addPlaybookRunUsers` I updated the name of that function to make a bit more sense in my PR: https://github.com/mattermost/mattermost-plugin-playbooks/pull/1521

crspeller (changes requested):
Nice work! I think we should get a @Phrynobatrachus pass on this. What do you think?

webapp/src/components/playbook_actions_modal.tsx
trilopin: > Nice work! I think we should get a @Phrynobatrachus pass on this. What do you think?

Agreed, I wanted to do a pure code review first but I do want @Phrynobatrachus to help here. 💯

server/app/playbook_run_service.go
trilopin: the code you mention adds botUserID and RunOwner to the channel, but not the run. Since we broke that promotion channel->run we have to explicitly add them.

Should I move the new code I did inside the function to hold the member-creation for bot&owner in all needed places?

trilopin:
> I think we should get a @Phrynobatrachus pass on this. What do you think?


Agreed, I wanted to do a pure code review first but I do want @Phrynobatrachus to help here. 💯

webapp/src/components/actions_modal_action.tsx
trilopin: simplified, it wasn't needed

webapp/src/components/playbook_actions_modal.tsx
trilopin: they update when modal is confirmed, onSave() internal handler

Local copies are saved in internal state

webapp/src/components/playbook_actions_modal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const [categorizeChannelEnabled, setCategorizeChannelEnabled] = useState(playbook.categorize_channel_enabled);
+    const [welcomeMessage, setWelcomeMessage] = useState(playbook.message_on_join);
+    const [categoryName, setCategoryName] = useState(playbook.category_name);
+    const archived = playbook.delete_at !== 0;
+
+    const onHide = () => {
+        setWelcomeMessageEnabled(playbook.message_on_join_enabled);
+        setWelcomeMessage(playbook.message_on_join);
+        dispatch(hidePlaybookActionsModal());
+    };
```
gbochora, line 37: Should we add here setting `categoryName` and `categorizeChannelEnabled` states?

webapp/src/components/run_actions_modal.tsx
```diff
@@ -31,9 +30,13 @@ const RunActionsModal = ({playbookRun, readOnly}: Props) => {
...
 
+    const [createChannelMemberEnabled, setCreateChannelMemberEnabled] = useState(playbookRun.create_channel_member_on_new_participant);
+    const [removeChannelMemberEnabled, setRemoveChannelMemberEnabled] = useState(playbookRun.remove_channel_member_on_removed_participant);
+
     const [channelIds, setChannelIds] = useState(playbookRun.broadcast_channel_ids);
     const [webhooks, setWebhooks] = useState(playbookRun.webhook_on_status_update_urls);
     const [isValid, setIsValid] = useState<boolean>(true);
+    const updateRun = useUpdateRun(playbookRun.id);
 
     const onHide = () => {
```
gbochora, line 41: The same is here. It would be better to reset `createChannelMemberEnabled` and `removeChannelMemberEnabled` states on hide.

webapp/src/components/playbook_actions_modal.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const onSave = () => {
+        if (welcomeMessage === '' && welcomeMessageEnabled) {
+            setWelcomeMessageEnabled(false);
+        }
+        if (categoryName === '' && categorizeChannelEnabled) {
+            setCategorizeChannelEnabled(false);
+        }
```
gbochora, line 45: I think these state changes will not be applied immediately, and `updatePlaybook` will be called with old values.

gbochora (approved):
Great work @trilopin! Looks good, I just left a few non-blocking comments.

webapp/src/components/playbook_actions_modal.tsx
trilopin: kept local state conditions + handled in var assignment

server/sqlstore/playbook.go
```diff
@@ -335,6 +341,7 @@ func (p *playbookStore) GetPlaybooks() ([]app.Playbook, error) {
 	defer p.store.finalizeTransaction(tx)
 
 	var playbooks []app.Playbook
+	// TOOD add new actions to numactions
```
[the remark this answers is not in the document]
trilopin: nope, good catch

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -1066,8 +1065,8 @@ describe('playbooks > edit', () => {
                     });
                 });
 
-                it('persists the category even if the toggle is off', () => {
-                    cy.get('#user-joins-channel-categorize').within(() => {
+                it.skip('persists the category even if the toggle is off', () => {
```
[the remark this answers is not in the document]
trilopin: oops, I forgot to remove the skip and fix this one. Thanks for noticing it.

trilopin:
merge blocked until https://github.com/mattermost/mattermost-plugin-playbooks/pull/1521 is merged (owner&invite-action management)

server/app/playbook_run_service.go
crspeller: Of course! 😵‍💫
0/5 but that section is a bit confusing.

Phrynobatrachus (approved):
Awesome work, thanks for all of the test updates!

gbochora (approved):
Looks good! 🥳

server/app/playbook_run_service.go
trilopin: I've renamed the method (again, sorry) and added the code for the Run membership.

abhijit-singh (approved):
Looks awesome @trilopin! One small request — The icon in the `Setup channel actions` button in the playbook outline seems too large. It should be 16px x 16 px and the button height should be 40 px as per the designs.

PR #1520, opened 2022-10-19
Read only view, channel members who are not participants

### LINKED ISSUE #1441: Add RHS screen for channel has a run but you can't access it
If the playbook is private, you will not have access to runs however you can still be in a run channel without being a participant. RHS screen is needed to show this state. 

[Figma Designs](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=37%3A125625)

### LINKED ISSUE #1442: Read only view for channel members who are not run participants.
If a user is a member of the run channel but not a participant there needs to be a read only view.

If you attempt any interaction with the RHS a toast should pop up saying that you need to join.

Note that there are separate tickets for the follow button and participate icon. 

[Figma Designs](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=275%3A279713)

crspeller:
## Summary
Adds a read-only view for channel members of a run channel who are not participants in the run.

Read only view:
image

Clicking on:
- Summary
- Owner dropdown
- Status update button
- Finish run button

Causes a toast to appear:
image

Clicking participate causes the participate first dialog to appear:
image

Note that is is currently impossible to be in the channel without being a participant.

Note for reviewers there is some refactoring of the toast code and and the checklist code. They are in separate commits. 

## Ticket Link

Fixes #1442

### DIFF
--- a/webapp/i18n/en.json
+++ b/webapp/i18n/en.json
@@ -441,6 +441,7 @@
   "q/Qo8l": "Private playbooks are only available in Mattermost Enterprise",
   "q48ca7": "Give feedback about Playbooks.",
   "q6f8x9": "Change since last update",
+  "qDxsQH": "Become a participant to interact with this run",
   "qGlwfc": "Start run",
   "qyJtWy": "Show less",
   "rDvvQs": "{completed, number} / {total, number} done",
--- a/webapp/src/components/backstage/follow_button.tsx
+++ b/webapp/src/components/backstage/follow_button.tsx
@@ -14,7 +14,8 @@ import {
     telemetryEvent,
 } from 'src/client';
 import {PlaybookRunEventTarget} from 'src/types/telemetry';
-import {ToastType, useToaster} from 'src/components/backstage/toast_banner';
+import {useToaster} from 'src/components/backstage/toast_banner';
+import {ToastStyle} from 'src/components/backstage/toast';
 import Tooltip from 'src/components/widgets/tooltip';
 import {useLHSRefresh} from 'src/components/backstage/lhs_navigation';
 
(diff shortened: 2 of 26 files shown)


trilopin (approved):
Very nice!, two non block questions:

- would make sense to trigger the toast when trying to check a task?
- should we wrap toast in a debounce to avoid several toasts from being added (video attached)

https://user-images.githubusercontent.com/4096774/197177178-57bc9566-93d4-4829-8753-3dbfdcf0acc5.mp4

trilopin:
Sometimes UI makes a weird effect with the scroll when moving from viewer to participant.

CleanShot 2022-10-21 at 12 53 39

crspeller:
> would it make sense to trigger the toast when trying to check a task?

0/5 The checklist already has a disabled state. When you disable a checkbox you can't get click events from it so I just left it as is. We could do something fancy here with a fake disabled state if we really want to.

> should we wrap toast trigger calls in a debounce to avoid several toasts from being added

Aww, but who doesn't want to watch that cascade of participate toasts? Will fix. 

> Sometimes UI makes a weird effect with the scroll when moving from viewer to participant.

Huh, you actually click cancel there so it's not when actually participating. I can reproduce. It seems to have something to do with the toasts. If you close the modal while the toast is animating, it causes it to happen. If you close the modal quick before the toast starts to animate or wait until it's done, then it doesn't happen.

gbochora (approved):
Looks great! 🎉

abhijit-singh (approved):
Looks great @crspeller! A few requests: 
1. In the toast to become a participant that pops up, the close icon does not seem to be working for me. It should be clickable and dismiss the toast. 
2. In the toast to become a participant that pops up, can we make the font-weight:600 for the participate button?
3. I do think we should trigger the toast when the user tries to check off a task as that is a key interaction in the RHS. I'd be okay to do that as a follow up though, doesn't have to be done in this PR.

crspeller:
@abhijit-singh Fixed 1 and 2. Made a ticket for 3. 

@trilopin Worked around the strange issue. It's hard to reproduce. I believe it's caused by a conflict between the modal and the transitioning toasts. If you close the modal while toasts are transitioning it's an issue. So I made the toast instantly close when the button is pressed and reduced the transition timeout.

PR #1521, opened 2022-10-19
Modify the invite members flow to invite participants

### LINKED ISSUE #1416: Modify existing participation flows
Participation currently means joining the linked channel. Places where users participate need to be modified so they participate rather than join the channel.

Known locations (search for more): 
- [x] Run detail page participate button [Figma Designs](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=92%3A118276)
- [ ] Playbook action to "invite members" to the channel should instead "invite participants"

crspeller:
## Summary
Simple change so that the invite members action becomes an invite participants action.
New action:
image

## Ticket Link
Fixes #1416


iomodo (commented):
Should we change `Members` to `Participants` in the selector as well?

server/app/playbook_run_service.go
```diff
@@ -388,7 +388,7 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
 			continue
 		}
 
-		_, err = s.pluginAPI.Channel.AddUser(playbookRun.ChannelID, userID, s.configService.GetConfiguration().BotUserID)
+		err := s.AddParticipants(playbookRun.ID, []string{userID}, s.configService.GetConfiguration().BotUserID)
```
trilopin, line 391: with this flow, we would add them only to the run if the action toggle is off. 

is that what we want? 
1/5 on this one, should we avoid creating channel members for participants created on run creation because of a flag described as "When a new participant joins the run"?

trilopin: would make sense to?
- participants invited from playbook -> add to channel and run
- new participants added after creating the run -> respect the toggle


Both scenarios (this one and yours) seem like a bit surprising to me...

gbochora (approved):

trilopin (approved):

abhijit-singh (approved):
Looks good. One suggestion based on the discussion — Let's change `members` to `people` and be consistent with channels. And when a few people are selected we can just have that label as `selected`.
image
Also, we can get rid of the `Search for people` heading in the dropdown that opens if it's not a big change. 

Channels reference:
Screenshot 2022-10-25 at 2 46 53 PM

crspeller:
@abhijit-singh Updated experience:
image

PR #1527, opened 2022-10-26
Mm 47677 morph v0.21.0 v0.25.0

gbochora:
## Summary
This PR adds morph migrations for v0.21.0-v0.25.0. 
NOTE: The migration v0.21.0-v0.22.0 is a little complex, it was written using go code, and now it's done using only SQL. The rest migrations are trivial.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-47677

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


crspeller (approved):

server/sqlstore/migrations/mysql/000049_add_last_status_update_at_to_ir_incident.up.sql
```diff
@@ -0,0 +1,25 @@
...
+PREPARE addColumnIfNotExists FROM @preparedStatement;
+EXECUTE addColumnIfNotExists;
+DEALLOCATE PREPARE addColumnIfNotExists;
+
+UPDATE IR_Incident as dest, 
+(
+    SELECT i.ID as ID, COALESCE(MAX(p.CreateAt), i.CreateAt) as LastStatusUpdateAt
+    FROM IR_Incident as i
+    LEFT JOIN IR_StatusPosts as sp on i.ID = sp.IncidentID
+    LEFT JOIN Posts as p on sp.PostID = p.Id
```
trilopin, line 21: I imagine there's no difference because you handle the coalesce+group but, wouldn't the Post join be INNER instead of LEFT?

trilopin (approved):
Cool!

server/sqlstore/migrations/mysql/000049_add_last_status_update_at_to_ir_incident.up.sql
gbochora: Aha, I think INNER join will have the same result, but I took this select from the existing migration and didn't think about changing it.

PR #1531, opened 2022-10-27
Decrease the noise of the daily digest message

shinnlok:
## Summary
Currently, we send a daily digest if a user has:

Tasks that are due today or are overdue
Tasks with no due date
After introducing the weekly digest message, we can decrease the noise of the daily digest by notifying users only if they have tasks due today or overdue. The weekly digest message will have full information, including tasks with no due date, so they will not be forgotten.

## Ticket Link
Fixes: https://github.com/mattermost/mattermost-server/issues/21503

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


lieut-data:
Oops, I'm sorry I missed this one @shinnlok! Thanks for the contribution :)

Issue #1417, opened 2022-08-16
Add ability to automatically add participants to the channel when inviting them


crspeller:
When inviting users to participate you may want to automatically add them to the run channel at the same time. Add a checkbox that will automatically invite the added users to the linked run channel if the automatic action is not enabled. If the automatic action is enabled it should display a message stating what will happen. 

[Figma Designs](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=135%3A218949)

PR #1534, opened 2022-10-28
1417 add participants automatically

gbochora:
## Summary
Add the ability to automatically add participants to the channel when inviting them.

https://user-images.githubusercontent.com/4368372/198704724-af0e18e0-c21d-4fd1-a0ac-0b7e9c53c713.mp4

## Ticket Link
https://github.com/mattermost/mattermost-plugin-playbooks/issues/1417

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/runs/rdp_rhs_participants_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_rhs_participants_spec.js
@@ -55,7 +55,7 @@ describe('runs > run details page > rhs > participants', () => {
         cy.apiRunPlaybook({
             teamId: testTeam.id,
             playbookId: testPublicPlaybook.id,
-            playbookRunName: 'the-run-name',
+            playbookRunName: 'the-run-name' + Date.now(),
             ownerUserId: testUser.id,
         }).then((playbookRun) => {
             testRun = playbookRun;
@@ -216,6 +216,7 @@ describe('runs > run details page > rhs > participants', () => {
 
                 // * Verify that no users were invited
                 cy.getFirstPostId().then((id) => {
+                    cy.contains('Someone').should('not.exist');
                     cy.get(`#postMessageText_${id}`)
                         .contains(`@${testViewerUser.username}`);
                 });
--- a/tests-e2e/cypress/integration/runs/rdp_rhs_participants_spec.js
+++ b/tests-e2e/cypress/integration/runs/rdp_rhs_participants_spec.js
@@ -55,7 +55,7 @@ describe('runs > run details page > rhs > participants', () => {
         cy.apiRunPlaybook({
             teamId: testTeam.id,
             playbookId: testPublicPlaybook.id,
-            playbookRunName: 'the run name',
+            playbookRunName: 'the-run-name',
             ownerUserId: testUser.id,
         }).then((playbookRun) => {
             testRun = playbookRun;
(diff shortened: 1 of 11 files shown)


webapp/src/components/backstage/playbook_runs/playbook_run/add_participant_modal.tsx
```diff
@@ -32,8 +34,10 @@ const AddParticipantsModal = ({playbookRun, id, title, show, hideModal}: Props)
     const dispatch = useDispatch();
     const [profiles, setProfiles] = useState<UserProfile[]>([]);
     const {addToRun} = useManageRunMembership(playbookRun.id);
-    const [addToChannel, setAddToChannel] = useState(false);
+    const [forceAddToChannel, setForceAddToChannel] = useState(false);
+    const [channel, meta] = useChannel(playbookRun.channel_id);
     const isChannelMember = useSelector(isCurrentUserChannelMember(playbookRun.channel_id));
+    const isPrivateChannelWithAccess = meta.error === null && channel?.type === General.PRIVATE_CHANNEL;
```
gbochora, line 40: `isChannelMember` is only valid when the channel is public; we don't have membership information for private channels. So, I'm checking if the current user can access the channel to handle the private channel case.

webapp/src/graphql/runs.graphql
```diff
@@ -12,8 +12,8 @@ mutation UpdateRun($id: String!, $updates: RunUpdates!) {
   updateRun(id: $id, updates: $updates)
 }
 
-mutation AddRunParticipants($runID: String!, $userIDs: [String!]!) {
-	addRunParticipants(runID: $runID, userIDs: $userIDs)
+mutation AddRunParticipants($runID: String!, $userIDs: [String!]!, $forceAddToChannel: Boolean = false) {
```
trilopin, line 15: if here is already defaulted to false, do we need to handle in the root resolver as a *bool? Could we simplify root resolved setting it as bool?

tests-e2e/cypress/integration/runs/rdp_rhs_participants_spec.js
```diff
@@ -122,25 +122,104 @@ describe('runs > run details page > rhs > participants', () => {
...
+                    cy.get(`#postMessageText_${id}`).within(() => {
+                        // just to wait until the username is fetched
+                        cy.contains('Someone').should('not.exist');
+                        cy.contains('You were added to the channel by @playbooks.');
+                        cy.contains(`@${testViewerUser.username}`).should('not.exist');
+                    });
+                });
+            });
+
+            it('join action disabled, checkbox selected', () => {
```
trilopin, line 189: this tests is failing when I run locally

trilopin (approved):
Code is fine and feature works!

Some non-blocking points:
- should we carry in the telemetry whether addParticipants imply channel members or not (taking into account the 3 possible values no/action/forced). Maybe this is more for your effort @crspeller 
- could we simplify root resolver making force flag not nullable (it already has a default false value)

webapp/src/graphql/runs.graphql
gbochora: I've tried that, and graphql fails to initialize:
`"error": "field "ForceAddToChannel": bool is not a pointer or a nullable type.`

I'll try one more time, maybe I'm missing something.

abhijit-singh (approved):
Looks good!

abhijit-singh:
@gbochora One small non-blocking request— Can we make the font-weight of the checkbox text "_Also add people to the channel linked to this run_" `normal` or `400` instead of the current `700`?

crspeller (approved):

PR #1535, opened 2022-10-28
[WIP] Archive and unarchive a playbook run

pradeepmurugesan:
## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


lieut-data:
Thanks again for this contribution, @pradeepmurugesan! We've since resolved this a different way.

PR #1536, opened 2022-11-01
[MM-47542] Add collections and topics support to playbooks

iomodo:
## Summary
PR as a part of threads everywhere feature adds support for collections and topics in playbooks.
Registers `run` collection type, also `task` and `status` topic types
Adds hooks
`UserHasPermissionToCollection`
`GetAllCollectionIDsForUser`
`GetAllUserIdsForCollection`
`GetCollectionMetadataByIds`
`GetTopicMetadataByIds`

This PR depends on https://github.com/mattermost/mattermost-server/pull/21446, so do not pay attention to the failed checks. Will update the `.mod` file after server PR will be merged

## Ticket Link

  Fixes: https://mattermost.atlassian.net/browse/MM-47542

## Checklist

- ~~[ ] Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/server/app/permissions_service.go
+++ b/server/app/permissions_service.go
@@ -90,7 +90,7 @@ func (p *PermissionsService) hasPermissionsToPlaybook(userID string, playbook Pl
 	return p.pluginAPI.User.HasPermissionToTeam(userID, playbook.TeamID, permission)
 }
 
-func (p *PermissionsService) hasPermissionsToRun(userID string, run *PlaybookRun, permission *model.Permission) bool {
+func (p *PermissionsService) HasPermissionsToRun(userID string, run *PlaybookRun, permission *model.Permission) bool {
 	// Check at run level
 	if p.pluginAPI.User.RolesGrantPermission(p.getRunRole(userID, run), permission.Id) {
 		return true
@@ -406,7 +406,7 @@ func (p *PermissionsService) RunManageProperties(userID, runID string) error {
 		return errors.Wrapf(err, "Unable to get run to determine permissions, run id `%s`", runID)
 	}
 
-	if p.hasPermissionsToRun(userID, run, model.PermissionRunManageProperties) {
+	if p.HasPermissionsToRun(userID, run, model.PermissionRunManageProperties) {
 		return nil
 	}
 
--- a/server/config/configuration.go
+++ b/server/config/configuration.go
@@ -31,6 +31,9 @@ type Configuration struct {
 
 	// AdminLogVerbose: set to include full context with admin log messages.
 	AdminLogVerbose bool
+
+	// ThreadsEverywhereEnabled is a feature flag on a server
+	ThreadsEverywhereEnabled bool
 }
 
 // Clone shallow copies the configuration. Your implementation may require a deep copy if
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -30,6 +30,11 @@ const (
 	updateMetricsTaskFrequency = 15 * time.Minute
 
 	metricsExposePort = ":9093"
+
+	RunCollectionType = "run"
+
+	StatusTopicType = "status"
+	TaskTopicType   = "task"
 )
 
 // These credentials for Rudder need to be populated at build-time,
@@ -69,6 +74,8 @@ type Plugin struct {
 	telemetryClient      TelemetryClient
 	licenseChecker       app.LicenseChecker
 	metricsService       *metrics.Metrics
+
+	collectionAndTopicTypes map[string][]string
 }
 
 type StatusRecorder struct {
@@ -217,6 +224,16 @@ func (p *Plugin) OnActivate() error {
 	mutex.Unlock()
 
 	p.permissions = app.NewPermissionsService(p.playbookService, p.playbookRunService, pluginAPIClient, p.config, p.licenseChecker)
+
+	p.collectionAndTopicTypes[RunCollectionType] = []string{StatusTopicType, TaskTopicType}
+	// register collections and topics
+	for collectionType, topicTypes := range p.collectionAndTopicTypes {
+		for _, topicType := range topicTypes {
+			if err := p.API.RegisterCollectionAndTopic(collectionType, topicType); err != nil {
+				return errors.Wrapf(err, "failed to register collection - %s and topic - %s", collectionType, topicType)
+			}
+		}
+	}
 
 	api.NewGraphQLHandler(
 		p.handler.APIRouter,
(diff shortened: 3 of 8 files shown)


server/app/permissions_service.go
```diff
@@ -85,6 +85,26 @@ func (p *PermissionsService) hasPermissionsToPlaybook(userID string, playbook Pl
...
+	// For now, everyone with access to the channel is a run admin
+	if p.pluginAPI.User.HasPermissionToChannel(userID, run.ChannelID, model.PermissionReadChannel) {
+		return []string{RunRoleMember, RunRoleAdmin}
+	}
+
+	return []string{}
+}
+
+
+func (p *PermissionsService) HasPermissionsToRun(userID string, run *PlaybookRun, permission *model.Permission) bool {
```
iomodo, line 98: Should we rethink this with regards to current multiple runs in multiple channels effort?

server/config/configuration.go
```diff
@@ -31,6 +31,9 @@ type Configuration struct {
 
 	// AdminLogVerbose: set to include full context with admin log messages.
 	AdminLogVerbose bool
+
+	// ThreadsEverywhereEnabled is a feature flag on a server
+	ThreadsEverywhereEnabled bool
```
iomodo, line 36: We can get rid of `ThreadsEverywhereEnabled` checks if it adds too much complexity. Server is doing this checks anyway and ideally should not call hooks if feature is disabled

server/sqlstore/playbook_run.go
```diff
@@ -1448,6 +1449,93 @@ func (s *playbookRunStore) updateParticipating(playbookRunID string, userIDs []s
...
+		return nil, errors.Wrap(err, "failed to query playbook run by runIDs")
+	}
+	return runs, nil
+}
+
+// GetTaskMetadataByIDs gets PlaybookRunIDs and TeamIDs from runs by taskIDs
+func (s *playbookRunStore) GetTaskMetadataByIDs(taskIDs []string) ([]app.TopicMetadata, error) {
+	tasks := make([]app.TopicMetadata, 0, len(taskIDs))
+	var tasksInDB []app.TopicMetadata
+	for _, taskID := range taskIDs {
```
iomodo, line 1492: Querying multiple tasks at once is possible, with the query like 
```sql
sq.Like{"ChecklistsJSON": fmt.Sprintf("%%\"%s\"%%", taskID1)} OR sq.Like{"ChecklistsJSON": fmt.Sprintf("%%\"%s\"%%", taskID2)} ...
```
but it's too heavy and we do not gain that much of a difference in performance.

lieut-data: As above, I think Squirrel can also just do this for us, so maybe still cleaner?

server/sqlstore/playbook_run.go
```diff
@@ -1448,6 +1449,93 @@ func (s *playbookRunStore) updateParticipating(playbookRunID string, userIDs []s
...
+		query := s.store.builder.
+			Select("ID AS RunID", "TeamID").
+			From("IR_Incident")
+
+		if s.store.db.DriverName() == model.DatabaseDriverMysql {
+			query = query.Where(sq.Like{"ChecklistsJSON": fmt.Sprintf("%%\"%s\"%%", taskID)})
```
lieut-data, line 1498: I think this is potentially brittle against a task name containing a matching ID. Propose that we use SQL to whittle down the runs, but then actually return the `ChecklistsJSON`, parse it, and find it "by hand". We follow a similar pattern in `GetRunsWithAssignedTasks` above.

(Having a dedicated table for `Tasks` obviously makes this 100% better later :))

server/sqlstore/playbook_run.go
```diff
@@ -1448,6 +1449,93 @@ func (s *playbookRunStore) updateParticipating(playbookRunID string, userIDs []s
...
+}
+
+// GetStatusMetadataByIDs gets PlaybookRunIDs and TeamIDs from runs by statusIDs
+func (s *playbookRunStore) GetStatusMetadataByIDs(statusIDs []string) ([]app.TopicMetadata, error) {
+	var statuses []app.TopicMetadata
+	query := s.store.builder.
+		Select("sp.PostID AS ID", "sp.IncidentID AS RunID", "i.TeamID AS TeamID").
+		From("IR_StatusPosts as sp").
+		Join("IR_Incident as i ON sp.IncidentID = i.ID").
+		Where(fmt.Sprintf("sp.PostID IN ('%s')", strings.Join(statusIDs, "', '")))
```
lieut-data, line 1532: Same as above, let's bind these programmatically. And now that I think more about it, these definitely /could/ be user-provided and thus result in a SQL injection.

server/sqlstore/playbook_run.go
```diff
@@ -1448,6 +1449,93 @@ func (s *playbookRunStore) updateParticipating(playbookRunID string, userIDs []s
...
+		// we have single task from db
+		tasksInDB[0].ID = taskID
+		tasks = append(tasks, tasksInDB[0])
+	}
+
+	return tasks, nil
+}
+
+// GetStatusMetadataByIDs gets PlaybookRunIDs and TeamIDs from runs by statusIDs
+func (s *playbookRunStore) GetStatusMetadataByIDs(statusIDs []string) ([]app.TopicMetadata, error) {
```
lieut-data, line 1526: nit: Should we name this as something like `GetStatusAsTopicMetadataByIDs` to clarify?

server/sqlstore/playbook_run.go
```diff
@@ -1448,6 +1449,93 @@ func (s *playbookRunStore) updateParticipating(playbookRunID string, userIDs []s
...
+		From("IR_Incident").
+		Where(fmt.Sprintf("ID IN ('%s')", strings.Join(runIDs, "', '")))
+	if err := s.store.selectBuilder(s.store.db, &runs, query); err != nil {
+		return nil, errors.Wrap(err, "failed to query playbook run by runIDs")
+	}
+	return runs, nil
+}
+
+// GetTaskMetadataByIDs gets PlaybookRunIDs and TeamIDs from runs by taskIDs
+func (s *playbookRunStore) GetTaskMetadataByIDs(taskIDs []string) ([]app.TopicMetadata, error) {
```
lieut-data, line 1489: nit: Should we name this as something like `GetTaskAsTopicMetadataByIDs` to clarify?

server/config/configuration.go
lieut-data: I'm :+1: for simplifying if we can.

server/app/permissions_service.go
lieut-data: I'm not sure: can you elaborate on what we'd need to change?

server/plugin.go
```diff
@@ -30,6 +30,11 @@ const (
 	updateMetricsTaskFrequency = 15 * time.Minute
 
 	metricsExposePort = ":9093"
+
+	RunCollectionType = "run"
```
lieut-data, line 34: nit: Can we document the purpose of this and below constants?

server/plugin.go
```diff
@@ -70,6 +75,8 @@ type Plugin struct {
 	telemetryClient      TelemetryClient
 	licenseChecker       app.LicenseChecker
 	metricsService       *metrics.Metrics
+
+	collectionAndTopicTypes map[string][]string
```
lieut-data, line 79: Why do we need to maintain this map ourselves? Could we just hard-code the two calls (adding more when necessary)?

server/plugin.go
```diff
@@ -395,3 +412,126 @@ func (p *Plugin) getErrorCounterHandler() func(next http.Handler) http.Handler {
...
+
+	var getTopicMetadataByIDs func(topicIds []string) ([]app.TopicMetadata, error)
+	switch topicType {
+	case StatusTopicType:
+		getTopicMetadataByIDs = p.playbookRunService.GetStatusMetadataByIDs
+	case TaskTopicType:
+		getTopicMetadataByIDs = p.playbookRunService.GetTaskMetadataByIDs
+	default:
+		return map[string]*model.TopicMetadata{}, errors.Errorf("topic type %s is not registered by playbooks", topicType)
+	}
```
lieut-data, line 505: I like this pattern more than the check against `collectionAndTopicTypes` above (and it would be more consistent if we just `switch`ed everywhere).

server/plugin.go
```diff
@@ -395,3 +412,126 @@ func (p *Plugin) getErrorCounterHandler() func(next http.Handler) http.Handler {
...
+			TeamId:         topic.TeamID,
+			CollectionId:   topic.RunID,
+		}
+	}
+
+	return topicsMetadata, nil
+}
+
+func mergeSlice(a, b []string) []string {
+	m := map[string]struct{}{}
```
lieut-data, line 525: super nit: Should we pre-initialize this with space for `len(a) + len(b)`?

lieut-data (changes requested):
Awesome complement to the ongoing work -- a few comments inline.

server/sqlstore/playbook_run.go
iomodo: Could you elaborate? I don't think arrays could be used in `sq.Like`.

server/sqlstore/playbook_run.go
```diff
@@ -1448,6 +1449,93 @@ func (s *playbookRunStore) updateParticipating(playbookRunID string, userIDs []s
...
+
+	var ids []string
+	if err := s.store.selectBuilder(s.store.db, &ids, query); err != nil {
+		return nil, errors.Wrap(err, "failed to query for playbook runs")
+	}
+	return ids, nil
+}
+
+// GetRunMetadataByIDs returns playbook runs metadata by passed run IDs.
+// Notice that order of passed ids and returned runs might not coincide
```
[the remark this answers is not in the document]
iomodo: I'd rather make sure that order is correct

server/sqlstore/playbook_run.go
lieut-data: You're right! I somehow thought that `sq.Like` would figure it out, but I realize it's not obvious whether it should "OR" or "AND" together such expressions, so we'd need a new concept in the library.

server/plugin.go
```diff
@@ -31,10 +31,15 @@ const (
 
 	metricsExposePort = ":9093"
 
-	RunCollectionType = "run"
-
+	// Topic represents a start of a thread. In playbooks we support 2 types of topics:
+	// status topic - indicating the start of the thread below status update and
+	// task topic - indicating the start of the thread below task(checklist item)
 	StatusTopicType = "status"
```
lieut-data, line 37: nit: I wonder if it makes sense to call these `TopicTypeStatus` and `TopicTypeTask` and `CollectionTypeRun`?

lieut-data (changes requested):
Nice! Just a few follow up comments.

server/sqlstore/playbook_run.go
```diff
@@ -1472,26 +1471,37 @@ func (s *playbookRunStore) GetPlaybookRunIDsForUser(userID string) ([]string, er
...
-	return runs, nil
+
+	runsMap := make(map[string]app.RunMetadata, len(runs))
+	for _, run := range runs {
+		runsMap[run.ID] = run
+	}
+	orderedRuns := make([]app.RunMetadata, len(runIDs))
+	for i, runID := range runIDs {
+		orderedRuns[i] = runsMap[runID]
+	}
```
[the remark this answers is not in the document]
iomodo: That was a conscious decision. Same thing is true for other `GetSomethingByIDs` methods. 
List you get is the same length you passed, order of returned elements is preserved. And if you pass wrong ID you get empty object(instead of fail).

lieut-data (approved):
🎉

calebroseland (approved):

Issue #1526, opened 2022-10-25
[Polish] Change run owner & task assignee dropdown items


Phrynobatrachus:
https://community.mattermost.com/core/pl/5huot5yk5pf1pqu9bxrgj1nmfw

> (Abhijit) I think we uncovered a blind spot because we'll probably need to update the owner change dropdown also. Instead of channel members and not in channel, should it be run participants and not participating?

![Image](https://user-images.githubusercontent.com/3723666/197841188-bc93c3dd-f5c5-445e-a6dc-fbab6bb5fe65.png)

Issue #1525, opened 2022-10-25
[Polish] Refresh within RHS panel on Add/Remove Participants


Phrynobatrachus:
The RHS overview section currently refreshes on those actions. The clickthrough panel does not until a full page refresh, so they can get out of sync visually. 

![Screenshot from 2022-10-25 13-06-42](https://user-images.githubusercontent.com/3723666/197849006-1cd219d4-211a-4be5-9e71-528c91e1bf54.png)

PR #1538, opened 2022-11-03
remove channel coupling in ProfileSelector/AssignTo

trilopin:
## Summary
Evolve ProfileSelector / AssignTo to become decoupled from channel.

Literals are now: run participants / not participating when groups are shown.
However it wasn't just a literals thing, we took as run members only those members that were also in the channel. The significant change is to take the participant_ids list to decide if they are members or not (and force `ensureProfiles` for them).
Some variables have been renamed in the ProfileSelector to run away from channel/run coupling.

Run list has a profile selector that is not linked to a run (but many). It was requesting runOwners without team_id which made the server to answer with far more users that it should. It's fixed now.

Some real time udapte problems have been solved too.

### UI

**RHS change owner**
Screenshot 2022-11-03 at 18 09 17

**RHS checklist**
Screenshot 2022-11-03 at 18 11 22

**RDP checklist**
Screenshot 2022-11-03 at 18 08 58

**RDP run owner**
Screenshot 2022-11-03 at 18 08 44

**Run filters**
Screenshot 2022-11-03 at 19 09 58

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1526
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1525

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] Unit tests updated


webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -90,6 +90,8 @@ const PlaybookRunDetails = () => {
 
     const queryParams = qs.parse(location.search, {ignoreQueryPrefix: true});
     const role = playbookRun?.participant_ids.includes(myUser.id) || playbookRun?.owner_user_id === myUser.id ? Role.Participant : Role.Viewer;
+
+    useEnsureProfiles(playbookRun?.participant_ids ?? []);
```
trilopin, line 94: I don't understand why this line makes follow/unfollow button in RDP-RHS fail (it doesn't even trigger onclick handler).

webapp/src/components/backstage/runs_list/filters.tsx
```diff
@@ -133,7 +133,7 @@ const Filters = ({fetchParams, setFetchParams, fixedPlaybook, fixedFinished}: Pr
     };
 
     async function fetchOwners() {
-        const owners = await fetchOwnersInTeam(fetchParams.team_id || '');
+        const owners = await fetchOwnersInTeam(currentTeamId);
```
trilopin, line 136: we were asking for all owners always because fetchParams.team_id was always empty

abhijit-singh (changes requested):
Thanks @trilopin, awesome work!
A couple of requests:
1. If I change the owner from the Channels RHS to someone who is not in the run, there's a modal saying that they will be added to the channel — that modal no longer makes sense and can be removed. I don't think we need a confirmation modal for adding people to the run but I can be convinced otherwise if anyone feels we need it. cc @itao 


2. (Non-blocking) Since we're touching this dropdown, can we make the `font-weight: 600` for the `Run participants` and `Not participating` headings in the dropdown. It seems to be incorrectly set to 500 right now.

trilopin:
> 1. If I change the owner from the Channels RHS to someone who is not in the run, there's a modal saying that they will be added to the channel — that modal no longer makes sense and can be removed. I don't think we need a confirmation modal for adding people to the run but I can be convinced otherwise if anyone feels we need it. cc @itao
       
Good point. If we currently do the change-owner from RDP, we already add it to the run but not to the channel if we action toggle is off. I'd do the same in both places.         

> 2. (Non-blocking) Since we're touching this dropdown, can we make the `font-weight: 600` for the `Run participants` and `Not participating` headings in the dropdown. It seems to be incorrectly set to 500 right now.

👍🏽

webapp/src/components/profile/profile_selector.tsx
```diff
@@ -47,15 +47,16 @@ interface Props {
...
-    getUsersInTeam: () => Promise<UserProfile[]>;
+    getAllUsers: () => Promise<UserProfile[]>;
     onSelectedChange?: (userType?: string, user?: UserProfile) => void;
     customControlProps?: any;
     placement?: Placement;
     className?: string;
     selectWithoutName?: boolean;
     customDropdownArrow?: React.ReactNode;
     onOpenChange?: (isOpen: boolean) => void;
+    memberUserIds: string[];
```
crspeller, line 58: Should this be `participantUserIds`?

webapp/src/components/profile/profile_selector.tsx
```diff
@@ -47,15 +47,16 @@ interface Props {
...
+    isMembersModeEnabled: boolean;
 }
 
 export default function ProfileSelector(props: Props) {
```
crspeller, line 62: Could we do the `useEnsureUsers` in here instead? (maybe it would fix that strange problem above)

crspeller (changes requested):

gbochora (approved):
Looks great! Thanks, @trilopin!

webapp/src/components/profile/profile_selector.tsx
trilopin: The problem does not happen anymore. 

I didn't do ensureProfiles here to avoid having to create an additional lock to prevent concurrent fetch (like we have for team/channels) at hooks/general. The thing is that even if we could do the lock lockProfilesInRunFetch, would we have to store duplicated prodilesByRun in redux?

Given all that complexity, it seemed easier to do it at the top level just once.

webapp/src/components/profile/profile_selector.tsx
trilopin: Could be. 

But I'm also wondering if we could do this component also run-agnostic. The only things that seem run-related are the literals for the group names. 

What if this component has the following props:
- handler for one master group of users 
- a subset of user ids (optional)
- group names (optional) with the literals

Would this make sense to you?

crspeller: Yes please. That is way better. We shouldn't have run/playbook specific terms in such a generic component.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
trilopin: it seems to be solved 🎉

webapp/src/components/profile/profile_selector.tsx
trilopin: it felt like that, but I'll double-check it.

webapp/src/components/datetime_selector.tsx
```diff
@@ -152,8 +152,8 @@ export const DateTimeSelector = ({
                 styles={selectStyles}
                 noOptionsMessage={() => <InvalidLabel>{formatMessage({defaultMessage: 'Invalid time duration'})}</InvalidLabel>}
                 onChange={onSelectedChange}
-                classNamePrefix='playbook-run-user-select' // TODO debt: rename
-                className='playbook-run-user-select' // TODO debt: rename
+                classNamePrefix='playbook-react-select'
```
trilopin, line 155: generic classname name since it was used/applied to any react-select use

tests-e2e/cypress/support/plugin/ui_commands.js
```diff
@@ -44,6 +44,8 @@ Cypress.Commands.add('startPlaybookRunWithSlashCommand', (playbookName, playbook
 // Selects Playbooks icon in the App Bar
 Cypress.Commands.add('getPlaybooksAppBarIcon', () => {
     cy.apiGetConfig(true).then(({config}) => {
+        // Wait for channel view being ready in the layout
+        cy.get('#channel_view').should('be.visible');
```
trilopin, line 48: @Phrynobatrachus thoughts about this? 

appbar was inside channel_view before, but now is outside. Without this line, when we click on appbar playbooks icon before Channelview render what triggers is the user menu

Phrynobatrachus: Yep I think this should help, I included it in #1557

abhijit-singh (approved):
Looks good to me! 

Just one thing I'm wondering is if we should add users to a run if a task is assigned to them. But it seems like we don't do that today so I'm not sure. We can discuss and pick it up later if needed.

trilopin:
> Just one thing I'm wondering is if we should add users to a run if a task is assigned to them. But it seems like we don't do that today so I'm not sure. We can discuss and pick it up later if needed.

True, I missed that point.

How do you imagine that? Just adding them to the run as participants or is an informative modal necessary if the assignee is not a participant yet?

webapp/src/components/profile/profile_selector.tsx
```diff
@@ -47,15 +52,27 @@ interface Props {
...
-    onSelectedChange?: (userType?: string, user?: UserProfile) => void;
+    onSelectedChange?: (user?: UserProfile) => void;
     customControlProps?: any;
     placement?: Placement;
     className?: string;
     selectWithoutName?: boolean;
     customDropdownArrow?: React.ReactNode;
     onOpenChange?: (isOpen: boolean) => void;
+
+    // isMembersModeEnabled: boolean;
```
crspeller, line 63: Remove?

crspeller (approved):
Nice!

PR #1540, opened 2022-11-04
1415 join run when you become participant

### LINKED ISSUE #1415: Add ability to join run channel when you become a participant
Can also be request access instead of actually joining the channel if it is private.
Should show a message if the action is enabled to add you automatically rather than an option.
Checkbox should be unchecked by default.

[Figma Designs](https://www.figma.com/file/oBERv3BlbjpcFMKgPAo7E5/%F0%9F%91%A5-Members-%E2%89%A0-Participants?node-id=289%3A215720)

gbochora:
## Summary
This PR updates become a participant flow. When a not participant tries to join the run:
- If the `join action` is enabled, the user will also become the linked channel member, whether the channel is public or private.
- If the `join action` is disabled and the linked channel is public, the user can decide whether to join the channel and select the checkbox.
- If the `join action` is disabled and the linked channel is private, the user can't join the channel, but he/she can send the join request.

https://user-images.githubusercontent.com/4368372/199952382-286e3795-afe4-40dd-b966-63b51d0eb9bc.mp4

## Ticket Link
Fixes #1415

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


server/app/playbook_run_service.go
```diff
@@ -2939,29 +2939,7 @@ func (s *PlaybookRunServiceImpl) participateActions(playbookRun *PlaybookRun, ch
...
-		msg := T("app.user.run.joined_run_channel_private_add_participant", data)
-		if isParticipateFlow {
-			msg = T("app.user.run.joined_run_channel_private_participate", data)
-		}
-
-		if _, err := s.poster.PostMessage(playbookRun.ChannelID, msg); err != nil {
-			logrus.WithError(err).Errorf("participateActions: failed to send message to private channel, userID '%s'", user.Id)
-		}
-		return
-	}
```
trilopin, line 2962: oh nice, this makes no sense anymore. I think we should manually remove those literals. shouldn't we?

webapp/src/components/rhs/rhs_run_details.tsx
```diff
@@ -72,7 +72,7 @@ const RHSRunDetails = () => {
         }
     }, [runDetailsStep]);
 
-    const {ParticipateConfirmModal, showParticipateConfirm} = useParticipateInRun(playbookRun?.id || '', 'channel_rhs');
+    const {ParticipateConfirmModal, showParticipateConfirm} = useParticipateInRun(playbookRun || {} as PlaybookRun, 'channel_rhs');
```
trilopin, line 75: wouldn't you accept undefined playbookRun in the hook instead of `{}  as PlaybookRun`?

trilopin (changes requested):
This modal of becoming participant can be triggered from channels RHS and, in that case, side-actions related to channel would make no sense in the modal. isn't it?

Screenshot 2022-11-10 at 10 41 46

gbochora:
> This modal of becoming participant can be triggered from channels RHS and, in that case, side-actions related to channel would make no sense in the modal. isn't it?
> 
Good catch! I've fixed the logic; if a user is already a channel member, it will not show an extra message.

trilopin (approved):
Cool 🎉

crspeller (approved):
Awesome!

Issue #1541, opened 2022-11-04
prepare core playbook and update at Playbook Editor


trilopin:
The model and store layer of ir_playbook will need a couple more attributes:
- channelid: should be filled with the channel id in case we want to default new runs to be linked to an existent channel 
- channelmode (better names accepted) enum to govern the radio. we could rely on the nullability of the channelid field, but that could be a problem later (if we don't want channel at all for example).

we need to migrate the database and extend the models and the webapp types.

**UI at PBE**
https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=321%3A169395

![Image](https://user-images.githubusercontent.com/4096774/199056038-8a0e2abe-793e-4396-91bb-2e9ce11693b1.png)

Radio button + input to have the possibility of choosing an existent channel.
When the new choice is checked, Configure channel and public/private selector should be disabled.

To discuss:
* if we attach runs to existing channels, don't we have to decouple the channel name from the run name? Does it mean we have to move from "Channel name template" to "Run name template" and make it available to both choices "existing and new channels"?

PR #1549, opened 2022-11-07
PBE run link to existing channel

trilopin:
## Summary
Add the ability to configure playbooks that can start runs linked to an existing channel.
This PR does not cover:
- multiple runs list in channels rhs
- ability to use different config than the one in PBE when starting a run

### Feature flag
There is a new feature flag:
* sever-side: `	LinkRunToExistingChannelEnabled bool` at Configuration
* webapp: `selectLinkRunToExistingChannelEnabled` selector and `useLinkRunToExistingChannelEnabled` hook available

The main idea is to merge to master (be able to deploy) as often as possible instead of working too much in integration branches

To enable it, in server config.json add `"linkruntoexistingchannelenabled": true,` to Playbooks config.

The feature flag is false (disabled) by default, so if it's not configured disabled will be assumed. It's expected to be set at true by the end of the V0. 

### API & GQL
The playbook query carries the new fields and `UpdatePlaybook` mutation accepts them.

### Service & Store
Two new fields in the playbook: `ChannelID` and `ChannelMode`.

The new fields have been created in the playbook, create-new-channel is the default behavior.
- (postgre) Removed the unique constraint for the channelID index in the ir_incident table.
- (MySQL) removed unique index (and kept the other index in channeled)

**Name field in ir_incident**
This field wasn't used and playbookrun store used channel.displayname instead.
Less than 2% of the runs had a non-empty name in the community server. That's why there's a migration to update all empty names by the matching channel display name.

### Playbook UI
The user can choose whether create a new channel per run or link an existing one. Available channels are those you have permission to.

Channel Selector has been modified to allow single channel selection (it was hardcoded top multi before)

It is considered valid to save without selecting a channel but telling that you'll link the run to an existing channel. It will be mandatory in the run-start.

https://user-images.githubusercontent.com/4096774/200348951-daf36d56-c9eb-409c-b3f4-f14a387eb256.mp4

**How it looks if feature flag is disabled (default)**
Screenshot 2022-11-10 at 16 54 26

## Channel selector
Already existing channel selector has been upgraded to:
- be usable for one channel (previously only multi)
- display options nicely: icon for private/public and some styles
Screenshot 2022-11-11 at 12 55 26
Screenshot 2022-11-11 at 12 54 58
Screenshot 2022-11-11 at 12 50 42

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1541

## Checklist

- [ ] ~~Telemetry updated~~
- [X] Gated by an experimental feature flag
- [X] Unit tests updated


server/api/graphql_root_playbook.go
```diff
@@ -122,6 +122,8 @@ func (r *PlaybookRootResolver) UpdatePlaybook(ctx context.Context, args struct {
 		IsFavorite                              *bool
 		CreateChannelMemberOnNewParticipant     *bool
 		RemoveChannelMemberOnRemovedParticipant *bool
+		ChannelID                               *string
```
crspeller, line 125: Just `ChannelID` confused me on first pass. Maybe call it `RunLinkChannelID`? Or something else more descriptive?

server/app/playbook_run_service.go
```diff
@@ -253,30 +259,30 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
...
 		playbookRun.ChannelID = channel.Id
 	} else {
-		existingPlaybookRunID, err := s.GetPlaybookRunIDForChannel(playbookRun.ChannelID)
-		if err != nil && !errors.Is(err, ErrNotFound) {
-			return nil, err
-		} else if existingPlaybookRunID != "" {
-			return nil, errors.Wrapf(ErrMalformedPlaybookRun, "playbook run %s already exists for channel %s", existingPlaybookRunID, playbookRun.ChannelID)
-		}
-
+		/*	PRE MULTIPLERUNS IN A CHANNEL
```
crspeller, line 269: We can just remove this.

server/sqlstore/playbook_run.go
```diff
@@ -179,7 +179,7 @@ func NewPlaybookRunStore(pluginAPI PluginAPIClient, sqlStore *SQLStore) app.Play
 
 	// When adding a PlaybookRun column #1: add to this select
 	playbookRunSelect := sqlStore.builder.
-		Select("i.ID", "c.DisplayName AS Name", "i.Description AS Summary", "i.CommanderUserID AS OwnerUserID", "i.TeamID", "i.ChannelID",
+		Select("i.ID", "i.Name AS Name", "i.Description AS Summary", "i.CommanderUserID AS OwnerUserID", "i.TeamID", "i.ChannelID",
```
crspeller, line 182: We can remove the join on the channels table right?

crspeller (changes requested):
Looking good!

gbochora (approved):
Great work, @trilopin 🥳

server/sqlstore/playbook_run.go
trilopin: I also created a ticket for the refactor of the digest queries  https://github.com/mattermost/mattermost-plugin-playbooks/issues/1571

server/api/graphql_root_playbook.go
trilopin: 2/5 if the playbook acts as a run template, wouldn't the same naming (for the fields that are copied) be the less surprising choice?

crspeller: That's a good point. I can see it both ways. I will leave it to your discretion.

server/app/playbook_run_service.go
```diff
@@ -253,30 +259,30 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
...
 		channel, err = s.pluginAPI.Channel.Get(playbookRun.ChannelID)
 		if err != nil {
 			return nil, err
 		}
 
-		playbookRun.Name = channel.Name
+	}
+
+	if playbookRun.Name == "" {
+		playbookRun.Name = pb.ChannelNameTemplate
```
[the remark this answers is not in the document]
trilopin: not really

server/sqlstore/migrations/future/postgres/0.61.down.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE IR_Playbook DROP COLUMN IF EXISTS ChannelID;
+ALTER TABLE IR_Playbook DROP COLUMN IF EXISTS ChannelMode;
+
+-- Drop unique constraint and kee the index
+ALTER TABLE IR_Incident  ADD CONSTRAINT ir_incident_channelid_key UNIQUE(ChannelID);
```
trilopin, line 5: I tried IF NOT EXISTS but apparently does not work here. thoughts @gbochora ?

abhijit-singh (approved):
Looks good!

server/sqlstore/migrations/future/postgres/0.61.down.sql
gbochora:
```suggestion
DO
$$
BEGIN
    IF NOT EXISTS (
        SELECT INDEXNAME FROM PG_INDEXES
        WHERE TABLENAME = 'ir_incident'
        AND INDEXNAME = 'ir_incident_channelid_key'
    ) THEN
        ALTER TABLE IR_Incident ADD CONSTRAINT ir_incident_channelid_key UNIQUE(ChannelID);
	END IF;
END
$$;

```

gbochora (commented):
I've left a few comments regarding morph migrations.

server/sqlstore/migrations/future/mysql/0.61.down.sql
```diff
@@ -0,0 +1,45 @@
...
+    'ALTER TABLE IR_Playbook DROP COLUMN ChannelMode;',
+    'SELECT 1;'
+));
+
+PREPARE dropColumnIfExists FROM @preparedStatement;
+EXECUTE dropColumnIfExists;
+DEALLOCATE PREPARE dropColumnIfExists;
+
+SET @preparedStatement = (SELECT IF(
+    EXISTS(
```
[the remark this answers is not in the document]
trilopin: oops

crspeller (approved):

PR #1554, opened 2022-11-08
MM-48007 Respect read only channel permissions.

crspeller:
## Summary
Respect read only channel permisions from server.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-48007

### DIFF
--- a/server/app/permissions_service.go
+++ b/server/app/permissions_service.go
@@ -449,6 +449,20 @@ func (p *PermissionsService) RunViewByChannel(userID, channelID string) error {
 	return p.RunView(userID, runID)
 }
 
+func (p *PermissionsService) RunUpdateStatus(userID string, playbookRun *PlaybookRun) error {
+	if !CanPostToChannel(userID, playbookRun.ChannelID, p.pluginAPI) {
+		return errors.Wrapf(ErrNoPermissions, "user %s cannot post to playbook run channel %s", userID, playbookRun.ChannelID)
+	}
+
+	for _, channelID := range playbookRun.BroadcastChannelIDs {
+		if !CanPostToChannel(userID, channelID, p.pluginAPI) {
+			return errors.Wrapf(ErrNoPermissions, "user %s cannot post to broadcast channel %s", userID, channelID)
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


Phrynobatrachus:
Hopefully resolved by https://github.com/mattermost/mattermost-plugin-playbooks/pull/1557 :crossed_fingers:

server/app/permissions_service.go
```diff
@@ -449,6 +449,20 @@ func (p *PermissionsService) RunViewByChannel(userID, channelID string) error {
...
 }
 
+func (p *PermissionsService) RunUpdateStatus(userID string, playbookRun *PlaybookRun) error {
+	if !CanPostToChannel(userID, playbookRun.ChannelID, p.pluginAPI) {
+		return errors.Wrapf(ErrNoPermissions, "user %s cannot post to playbook run channel %s", userID, playbookRun.ChannelID)
+	}
+
+	for _, channelID := range playbookRun.BroadcastChannelIDs {
+		if !CanPostToChannel(userID, channelID, p.pluginAPI) {
+			return errors.Wrapf(ErrNoPermissions, "user %s cannot post to broadcast channel %s", userID, channelID)
```
lieut-data, line 459: Should we consider failing these separately, allowing the status update to be posted to channel (later, a dedicated topic) even if the broadcast channel update can't be posted?

lieut-data (approved):

server/app/permissions_service.go
crspeller: Possibly. We would need designs for what exactly to tell the user, since not posting to some of the broadcast channels would be unexpected.

lieut-data: Do we have designs for what happens now when the request fails? I'd be fine with just logging a warning and skipping the functionality itself if that made sense. (Probably less confusing than the status update feature itself being non-functional.)

crspeller: No designs for current functionality. No permissions case isn't handled since in the channel model you would never have access to press the button without permission to post a status update. 

We should probably think though the design here and implement the preferred solution outside this PR.

oh6hay (approved):
LGTM :+1:

PR #1555, opened 2022-11-09
1553 read only rhs toast on checklist

### LINKED ISSUE #1553: Pop up read only RHS toast when user attempts to check off a checklist item

gbochora:
## Summary
This PR adds toast when a channel member with read-only access to the run tries to change the checklist state.

https://user-images.githubusercontent.com/4368372/200803827-ff705ebd-d090-4ecd-bf88-be2ee675e828.mp4

## Ticket Link
Fixes #1553

## Checklist

~~- [ ] Telemetry updated~~
~~- [ ] Unit tests updated~~

### DIFF
--- a/webapp/src/components/checklist_item/inputs.tsx
+++ b/webapp/src/components/checklist_item/inputs.tsx
@@ -1,7 +1,7 @@
 
 import React, {useRef, useState} from 'react';
 import {useUpdateEffect} from 'react-use';
-import styled from 'styled-components';
+import styled, {css} from 'styled-components';
 import {useIntl} from 'react-intl';
 import {ClientError} from '@mattermost/client';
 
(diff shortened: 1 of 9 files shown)


crspeller (approved):
Nice :)

trilopin:
it's triggering the toast when the run is finished, shouldn't be completely disabled in that case?

webapp/src/components/checklist_item/inputs.tsx
```diff
@@ -30,6 +31,10 @@ export const CheckBoxButton = (props: CheckBoxButtonProps) => {
     //     Additionally, we prevent the user from clicking multiple times
     //     and leaving the item in an unknown state
     const handleOnChange = async () => {
+        if (props.readOnly) {
+            props.onChange(isChecked ? ChecklistItemState.Closed : ChecklistItemState.Open);
```
trilopin, line 35: do we have to invoke this onChange without changing it at all? can we skip this?

trilopin (approved):
Cool

I left a couple of non-blocking comments

webapp/src/components/checklist_item/inputs.tsx
gbochora: Yep, I've passed `onViewerModeInteract` to the `CheckBoxButton` component, so there is no need to handle click in the parent component.

PR #1556, opened 2022-11-09
avoid checklistitem reset while editing

trilopin:
## Summary
The component gets rendered again when losing focus, values are obviously different (props have the values before the edition), and were forced to the previous values.

Split useEffect with smaller deps, so useUpdateEffect is not called again with no sense.

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-48308

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):
Not sure this is how this should be done, but refactoring this component can take place independent of the bug fix.

trilopin:
> Not sure this is how this should be done, but refactoring this component can take place independent of the bug fix.

IMHO makes no sense to resetting the component in a re-render and the code added is valid. However, I agree we shouldn't stop here and check the reason why it re-renders with focus changes.

calebroseland (changes requested):
Having this single update effect with the `isEditing` gate for updating all 5 properties in-tandem could lead to odd edge cases. 

Let's say two people are collaborating on a checklist:
- user-1 clicks edit on task-a, and starts drafting a change to the description
- user-2 assigns task-a to user-1, and goes on to something else
- user-1 finishes their changes and clicks save on task-a

User-1 would be unaware that they should be assigned to task-a, but user-2's change was overwritten by user-1's stale `assigneeID` value. 

Under this pattern, this grows even more likely with `command`, `dueDate`, and potentially additions later on. 


I'd propose 2 options:

1. Increase the update effect granularity—one for each property, as suggested above, or
2. Use `useProxyState` to mediate the incoming props, local state, and side effects (more on that later). It has a built in update effect that keeps its state up-to-date from any incoming prop changes.

e.g. 
```diff
-    const [titleValue, setTitleValue] = useState(props.checklistItem.title);
+   const [titleValue, setTitleValue] = useProxyState(props.checklistItem.title);
```
I'd recommend `useProxyState` for reasons explained below.

### UX (for later/followup)

I'm also noticing that Assignee, Slash Command, and Due Date all bypass the Edit-Save/Cancel flow. This seems to break the UX expectation that if I click Edit, change the assignee, then click cancel—that my change would be discarded—but as of now these three properties get saved immediately when they're changed regardless of `isEditing`
CleanShot 2022-11-09 at 13 25 07@2x

We could swap the Save/Cancel buttons to make it more obvious, but I'm not sure about this.
CleanShot 2022-11-09 at 13 45 33@2x


But I'd like to propose another solution:

`useProxyState` handles update effect, but it also fires a side effect when it's updated from the state setter. So this could be used to bring those 3 properties into the Edit-Save/Cancel flow.

E.g. for `assigneeID`, something like:

```tsx
    const saveAssignee = async (userId = '') => {
        if (props.playbookRunId) {
            const response = await setAssignee(props.playbookRunId, props.checklistNum, props.itemNum, userId);
            if (response.error) {
                console.log(response.error); // eslint-disable-line no-console
            }
        } else {
            const newItem = {...props.checklistItem};
            newItem.assignee_id = userId;
            props.onUpdateChecklistItem?.(newItem);
        }
    };

    const [assigneeID, setAssigneeID] = useProxyState(props.checklistItem.assignee_id, (userId) => {
        if (props.newItem || isEditing) {
            return;
        }
        saveAssignee(userId);
    }, 0);

    const onAssigneeChange = (userType?: string, user?: UserProfile) => {
        setAssigneeID(user?.id);
    };

    //... reset onCancel
    setAssigneeID(props.checklistItem.assignee_id);
    //... persist onSave
    saveAssignee(assigneeID);
```

Rinse/repeat as applicable for `command`, `dueDate`, etc. and that should straighten out the Edit-Save/Cancel flow IMHO.

trilopin:
@calebroseland Thougths on merging this with just useEffect split (which works fine) to solve the urgent part while I continue processing the rest of your feedback?

calebroseland (approved):
> @calebroseland Thougths on merging this with just useEffect split (which works fine) to solve the urgent part while I continue processing the rest of your feedback?

💯 perfectly fine with me

trilopin:
Follow-up ticket https://mattermost.atlassian.net/browse/MM-48487

PR #1557, opened 2022-11-09
e2e: Flake fixes

Phrynobatrachus:
The test failures in `channels > RHS` specs from the past few days appear to be similar to the ones fixed in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1510 - namely that the specs were trying to perform some action (clicking) while the DOM was still unstable. This PR adds some `should()` guards and API `wait()`s to avoid that.


crspeller (approved):

trilopin:
THanks @Phrynobatrachus ❤️

Issue #1560, opened 2022-11-09
Remove ChannelID unique constraint


crspeller:
Multiple runs can not share the same channel ID until the unique constraint is removed. Make a migration to remove the constraint.

crspeller:
Fixed by #1549

PR #1561, opened 2022-11-09
Adding channel filtering support to runs listing graphql API

### LINKED ISSUE #1545: run list by channel server side (GQL)
Prepare service and store layer to be able to get all the runs linked to a channeled.

At graphql side we have to decide if we evolve the runs root resolver to accept channelID as criteria (and make optional participantOrFollowerID?) or to create another root resolver.

Current:
```graphql
	runs (participantOrFollowerID: $userID, teamID: $teamID, sort: "name", statuses: ["InProgress"]){
		id
		name
		isFavorite
		playbookID
		ownerUserID
		participantIDs
		metadata { followers }
	}
``` 

```go
func (r *RunRootResolver) Runs(ctx context.Context, args struct {
	TeamID                  string `url:"team_id,omitempty"`
	Sort                    string
	Statuses                []string
	ParticipantOrFollowerID string `url:"participant_or_follower,omitempty"`
}) ([]*RunResolver, error) {
...
```

crspeller:
## Summary
Adds the ability to list runs by channel using the GraphQL API. 
Note that it's currently impossible to get in a state where this API will actually return multiple results. The unique constraint needs to be removed, see #1560 
There should be no problem with committing this directly to master. 

There was some discussion in the ticket around making `participantOrFollowerID` optional, however it already has a default in the schema. 

I also think the changes here are simple enough not to warrent a new root resolver for this case.

## Ticket Link
Fixes #1545

### DIFF
--- a/server/api/graphql_root_run.go
+++ b/server/api/graphql_root_run.go
@@ -34,10 +34,11 @@ func (r *RunRootResolver) Run(ctx context.Context, args struct {
 }
 
 func (r *RunRootResolver) Runs(ctx context.Context, args struct {
-	TeamID                  string `url:"team_id,omitempty"`
+	TeamID                  string
 	Sort                    string
 	Statuses                []string
-	ParticipantOrFollowerID string `url:"participant_or_follower,omitempty"`
+	ParticipantOrFollowerID string
+	ChannelID               string
 }) ([]*RunResolver, error) {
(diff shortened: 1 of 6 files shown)


gbochora (approved):
Nice!

trilopin (approved):
well done, small but important change!

PR #1562, opened 2022-11-09
Remove channelid unique constraint.

### LINKED ISSUE #1560: Remove ChannelID unique constraint

crspeller:
## Summary
Removes the unique constraint restricting a one to one channel to run mapping. 

## Ticket Link
Fixes #1560

### DIFF
--- a/server/sqlstore/migrations.go
+++ b/server/sqlstore/migrations.go
@@ -2393,4 +2393,20 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.60.0"),
+		toVersion:   semver.MustParse("0.61.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DatabaseDriverMysql {
+				if err := dropIndexIfExists(e, sqlStore, "IR_Incident", "ChannelID"); err != nil {
+					return errors.Wrapf(err, "failed to remove channelid uniqueness constraint")
+				}
+			} else {
+				if _, err := e.Exec(`ALTER TABLE IR_Incident DROP CONSTRAINT IF EXISTS ir_incident_channelid_key`); err != nil {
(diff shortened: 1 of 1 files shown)


crspeller:
@trilopin Not sure if we want to hold this one behind a feature flag or not. I don't think this will create any problems on it's own so it should be safe for master.

server/sqlstore/migrations.go
```diff
@@ -2393,4 +2393,20 @@ var migrations = []Migration{
 			return nil
 		},
 	},
+	{
+		fromVersion: semver.MustParse("0.60.0"),
+		toVersion:   semver.MustParse("0.61.0"),
+		migrationFunc: func(e sqlx.Ext, sqlStore *SQLStore) error {
+			if e.DriverName() == model.DatabaseDriverMysql {
+				if err := dropIndexIfExists(e, sqlStore, "IR_Incident", "ChannelID"); err != nil {
```
gbochora, line 2401: Could we use the `dropIndexIfExists` function for both databases?

gbochora (approved):
Looks good! 🎉

trilopin (commented):
I've already included it as part of the core changes https://github.com/mattermost/mattermost-plugin-playbooks/pull/1549/files#diff-9ba9301f962251f04d874eb7f9538c18790293d063cf6707300687cb49301881

would you still keep separated migrations or try to have one with all changes?

trilopin:
> @trilopin Not sure if we want to hold this one behind a feature flag or not. I don't think this will create any problems on it's own so it should be safe for master.

If we move ahead with the featureFlag, one thing we could move the constraint to application and check with featureflag:
- remove unique constraint
- create flag linkRunTOExitingChannelEnabled, default = false
- if linkRunTOExitingChannelEnabled is false, in service layer CreatePlaybookRun: check for previous runs for the same channelID, if the results are not empty, return 400 error
 

thoughts?

crspeller:
Talked offline. Going to close this in favor of getting #1549 in quicker.

PR #1563, opened 2022-11-09
useProxyState: make onChange optional

calebroseland:
## Summary
Simplify `useProxyState`, make `onChange` optional. (Context: #1556)

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


trilopin (approved):
Sorry for the delay.

If I understood well, making optional onChange would have the same effect than useState+UseUpdateEffectand and leave space for future potential debounced side-effect. Is that correct?

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -77,7 +77,7 @@ const ChecklistList = ({
 
     const updatePlaybook = useUpdatePlaybook(inPlaybook?.id);
     const [playbook, setPlaybook] = useProxyState(inPlaybook, useCallback((updatedPlaybook) => {
```
trilopin, line 79: I wonder if that useCallback can be removed since we already have one inside useProxyState for `onChangeDebounced`.

calebroseland: @trilopin the `onChangeDebounced` useCallback is downstream of/dependent on this one—so if we don't keep this one's identity stable, it'll recreate and I think the cycle wouldn't work properly.

calebroseland:
> If I understood well, making optional onChange would have the same effect as useState+UseUpdateEffect and also leave space for future potential debounced side-effect. Is that correct?

That's right, @trilopin. 👍 The benefit of having the side-effect built into the logic like this, is that you can differentiate between an intentional `setState` or an incoming prop-change. You could of course do the same thing with a custom set function wrapper with a plain `useState`/`useUpdateEffect`, but having it standardized can reduce quite a lot of boilerplate.

PR #1564, opened 2022-11-10
Morph v0.25.0 v0.30.0

gbochora:
## Summary
This PR adds morph migration files for v0.25.0-v0.30.0. Minor refactoring: moved helper functions to the utils file.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-47954

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


crspeller (approved):

trilopin (changes requested):
I stopped reviewing until I solve some doubts.

do we expect to run all morph migrations at least once over in-use databases? 

if yes: 
- wouldn't some updates overwrite some important data?
- wouldn't some updates be just the initial load of a just-created field? 

let me know if I'm reading this wrongly and I'll move ahead with the rest of the review.

gbochora:
> do we expect to run all morph migrations at least once over in-use databases?
No, I plan to run only new migrations. Morph stores the list of migrations that have been run in the `IR_db_migrations` table and doesn't rerun them. So, before running the first morph migration, I'll check the current legacy migration version and, based on that information, write corresponding morph migrations versions in the `IR_db_migrations` table.

trilopin:
> No, I plan to run only new migrations. Morph stores the list of migrations that have been run in the `IR_db_migrations` table and doesn't rerun them. So, before running the first morph migration, I'll check the current legacy migration version and, based on that information, write corresponding morph migrations versions in the `IR_db_migrations` table.

So we could say that the last migration in the old(current) migration system will be to set the right status into the morph migrations data. is that what you have in mind?

Thanks for explaining.

gbochora:
> So we could say that the last migration in the old(current) migration system will be to set the right status into the morph migrations data. is that what you have in mind?
Yes, that's correct.

trilopin (approved):

PR #1567, opened 2022-11-11
playbooks-mpa

gbochora:
## Summary
Multi-product architecture, server part.

## Ticket Link
No ticket

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


lieut-data:
Thank you again, @gbochora for all this effort! We're close to effectively merging this into the monorepo, but we'll be keeping `master` as a plugin for backwards compatibility / maintenance.

PR #1570, opened 2022-11-14
MM 44807 Adjust outline hover effects

m-ripper:
## Summary
The hover effects in the playbook outline are now more consistent.
Only section headers with a right header element are highlighted on hovering.

msedge_ZvIexJGS9g

## Ticket Link
[MM-44807](https://mattermost.atlassian.net/browse/MM-44807)

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbook_editor/outline/outline.tsx
+++ b/webapp/src/components/backstage/playbook_editor/outline/outline.tsx
@@ -72,6 +72,7 @@ const Outline = ({playbook, refetch}: Props) => {
             <Section
                 id={'status-updates'}
                 title={formatMessage({defaultMessage: 'Status Updates'})}
+                hasSubtitle={true}
                 hoverEffect={true}
                 headerRight={(
                     <HoverMenuContainer data-testid={'status-update-toggle'}>
@@ -118,6 +119,7 @@ const Outline = ({playbook, refetch}: Props) => {
             <Section
                 id={'retrospective'}
                 title={formatMessage({defaultMessage: 'Retrospective'})}
+                hasSubtitle={retrospectiveAccess && !playbook.retrospective_enabled}
                 hoverEffect={true}
                 headerRight={(
                     <HoverMenuContainer>
--- a/webapp/src/components/backstage/playbook_editor/outline/outline.tsx
+++ b/webapp/src/components/backstage/playbook_editor/outline/outline.tsx
@@ -88,6 +88,16 @@ const Outline = ({playbook, refetch}: Props) => {
                         />
                     </HoverMenuContainer>
                 )}
+                onHeaderClick={() => {
+                    if (archived) {
+                        return;
+                    }
+                    updatePlaybook({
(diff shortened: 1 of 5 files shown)


abhijit-singh (changes requested):
Thanks @m-ripper! A couple of requests:
1. Clicking on the copy link icon on the left should not toggle the section on and off. Ideally, it should just copy the link and not trigger the toggle. Also, once you copy link, the hover state gets stuck.
2. Can we reduce the space between the header and the subtitle? The header section seems to have a margin-bottom of 10px currently which can be reduced to 2px.

m-ripper:
> Thanks @m-ripper! A couple of requests:
> 
> 1. Clicking on the copy link icon on the left should not toggle the section on and off. Ideally, it should just copy the link and not trigger the toggle. Also, once you copy link, the hover state gets stuck.
> 2. Can we reduce the space between the header and the subtitle? The header section seems to have a margin-bottom of 10px currently which can be reduced to 2px.

Thanks @abhijit-singh for the feedback.

Regarding 2.: Should this affect all headers or only those with a subtitle?

abhijit-singh:
> Regarding 2.: Should this affect all headers or only those with a subtitle?

Just the ones with a subtitle, which will be `Status updates` and `Retrospective` today.

webapp/src/components/backstage/playbook_editor/outline/section_retrospective.tsx
```diff
@@ -61,7 +65,7 @@ const SectionRetrospective = ({playbook, refetch}: Props) => {
                 </BackstageSubheader>
                 <StyledSelect
                     value={retrospectiveReminderOptions.find((option) => option.value === playbook.retrospective_reminder_interval_seconds)}
-                    onChange={(option: {label: string; value: number;}) => {
+                    onChange={(option: { label: string; value: number; }) => {
```
iomodo, line 68: extra spaces

webapp/src/components/backstage/playbook_editor/outline/outline.tsx
```diff
@@ -88,6 +88,16 @@ const Outline = ({playbook, refetch}: Props) => {
                         />
                     </HoverMenuContainer>
                 )}
+                onHeaderClick={() => {
```
iomodo, line 92: Could we create a single function to make sure that  toggle `onChange` and `onHeaderClick` does the same thing

webapp/src/components/backstage/playbook_editor/outline/outline.tsx
```diff
@@ -122,6 +132,14 @@ const Outline = ({playbook, refetch}: Props) => {
...
```
iomodo, line 137: same here

trilopin (approved):
🎉 LGTM

iomodo (approved):
Looks great @m-ripper, thanks <3

abhijit-singh (approved):

PR #1572, opened 2022-11-14
MM-47235: Fix styling of hr in status updates

lieut-data:
## Summary
An upstream change to CSS broke the styling for status updates:
image

Restore to the expected design:
CleanShot 2022-11-14 at 11 31 20@2x

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-47235


webapp/src/components/update_post.tsx
```diff
@@ -117,10 +117,11 @@ const StyledPostText = styled(PostText)`
 `;
 
 const Separator = styled.hr`
-    && {
+    &&& {
```
lieut-data, line 120: More specificity to overcome: 
image

webapp/src/components/update_post.tsx
```diff
@@ -117,10 +117,11 @@ const StyledPostText = styled(PostText)`
...
         border: none;
         height: 1px;
         background: rgba(var(--center-channel-color-rgb), 0.61);
+        margin: 0;
```
lieut-data, line 124: Override new Bootstrap scaffolding:
image

webapp/src/components/update_post.tsx
```diff
@@ -117,10 +117,11 @@ const StyledPostText = styled(PostText)`
 `;
 
 const Separator = styled.hr`
-    && {
+    &&& {
         border: none;
         height: 1px;
         background: rgba(var(--center-channel-color-rgb), 0.61);
```
calebroseland, line 123: Seemed a little bolder than the usual divider, probably typo during feature—should be 0.16 per [designs](https://www.figma.com/file/OJeUZDfESLlXgblbn7Beac/Better-Status-Updates?node-id=1027%3A333472&t=2OvhTlqVtl7MFdOs-4).

```suggestion
        background: rgba(var(--center-channel-color-rgb), 0.16);
```

calebroseland (approved):
LGTM! Though it would be nice to fix the bg color while we're here.

webapp/src/components/update_post.tsx
calebroseland, line 124: Though maybe this, too:

```suggestion
        margin: 12px 0;
```

CleanShot 2022-11-14 at 10 00 27@2x
CleanShot 2022-11-14 at 10 00 22@2x

abhijit-singh (changes requested):
There seems to be a 20% opacity applied on top of the `background: rgba(var(--center-channel-color-rgb), 0.16);`, which is making the line very light. 

I think we can either add an opacity:1 attribute to this instance or change the color to  `background: rgba(var(--center-channel-color-rgb), 0.8);`.
image

lieut-data:
Wow, good eye, @abhijit-singh! Now looks like:

image

abhijit-singh (approved):
💯

PR #1575, opened 2022-11-15
MM-48449: avoid race with /playbook

lieut-data:
## Summary
If a task slash command uses `/playbook ...` which in turn modifies the run (e.g. `/playbook owner @username`), the update from that slash command gets overwritten by the code invoking the slash command to record the execution of same.

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-48449

## Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -1716,6 +1716,13 @@ func (s *PlaybookRunServiceImpl) RunChecklistItemSlashCommand(playbookRunID, use
 		return "", errors.Wrap(err, "failed to run slash command")
 	}
 
+	// Fetch the playbook run again, in case the slash command actually changed the run
+	// (e.g. `/playbook owner`).
+	playbookRun, err = s.store.GetPlaybookRun(playbookRunID)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


server/app/playbook_run_service.go
```diff
@@ -1716,6 +1716,13 @@ func (s *PlaybookRunServiceImpl) RunChecklistItemSlashCommand(playbookRunID, use
 		return "", errors.Wrap(err, "failed to run slash command")
 	}
 
+	// Fetch the playbook run again, in case the slash command actually changed the run
+	// (e.g. `/playbook owner`).
+	playbookRun, err = s.store.GetPlaybookRun(playbookRunID)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to retrieve playbook run after running slash command")
+	}
```
lieut-data, line 1724: This is less elegant than I'd like, e.g. atomically updating just the affected rows in the database, but I believe we're tackling some of that in other efforts.

m-ripper (approved):
Looks good!

gbochora (approved):
Nice!

PR #1576, opened 2022-11-16
respect fullname setting at getOwners

trilopin:
## Summary
Honour PrivacySettings.ShowFullName value at getOwners endpoint.

Added client function and tests.

It's a bit surprising to get a nil value when is false from GetConfig() server calls, but it will be investigated separately.

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-48008

## Checklist
- [ ] ~~Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -498,13 +498,6 @@ type OwnerInfo struct {
 	Nickname  string `json:"nickname"`
 }
 
-func (o *OwnerInfo) Sanitize(options map[string]bool) {
-	if len(options) != 0 && !options["fullname"] {
-		o.FirstName = ""
-		o.LastName = ""
-	}
-}
-
 // DialogState holds the start playbook run interactive dialog's state as it appears in the client
 // and is submitted back to the server.
 type DialogState struct {
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -498,6 +498,13 @@ type OwnerInfo struct {
 	Nickname  string `json:"nickname"`
 }
 
+func (o *OwnerInfo) Sanitize(options map[string]bool) {
+	if len(options) != 0 && !options["fullname"] {
+		o.FirstName = ""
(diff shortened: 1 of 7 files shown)


server/app/playbook_run_service.go
```diff
@@ -1334,6 +1334,23 @@ func (s *PlaybookRunServiceImpl) GetOwners(requesterInfo RequesterInfo, options
 	if err != nil {
 		return nil, errors.Wrap(err, "can't get owners from the store")
 	}
+
+	// ShowFullName is coming as nil when setting is set to false
+	// TODO: further investigation
```
crspeller, line 1339: I like to have these with actual tickets otherwise when I read this I have no idea what the status on this TODO is.

server/app/playbook_run.go
```diff
@@ -498,6 +498,13 @@ type OwnerInfo struct {
 	Nickname  string `json:"nickname"`
 }
 
+func (o *OwnerInfo) Sanitize(options map[string]bool) {
+	if len(options) != 0 && !options["fullname"] {
```
crspeller, line 502: Is there a reason to make this complicated? Why not just leave should we or should we not sanitize for the caller to decide?

crspeller (changes requested):

server/app/playbook_run.go
trilopin: I followed the same pattern that is used at mattermost-server with Users where more fields can be sanitized based on config. If we don't plan to add more info here, we can just clean it in the getowners service call.

trilopin: I've simplified it since we have only one field and GetSanitizeOptions is not usable until we figure out what's happening with nil return.

crspeller (approved):

lieut-data (approved):

oh6hay (approved):
LGTM :+1:

PR #1578, opened 2022-11-16
WS updates run for non channel members

trilopin:
## Summary
Evolve run websocket approach:
- move all methods that are sending channel-broadcast to individual users messages
- remove channel based broadcast to avoid permissions/security issues (what if a member of the channel is not a run participant)
- optional parameters to enable adding more userIDs to be notified (like in add/moreve participants)
- avoid double getRun from database: sometimes we have an updated run and we can use it, the rest of the time we just get a fresh one.

Demo with two users that are participants but not channel members.

https://user-images.githubusercontent.com/4096774/202272512-fc0254a7-3377-48d1-a606-7f63bc182f81.mp4

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-48337

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -690,10 +690,7 @@ func (s *PlaybookRunServiceImpl) AddPostToTimeline(playbookRunID, userID, postID
 	}
 
 	s.telemetry.AddPostToTimeline(playbookRunModified, userID)
-
-	if err = s.sendPlaybookRunUpdatedWS(playbookRunModified); err != nil {
-		logrus.WithError(err).Error("failed to send websocket event")
-	}
+	s.sendPlaybookRunUpdatedWS(playbookRunModified)
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


server/app/playbook_run_service.go
```diff
@@ -2627,25 +2587,60 @@ func (s *PlaybookRunServiceImpl) newAddToTimelineDialog(playbookRuns []PlaybookR
...
+	playbookRun := sendWSOptions.PlaybookRun
+	if playbookRun == nil {
+		playbookRun, err = s.store.GetPlaybookRun(playbookRunID)
+		if err != nil {
+			logrus.WithError(err).Error("failed to retrieve playbook run when sending websocket")
+			return
+		}
+	}
+
+	uniqueUserIDs := make(map[string]bool, 0)
```
lieut-data, line 2630: Should we pre-allocate to `1 + len(sendWSOptions.AdditionalUserIDs) + len(playbookRun.ParticipantIDs)`? It's true it might be too big, but this approximates it most of the time.

server/app/playbook_run_service.go
```diff
@@ -2627,25 +2587,60 @@ func (s *PlaybookRunServiceImpl) newAddToTimelineDialog(playbookRuns []PlaybookR
...
+	uniqueUserIDs[playbookRun.OwnerUserID] = true
+	if len(sendWSOptions.AdditionalUserIDs) > 0 {
```
lieut-data, line 2632: nit: this conditional is redundant, given the `range` will simply do nothing if the slice is empty.

lieut-data (approved):
I love this! Super elegant -- three non-blocking comments below.

server/app/playbook_run_service.go
trilopin: nice catch, it was part of the previous approach, not needed anymore

crspeller (approved):

PR #1582, opened 2022-11-16
MM-48471-remove-help-link

calebroseland:
## Summary
Remove markdown help link
## Ticket Link
- Fixes https://mattermost.atlassian.net/browse/MM-48471

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


lieut-data (approved):
Agree with @trilopin, though I also share the concern that the help link as it exists today is broken, so maybe this the best middle ground until we have a common components from the threads everywhere effort?

calebroseland:
@lieut-data @trilopin, the help-pages fix is an equally simple one-liner css change, so if it's desired to keep it just a little while longer, that's a perfectly fine option, as well.

lieut-data:
Happy to defer to your judgment, @calebroseland -- thanks for owning this :)

PR #1583, opened 2022-11-17
enable empty checklist as droppable are for tasks

trilopin:
## Summary
Addtask block now is inside droppable area, so drop is accessible even with no items inside.

Addtask is not dragable, and the behavior will be the same. tasks can not be dropped below addtask.

Approaches that I tried and didn't work:
- placeholder: is already in place and is only activated when you drag items from that list
- droppable.snapshot.isDraggingOver: did not trigger because droppable height was 0 

https://user-images.githubusercontent.com/4096774/202436165-4a295d7a-38fc-4c68-9a36-ac19f8f858f4.mp4

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-48309

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/checklist/generic_checklist.tsx
+++ b/webapp/src/components/checklist/generic_checklist.tsx
@@ -75,6 +75,7 @@ const GenericChecklist = (props: Props) => {
     const keys = generateKeys(props.checklist.items.map((item) => props.id + item.title));
 
     return (
+
         <Droppable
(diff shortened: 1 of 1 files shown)


gbochora (approved):
🎉

calebroseland (approved):
Nice and straightforward!

PR #1584, opened 2022-11-17
1543 start a run modal upgrade

### LINKED ISSUE #1543: start-a-run modal
[Figma link](https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=432%3A201700)

For this task, we assume that we always have playbook id, so there's no back flow to playbooks list and we don't cover the advanced-configuration flow.

The fields that appear are:
- name template (field with initial focus)
- choice of existing/new channel 
- channel (if existing channel flow is checked)

The main idea is to re-use the current modal (first screenshot) and add the new options (second and third screenshots)

Image

Image

Image

gbochora:
## Summary
This PR updates the `Run a Playbook` modal UI. When the `linkRunToExistingChannelEnabled` feature flag is enabled, the modal shows channel configuration, and the user can modify it.
When a user runs a playbook from the command line or action menu, the modal doesn't show channel configs but starts a run based on the playbook configuration.

https://user-images.githubusercontent.com/4368372/202961712-76901f29-ea09-4772-9282-72ba11b0a42c.mp4

## Ticket Link
Fixes #1543 

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


crspeller (approved):

webapp/src/components/modals/run_playbook_modal.tsx
```diff
@@ -84,11 +104,90 @@ const RunPlaybookModal = ({
             });
     };
 
+    const channelConfigSection = (
```
trilopin, line 107: Given that this code is temporary, it might not be worth refactoring it.

For the next step of doing the final modal, I would look for ways to share this piece of code between pbe and modal.

trilopin (changes requested):
Shouldn't the context of the choices not be shown at the same time? From the designs, we just see the channel selector or the public/private selector depending on which choice we selected.

Screenshot 2022-11-18 at 12 06 49

Screenshot 2022-11-18 at 12 06 45

webapp/src/components/modals/run_playbook_modal.tsx
gbochora: Good point!

server/app/playbook_run_service.go
```diff
@@ -227,12 +227,6 @@ func (s *PlaybookRunServiceImpl) sendWebhooksOnCreation(playbookRun PlaybookRun)
 
 // CreatePlaybookRun creates a new playbook run. userID is the user who initiated the CreatePlaybookRun.
 func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb *Playbook, userID string, public bool) (*PlaybookRun, error) {
-
-	// TODO: forced until start-a-run modal can overwrite it
-	if pb != nil && pb.ChannelMode == PlaybookRunLinkExistingChannel {
-		playbookRun.ChannelID = pb.ChannelID
```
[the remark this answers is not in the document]
gbochora: Slash and post both use the same endpoint(`createPlaybookRunFromDialog`), and when the run object is created, I set the `ChannelID:   playbook.GetRunChannelID(),`. I assumed that if `ChannelID` is empty, we must create a new channel. Otherwise, we need to run in the existing channel. This is not entirely correct; the `ChannelID` might be not-empty, but at the same time, `ChannelMode` can be set to `Create a new channel`.
So, you are right; I need to add the `ChannelMode` check. Thanks, great catch!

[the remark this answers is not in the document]
gbochora: It turned out that I already did the `ChannelMode` check in the `playbook.GetRunChannelID()` function, so there is no need t modify anything.

server/api/playbook_runs.go
```diff
@@ -252,15 +253,23 @@ func (h *PlaybookRunHandler) createPlaybookRunFromDialog(c *Context, w http.Resp
...
+	if err != nil {
+		h.HandleErrorWithCode(w, c.logger, http.StatusInternalServerError, "unable to get playbook", err)
+		return
+	}
+
 	playbookRun, err := h.createPlaybookRun(
 		app.PlaybookRun{
 			OwnerUserID: request.UserId,
 			TeamID:      request.TeamId,
+			ChannelID:   playbook.GetRunChannelID(),
```
trilopin, line 266: shouldn't we use the channelid the user configured? 

I tried to do the following:
- configure playbook p1 to link to c1
- start a run modal, but configure to channel c2

run is still created linked to c1

trilopin (changes requested):
check the comment about changing existing channel during run creation

a couple of server e2e test on this matter would be super useful given how core is this behavior

abhijit-singh (changes requested):
Great work @gbochora! Haven't been able to test this locally (issues with my local dev environment that I haven't been able to solve yet) so just going by the screen recording. I had a couple of requests:
1. Let's remove the text below the run name input box that says `A channel will be created with this name`. I would suggest removing it irrespective of the choice made in the radio options below it. 
2. The spacing between the help text below run name and the radio options seems a little tight right now. With the help text gone, can we ensure that there's at least 24px space between the run name input box and the radio option to `Link run to existing channel`.

server/api/playbook_runs.go
gbochora: This endpoint is for the run dialog, and at this stage, users can't configure the channel in the UI. Because of that, I set the channel from the playbook configs. 
The bug you described was happening because of this code in the `PlaybookRunServiceImpl > CreatePlaybookRun`:
```
	if pb != nil && pb.ChannelMode == PlaybookRunLinkExistingChannel {
		playbookRun.ChannelID = pb.ChannelID
```

gbochora:
@trilopin Fixed changing existing channel issue, added server e2e tests.

@abhijit-singh resolved you requests, this is how modal looks like now:

image

trilopin (approved):
Thanks @gbochora for addressing all feedback


I've noticed that the modal is still submitted if I click Enter when I'm at the template input and there no channel filled. it would be great to cover that edge case

abhijit-singh (approved):
Looks good! Thanks @gbochora!

Issue #1586, opened 2022-11-18
rename a run


trilopin:
![Image](https://user-images.githubusercontent.com/4096774/202693389-ae9445f6-2e5e-4360-ba61-1c980155c8cf.png)
![Image](https://user-images.githubusercontent.com/4096774/202693390-945cb3ee-4b9f-4aac-9fab-362dd92fe711.png)

PR #1587, opened 2022-11-18
Gh 1586 rename run graphql

trilopin:
## Summary
Implement Run rename feature:
- inline edit via TextEdit component
- hides context menu and closest icons when edit mode is enabled
- implemented in graphql

I also migrated the summary update from rest to graphql because why not.

CleanShot 2022-11-18 at 17 51 50

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1586

## Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


gbochora (changes requested):
Great work @trilopin! The only blocker is the permission check; it breaks the favorite/unfavorite functionality when a user is not a participant.

server/api/graphql_root_run.go
```diff
@@ -100,7 +103,7 @@ func (r *RunRootResolver) UpdateRun(ctx context.Context, args struct {
 	}
 	userID := c.r.Header.Get("Mattermost-User-ID")
 
-	if err := c.permissions.RunView(userID, args.ID); err != nil {
+	if err := c.permissions.RunManageProperties(userID, args.ID); err != nil {
```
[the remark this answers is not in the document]
trilopin: Sorry, I've just reverted to RunView before seeing your comment.

Agree on confusing permission, fav/unfav should be a different mutation than UpdateRun since we are not updating a run but a run-user relationship (plus the permission confussion)

gbochora (approved):
🎉

server/app/playbook_run_service.go
```diff
@@ -2259,23 +2259,23 @@ func (s *PlaybookRunServiceImpl) ChangeCreationDate(playbookRunID string, creati
...
-
-	playbookRun, err = s.store.UpdatePlaybookRun(playbookRun)
-	if err != nil {
-		return errors.Wrap(err, "failed to update playbook run")
-	}
-
-	s.sendPlaybookRunUpdatedWS(playbookRunID, WithPlaybookRun(playbookRun))
-	return nil
-}
+// func (s *PlaybookRunServiceImpl) UpdateDescription(playbookRunID, description string) error {
```
crspeller, line 2262: Remove.

webapp/src/components/backstage/playbook_runs/playbook_run/header.tsx
```diff
@@ -53,28 +56,54 @@ export const RunHeader = ({playbookRun, playbookRunMetadata, isFollowing, hasPer
...
-                data-testid={'rhs-header-button-run-actions'}
-            />
-            <StyledCopyLink
-                id='copy-run-link-tooltip'
-                to={getSiteUrl() + '/playbooks/runs/' + playbookRun?.id}
-                tooltipMessage={formatMessage({defaultMessage: 'Copy link to run'})}
-            />
+
+            <TextEdit
+                disabled={playbookRun.current_status !== PlaybookRunStatus.InProgress}
```
crspeller, line 61: Should this also be disabled if we are not a participant?

crspeller (changes requested):
Just a couple things.

webapp/src/components/backstage/playbook_runs/playbook_run/header.tsx
trilopin: I can add that to the condition if you want, but it's already unreachable because as viewer the context menu doesn't have the rename option.

abhijit-singh (approved):
Looks good! Thanks @trilopin!

webapp/src/components/backstage/playbook_runs/playbook_run/header.tsx
crspeller: Ah right. It thought there was a direct way to access it.

crspeller (approved):

PR #1591, opened 2022-11-21
guarantee profiles load from redux for profileselector

trilopin:
## Summary
If run participants don't come in first page of useProfilesInTeam, they won't be even displayed in the assignto/profileSelector (no matter ensureProfiles). Data is in redux but not in component.

Force retrieval from redux in profileSelector.

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-48514

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/profile/profile_selector.tsx
+++ b/webapp/src/components/profile/profile_selector.tsx
@@ -2,11 +2,11 @@
 // See LICENSE.txt for license information.
 
 import React, {useEffect, useState} from 'react';
-import {useSelector} from 'react-redux';
+import {useSelector, useStore} from 'react-redux';
 import {useIntl} from 'react-intl';
 import ReactSelect, {ActionTypes, ControlProps, StylesConfig} from 'react-select';
 
-import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
+import {getCurrentUserId, makeGetProfilesByIdsAndUsernames} from 'mattermost-redux/selectors/entities/users';
 import {GlobalState} from '@mattermost/types/store';
 import {UserProfile} from '@mattermost/types/users';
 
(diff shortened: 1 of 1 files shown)


webapp/src/components/profile/profile_selector.tsx
```diff
@@ -113,7 +119,6 @@ export default function ProfileSelector(props: Props) {
         const subsetUserIds = props.userGroups?.subsetUserIds || [];
         const allUsers = await props.getAllUsers();
```
lieut-data, line 120: `getAllUsers` confused me for a while, until I realized our pattern of `useEnsureProfiles` before using this component. I presume we lifted that out to avoid multiple calls to ensure these user profiles?

webapp/src/components/profile/profile_selector.tsx
```diff
@@ -113,7 +119,6 @@ export default function ProfileSelector(props: Props) {
...
         const usersNotInSubset = allUsers.filter((user) => !subsetUserIds.find((userId) => userId === user.id));
```
lieut-data, line 121: I guess we have a similar "problem" here, in that if the user not in the subset isn't loaded by the first page of results, we can never find them?

webapp/src/components/profile/profile_selector.tsx
```diff
@@ -85,6 +85,12 @@ export default function ProfileSelector(props: Props) {
         setOpen(!isOpen);
     };
 
+    // props.userGroups?.subsetUserIds are not guaranteed to be in the page returned by props.getAllUsers
+    // but they're expected to be at redux
```
lieut-data, line 89: See below, but it took me a while to understand why this was true.

lieut-data (approved):
A handful of comments as I sought to better understand this component. Is there an opportunity for an e2e test to help ensure this behaviour going forward?

webapp/src/components/profile/profile_selector.tsx
trilopin: yes, given the fact that we don't have duplicated lists in the store for usersInRun as we have for userInchannel, I discarded the option of creating a new lock and useProfilesInRun. I rely on ensureProfiles on the top component (playbook_run.tsx, rhs_about.tsx and task_inbox) to avoid multiple concurrent calls.

webapp/src/components/profile/profile_selector.tsx
trilopin: Yes, that's why we planned to move this to server search. 

The problem also happened before for users not in the channel and not on the first page of the team. But this problem is a bit less important than not even showing the ones that are participating.

gbochora (approved):
Nice!

PR #1595, opened 2022-11-23
leave run and add participants on run creation fixes

trilopin:
### Leave run
More accurate message in confirm modal when leaving run and while not being a follower.
CleanShot 2022-11-23 at 16 48 05

### start run while inviting multiple users

* Merge timeline events for invite-users on run creation into one only add-to-run event.
* Check if invited users are part of the team (added e2e)
* Several defensive measures to avoid wrong upserts with empty lists that would end as an error.
* do not make followers those users who failed to be added 

https://user-images.githubusercontent.com/4096774/203643666-e6e4e6b2-8f04-4b9c-a97a-ac48b30e5d36.mp4

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-48594
Fixes  https://mattermost.atlassian.net/browse/MM-48601

## Checklist
- [ ] ~~Telemetry updated~~
- [X] Unit tests updated


trilopin:
here we have another case of add-means-follow additional complexity coming from LHS @lieut-data

gbochora (approved):
Looks good! Just a few optional comments.

server/app/playbook_run_service.go
```diff
@@ -379,20 +379,13 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb
...
-			usersFailedToInvite = append(usersFailedToInvite, userID)
-			continue
-		}
-
-		err := s.AddParticipants(playbookRun.ID, []string{userID}, s.configService.GetConfiguration().BotUserID, false)
-		if err != nil {
-			usersFailedToInvite = append(usersFailedToInvite, userID)
-			continue
-		}
+	usersFailedToInvite, err := s.AddParticipants(playbookRun, invitedUserIDs, s.configService.GetConfiguration().BotUserID, false)
```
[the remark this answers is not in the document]
trilopin: I wanted the team check to be there for different cases of addParticipants than start-run.

server/api/graphql_root_run.go
```diff
@@ -205,11 +205,24 @@ func (r *RunRootResolver) AddRunParticipants(ctx context.Context, args struct {
 		}
 	}
 
-	if err := c.playbookRunService.AddParticipants(args.RunID, args.UserIDs, userID, args.ForceAddToChannel); err != nil {
+	run, err := c.playbookRunService.GetPlaybookRun(args.RunID)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get run")
+	}
+
+	failed, err := c.playbookRunService.AddParticipants(run, args.UserIDs, userID, args.ForceAddToChannel)
```
lieut-data, line 213: I'd love to see the service log a warning per user vs. having to give back `failed` to a client who "probably doesn't care".

server/app/playbook_run_service.go
```diff
@@ -2976,18 +2969,27 @@ func (s *PlaybookRunServiceImpl) leaveActions(playbookRun *PlaybookRun, userID s
...
 
-func (s *PlaybookRunServiceImpl) AddParticipants(playbookRunID string, userIDs []string, requesterUserID string, forceAddToChannel bool) error {
+func (s *PlaybookRunServiceImpl) AddParticipants(playbookRun *PlaybookRun, userIDs []string, requesterUserID string, forceAddToChannel bool) ([]string, error) {
+	usersFailedToInvite := make([]string, 0)
+	usersToInvite := make([]string, 0)
+
 	if len(userIDs) == 0 {
-		return nil
+		return usersFailedToInvite, nil
 	}
```
lieut-data, line 2978: nit: I feel like this would read more clearly as `return nil, nil` and located at the top of the method.

lieut-data (changes requested):
Thanks, @trilopin! A few thoughts on how we might organize differently: let me know if I understood the changes correctly :)

server/app/playbook_run_service.go
```diff
@@ -2997,32 +2999,35 @@ func (s *PlaybookRunServiceImpl) AddParticipants(playbookRunID string, userIDs [
...
 
 	// ws send run
-	userIDs = append(userIDs, requesterUserID)
-	s.sendPlaybookRunUpdatedWS(playbookRunID, WithAdditionalUserIDs(userIDs))
+	if len(usersToInvite) > 0 {
+		usersToNotify := usersToInvite
+		usersToNotify = append(usersToNotify, requesterUserID)
+		s.sendPlaybookRunUpdatedWS(playbookRun.ID, withAdditionalUserIDs(usersToNotify))
+	}
```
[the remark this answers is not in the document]
trilopin: I don't think so: if you couldn't invite anyone, the run has the same state as before the action, and therefore, there's no need to send the websocket.

server/api/graphql_root_run.go
trilopin: I could agree on the log at service level, but we still have to return the failed (or the succeeded) items to do proper follow (don't make followers the users who failed to participate).

One more time the underlying question appears: should we keep the point "participate implies following"? Most of the weird data pass/return is because of having different actions that are linked.

Another point is, if we do believe that "participate implies following", we may have to do that in the service participation code (addparticipants or participateActions). This would be a bit more self-contained.

server/api/graphql_root_run.go
```diff
@@ -205,11 +205,24 @@ func (r *RunRootResolver) AddRunParticipants(ctx context.Context, args struct {
...
 	for _, userID := range args.UserIDs {
+		for _, f := range failed {
+			if f == userID {
+				continue OUTER
+			}
+		}
+
 		if err := c.playbookRunService.Follow(args.RunID, userID); err != nil {
 			return "", errors.Wrap(err, "failed to make participant to unfollow run")
 		}
```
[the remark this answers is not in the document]
trilopin: oh, you are saying sth similar to what I answered. Agreed.

 Let's decide to stop "following on participate" or "following as part of participate code".

server/api/graphql_root_run.go
lieut-data: I'm inclined to preserve the existing semantics that participate == follow, at least until we revisit this holistically. Does that simplify the immediate approach, @trilopin?

trilopin: agreed, at least until we rethink LHS is the best choice.

in this scenario, I'm going to move the follow side-effect when participating to the service, cleaning the interface/data.

server/app/playbook_run_service.go
```diff
@@ -2977,50 +2946,75 @@ func (s *PlaybookRunServiceImpl) leaveActions(playbookRun *PlaybookRun, userID s
...
 		return err
 	}
 
 	// ws send run
-	userIDs = append(userIDs, requesterUserID)
-	s.sendPlaybookRunUpdatedWS(playbookRunID, WithAdditionalUserIDs(userIDs))
+	if len(usersToInvite) > 0 {
+		usersToNotify := usersToInvite
+		usersToNotify = append(usersToNotify, requesterUserID)
+		s.sendPlaybookRunUpdatedWS(playbookRun.ID, withAdditionalUserIDs(usersToNotify))
```
lieut-data, line 3016: As an alternate to my comment above, would it make sense to just pass another `withAdditionalUserIDs` for the `requesterID`? (Does the abstraction here handle that itself?)

server/app/playbook_run_service.go
```diff
@@ -2977,50 +2946,75 @@ func (s *PlaybookRunServiceImpl) leaveActions(playbookRun *PlaybookRun, userID s
...
 			}
 		}
 		users = append(users, user)
+
+		// Configured actions
 		s.participateActions(playbookRun, channel, user, requesterUser, forceAddToChannel)
+
+		// Participate implies following the run
+		if err := s.Follow(playbookRunID, userID); err != nil {
+			return errors.Wrap(err, "failed to make participant to unfollow run")
```
lieut-data, line 3003:
```suggestion
			return errors.Wrap(err, "failed to make participant to follow run")
```

lieut-data (approved):
Nice cleanup, @trilopin! Just one proposed change below around slice semantics.

server/app/playbook_run_service.go
trilopin: nice point, the abstraction wasn't prepared to be called twice but it was really easy to handle it

server/app/playbook_run_service.go
```diff
@@ -2593,7 +2593,10 @@ type RunWSOption func(options *RunWSOptions)
 
 func withAdditionalUserIDs(additionalUserIDs []string) RunWSOption {
 	return func(options *RunWSOptions) {
-		options.AdditionalUserIDs = additionalUserIDs
+		if options.AdditionalUserIDs == nil {
+			options.AdditionalUserIDs = make([]string, 0)
+		}
```
lieut-data, line 2598: I think this is redundant, right? `append(nil, additionalUserIDs...)` should work fine.

lieut-data (approved):

server/app/playbook_run_service.go
trilopin: true, I'm a bit defensive with uninitialized slices/maps, sometimes unnecessarily :)

gbochora (approved):
Nice improvements!

Issue #1546, opened 2022-11-04
Start a run buttons for RHS runs list


trilopin:
For this issue, it's required to have finished the playbook-list step in the start-a-run modal.

[Figma designs](https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=523%3A215564)

![Image](https://user-images.githubusercontent.com/4096774/199214577-dcbc0c4b-2233-4317-ad5a-6e6c4f89ae39.png)

At Channels RHS, Add header with start-a-run CTA that triggers the start-a-run modal in the "Select a playbook" step (since we don't have playbook id).

There is also a start run button on the no active runs screen.

PR #1598, opened 2022-11-24
Multiple runs in a channel: Channel RHS runs list

### LINKED ISSUE #1544: channels RHS: runs list
[Figma designs](https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=523%3A215564)

server side done at https://github.com/mattermost/mattermost-plugin-playbooks/issues/1545

**out of task scope**
The following points will be covered in other issues:
- link run to another different channel 
- multiple sort options component

**Run card UI**
Image

Card with the following info:
- run name
- run owner
- run participants 
- playbook name
- last update info if not empty / status update due date if overdue
- progress bar (length based on completeness and color in due/overdue)
- dotmenu: link to run details 
 

**Runs list**

List of all runs linked to this channel:
- first list, all in-progress runs in alphabetical order
- second list: the last N finished runs sorted by updated at desc, if there are more finished runs we must offer a "show more" button to get the next batch.

Image

### LINKED ISSUE #1607: Handle no runs case for finished runs
When the playbooks RHS runs list has no finished runs it will display a blank list. We should have some message there indicating there are no finished runs.

Designs not available yet.

crspeller:
## Summary

Runs list in the channels RHS supporting multiple runs in a single channel.

#### Selecting filtering and sorting for multiple runs:
selecting_filtering

#### Pagination:
pagination

#### Only finished runs:
onlyfinished

#### Rules for auto display:
No Runs Ever -> RHS Home
Only Finished Runs -> Runs list blank state
Single active run (ignoring finished) -> Details page for that run (back button goes to runs list)
Multiple active runs -> Runs list

### Graphql Pagination
GraphQL pagination has been implemented for runs list. The pattern comes from https://graphql.org/learn/pagination/ which is often referred to as Relay style as it's the preferred style for the Relay GraphQL client. It's probably over complicated for our use case at the moment, but it's the closest thing to a standard pagination style GraphQL has and it made it really easy to make Apollo work with it.
Everything is called "cursors" but the actual implementation is still just offset and limit by a different name.

### RHS Title remote render
I didn't like the way we where handling the awkwardness of having to render the title of the RHS separate from the body. In this new style we use a portal to wave the problem away with a simple wrapper, this way each RHS page can render its title itself. We also avoid abusing redux for this. Have a look in rhs_title_remote_render.tsx to see how it's done.

## Follow up tickets
TODO: Make these tickets. 
- [ ] Give feedback button in RHS runs list #1600 
- [ ] Dot menus on RHS runs list items #1601
- [ ] Start run buttons for RHS runs list #1546 
- [ ] Fancy title for RHS runs list #1602
- [ ] Switch sort icon and other improvements for RHS list sorting #1603
- [ ] Progress bar on items in RHS runs list #1604
- [ ] Loading states for RHS runs list #1605
- [ ] GraphQL migration product runs list to paginated API #1606
- [ ] No runs case for Finished runs. Fixes #1607

## Ticket Link
Fixes #1544


server/api/graphql.go
```diff
@@ -62,7 +62,7 @@ func NewGraphQLHandler(
 
 	if !configService.IsConfiguredForDevelopmentAndTesting() {
 		opts = append(opts,
-			graphql.MaxDepth(4),
+			graphql.MaxDepth(8),
```
crspeller, line 65: Queries are deeper now so had to bump this limit.

webapp/package.json
```diff
@@ -103,7 +103,7 @@
     "webpack-dev-server": "4.7.2"
   },
   "dependencies": {
-    "@apollo/client": "3.5.10",
+    "@apollo/client": "3.7.1",
```
crspeller, line 106: Ran into what I thought was an Apollo bug, turns out it was not but I kept the update.

webapp/src/graphql/runs.graphql
```diff
@@ -8,6 +8,73 @@ query Run($id: String!) {
     }
 }
 
+fragment RHSRunFields on Run {
```
crspeller, line 11: Finally a use for fragments!

webapp/src/index.tsx
```diff
@@ -317,7 +324,15 @@ export default class Plugin {
         };
         const graphqlClient = new ApolloClient({
             link: new HttpLink({fetch: graphqlFetch}),
-            cache: new InMemoryCache(),
+            cache: new InMemoryCache({
+                typePolicies: {
+                    Query: {
+                        fields: {
+                            runs: relayStylePagination(['teamID', 'sort', 'direction', 'statuses', 'participantOrFollowerID', 'channelID']),
```
crspeller, line 331: This I don't like. You have to list here all the fields that should make two queries unique. Otherwise the default relay implementation for Apollo combines them. 

You can also do the paging yourself, but this makes it so much easier on the component side.

webapp/src/rhs_title_remote_render.tsx
```diff
@@ -0,0 +1,30 @@
...
+    );
+};
+
+export const RHSTitleRemoteRender = (props: {children: React.ReactNode}) => {
+    const container = document.getElementById(placeholderID);
+    if (!container) {
+        return null;
+    }
+
+    return ReactDOM.createPortal(props.children, container);
```
crspeller, line 29: Here is the magic. Just create a portal you can render too. That's it.

webapp/src/components/checklist_item/assign_to.tsx
```diff
@@ -27,7 +27,7 @@ const AssignTo = (props: AssignedToProps) => {
     const [profileSelectorToggle, setProfileSelectorToggle] = useState(false);
 
     const resetAssignee = () => {
-        props.onSelectedChange();
+        props.onSelectedChange?.();
```
trilopin, line 30: why is this change neccesary?

webapp/src/components/rhs/rhs_main.tsx
```diff
@@ -7,81 +7,161 @@ import {useDispatch, useSelector} from 'react-redux';
...
+        });
+    };
+
+    const getMoreFinished = () => {
+        finishedResult.fetchMore({
+            variables: {
+                after: finishedResult.data?.runs.pageInfo.endCursor,
+            },
+        });
+    };
```
trilopin, line 61: this is where all prev effort pays off 🎉

webapp/src/components/rhs/rhs_run_list.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+import Scrollbars from 'react-custom-scrollbars';
+
+import {DateTime} from 'luxon';
+
+import Profile from 'src/components/profile/profile';
+
+import DotMenu, {DotMenuButton, DropdownMenuItem, TitleButton} from 'src/components/dot_menu';
+
+import {SecondaryButton, TertiaryButton} from '../assets/buttons';
```
trilopin, line 18: optional nit: 

I'd want all imports to be absolute. They are more resilient to refactors or code re-use. They also. are more dense in blocks and some blank lines are not needed.

webapp/src/components/rhs/rhs_run_list.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+import {RHSTitleRemoteRender} from 'src/rhs_title_remote_render';
+
+import ClipboardChecklist from '../assets/illustrations/clipboard_checklist_svg';
+
+import {UserList} from './rhs_participants';
+import {RHSTitleText} from './rhs_title_common';
+
+interface PlaybookToDisplay {
```
trilopin, line 27: optional nit:

I'd like to put styled components after the actual component.

webapp/src/components/rhs/rhs_run_list.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+    return (
+        <>
+            <RHSTitleRemoteRender>
+                <RHSTitleText>
+                    {/* product name; don't translate */}
+                    {/* eslint-disable-next-line formatjs/no-literal-string-in-jsx */}
+                    {'Playbooks'}
+                </RHSTitleText>
+            </RHSTitleRemoteRender>
```
trilopin, line 165: I like the trick with the portalling, the previous title was very complex and now it's cleaner with even more functionality covered. ❤️

webapp/src/slash_command.ts
```diff
@@ -52,10 +47,6 @@ export function makeSlashCommandHook(store: Store) {
                 store.dispatch(toggleRHS());
             }
 
-            if (inPlaybookRunChannel(state) && currentRHSState(state) !== RHSState.ViewingPlaybookRun) {
-                store.dispatch(setRHSViewingPlaybookRun());
```
trilopin, line 56: why is this removed?

trilopin (changes requested):
What an effort here, love the cursor pattern and the portalling for the titles.

I left some questions or comments about nits.

From manual QA I've noticed the following details: 

1. I miss the blue check in the sort dropdown to indicate the active choice.

Screenshot 2022-11-24 at 12 39 04

CleanShot 2022-11-24 at 12 39 29

2. last update in card only reflect timeline changes but some changes don't have any event (update summary for example)

3. could we have missing zero case for empty lists (finished)

or maybe that option should be disabled.
 
Screenshot 2022-11-24 at 12 48 10


4. finish run from rhs

https://user-images.githubusercontent.com/4096774/203777925-7c869414-fce3-4da5-8805-0b5216112966.mp4

webapp/src/index.tsx
crspeller: Agreed.

server/api/graphql_run.go
```diff
@@ -133,6 +134,85 @@ func (r *RunResolver) Metadata(ctx context.Context) (*MetadataResolver, error) {
...
 
+func (r *RunResolver) Playbook(ctx context.Context) (*PlaybookResolver, error) {
+	return getGraphqlPlaybook(ctx, r.PlaybookID)
+}
+
+func (r *RunResolver) LastUpdatedAt(ctx context.Context) float64 {
+	if len(r.PlaybookRun.TimelineEvents) < 1 {
+		return float64(r.PlaybookRun.CreateAt)
+	}
+	return float64(r.PlaybookRun.TimelineEvents[len(r.PlaybookRun.TimelineEvents)-1].EventAt)
```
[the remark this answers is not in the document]
crspeller: Unfortunately Run.UpdateAt isn't a thing. This goes with your observation above that the summary updating doesn't reflect on this value. 

There is actually one for just the summary so we could fix that, but it wouldn't fix it for stuff like editing a checklist item which doesn't have an event. 

We should probably think though what events we want to count here so I will make a follow up ticket for this.

server/api/schema.graphqls
```diff
@@ -218,10 +228,23 @@ type Run {
...
 }
 
+type RunConnection {
+	totalCount: Int!
+	edges: [RunEdge!]!
+	pageInfo: PageInfo!
+}
+
+type RunEdge {
+	cursor: String!
```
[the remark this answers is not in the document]
crspeller: Its stored as an integer because it is just the page number. It's always the page number for us. It's returned as a string by GraphQL because it's supposed to be opaque to the client. (actually there was a recommendation to base64 encode it). That way you can change paging patterns and the client is none the wiser.

webapp/src/components/checklist_item/assign_to.tsx
crspeller: I don't remember 🤷

webapp/src/components/rhs/rhs_main.tsx
crspeller: Yup!

webapp/src/components/rhs/rhs_run_list.tsx
crspeller: Agreed. Wish there was an eslint rule that would cover this. When these get auto imported it does relative for some reason.

webapp/src/slash_command.ts
crspeller: No need to set the state in redux anymore. Handled internally.

trilopin:
> 1. I miss the blue check in the sort dropdown to indicate the active choice.
> 
> Screenshot 2022-11-24 at 12 39 04

I've just noticed that the title "SORT RUNS BY" is also missing there.

crspeller:
@trilopin 

1. Will add a separate ticket for that. 
2. See inline comment.
3. Separate ticket.
4. Can't reproduce this. Is there something specific to your environment causing that? I can finish runs no problem admins and non-admins. The one thing I did notice doing this is that if the action to remove you from a channel you are in is enabled for a run in a channel with multiple runs, it will remove you from the channel despite being in other runs! This is bad.

trilopin:
> @trilopin
> 
> 1. Will add a separate ticket for that.

ok

> 2. See inline comment.

oh true, there's a bigger pending conversation here. I queued item for next sync

> 3. Separate ticket.

ok

> 4. Can't reproduce this. Is there something specific to your environment causing that? I can finish runs no problem admins and non-admins. The one thing I did notice doing this is that if the action to remove you from a channel you are in is enabled for a run in a channel with multiple runs, it will remove you from the channel despite being in other runs! This is bad.

I'll try again later today

🎉

trilopin:
More feedback:

1. finished runs do nothing when clicked
CleanShot 2022-11-25 at 12 41 49

2. slash commands interaction is erratic

- messages in the channel are not accurate (mentions to "this run" while we have 9)
- don't give context about which run is affected
- finish through slash command while seeing the list works for one run (don't know which one)

should we block commands while the list is being showed and keep them just for when the run details is what we see in RHS?

CleanShot 2022-11-25 at 12 39 40

crspeller:
1. Fixed. 

2. Made a draft ticket in the project. (this one needs more discussion)

tests-e2e/cypress/integration/channels/rhs/list_spec.js
```diff
@@ -0,0 +1,103 @@
...
+        cy.apiLogin(testUser);
+
+        // # Size the viewport to show the RHS without covering posts.
+        cy.viewport('macbook-13');
+
+        // # Navigate directly to the application and the playbook run channel
+        cy.visit(`/${testTeam.name}/channels/${testChannel.name}`);
+
+        // # Wait because cypress
+        cy.wait(1000);
```
trilopin, line 77: optional: I'd try to replace time based waits with element based waits.

in other places we have:
```js
                // # Wait until the channel loads enough to show the post textbox.
                cy.get('#post-create').should('exist');
```

webapp/src/components/rhs/rhs_main.tsx
```diff
@@ -7,81 +7,168 @@ import {useDispatch, useSelector} from 'react-redux';
...
 
-import {receivedTeamPlaybookRuns, setRHSOpen, setRHSViewingPlaybookRun, setRHSViewingList} from 'src/actions';
-import RHSHome from 'src/components/rhs/rhs_home';
-import {currentRHSState, inPlaybookRunChannel} from 'src/selectors';
-import {RHSState} from 'src/types/rhs';
-import RHSWelcomeView from 'src/components/rhs/rhs_welcome_view';
+import {setRHSOpen} from 'src/actions';
 import RHSRunDetails from 'src/components/rhs/rhs_run_details';
-import {fetchPlaybookRunByChannel} from 'src/client';
+
```
trilopin, line 12: would make sense to add here the algorithm for display in a comment?

Rules for auto display (from your pr desc):
- No Runs Ever -> RHS Home
- Only Finished Runs -> Runs list blank state
- Single active run (ignoring finished) -> Details page for that run (back button goes to runs list)
- Multiple active runs -> Runs list

trilopin (approved):
great job 🎉

gbochora (approved):
Massive amount of work and lots of great stuff! I've left a few questions and optional suggestions.

1. After a run is started in the channel, there is no longer way to navigate to RHS > Your Playbooks list page. Are we ok with that? It doesn't seem like a huge loss, but I still wanted to ensure it.
2. If a channel has a single run and it's finished, should we, by default, show the run details in the RHS?
3. Should we make the alphabetic sort ascending? Now it sorts descending.
image

server/api/graphql_root_run.go
```diff
@@ -36,10 +36,13 @@ func (r *RunRootResolver) Run(ctx context.Context, args struct {
 func (r *RunRootResolver) Runs(ctx context.Context, args struct {
 	TeamID                  string
 	Sort                    string
+	Direction               string
 	Statuses                []string
 	ParticipantOrFollowerID string
 	ChannelID               string
-}) ([]*RunResolver, error) {
+	First                   *int32
+	After                   *string
```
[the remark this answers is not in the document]
crspeller: It's the spec: https://graphql.org/learn/pagination/
The idea is to make it opaque to the client. So you can use any pagination pattern.

webapp/src/components/rhs/rhs_main.tsx
```diff
@@ -7,81 +7,168 @@ import {useDispatch, useSelector} from 'react-redux';
...
+    const getMoreFinished = () => {
+        finishedResult.fetchMore({
+            variables: {
+                after: finishedResult.data?.runs.pageInfo.endCursor,
+            },
+        });
+    };
+
+    const error = inProgressResult.error || finishedResult.error;
+    const fetchFinished = () => (true);
```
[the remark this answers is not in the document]
crspeller: Oh. That's a remnant from some performance optimization I was going to do. Removed.

crspeller:
1. I believe so. @abhijit-singh can confirm. 
2. I don't think so. It does make it a bit more awkward to get to, but I guess we are assuming you would rather show a new run. Thoughts @abhijit-singh ?
3. Certainly. Fixed.

abhijit-singh:
Amazing work! Appreciate all the details and interactions, and thank you for making follow up tickets for pending items. 

To answer a few questions from the comments: 
1. > there is no longer way to navigate to RHS > Your Playbooks list page.

Yep that's not a problem. We'll be able to get rid of it completely with the upcoming `Checklists` effort.

2. > If a channel has a single run and it's finished, should we, by default, show the run details in the RHS?

The first thing we'd want you to do is `Start a new run` in this case. We could possibly make it easier to get to the finished run but let's do that as an iteration if we feel the need.

abhijit-singh (approved):
Looks good to go ahead! Just one non-blocking request apart from the follow-up items already listed — the font of the title `Runs in progress`/`Finished runs` seems incorrect. It should be 
```
/* Heading 200 */
font-family: 'Metropolis';
font-style: normal;
font-weight: 600;
```

crspeller:
@abhijit-singh Ahh I didn't notice the font changed. Will fix in follow up PR: https://github.com/mattermost/mattermost-plugin-playbooks/pull/1609

webapp/src/components/rhs/rhs_run_list.tsx
trilopin: Screenshot 2022-11-30 at 11 54 12

if you use vscode, you can tell that auto-imported files are absolute from the source configured in tsconfig.

trilopin: Also, proposal for eslint rule https://github.com/mattermost/mattermost-plugin-playbooks/pull/1624

PR #1611, opened 2022-11-25
MM-47749: Task Actions

ashishbhate:
## Summary

- Task Actions, https://www.figma.com/file/tnkgFLUnEijcjF8Tq1v4Bu/%F0%9F%A6%8A-Playbooks-%3C%3E-Gitlab?node-id=96%3A77967&t=pxjdUUoumCWbTNXN-0
  This PR adds the Task Actions feature. For more details about the feature see the linked figma designs and Jira ticket. For more details about the implementation see the next comment.

## Ticket Link

- https://mattermost.atlassian.net/browse/MM-47749

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


ashishbhate:
### Notes about Task Actions:

1. A `TaskAction` is made up of a `Trigger` and `Actions`.  When a `Trigger` is triggered, `Actions` are executed". The backend supports having more than one action executed by a trigger, but currently we only have one type of action implemented and the UI only supports a single action (more on this later)

```Go
type TaskAction struct {
	Trigger Trigger  `json:"trigger"`
	Actions []Action `json:"actions"`
}
```

2. A checklist item (aka task) can have multiple `TaskActions`. Currently the UI only shows one `TaskAction`. All the supporting code already exists. This was done to speed up the implementation/delivery of this feature.

3. A checklist item (aka task) can have multiple `TaskActions`, and each `TaskAction` can have multiple `Actions`. This makes `TaskActions` fully composable, and flexible enough to execute complex, and conditional logic. 

6. One type of trigger has been implemented: `keywords_by_users` - This `Trigger` is triggered when a message containing any of the specified "keywords" is posted by the any of the specified "users". 

7. One type of action has been implemented: `mark_item_as_done` - This `Action` marks a checklist item as done. 

8. `Triggers` and `Actions` share a simple structure (implemented with stronger more specific types on the backend for safety)

```Go
struct {
	Type string `json:"type"`
	Payload string `json:"payload"`
}
```

The `Type` identifies the type of the action or trigger.
The `Payload` contains action/trigger specific configuration/data


### Notes about this PR/Implementation

1. The frontend definitely needs polish. I've tried to reuse some components from "Channel Actions". The structure and implementation of these components are a close match, but don't fit "Task Actions" exactly, so you will notice some oddities in the way the components are laid out. Some of these existing components need tweaking to make them more reusable and stylistically configurable. I've only made small changes where absolutely required. Other places require larger changes that can come in follow-up PRs (e.g. the sizes of the selected user labels in the Profile Autocomplete input)

2. The GraphQL types could possibly be implemented better in terms of Nullability. I struggled to get the three type systems (Typescript, GraphQL, and Go) to play well together. 

3. Tests! There are none unfortunately. I'm working on tests for the backend. I will push these as the PR is being reviewed. I would love some assistance with tests on the frontend.

4. Frontend coding style - My VSCode isn't linting or re-formatting frontend code correctly. I'm not sure why. `make check-style` passes, but feel free to leave comments on the code style that I may have missed.

### Things left to do

- [ ] Support for Message Attachments - WIP
- [ ] Tests - WIP coming soon
- [ ] Telemetry - @lieut-data I still haven't figured out telemetry (or metrics). Is this a blocker for this PR?
- [ ] Feature Flag - @lieut-data Does this feature need to go behind a feature flag?

server/api/graphql.go
```diff
@@ -62,7 +62,7 @@ func NewGraphQLHandler(
 
 	if !configService.IsConfiguredForDevelopmentAndTesting() {
 		opts = append(opts,
-			graphql.MaxDepth(4),
+			graphql.MaxDepth(6),
```
lieut-data, line 65: cc @crspeller, what are the wider ramifications of allowing this?

server/api/graphql_playbook.go
```diff
@@ -113,20 +114,86 @@ func (r *ChecklistItemResolver) DueDate() float64 {
...
+	State            string            `json:"state"`
+	StateModified    float64           `json:"state_modified"`
+	AssigneeID       string            `json:"assignee_id"`
+	AssigneeModified float64           `json:"assignee_modified"`
+	Command          string            `json:"command"`
+	CommandLastRun   float64           `json:"command_last_run"`
+	Description      string            `json:"description"`
+	LastSkipped      float64           `json:"delete_at"`
+	DueDate          float64           `json:"due_date"`
+	TaskActions      *[]app.TaskAction `json:"task_actions"`
```
lieut-data, line 198: Is `TaskActions` as a pointer to a slice necessary?

server/api/graphql_root_playbook.go
```diff
@@ -237,6 +237,9 @@ func (r *PlaybookRootResolver) UpdatePlaybook(ctx context.Context, args struct {
 	// Not optimal graphql. Stopgap measure. Should be updated seperately.
 	if args.Updates.Checklists != nil {
 		cleanUpUpdateChecklist(*args.Updates.Checklists)
+		if err := validateUpdateTaskActions(*args.Updates.Checklists); err != nil {
+			return "", errors.Wrapf(err, "failed to marshal checklist in graphql json for playbook id: '%s'", args.ID)
```
lieut-data, line 241: Should we update the error message here to indicate the failure to validate vs. marshal?

server/api/graphql_root_playbook.go
```diff
@@ -475,3 +478,25 @@ func cleanUpUpdateChecklist(checklists []UpdateChecklist) {
 		}
 	}
 }
+
+// validateUpdateTaskActions validates the taskactions in the given checklist
+// NOTE: Any changes to this function must be made to function 'validateTaskActions' for the REST endpoint.
+func validateUpdateTaskActions(checklists []UpdateChecklist) error {
+	for listIndex := range checklists {
```
lieut-data, line 485: Do we need to iterate using the indices, or would it be clearer to just use the values directly?

server/api/playbooks.go
```diff
@@ -252,6 +260,22 @@ func cleanUpChecklist(checklists []app.Checklist) {
 	}
 }
 
+// validateTaskActions validates the taskactions in the given checklist
+// NOTE: Any changes to this function must be made to function 'validateUpdateTaskActions' for the GraphQL endpoint.
```
lieut-data, line 264: cc @crspeller / @trilopin, it looks like Ashish had to make changes to both our GraphQL and REST endpoints, adding complexity. Thoughts on creating tasks to migrate more of our REST code to GraphQL to avoid this?

server/app/playbook.go
```diff
@@ -253,6 +253,9 @@ type ChecklistItem struct {
 	// of the checklist item. 0 if not set.
 	// Playbook can have only relative timstamp, run can have only absolute timestamp.
 	DueDate int64 `json:"due_date" export:"due_date"`
+
+	// TaskActions is an array of all the task actions associated with this task.
+	TaskActions []TaskAction `json:"task_actions" export:"-"`
```
lieut-data, line 258: Is `export` omitted here on purpose? I guess there may be explicit `UserIDs` that won't match? cc @crspeller, thoughts?

server/app/playbook_run_service.go
```diff
@@ -3468,6 +3489,69 @@ func (s *PlaybookRunServiceImpl) dmPostToUsersWithPermission(users []string, pos
...
+	// get playbook from post
+	runID, err := s.GetPlaybookRunIDForChannel(post.ChannelId)
+	if err != nil {
+		if errors.Is(err, ErrNotFound) {
+			return
+		}
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"post_id":    post.Id,
+			"channel_id": post.ChannelId,
+		}).Error("unable retrieve run ID from post")
```
lieut-data, line 3502: I'm hesitant about logging anything on `MessageHasBeenPosted`: this could easily overwhelm the server logs if the issue is somehow Playbooks specific. Thoughts? Would it make sense to only `Debug` this instead?

server/app/playbook_run_service.go
```diff
@@ -3468,6 +3489,69 @@ func (s *PlaybookRunServiceImpl) dmPostToUsersWithPermission(users []string, pos
...
+		return
+	}
+
+	for checklistNum, checklist := range run.Checklists {
+		for itemNum, item := range checklist.Items {
+			for _, ta := range item.TaskActions {
+				if ta.Trigger.Type == KeywordsByUsersTriggerType {
+					t, err := NewKeywordsByUsersTrigger(ta.Trigger)
+					if err != nil {
+						logrus.WithError(err).Error("unable to decode trigger")
```
lieut-data, line 3520: Can we include some metadata with this log that identifies the task action in question?

server/app/playbook_run_service.go
```diff
@@ -3468,6 +3489,69 @@ func (s *PlaybookRunServiceImpl) dmPostToUsersWithPermission(users []string, pos
...
+						return
+					}
+					if t.IsTriggered((post)) {
```
lieut-data, line 3523: Is `((post))` intentional?

server/app/playbook_run_service.go
```diff
@@ -3468,6 +3489,69 @@ func (s *PlaybookRunServiceImpl) dmPostToUsersWithPermission(users []string, pos
...
+						err := s.doActions(ta.Actions, runID, post.UserId, ChecklistItemStateClosed, checklistNum, itemNum)
+						if err != nil {
+							logrus.WithError(err).Error("can't process task actions")
```
lieut-data, line 3526: As above, can we include some metadata in the log?

server/app/playbook_run_service.go
```diff
@@ -3468,6 +3489,69 @@ func (s *PlaybookRunServiceImpl) dmPostToUsersWithPermission(users []string, pos
...
+	for _, action := range taskActions {
+		if action.Type == MarkItemAsDoneActionType {
+			a, err := NewMarkIteamAsDoneAction(action)
+			if err != nil {
+				return errors.Wrapf(err, "unable to decode action")
+			}
+			if a.Payload.Enabled {
+				if err := s.ModifyCheckedState(runID, userID, ChecklistItemStateClosed, checklistNum, itemNum); err != nil {
+					if err != nil {
+						return errors.Wrapf(err, "can't mark iteam as done")
```
lieut-data, line 3546:
```suggestion
						return errors.Wrapf(err, "can't mark item as done")
```

server/app/task_actions.go
```diff
@@ -0,0 +1,153 @@
...
+	ValidTaskTriggerTypes = []TaskTriggerType{
+		KeywordsByUsersTriggerType,
+	}
+
+	ValidTaskActionTypes = []TaskActionType{
+		MarkItemAsDoneActionType,
+	}
+)
+
+// Triggers
```
lieut-data, line 51: nit: This comment seems incomplete?

lieut-data (changes requested):
Nice work, @ashishbhate! I've focused on the server-side changes with my review. A few comments for discussion.

server/api/graphql.go
ashishbhate: Specifying a lower MaxDepth is a good idea to prevent abuse and keep the application resource usage under control (deeper queries mean more work for the app). 

UUIC the graphQL AST is parsed and parts of the queries requesting data deeper than the max depth are like a no-op.

Without this bump we can't get TaskActions when requesting Playbook Runs from the GraphQL query.

server/app/playbook.go
ashishbhate: Yep, there is comment at the top of this struct

```
// The tag export supports the export/import feature. If the field makes sense for export, the value should be
// the JSON name of the item in the export format. If the field should not be exported the value should be "-".
// Fields should be exported if they are not server specific like InvitedUserIDs or are tracking metadata like CreateAt.
```

UserIds are server specific, and could be present in task actions so I opted not to export the field.

server/app/task_actions.go
ashishbhate: Nope, just indicates that this section of the file contains triggers. There's a similar comment on line 105 for actions and 128 for validators. 
0/5 if they're helpful or not

server/api/graphql_playbook.go
ashishbhate: This is a consequence of the GraphQL nullability issue I've spoken about. This needs to be a pointer for the null GraphQL type for the updator.
I think I might know how to make all these types work better together, but I really haven't had the time to try things out. Would you be open to this coming as a later improvement?

server/app/playbook_run_service.go
ashishbhate: Nope, bad code clean-up on refactor on my part :)

server/api/graphql_playbook.go
lieut-data: I'm fine with that, as I believe the change is all internal to the server.

server/app/playbook_run_service.go
lieut-data: Thanks for deep diving there: indeed it looks like my fears are unfounded given an existing pattern already exists. Let's leave it as you have it :)

server/api/graphql.go
crspeller: What he said + I already bumped the value higher in my PR: https://github.com/mattermost/mattermost-plugin-playbooks/pull/1598/files#diff-2db6c89c7fc8f9ca00844febd6f949939b62019f64b13251eff2af8bbb33822fL65

server/api/playbooks.go
crspeller: Yes we should. I want to deprecate the REST and use GraphQL api for most things.

server/app/playbook.go
crspeller: Agree the userID are server specific and won't export. It would be nice if we defined specific export behavior for this field so cross server applicable stuff would transfer. Not sure if that is possible, not familiar enough with this feature.

webapp/src/components/checklist_item/task_actions_modal.tsx
```diff
@@ -0,0 +1,234 @@
...
+        newTaskAction.trigger = {
+            type: KeywordsByUsersTriggerType,
+            payload: JSON.stringify(newTriggerPayload),
+        };
+
+        const newActionPayload = {enabled: newIsEnabled};
+        newTaskAction.actions = [{
+            type: MarkItemAsDoneActionType,
+            payload: JSON.stringify(newActionPayload),
+        }];
```
calebroseland, line 89: To reduce type casting, improve readability IMHO (2/5)

```suggestion
        const newTaskAction: TaskAction = {
            trigger: {
                type: KeywordsByUsersTriggerType,
                payload: JSON.stringify({keywords: newKeywords, user_ids: newUserIDs}),
            },
            actions: [
                {
                    type: MarkItemAsDoneActionType,
                    payload: JSON.stringify({enabled: newIsEnabled}),
                },
            ],
        };
```

webapp/src/components/checklist_item/task_actions_modal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const emptyTask = {} as TaskActionType;
+    const taskAction = (taskActions && (taskActions.length > 0)) ? taskActions[0] : emptyTask;
+
+    // we only have one kind of trigger
+    const triggerType = taskAction.trigger?.type ? taskAction.trigger.type : KeywordsByUsersTriggerType;
+    const triggerPayload = taskAction.trigger?.payload ? JSON.parse(taskAction.trigger.payload) : keywordsTriggerEmptyPayload;
+
+    // we currently only support one action per trigger
+    const actionType = (taskAction.actions?.length > 0 && taskAction.actions[0]?.type) ? taskAction.actions[0].type : MarkItemAsDoneActionType;
+    const actionPayload = (taskAction.actions?.length > 0 && taskAction.actions[0]?.payload) ? JSON.parse(taskAction.actions[0].payload) : markAsDoneEmptyPayload;
```
calebroseland, line 53: IMHO, it would be nice to see this pre-parsed/decoded at a higher level—but that may not be realistic for this MVP.

So for containment, could we wrap this parsing logic in a func or custom hook? (0/5 on the de-structuring)

```suggestion
    const {keywords, user_ids: userIds} = keywordsTriggerPayloadFromTaskAction(taskAction);
    const {enabled} = markAsDonePayloadFromTaskAction(taskAction);
```

```tsx
type KeywordsTriggerPayload = {keywords: string[]; user_ids: string[];}
export const keywordsTriggerEmptyPayload: KeywordsTriggerPayload = {keywords: [], user_ids: []};

type MarkAsDonePayload = {enabled: boolean;};
export const markAsDoneEmptyPayload: MarkAsDonePayload = {enabled: false};

const keywordsTriggerPayloadFromTaskAction = (taskAction: TaskAction): KeywordsTriggerPayload => {
    // we only have one kind of trigger
    const triggerType = taskAction.trigger?.type ? taskAction.trigger.type : KeywordsByUsersTriggerType;
    const triggerPayload: KeywordsTriggerPayload = taskAction.trigger?.payload ? JSON.parse(taskAction.trigger.payload) : keywordsTriggerEmptyPayload;

    return triggerPayload;
};

const markAsDonePayloadFromTaskAction = (taskAction: TaskAction): MarkAsDonePayload => {
    // we currently only support one action per trigger
    const actionType = (taskAction.actions?.length > 0 && taskAction.actions[0]?.type) ? taskAction.actions[0].type : MarkItemAsDoneActionType;
    const actionPayload: MarkAsDonePayload = (taskAction.actions?.length > 0 && taskAction.actions[0]?.payload) ? JSON.parse(taskAction.actions[0].payload) : markAsDoneEmptyPayload;
    return actionPayload;
};

// const TaskActionsModal(...) => { ...
```

calebroseland (changes requested):
Overall looks pretty good to me! Just a few nits, but nothing major.


Long-term, it would be nice to see the TS typing for `TaskAction` mature. Perhaps generics could be added down the line to increase clarity and `trigger`-`action[]` relationships.

calebroseland:
Pushed a small polish/fix, please feel free to rewind/revert/revise if necessary.

1). Task Actions disabled when task is skipped, run is finished, or if marked readonly (Copy-pasted from the other controls)
2). Added hover state
CleanShot 2022-11-28 at 23 14 41@2x

calebroseland:
Not crucial, could be done later, but nice a UI nicety would be to have pre-filled options for the `Posted by` option. 

Example from Assignee dropdown shown.

image

webapp/src/components/backstage/profile_autocomplete.tsx
```diff
@@ -143,13 +145,14 @@ const ProfileAutocomplete = (props: Props) => {
     return (
         <StyledAsyncSelect
             id={'profile-autocomplete'}
-            autoFocus={true}
+            autoFocus={!props.disableAutoFocus && true}
```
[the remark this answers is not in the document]
ashishbhate: TIL about `??` the "nullish coalescing operator"

webapp/src/components/checklist_item/task_actions_modal.tsx
```diff
@@ -0,0 +1,234 @@
...
+import {getUsers} from 'mattermost-redux/selectors/entities/common';
+import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
+
+import ActionsModal from 'src/components/actions_modal';
+import Action from 'src/components/actions_modal_action';
+import ProfileAutocomplete from 'src/components/backstage/profile_autocomplete';
+import Trigger, {TriggerKeywords} from 'src/components/actions_modal_trigger';
+import GenericModal from 'src/components/widgets/generic_modal';
+import {getRun} from 'src/selectors';
+import {TaskAction as TaskActionType} from 'src/types/playbook';
```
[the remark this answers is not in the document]
ashishbhate: We don't necessarily need it, but there is a variable called `taskAction` and I wanted make it easier to differentiate between the two. 
I'm 0/5. Let me know which way you prefer it.

ashishbhate:
@lieut-data I've pushed the server review suggestions

@calebroseland Thanks so much for the quick review and also for pushing your improvements.
I'm keeping track of all the UI polish issues. The feature team has a sync meeting tomorrow (Wednesday). Based on the discussions tomorrow, I'll have a better understanding of how much of the UI polish is doable now vs punting them for later.

server/app/playbook.go
lieut-data: Yeah, I think it would have to be an all or nothing here, since a trigger that "loses" its user id filter might be annoying, as would a partially exported set of task actions for triggers without user ids.

lieut-data (approved):
Looks awesome, @ashishbhate! I'm aiming to help contribute E2E tests starting this afternoon my time, but this PR is good to go from my perspective.

server/api/graphql_playbook.go
crspeller, line 198: If we modified the types we wouldn't have to have it like this. You can see it working on the struct above.

crspeller (changes requested):

server/api/schema.graphqls
```diff
@@ -89,6 +89,22 @@ input ChecklistItemUpdates {
 	command: String!
 	commandLastRun: Float!
 	dueDate: Float!
+	taskActions: [TaskActionUpdates!]
```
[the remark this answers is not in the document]
ashishbhate: IIRC, I did this way to prevent empty task actions json arrays being stored in the DB for every checklist for every checklist item update.

webapp/src/client.ts
```diff
@@ -408,6 +409,14 @@ export async function clientSetChecklistItemCommand(playbookRunID: string, check
     return data;
 }
 
+export async function clientSetChecklistItemTaskActions(playbookRunID: string, checklistNum: number, itemNum: number, taskActions: TaskAction[]) {
+    const data = await doPut(`${apiUrl}/runs/${playbookRunID}/checklists/${checklistNum}/item/${itemNum}/taskactions`,
```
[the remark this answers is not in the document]
ashishbhate: Because that is the existing pattern in the code - webapp/src/components/checklist_item/checklist_item.tsx

It appears Playbooks and runs are fetched through GraphQL, but are sometimes updated with GraphQL, and some other parts are updated through REST. To keep it consistent with the rest of the code where this function is used (see file above), where the due date, command, etc. are all updated through the REST API, I implemented the REST API as well.

Also, I have no idea about the usage of the Playbooks API. Since there were REST endpoints to update other parts of a checklist item, it made sense to write it for task actions as well for the sake of completeness.  

I didn't want to break any existing patterns or set new GraphQL patterns for an unfamiliar code base that it going through a piecemeal migration from REST to GraphQL.

server/api/schema.graphqls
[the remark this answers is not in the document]
crspeller: Yes the way checklist items are stored is a bit awkward. I see why it's like this.

webapp/src/client.ts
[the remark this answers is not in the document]
crspeller: I understand following existing patterns. Could we have your help removing the REST API you added and just use the GraphQL one you added instead? GraphQL is the way forward for us now.

server/api/graphql_playbook.go
crspeller: Never mind I see the difference here now.

webapp/src/client.ts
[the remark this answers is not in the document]
ashishbhate: Yep, I will update to use the GraphQL updator, and remove the new REST endpoint for task actions.

This PR also includes an addition to a validator in the the `createPlaybook` REST endpoint. Let me know if this should be removed too.

ashishbhate:
@crspeller I've removed the REST endpoint. TaskActions is fully GraphQL now (both on Playbooks and Run).

crspeller (approved):
Thanks!

calebroseland:
Fixed: 

| before | after |
|---|---|
|  CleanShot 2022-11-30 at 16 02 07 | CleanShot 2022-11-30 at 16 07 29 |

calebroseland (approved):

webapp/src/types/telemetry.ts
```diff
@@ -6,7 +6,7 @@
 // Event/Telemetry inventory available at https://docs.google.com/spreadsheets/d/15VBD2i-v7JX11H80beJj64wU8lqyMAm1UrDIjKjx63o/edit#gid=374475626
 
 export enum GeneralViewTarget {
-    TaskInbbox = 'task_inbox',
+    TaskInbox = 'task_inbox',
```
ashishbhate, line 9: fixed a small typo along the way

abhijit-singh (approved):
Looks awesome! Thanks @ashishbhate!

lieut-data:
Wow, nice work everyone! This is a huge effort coming together :D

DHaussermann:
**Test Summary** 
- General 
  - Task actions can be added to a playbooks
  - Tasks actions can be added to individual runs 
  - Mark as done must be true for the task to be completed 

- Keywords Testing
  - Single word
  - Multiple words
  - Case sensitive 

- Respect user scope
  - Team member not in channel cannot be selected as users on the task action
  - Only current channel members can be assigned to an action task
  - All current channel members can be assigned even if they are not yet "participants" of the playbook run

-- issue --
1. Post action is still not getting saved on first save attempt - Must save twice.
2. String evaluation is not formatting agnostic "Published release `v1.5.0`"
3. Affected by autolink re-writting your posts  "Close Jira ticket MM-1234 " the "MM-1234" actually get's re-written as "[MM-1234](https://mattermost.atlassian.net/browse/MM-"  0/5 Can this ignore `[]` to avoid m,ost common cases of Autolink 
4. Task actions can add assignees at the playbook level. 0/5 Do we want this? You can assign users on the team who may not be in the channel when a run is created   
-- UX --
5. It would be ideal if there were some kind of feedback that I have just marked a task as done in the thread. "Release cut v1.5.0"  ""Release cut v1.5.0 has failed" could result in someone marking a task done by mistake. There no way to know if the RHS is closed.   
6. To add multiple strings at once I need to move the focus out and back in or hit return. (unless there's a more obvious way I've missed) Otherwise it adds a multiword string. Is this how this type of control normally works?  

**Notes**:
- Case insensitive opt-in would be a nice touch as future enhancement. 
- Why does this ignore `-` character? For strings like `followup` Vs `follow-up`  will match. Are there excempted characters? If so see 3.

PR #1612, opened 2022-11-25
Loading states for runs RHS

### LINKED ISSUE #1605: Loading states for RHS runs list
RHS should have loading states for when pagination is happening and initial load. Some delay can be experienced on slow connections.

crspeller:
## Summary
Adds loading states for playbooks RHS

Gifs below on simulated slow connection. 

#### General load
mainloadslow

#### Show More
showmore

## Ticket Link
Fixes #1605

### DIFF
--- a/webapp/src/components/assets/loading_spinner.tsx
+++ b/webapp/src/components/assets/loading_spinner.tsx
@@ -1,0 +1,79 @@
+
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+const LoadingSpinner = (props: {className?: string}) => (
+    <svg
+        width='32'
+        height='32'
+        color='var(--button-bg)'
+        viewBox='0 0 20 20'
+        fill='none'
+        className={props.className}
+    >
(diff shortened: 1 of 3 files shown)


webapp/src/components/assets/loading_spinner.tsx
```diff
@@ -0,0 +1,79 @@
+
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+const LoadingSpinner = (props: {className?: string}) => (
```
trilopin, line 7: we have also src/components/assets/icons/spinner.tsx. Since it's the only thing we couldn't migrate from font-awesome, I'd be inclined to use yours and remove that one.

trilopin (approved):
great job 🎉 

Please, consider creating a ticket to migrate older spinners to this one.

webapp/src/components/assets/loading_spinner.tsx
crspeller: #1622

gbochora (approved):
Cool!

abhijit-singh (approved):
💯

PR #1623, opened 2022-11-29
Owner display and update on finish

### LINKED ISSUE #1619: channels-rhs: live update when finish a run
if we finish a run inside channels rhs, we must refetch the data or it's will appear as in-progress when clicking in back icon

very end of the video
https://user-images.githubusercontent.com/4096774/204625154-2e5d541b-a44a-4375-8ad9-b51e9b4b0128.mp4

### LINKED ISSUE #1620: owner shouldn't be displayed as participant in run card
internally the owner is a participant, but visually shouldn't be displayed twice

Image

crspeller:
## Summary
Two runs list fixups.

- Remove owner from participants display. The run owner will no longer be displayed in the list of participants on the runs list card.
- Refetch when back clicked. This fixes the case where you finish the run (or change some other visible property) and navigate back to the runs list using the back button.

## Ticket Link
Fixes #1620
Fixes #1619

### DIFF
--- a/webapp/src/components/rhs/rhs_main.tsx
+++ b/webapp/src/components/rhs/rhs_main.tsx
@@ -62,6 +62,11 @@ const useFilteredSortedRuns = (channelID: string, listOptions: RunListOptions) =
 
     const error = inProgressResult.error || finishedResult.error;
 
+    const refetch = () => {
+        inProgressResult.refetch();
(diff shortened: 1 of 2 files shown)


trilopin (approved):
👍🏽

gbochora (approved):
🎉

PR #1624, opened 2022-11-30
[Proposal] add eslint warning for relative imports

trilopin:
## Summary
Warn if imports are relative.

bad
```ts
import {X} from '../module'
````

Good
```ts
import {X} from 'src/components/folder/module'
```

Motivations:
- Increase awareness on folder/context/parent-component
- Increase how easy is to refactor code from one module to another
- Increase how easy is to copy code from one module to another 

Also, at vscode autoimports can be configured as non-relative. 
Screenshot 2022-11-30 at 11 54 12

## Ticket Link
Nope

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/.eslintrc.json
+++ b/webapp/.eslintrc.json
@@ -629,7 +629,7 @@
     "unused-imports/no-unused-imports": 2,
     "no-relative-import-paths/no-relative-import-paths": [
       "warn",
-      { "allowSameFolder": false }
+      { "allowSameFolder": true }
     ]
   },
   "overrides": [
--- a/webapp/.eslintrc.json
+++ b/webapp/.eslintrc.json
@@ -10,7 +10,8 @@
     "import",
     "formatjs",
     "@typescript-eslint",
-    "unused-imports"
+    "unused-imports",
+    "no-relative-import-paths"
   ],
   "env": {
     "browser": true,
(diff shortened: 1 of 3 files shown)


trilopin:
any thoughts on this? @calebroseland @crspeller

webapp/.eslintrc.json
```diff
@@ -625,7 +626,11 @@
     "formatjs/no-multiple-plurals": 2,
     "formatjs/enforce-placeholders": 2,
     "formatjs/no-literal-string-in-jsx": 2,
-    "unused-imports/no-unused-imports": 2
+    "unused-imports/no-unused-imports": 2,
+    "no-relative-import-paths/no-relative-import-paths": [
+      "warn",
+      { "allowSameFolder": false }
```
calebroseland, line 632: This would be the one relative path that I would keep—it's nice IMO to have a distinction for sibling files. But I agree 5/5: everything else should be absolute—`../../../../` is not readable.


```suggestion
      { "allowSameFolder": true }
```

trilopin:
>< @trilopin Nice! After your comments I looked briefly for one.
> I see it supports --fix could we do a one time fix and then set it to error?

I'd be happy to run the --fix and convert it into an error. Since it's going to be a huge (easy tho) amount of changes that can potentially cause conflicts on current PRs, I'd do it in a second pr to make sure everyone is aware (will raise the topic in the next sync meeting).

crspeller (approved):

calebroseland (approved):
Thanks for doing this!

PR #1615, opened 2022-11-29
Refactor code to move out fav/unfav from UpdateRun

jecepeda:
## Summary

* Add new GraphQL mutation to fav/unfav a run
* Remove IsFavorite from UpdateRun mutation
* Add tests in server to test the new functionality
* Update schema
* Change webapp code to use the new mutation

## Ticket Link

Closes https://github.com/mattermost/mattermost-server/issues/21706

## Checklist

- [ ] ~Telemetry updated~
- [ ] ~Gated by experimental feature flag~
- [x] Unit tests updated

PR #1632, opened 2022-12-01
Add Graphql Dataloader to favorites & playbooks

jecepeda:
## Summary

After doing [this PR](https://github.com/mattermost/mattermost-plugin-playbooks/pull/1615) I noticed that playbooks don't use any dataloader. If we don't use dataloaders, we will do N+1 when obtaining nested substructures.

This PR covers two data structures that are nested:
- `isFavorite`, as it's the last piece of code I worked on.
- `playbook` inside a playbook run. This is a more general case that can be applied to the rest of the PR.

Added:
* Add favorite dataloader to the main graphql function
* Add playbook dataloader to the main graphql function
* Create GraphQl loader for isFavorite on runs
* Create GraphQL loader for playbooks on runs
* Create new category service function
* Add tests that take multiple favorite data/ playbook data in order to test the correct functionality.

## Concerns

When adding the loader for playbooks, I've used the sqlstore function called `GetPlaybooksForTeam`. This function takes nested structures. 
If `mattermost-plugin-playbooks` is going to adopt graphql as the main query strategy, repository functions should not take sub tables or other inner structures, in favor of using dataloader. This will reduce database load in case we're not showing subtable data.

## Ticket Link

No ticket link

## Checklist

- [ ] Telemetry updated
- [x] Unit tests updated


server/app/playbook_run_service.go
```diff
@@ -3540,9 +3540,7 @@ func (s *PlaybookRunServiceImpl) doActions(taskActions []Action, runID string, u
 			}
 			if a.Payload.Enabled {
 				if err := s.ModifyCheckedState(runID, userID, ChecklistItemStateClosed, checklistNum, itemNum); err != nil {
-					if err != nil {
-						return errors.Wrapf(err, "can't mark item as done")
-					}
+					return errors.Wrapf(err, "can't mark item as done")
```
jecepeda, line 3543: Running golangci on master gave me this error: `nilness: tautological condition: non-nil != nil`

Basically we're already checking on 3542 `if err != nil`

server/api/resolver.go
```diff
@@ -0,0 +1,40 @@
...
+const loaderBatchCapacity = 200
+
+type loaderKey int
+
+const (
+	favoritesCtx loaderKey = iota
+	playbookCtx
+)
+
+func getFavoritesLoader(ctx context.Context) (*dataloader.Loader, error) {
```
trilopin, line 19: would make sense to rename this file as server/api/grahql_dataloader.go?

server/api/resolver_favorite.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (data *favoriteInfo) String() string {
+	return fmt.Sprintf("fav://%s-%s-%s-%s", data.TeamID, data.UserID, data.ID, data.Type)
+}
+
+func (data *favoriteInfo) Raw() interface{} {
+	return data
+}
+
+func graphQLRunFavoritesLoader(ctx context.Context, keys dataloader.Keys) []*dataloader.Result {
```
trilopin, line 26: this loader seems to work for both playbook favorites and run favorites since favoriteInfo includes app.CategoryItemType. could we rename this func to remove the Run keyword from it?

server/api/resolver_favorite.go
```diff
@@ -0,0 +1,67 @@
+package api
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/graph-gophers/dataloader/v6"
+	"github.com/mattermost/mattermost-plugin-playbooks/server/app"
+)
+
```
trilopin, line 10: woudl make sense to rename this file as server/api/graphql_loader_favorite.go?

server/api/graphql_run.go
```diff
@@ -93,25 +93,31 @@ func (r *RunResolver) TimelineEvents() []*TimelineEventResolver {
 }
 
 func (r *RunResolver) IsFavorite(ctx context.Context) (bool, error) {
+	loader, err := getFavoritesLoader(ctx)
```
trilopin, line 96: could you please do the same for graphql_playbook.go IsFavorite function?

trilopin (changes requested):
Really good job @jecepeda, it's amazing to have your help here.

I've added some comments, requests and doubts.

💯

server/api/graphql_run.go
jecepeda: Sure!

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/golang/mock v1.6.0
 	github.com/gorilla/mux v1.8.0
+	github.com/graph-gophers/dataloader/v6 v6.0.0
```
crspeller, line 17: Could we use v7 instead of v6? I have a in progress work and I am using v7. It supports generic so the interface is more type safe.

server/api/graphql.go
```diff
@@ -135,6 +136,13 @@ func (h *GraphQLHandler) graphQL(c *Context, w http.ResponseWriter, r *http.Requ
 	reqCtx := r.Context()
 	reqCtx = context.WithValue(reqCtx, ctxKey{}, graphQLContext)
 
+	// dataloaders
+	favoritesLoader := dataloader.NewBatchedLoader(graphQLRunFavoritesLoader, dataloader.WithBatchCapacity(loaderBatchCapacity))
```
crspeller, line 140: Can we add these to the existing graphQLContext rather than individual items on the context? 

Like this:
```go
	postsLoader := dataloader.NewBatchedLoader(postsBatchLoader, dataloader.WithBatchCapacity[string, *model.Post](batchCapacity))

	graphQLContext := &GraphQLContext{
		r:                  r,
		playbookService:    h.playbookService,
		playbookRunService: h.playbookRunService,
		categoryService:    h.categoryService,
		pluginAPI:          h.pluginAPI,
		logger:             c.logger,
		config:             h.config,
		permissions:        h.permissions,
		playbookStore:      h.playbookStore,
		licenceChecker:     h.licenceChecker,
		postsLoader:        postsLoader,
	}

```

server/api/resolver.go
```diff
@@ -0,0 +1,40 @@
...
+	l, ok := ctx.Value(favoritesCtx).(*dataloader.Loader)
+	if !ok {
+		return nil, errors.New("no favorite loader found in context")
+	}
+	return l, nil
+}
+
+func getPlaybookLoader(ctx context.Context) (*dataloader.Loader, error) {
```
crspeller, line 27: These become redundant given the suggestion above.

crspeller (changes requested):

jecepeda:
Hey @crspeller @trilopin I just finished your requested changes. I'll take a look now in case there're any errors and/or linting errors, but I generally think this is ready for another round of reviews.

server/api/graphql_loader_playbook.go
```diff
@@ -0,0 +1,87 @@
...
+	playbooksByID := make(map[string]*app.Playbook)
+	for i := range playbookResult.Items {
+		playbooksByID[playbookResult.Items[i].ID] = &playbookResult.Items[i]
+	}
+
+	for i, playbookInfo := range keys {
+		playbook, ok := playbooksByID[playbookInfo.ID]
+		if !ok {
+			result[i] = &dataloader.Result[V]{
+				Error: fmt.Errorf("playbook %q not found", playbookInfo.ID),
```
trilopin, line 64: This error can be triggered when we try to execute a nested query with runs and their playbooks but we lack permissions for that graph navigation. 

There's a valid scenario where a user can have access to runs, but not the playbooks. related to those runs. That's what happens in the current e2e failures at command_spec.js.

For those cases, I wonder if we can just nest an empty PlaybookResolver instead of failing (and handle at the webapp side the different actions as we already do).

cc @crspeller on funny scenarios nesting resources with different permission levels 🎉

crspeller: We need the equivalent of:
```go
if !errors.Is(err, app.ErrNoPermissions) {
    c.logger.WithError(err).Error("error retrieving playbook of run")
}
return nil, nil
```
from the previous code. Just ignore it if it's a permissions error and return nil.

crspeller (changes requested):
Other than the permissions thing it looks good.

server/api/graphql_loader_playbook.go
```diff
@@ -0,0 +1,85 @@
...
+	ID     string
+}
+
+func (data *playbookInfo) String() string {
+	return fmt.Sprintf("playbook://%s-%s-%s", data.UserID, data.TeamID, data.ID)
+}
+
+func (data *playbookInfo) Raw() interface{} {
+	return data
+}
```
trilopin, line 23: Are Raw / String functions still needed ?

trilopin (approved):
Huge job @jecepeda 

I left a couple of non-blocking comments

server/api/graphql_loader_playbook.go
trilopin: Those permissions are handled not by permission service but with buildPermissionExpr at the store layer, so I guess we just have to ignore non-returned playbooks

crspeller (approved):

PR #1634, opened 2022-12-01
Make playbook retrieval optional for case where user has no permissions.

crspeller:
## Summary

Make playbook retrieval optional for runs in GraphQL for the case where the user has permissions to the run but not the playbook that created it. 

Shows as blank in the RHS:
image

### DIFF
--- a/server/api/graphql_run.go
+++ b/server/api/graphql_run.go
@@ -135,11 +135,12 @@ func (r *RunResolver) Metadata(ctx context.Context) (*MetadataResolver, error) {
 }
 
 func (r *RunResolver) Playbook(ctx context.Context) (*PlaybookResolver, error) {
-	if val, err := getGraphqlPlaybook(ctx, r.PlaybookID); err != nil {
+	val, err := getGraphqlPlaybook(ctx, r.PlaybookID)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


trilopin (approved):
LGTM

server/api/graphql_run.go
```diff
@@ -135,7 +135,12 @@ func (r *RunResolver) Metadata(ctx context.Context) (*MetadataResolver, error) {
 }
 
 func (r *RunResolver) Playbook(ctx context.Context) (*PlaybookResolver, error) {
-	return getGraphqlPlaybook(ctx, r.PlaybookID)
+	val, err := getGraphqlPlaybook(ctx, r.PlaybookID)
+	if err != nil {
+		return nil, nil
+	}
```
lieut-data, line 141: It feels weird to just swallow this error. Might we check explicitly for a permissions error and ignore that explicitly?

lieut-data (approved):
E2E tests? 🙏🏼

server/api/graphql_run.go
crspeller: Yes it does feel strange. We can explicit check permissions errors for this reason. Decided to log otherwise.

PR #1639, opened 2022-12-02
MM-46961 Implement pre-assigning of users to tasks in playbooks

m-ripper:
## Summary
So far, only playbook runs could have assignees. With this PR, users can be assigned to playbooks.\
This means, that a new run of the playbook will automatically invite the user and assign the user to the task that was pre-assigned.  

When invitations are disabled while there are task with pre-assignees, a confirm modal will pop up that will ask the user to confirm that all pre-assignments will automatically be cleared when confirmed. 

Similar to that, a confirm modal will pop up if a pre-assignee is getting removed from the invite list, that will ask the user to confirm that all tasks with this assignee will be unassigned when confirmed.

[simplescreenrecorder-2022-12-02_16.50.04.webm](https://user-images.githubusercontent.com/8669935/205332364-5b8bef23-ef2e-44a8-bebd-0952f173c743.webm)

In the backend, the interfaces `ChecklistCommon`, and `ChecklistItemCommon` to access `app.Checklist` and `api.UpdateChecklist`  as well as `app.ChecklistItem` and `api.UpdateChecklistItem` respectively were implemented. 

These interfaces make it easier to implement methods that can handle both data structures, such as the new `app.CleanUpChecklists` which replaces the old `cleanUpChecklist` functions.  

The biggest disadvantage is the need of getters and setters because fields are currently not accessible via generics, but those are also easier to maintain than business logic. 
 
 

## Ticket Link
[MM-46961](https://mattermost.atlassian.net/browse/MM-46961)

## Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -76,26 +69,38 @@ const ChecklistList = ({
...
+            // Append all assignees found in the updated checklists and clear duplicates
+            const invitedUsers = new Set([...updatedPlaybook.invited_user_ids, ...getDistinctAssignees(updatedPlaybook.checklists)]);
+            if (invitedUsers.size && !updatedPlaybook.invite_users_enabled) {
+                updates.inviteUsersEnabled = true;
+            }
+
+            // Only update the userIds when new assignees were added
+            if (invitedUsers.size > updatedPlaybook.invited_user_ids.length) {
+                updates.invitedUserIDs = [...invitedUsers];
+            }
```
gbochora, line 102: This will enable `Invite participants` on any change, e.g., when I set task duration.

https://user-images.githubusercontent.com/4368372/207202642-1d7e9151-137d-44e5-837d-bb69a135d1d4.mp4

webapp/src/components/backstage/playbook_edit/automation/invite_users.tsx
```diff
@@ -22,31 +23,93 @@ interface Props {
...
+                message={formatMessage(
+                    {defaultMessage: 'The user <i>{name}</i> is pre-assigned to one or more tasks. Not automatically inviting this user will clear their pre-assignments.{br}{br}Are you sure you want to stop inviting this user as a member of the run?'},
+                    {br: <br/>, i: (x: React.ReactNode) => <i>{x}</i>, name: userToRemove?.username}
+                )}
+                confirmButtonText={formatMessage({defaultMessage: 'Remove user'})}
+                onConfirm={() => {
+                    if (!userToRemove?.userId || !userToRemove?.username) {
+                        setUserToRemove(null);
+                        return;
+                    }
```
gbochora, line 107: Do we need this code? The modal is visible only if this condition is false.

tests-e2e/cypress/integration/playbooks/edit_spec.js
```diff
@@ -66,6 +66,77 @@ describe('playbooks > edit', () => {
...
+
+                    // * Verify that the root of the assignee select menu exists
+                    cy.focused()
+                        .parents('.playbook-react-select')
+                        .should('exist');
+                });
+
+                cy.get('#actions').within(() => {
+                    cy.get('#invite-users').within(() => {
+                        // #*Verify invitations are enabled and a single user is invited
```
gbochora, line 130: nit:
```suggestion
                        // * Verify invitations are enabled and a single user is invited
```

gbochora (changes requested):
Great work, @m-ripper! Sorry for the delayed review.
The only blocking comment is about enabling invite participants on checklist edit.

calebroseland (changes requested):
Nice work, @m-ripper! A couple things, must mostly just some thoughts for potential/future optimizations.

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -77,27 +70,39 @@ const ChecklistList = ({
...
```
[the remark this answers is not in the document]
m-ripper: That's a great catch! I agree that it can be problematic if we wanted to use a debounced pattern in some cases, but even changing this to a deep comparison could be causing issues in a debounced pattern, therefore I will leave it as is and gladly update it if we change the pattern.

abhijit-singh (approved):
Great work @m-ripper! Looks awesome!

calebroseland (approved):
Thanks, @m-ripper!

gbochora (commented):
Looks good! Thanks, @m-ripper!
I have only one UX concern. It's unclear that assigning a user to a task enables the `Invite participants` action, and this might be a security risk. 
For example:
- user has a playbook, invite action is enabled, and ten users are selected 
- user disables invite action. NOTE: the selected list is not cleared, we still store it
- user sets a pre-assignee on a task, and it automatically enables invite action

As a result, when someone starts a run, ten users selected before will also be invited, and the user has no idea about that.

I think we should either inform users using toast that actions have been enabled or clear the selected list when disabling action.

@abhijit-singh, what do you think about this?

gbochora (approved):

lieut-data:
@gbochora and I synced this morning on this topic -- I'm inclined to leave the existing experience "as-is". While it might be surprising to enable the feature and add all these other users, only someone with access to the playbook could have configured those users in the first place, so it's not likely to be a security issue. @gbochora, that makes sense, I'd propose we approve and merge this.

lieut-data:
Haha, GitHub just updated to show the fact that we've already merged it :p -- 🎉

Issue #1621, opened 2022-11-29
Slash commands compatibility


trilopin:
Slash command `playbook list` (currently failing) should navigate the playbooks backstage run list.

Explore potential fixes for slash commands that are run-related (info, timeline check checkadd checkremove owner update). One solution that was commented was to autocomplete with run list when there's more than one.

Check messages, in case make no sense to refer one run in singular.

PR #1642, opened 2022-12-05
1621 slash commands compatibility

gbochora:
## Summary
1) This PR adds multiple runs compatibility for commands:
- `/playbook check`
- `/playbook checkadd`
- `/playbook checkremove`
- `/playbook finish`
- `/playbook owner`
- `/playbook info`
- `/playbook timeline`
If there are multiple runs in the channel, a user must provide(choose from autocomplete) the run number. However, when there is a single run in the channel, behavior is the same as before, a user doesn't need to provide the run number.

2) Separating channel members and run participants created several minor security issues. Channel members with no access to the run could view or modify it. I've added permission checks to fix it.

3) Users could change finished runs using commands, now, we allow modifying only in-progress runs.

4) Fixed RHS > Finish run functionality when there are multiple runs in the channel

5) E2E tests are coming soon.

https://user-images.githubusercontent.com/4368372/205587538-8d67b95b-0606-4868-ae7e-8c941a9d1129.mp4

NOTE:
A few commands still don't work:
- `/playbook list`
- `/playbook info` 
- `/playbook update`
It seems that these commands are intercepted on the mm server side, so I'll address them in a separate PR.

## Ticket Link
#1621 

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


server/api/playbook_runs.go
```diff
@@ -623,18 +624,23 @@ func (h *PlaybookRunHandler) getPlaybookRunByChannel(c *Context, w http.Response
...
+	if len(playbookRunIDs) > 1 {
+		h.HandleErrorWithCode(w, c.logger, http.StatusBadRequest, "multiple runs in the channel", nil)
+		return
+	}
+
+	if err := h.permissions.RunView(userID, playbookRunIDs[0]); err != nil {
 		c.logger.WithFields(logrus.Fields{
 			"user_id":         userID,
-			"playbook_run_id": playbookRunID,
+			"playbook_run_id": playbookRunIDs[0],
```
lieut-data, line 635: nit: could we declare a `playbookRunID := playbookRunIDs[0]` after the query to a) avoid changes below, and b) clearly indicate with a comment why we're picking only one.

server/api/playbook_runs.go
```diff
@@ -1085,17 +1091,14 @@ func (h *PlaybookRunHandler) getChecklistAutocompleteItem(c *Context, w http.Res
...
-	}
-
-	playbookRunID, err := h.playbookRunService.GetPlaybookRunIDForChannel(channelID)
+	playbookRunIDs, err := h.playbookRunService.GetPlaybookRunIDsForChannel(channelID, &userID)
 	if err != nil {
-		h.HandleError(w, c.logger, err)
+		h.HandleErrorWithCode(w, c.logger, http.StatusNotFound, "Not found",
+			errors.Errorf("playbook run for channel id %s not found", channelID))
 		return
 	}
```
lieut-data, line 1099: Should we distinguish `app.ErrNotFound` in case there is some other issue at play?

server/api/playbook_runs.go
```diff
@@ -1109,17 +1112,35 @@ func (h *PlaybookRunHandler) getChecklistAutocomplete(c *Context, w http.Respons
 	channelID := query.Get("channel_id")
 	userID := r.Header.Get("Mattermost-User-ID")
 
-	if !h.PermissionsCheck(w, c.logger, h.permissions.RunViewByChannel(userID, channelID)) {
+	playbookRunIDs, err := h.playbookRunService.GetPlaybookRunIDsForChannel(channelID, &userID)
+	if err != nil {
+		h.HandleErrorWithCode(w, c.logger, http.StatusNotFound, "Not found",
```
lieut-data, line 1117: Same as above, should we check for other types of errors?

server/api/playbook_runs.go
```diff
@@ -1109,17 +1112,35 @@ func (h *PlaybookRunHandler) getChecklistAutocomplete(c *Context, w http.Respons
...
+}
+
+func (h *PlaybookRunHandler) getChannelRunsAutocomplete(c *Context, w http.ResponseWriter, r *http.Request) {
+	query := r.URL.Query()
+	channelID := query.Get("channel_id")
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	playbookRunIDs, err := h.playbookRunService.GetPlaybookRunIDsForChannel(channelID, &userID)
+	if err != nil {
+		h.HandleErrorWithCode(w, c.logger, http.StatusNotFound, "Not found",
```
lieut-data, line 1138: Same as above, should we check for other types of errors?

server/app/playbook_run.go
```diff
@@ -609,9 +609,9 @@ type PlaybookRunService interface {
 	// GetPlaybookRunMetadata gets ancillary metadata about a playbook run.
 	GetPlaybookRunMetadata(playbookRunID string) (*Metadata, error)
 
-	// GetPlaybookRunIDForChannel get the playbookRunID associated with this channel. Returns ErrNotFound
-	// if there is no playbook run associated with this channel.
-	GetPlaybookRunIDForChannel(channelID string) (string, error)
+	// GetPlaybookRunIDsForChannel get the playbookRunIDs list associated with this channel. Returns ErrNotFound
```
lieut-data, line 612: nit: I wonder if it would be cleaner if we just returned an empty list instead of an error?

server/app/playbook_run.go
```diff
@@ -609,9 +609,9 @@ type PlaybookRunService interface {
...
+	// if there is no playbook run associated with this channel. If userID is not null, returns only runs that the user has access.
```
lieut-data, line 613: nit: would it make sense to distinguish with a `GetPlaybookRunIDsForChannelByUser` vs. having to pass a pointer?

server/app/playbook_run_service.go
```diff
@@ -1280,14 +1280,26 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
...
-		return "", err
+		return nil, err
+	}
+
+	if userID == nil {
+		return playbookRunIDs, nil
+	}
+
+	runIDsUserHasAccess := make([]string, 0)
+	for _, id := range playbookRunIDs {
```
lieut-data, line 1296: nit: perhaps `playbookRunID` instead of `id` for clarity?

server/app/playbook_run_service.go
```diff
@@ -2063,40 +2075,82 @@ func (s *PlaybookRunServiceImpl) MoveChecklistItem(playbookRunID, userID string,
...
 }
 
-// GetChecklistAutocomplete returns the list of checklist items for playbookRunID to be used in autocomplete
-func (s *PlaybookRunServiceImpl) GetChecklistItemAutocomplete(playbookRunID string) ([]model.AutocompleteListItem, error) {
-	playbookRun, err := s.store.GetPlaybookRun(playbookRunID)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve playbook run")
+// GetRunsAutocomplete returns the list of runs to be used in autocomplete
+func (s *PlaybookRunServiceImpl) GetRunsAutocomplete(playbookRunIDs []string) ([]model.AutocompleteListItem, error) {
+	if len(playbookRunIDs) == 1 {
```
lieut-data, line 2139: Given the result is the same, should we check `<= 1`?

server/app/playbook_run_service.go
```diff
@@ -2547,11 +2601,16 @@ func (s *PlaybookRunServiceImpl) newAddToTimelineDialog(playbookRuns []PlaybookR
 		}
 	}
 
-	defaultPlaybookRunID, err := s.GetPlaybookRunIDForChannel(post.ChannelId)
+	defaultPlaybookRunIDs, err := s.GetPlaybookRunIDsForChannel(post.ChannelId, nil)
```
lieut-data, line 2604: Given the `nil` user, can a user without access to some runs in a channel glean information about them here?

server/command/command.go
```diff
@@ -383,44 +414,125 @@ func (r *Runner) actionAddChecklistItem(args []string) {
...
+	index := 0
+	if multipleRuns {
+		if run, err = strconv.Atoi(args[index]); err != nil {
+			r.postCommandResponse("Error parsing the first argument. Must be a number.")
+			return
+		}
+		if run < 0 || run > len(playbookRunIDs) {
+			r.postCommandResponse("Invalid run number")
+			return
+		}
```
lieut-data, line 439: Should this be `run >= len(playbookRunIDs)` given 0-indexing?

server/sqlstore/playbook_run.go
```diff
@@ -811,22 +811,25 @@ func (s *playbookRunStore) GetTimelineEvent(playbookRunID, eventID string) (*app
...
-// GetPlaybookRunIDForChannel gets the playbook run ID associated with the given channel ID.
-func (s *playbookRunStore) GetPlaybookRunIDForChannel(channelID string) (string, error) {
+// GetPlaybookRunIDsForChannel gets the playbook run IDs list associated with the given channel ID.
+func (s *playbookRunStore) GetPlaybookRunIDsForChannel(channelID string) ([]string, error) {
 	query := s.queryBuilder.
 		Select("i.ID").
 		From("IR_Incident i").
-		Where(sq.Eq{"i.ChannelID": channelID})
+		Where(sq.Eq{"i.ChannelID": channelID}).
+		Where(sq.Eq{"i.CurrentStatus": app.StatusInProgress}).
```
lieut-data, line 820: This condition probably makes sense in most contexts, but are all callers aligned on this? (Previously they would have matched with the playbook run regardless of whether it was finished.)

lieut-data (changes requested):
Wow, there's a ton of improvements here -- great deep-dive, @gbochora! A few comments.

lieut-data:
@gbochora, heads up that I've merged master with your PR, as well as fixed https://github.com/mattermost/mattermost-plugin-playbooks/issues/1648 as part of this update (including an e2e test).

server/api/playbook_runs.go
trilopin: Can't we just remove this endpoint?

It's called by client.ts `fetchPlaybookRunByChannel` which has the following references:
- handleWebsocketUserAdded: refetch playbook run when receive WebsocketEvents.USER_ADDED
- getPlaybookRunFromStatusUpdate/handleWebsocketPostEditedOrDeleted: not sure about this one but looking at the code seems related to failure reported at https://github.com/mattermost/mattermost-plugin-playbooks/issues/1647
- handleWebsocketChannelUpdated: refetch playbookrun when receive WebsocketEvents.CHANNEL_UPDATED


Since we changed all run websocket events to be channel agnostic, can't we get rid off all this listeners?

server/api/playbook_runs.go
trilopin: agree, we shouldn't treat a potential sql error as a not-found error. 

silly question: should not-found be treated as error or we can just return an empty list?

server/app/permissions_service.go
```diff
@@ -423,15 +423,6 @@ func (p *PermissionsService) runManagePropertiesWithPlaybookRun(userID string, r
...
 }
 
-func (p *PermissionsService) RunManagePropertiesByChannel(userID, channelID string) error {
-	runID, err := p.runService.GetPlaybookRunIDForChannel(channelID)
-	if err != nil {
-		return errors.Wrapf(err, "Unable to get playbookRunID to determine permissions, channel id `%s`", channelID)
-	}
-
-	return p.RunManageProperties(userID, runID)
-}
```
trilopin, line 433: nice removal

server/api/playbook_runs.go
trilopin: if we keep this code, couldn't we remove the permission check since the function is returning those runs with access?

server/app/playbook_run.go
trilopin: 💯 to this if it's not very hard to handle at client side.

server/app/playbook_run_service.go
```diff
@@ -2084,40 +2096,82 @@ func (s *PlaybookRunServiceImpl) MoveChecklistItem(playbookRunID, userID string,
...
 // GetChecklistAutocomplete returns the list of checklist items for playbookRunID to be used in autocomplete
-func (s *PlaybookRunServiceImpl) GetChecklistAutocomplete(playbookRunID string) ([]model.AutocompleteListItem, error) {
-	playbookRun, err := s.store.GetPlaybookRun(playbookRunID)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve playbook run")
+func (s *PlaybookRunServiceImpl) GetChecklistAutocomplete(playbookRunIDs []string) ([]model.AutocompleteListItem, error) {
+	ret := make([]model.AutocompleteListItem, 0)
+
+	for j, id := range playbookRunIDs {
+		playbookRun, err := s.store.GetPlaybookRun(id)
```
trilopin, line 2103: n+1 query, Too much for this PR but I'd create a ticket to track this. It would be great to be able to use GetPlaybookRuns with a list of ids (also would cover the permission thing)

server/app/playbook_run_service.go
trilopin: I don't understand why this userID is nil

server/sqlstore/playbook_run.go
trilopin: GetPlaybookRuns uses app.PlaybookRunFilterOptions which has already channelID support (also has builtin permissions support and can avoid the RunView check later)

It may return more data than you need but I prefer that than creating more API surface (and we'll unnest it as we make progress with graphql)

trilopin (changes requested):
Nice cleanup

I left some comments ❤️

server/sqlstore/playbook_run.go
gbochora: Only the `/playbook info` command is not aligned on this, it needs all runs instead of only in-progress runs. However, the `/playbook info` command currently doesn't work for multiple runs. I could not fix it in this PR because it's handled on the mm server side. I've registered a ticket #1658

trilopin (changes requested):
Nice progress and tests. I left some comments.


Screenshot 2022-12-07 at 12 58 07

Check suggestion.undefined and the help that mentions "the run".

server/app/playbook_run_service.go
```diff
@@ -3478,8 +3525,21 @@ func (s *PlaybookRunServiceImpl) dmPostToUsersWithPermission(users []string, pos
...
+			IsAdmin: true,
+		},
+
+		PlaybookRunFilterOptions{
+			ChannelID: post.ChannelId,
+			Statuses:  []string{StatusInProgress},
+			Page:      0,
+			PerPage:   1000,
+		},
+	)
```
lieut-data, line 3541: I think we need to restore `GetPlaybookRunIDsForChannel` for this use case -- we can't be invoking a super complex query like `GetPlaybookRuns` on every message post.

lieut-data (changes requested):

server/app/playbook_run_service.go
lieut-data: Fixed!

trilopin: I understood why after @lieut-data added context, sorry for the noise @gbochora

server/sqlstore/playbook_run.go
```diff
@@ -811,22 +811,25 @@ func (s *playbookRunStore) GetTimelineEvent(playbookRunID, eventID string) (*app
 	return &event, nil
 }
 
-// GetPlaybookRunIDForChannel gets the playbook run ID associated with the given channel ID.
-func (s *playbookRunStore) GetPlaybookRunIDForChannel(channelID string) (string, error) {
+// GetPlaybookRunIDsForChannel gets the playbook run IDs list associated with the given channel ID.
+func (s *playbookRunStore) GetPlaybookRunIDsForChannel(channelID string) ([]string, error) {
```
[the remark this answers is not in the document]
gbochora: - Fixed failing e2e tests. It was caused because of the incorrect marshaling of the PlaybookRun object. 
- Fixed finish command help text
- Can't reproduce `SUGGESTION.UNDEFINED`, I used to reproduce but after `make clean`, it shows the correct text.

gbochora:
- Fixed failing e2e tests. It was caused because of the incorrect marshaling of the PlaybookRun object. 
- Fixed finish command help text
- Can't reproduce `SUGGESTION.UNDEFINED`, I used to reproduce but after `make clean`, it shows the correct text.

lieut-data (approved):

server/api/playbook_runs.go
```diff
@@ -652,7 +652,7 @@ func (h *PlaybookRunHandler) getPlaybookRunByChannel(c *Context, w http.Response
 		return
 	}
 
-	ReturnJSON(w, playbookRun, http.StatusOK)
+	ReturnJSON(w, &playbookRun, http.StatusOK)
```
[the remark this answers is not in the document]
gbochora: The `MarshalJSON` function is defined for `*PlaybookRun`: `func (r *PlaybookRun) MarshalJSON() ([]byte, error)`

When we pass the `PlaybookRun` object to the `ReturnJSON`, it calls `json.Marshal` on it, and as result, the standard marshaling function is called. Because of that, list fields in JSON  have null values instead of empty arrays. The null array was causing a javascript error in the mm web app.

trilopin (approved):
Thanks @gbochora , good refactor!

Issue #1644, opened 2022-12-05
flag not enabled right after login


trilopin:
upon creating a fresh account, then creating a new team, then creating a new playbook—the options for multiple runs in a channel in Outline don’t show up until a manual refresh 

https://user-images.githubusercontent.com/4096774/205692339-5177357a-3366-4105-a28f-fafd7fa04945.mp4

Settings are being fetched when still no auth, returning a 401. There's no retry after (unless reload page)
![Image](https://user-images.githubusercontent.com/4096774/205692422-4783fd43-4573-4ef7-a2cc-38ff2be432b3.png)

PR #1650, opened 2022-12-05
ensure settings are fetched after login

trilopin:
## Summary
Plugin settings are needed, among other things, to handle feature flags.
When a user initializes app unlogged, settings call returns a 401 (obviously) and is not refetched anymore.

Added hook to ensure settings are fetched, if empty will retry the fetch. Triggered in two main points backstage and main rhs.

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1644

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/hooks/general.ts
+++ b/webapp/src/hooks/general.ts
@@ -51,7 +51,6 @@ import {getProfileSetForChannel,
 } from 'src/selectors';
 import {
     fetchPlaybookRuns,
-    fetchGlobalSettings,
     clientFetchPlaybook,
     fetchPlaybookRunStatusUpdates,
     fetchPlaybookRun,
@@ -60,7 +59,6 @@ import {
 } from 'src/client';
 import {isCloud} from 'src/license';
 import {resolve} from 'src/utils';
-import {actionSetGlobalSettings} from 'src/actions';
 
 export type FetchMetadata = {
     isFetching: boolean;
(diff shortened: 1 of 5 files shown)


webapp/src/hooks/general.ts
```diff
@@ -290,6 +292,14 @@ export function useLinkRunToExistingChannelEnabled() {
     return useSelector(selectLinkRunToExistingChannelEnabled);
 }
 
+export async function useEnsureSettings() {
+    const dispatch = useDispatch();
+    const hasGlobalSettings = useSelector((state: GlobalState) => Boolean(globalSettings(state)));
+    if (!hasGlobalSettings) {
+        dispatch(actionSetGlobalSettings(await fetchGlobalSettings()));
+    }
```
lieut-data, line 300: Is there any risk of this running multiple times in a quick burst?

lieut-data (approved):

webapp/src/hooks/general.ts
trilopin: not a huge burst, but I can see 2 calls indeed. maybe a simple lock is a good idea

webapp/src/index.tsx
```diff
@@ -287,10 +288,6 @@ export default class Plugin {
             lastActivityTime = now;
         };
         document.addEventListener('click', this.activityFunc);
-
-        // We have user activity right now because the plugin is loading
-        // so fire the first connect event.
-        notifyConnect();
```
trilopin, line 293: This was also failing if user was unlogged. Moved to LoginHook too.
 
Screenshot 2022-12-06 at 06 06 46

trilopin:
@lieut-data @gbochora I followed @mickmister suggestion to simplify initial approach. No need to handle locks to avoid burst in this new approach:

LoginHook component is a new RootComponent which handle the side effects that should be triggered after login. It gets executed in pure login and full reload.

gbochora (approved):
Cool!

webapp/src/index.tsx
```diff
@@ -312,6 +309,8 @@ export default class Plugin {
 
         this.doRegistrations(registry, store, graphqlClient);
 
+        // https://mattermost.atlassian.net/browse/MM-48872
+        // This is handled by LoginHook, but it doesn't seem compatible with e2e tests.
```
lieut-data, line 313: Can you elaborate more on what's going on here? Just worried that we might be missing a case if e2e tests don't work.

trilopin: From a pure user perspective, we can remove the piece in initizialize related to fetch settings since the new LoginHook triggers in both unlogged->log and full-refresh-while-logged operations.

But, I was very surprised to see that E2E wasn't fetching settings (maybe is when they already exist or based on some weird fake navigation), so I kept the code in initzialize and create the ticket to investigate more.

Issue #1654, opened 2022-12-06
missing run name in finish run confirmation modal


trilopin:
![Image](https://user-images.githubusercontent.com/4096774/205852361-07e4fa06-0e2a-4431-be77-7c4de26478ca.png)

PR #1655, opened 2022-12-06
1654 Include run name in finish run modal

m-ripper:
## Summary
When finishing a run, the confirmation modal now also displays which run is affected. This is especially useful for multiple runs in a channel.

[Screencast from 06.12.2022 15:23:00.webm](https://user-images.githubusercontent.com/8669935/205937494-cb9ec7f5-847a-4788-a009-00d28824028a.webm)

## Ticket Link
Fixes: #1654 

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


server/app/playbook_run_service.go
```diff
@@ -2380,12 +2380,13 @@ func (s *PlaybookRunServiceImpl) GetSchemeRolesForChannel(channel *model.Channel
 	return model.ChannelGuestRoleId, model.ChannelUserRoleId, model.ChannelAdminRoleId
 }
 
-func (s *PlaybookRunServiceImpl) newFinishPlaybookRunDialog(outstanding int) *model.Dialog {
-	message := "Are you sure you want to finish the run for all participants?"
+func (s *PlaybookRunServiceImpl) newFinishPlaybookRunDialog(playbookRun *PlaybookRun, outstanding int) *model.Dialog {
```
trilopin, line 2383: please create a ticket to make this dialog translatable and link the ticket in a comment here

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -431,6 +419,22 @@ export const outstandingTasks = (checklists: Checklist[]) => {
     }
     return count;
 };
+export const useFinishRunConfirmationMessage = (run: PlaybookRun | null | undefined) => {
```
trilopin, line 422: I understand that you share this at finish & post update actions. I would see it more accurate to place the useFinishRunConfirmationMessage in the finish file and import it here.

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -431,6 +419,22 @@ export const outstandingTasks = (checklists: Checklist[]) => {
...
+    const {formatMessage} = useIntl();
+    const outstanding = outstandingTasks(run?.checklists || []);
+    const values = {
+        i: (x: React.ReactNode) => <i>{x}</i>,
+        runName: run?.name || '',
+    };
```
trilopin, line 428: I don't understand, couldn't you just pass {runName} to the formatMessage?

webapp/src/components/modals/update_run_status_modal.tsx
```diff
@@ -431,6 +419,22 @@ export const outstandingTasks = (checklists: Checklist[]) => {
...
+    let confirmationMessage = formatMessage({defaultMessage: 'Are you sure you want to finish the run <i>{runName}</i> for all participants?'}, values);
```
trilopin, line 429: unsure about how <i> plays with weblate, could it be troublesome?

trilopin (approved):
Thanks @m-ripper, good job 

All my comments are optional or questions.

webapp/src/components/modals/update_run_status_modal.tsx
m-ripper: I am not sure if this is required but I thought it's a good idea to highlight the run name by making it italic so that it stands out, but I am open to discuss this.

webapp/src/components/modals/update_run_status_modal.tsx
m-ripper: Currently, there are other translations similar to that, for example this: `Currently running the <strong>{playbookTitle}</strong> playbook` so I assume it should work, if these existed before already, but it's probably a good idea to take a closer look.

webapp/src/components/modals/update_run_status_modal.tsx
trilopin: ok, so the I value you pass it's for the i tag you uses in the string..TIL

I'm ok with this I would double-check with someone at translations team. If it's too complex for them we could replace with regular quotes.

webapp/src/components/modals/update_run_status_modal.tsx
trilopin: ok, never paid attention to that. go ahead

calebroseland (approved):
LGTM

Issue #1656, opened 2022-12-06
Force feature flag on for everyone


lieut-data:
Let's bring multiple runs in a channel out from behind the feature flag by forcing it on for everyone. We'll go through and rip out the related code in v7.7.

PR #1657, opened 2022-12-06
Force enable link run feature flag

lieut-data:
## Summary
Force the feature flag for linking multiple runs to a channel to be `true`, regardless of its actual setting. We'll rip this out altogether in an upcoming release.

## Ticket Link
Fixes: https://github.com/mattermost/mattermost-plugin-playbooks/issues/1656

## Checklist
- [x] Unit tests updated


trilopin:
I'm surprised you haven't changed those specs where we do 
 `cy.apiEnsureFeatureFlag('linkruntoexistingchannelenabled', false).then(({prevValue})`. 

Do they still pass?

trilopin (approved):
Thanks @lieut-data

PR #1660, opened 2022-12-07
TaskActions: Allow searching all users

ashishbhate:
## Summary

- Task Actions user search now allows searching all users in the installation

## Ticket Link

- N/A

## Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/webapp/src/components/checklist_item/task_actions_modal.tsx
+++ b/webapp/src/components/checklist_item/task_actions_modal.tsx
@@ -6,7 +6,6 @@ import styled from 'styled-components';
 
 import {searchProfiles} from 'mattermost-webapp/packages/mattermost-redux/src/actions/users';
 import {getUsers} from 'mattermost-redux/selectors/entities/common';
-import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
 
 import ActionsModal from 'src/components/actions_modal';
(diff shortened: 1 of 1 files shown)


lieut-data (approved):

lieut-data (approved):

DHaussermann:
@ashishbhate and @lieut-data I just want to confirm the expected behavior.

I did some testing and I'm seeing that all users on the server can now be targeted by task actions. This differs from other Playbooks functionality where the scope is team specific.  

Based on the summary comment this was the intention and did does allow us to target bots with no team membership.

If this is the desired behavior - LGTM! Please merge.

**note** @ashishbhate there is a UI issue where the user list is dropping down below the frame of the modal so only top 2 users are readable. I saw this last Friday but, it disappeared when I tried to take a screenshot. I thought it was fixed. I'll open a. new issue to address this.

lieut-data:
> I did some testing and I'm seeing that all users on the server can now be targeted by task actions. This differs from other Playbooks functionality where the scope is team specific.

Yes, this is now the intended behaviour. In all other cases, the restriction to team and/or run is "organic", in that it doesn't make sense to even allow configuring otherwise. This is definitely a new case.

Issue #1600, opened 2022-11-24
Add give feedback button to RHS runs list


crspeller:
In the designs there is a give feedback button at the bottom of the RHS runs list. 

https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=786%3A222641&t=Maojsm2eha9PbGNo-4

PR #1666, opened 2022-12-08
add give feedback to run rhs list

trilopin:
## Summary
CleanShot 2022-12-08 at 17 01 01

Add give feedback button to Channels RHS run list.

it's not positioned fixed, it appears at the end of the list.

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1600

## Checklist

- [ ] ~~Telemetry updated~~
- [X] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/channels/rhs/list_spec.js
+++ b/tests-e2e/cypress/integration/channels/rhs/list_spec.js
@@ -89,7 +89,7 @@ describe('channels > rhs > runlist', () => {
         cy.get('[data-testid="dropdownmenu"] > :nth-child(2)').click();
 
         // * Verify displayed options
-        cy.get('[data-testid="rhs-runs-list"]').children().should('have.length', numFinishedRuns);
+        cy.get('[data-testid="rhs-runs-list"]').children().should('have.length', numFinishedRuns + 1);
     });
 
     it('can click though', () => {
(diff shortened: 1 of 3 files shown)


m-ripper (approved):
Looks good! The only thing that could potentially be improved in my opinion, is the hover color of the button. It could be the same color as the feedback button in the playbook run view and the details in the Channels RHS.

gbochora (commented):
Cool! 
Two comments:
- Should we show the `Give feedback` on the black list view when there are no `in-progress/finished` runs?
- We show the `Give feedback` button on the run details view. Was it intentional? Because I couldn't find it in the designs.

trilopin:
> We show the Give feedback button on the run details view. Was it intentional? Because I couldn't find it in the designs.

It was already there (with some subtle differences like being 14px). @abhijit-singh any change you may want to do on this matter?

> One request — Can we make the Give feedback button stick to the bottom when there aren't many runs? It can behave as it is now when the list of runs scrolls off the screen.

Pushed changes to make it fixed (also spaced differently the zero cases that were top-aligned)
CleanShot 2022-12-13 at 11 09 14

trilopin:
Pushed changes:
- if there're enough runs feedback button will be after the last run and will be scrollable
- if there are lesser runs than visible space, the feedback button will be at bottom

additionally, I centered vertically the content for zero cases 

https://user-images.githubusercontent.com/4096774/207417472-5194a79b-d971-490d-8e89-fc7a222b77e9.mp4

abhijit-singh (approved):
💯 

Thank you @trilopin!

PR #1669, opened 2022-12-09
Use proxy state item

trilopin:
## Summary
Follow up https://github.com/mattermost/mattermost-plugin-playbooks/pull/1563 and https://github.com/mattermost/mattermost-plugin-playbooks/pull/1556

**This is a POC to experiment/learn with state management** 

An scenario we usually play at: 
- component state and server state
- there're different interactions: live local save + cta to remote save, local&server save at the same time, double-save with control + modal.
- updates sometimes rely on receiving WebSocket messages

Some of the problems we deal with:
- you can cancel and changes are saved anyway
- you have to click in several different save cta to get it really saved
- if WS fail UI loses the ability to update, if the server is slow to send WS, UI takes time to update

I followed @calebroseland suggestions and played a bit with useProxyState. This code assumes that useProxyState can be used without onChange param.

Handle local state which triggers also a remote side effect to effectively persist in the server.

Several doubts and next steps are added inline as comments.

## Ticket Link
Related https://mattermost.atlassian.net/browse/MM-48857

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/tests-e2e/cypress/integration/playbooks/edit/task_actions_spec.js
+++ b/tests-e2e/cypress/integration/playbooks/edit/task_actions_spec.js
@@ -75,6 +75,12 @@ describe('playbooks > edit > task actions', () => {
             // Save the dialog
             cy.findByTestId('modal-confirm-button').click();
 
+            // Save the edit to trigger remote update
+            cy.findByTestId('checklist-item-save-button').click();
+
+            // Edit again
+            editTask();
+
             // Verify no actions are configured
             cy.findByText('Task Actions').should('exist');
             cy.apiGetPlaybook(testPlaybook.id).then((playbook) => {
@@ -105,6 +111,12 @@ describe('playbooks > edit > task actions', () => {
 
             // Save the dialog
             cy.findByTestId('modal-confirm-button').click();
+
+            // Save the edit to trigger remote update
+            cy.findByTestId('checklist-item-save-button').click();
+
+            // Edit again
+            editTask();
 
             // Verify configured actions
             cy.findByText('1 action');
@@ -139,6 +151,12 @@ describe('playbooks > edit > task actions', () => {
             // Save the dialog
             cy.findByTestId('modal-confirm-button').click();
 
+            // Save the edit to trigger remote update
+            cy.findByTestId('checklist-item-save-button').click();
+
+            // Edit again
+            editTask();
+
             // Verify configured actions
             cy.findByText('1 action');
             cy.apiGetPlaybook(testPlaybook.id).then((playbook) => {
@@ -166,6 +184,12 @@ describe('playbooks > edit > task actions', () => {
 
             // Save the dialog
             cy.findByTestId('modal-confirm-button').click();
+
+            // Save the edit to trigger remote update
+            cy.findByTestId('checklist-item-save-button').click();
+
+            // Edit again
+            editTask();
 
             // Verify configured actions
             cy.findByText('1 action');
@@ -196,6 +220,12 @@ describe('playbooks > edit > task actions', () => {
             // Save the dialog
             cy.findByTestId('modal-confirm-button').click();
 
+            // Save the edit to trigger remote update
+            cy.findByTestId('checklist-item-save-button').click();
+
+            // Edit again
+            editTask();
+
             // Re-open the dialog
             cy.findByText('1 action').click();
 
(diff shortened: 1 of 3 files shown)


webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -99,40 +92,16 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
-        if (props.newItem) {
-            return;
-        }
+    // assignee state management
+    const saveAssignee = async (userId = '') => {
         if (props.playbookRunId) {
             const response = await setAssignee(props.playbookRunId, props.checklistNum, props.itemNum, userId);
             if (response.error) {
-                console.log(response.error); // eslint-disable-line no-console
+                console.error(response.error); // eslint-disable-line no-console
```
trilopin, line 104: Would make sense to revert the local change when the remote has errored?

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -413,6 +429,9 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
                                 command,
                                 description: descValue,
                             });
+                            saveAssignee(assigneeID);
+                            saveDueDate(dueDate);
+                            saveTaskActions(taskActions);
```
trilopin, line 434: This works perfectly but it's a bit ugly, 3 requests to make updates even if only 1 param has changed.

I'd be inclined to create the updateChecklistItem as a GQL mutation (as updateRun or updatePlaybook) and accept optionally all fields in one only mutation.

tests-e2e/cypress/integration/playbooks/edit/task_actions_spec.js
```diff
@@ -75,6 +75,12 @@ describe('playbooks > edit > task actions', () => {
             // Save the dialog
             cy.findByTestId('modal-confirm-button').click();
 
+            // Save the edit to trigger remote update
+            cy.findByTestId('checklist-item-save-button').click();
+
+            // Edit again
+            editTask();
```
trilopin, line 82: For the task actions modal, the experience is still weird because we have to click two Save CTA to make it really persist (before it saved in first no matter what you clicked in the second 😂 ).

I fixed the e2e just because I wanted them in green, but we need to think better about this one.

trilopin: oh, run task actions e2e are still red :(

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -162,8 +133,28 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
+    const [command, setCommand] = useProxyState(props.checklistItem.command, (cmd: string) => {
+        if (props.newItem || isEditing) {
+            return;
+        }
+        saveCommand(cmd);
+    }, 0);
+
+    useUpdateEffect(() => {
+        setTaskActions(taskActions);
+    }, [props.checklistItem.task_actions]);
```
calebroseland, line 155: Is this needed with `useProxyState(props.checklistItem.task_actions ...` below?

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -175,12 +166,20 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
...
+    const [taskActions, setTaskActions] = useProxyState(props.checklistItem.task_actions, (actions) => {
+        if (props.newItem || isEditing) {
+            return;
+        }
+        saveTaskActions(actions);
+    });
+    const saveTaskActions = async (newTaskActions: TaskActionType[]) => {
         if (props.newItem) {
             return;
         }
```
calebroseland, line 181: Same here with this above

```tsx
        if (props.newItem || isEditing) {
            return;
        }
        
```

tests-e2e/cypress/integration/playbooks/edit/task_actions_spec.js
calebroseland: > For the task actions modal, the experience is still weird because we have to click two Save CTA to make it really persist (before it saved in first no matter what you clicked in the second 😂 ).

+1

calebroseland (changes requested):
This is great, @trilopin! I pushed 1 change, fixed the e2e for that other spec, and there's just 2 questions, otherwise I think this is looking pretty good on the technical side. (I agree some things could be better on the UX flows)

lieut-data:
Closing this out for now as we revisit the feature set for Playbooks.

Issue #1670, opened 2022-12-09
"View all" runs shouldn't filter to "My runs only"


lieut-data:
Ooh, I just remembered a feature request: "View all" shouldn't limit to my runs by default now that we have LHS:

![image](https://user-images.githubusercontent.com/1023171/206747552-7a8115df-4e57-4d81-9f97-c41056daa4b2.png)

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/modxeqahatdszfra9pcqqyxrme) *by @lieut-data*._

PR #1672, opened 2022-12-09
make view all runs the default filter at run list

trilopin:
## Summary
"My Runs Only" is disabled by default when visiting run lsit

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1670

## Checklist
- [ ] ~~Telemetry updated~~
- [X] Unit tests updated


lieut-data (approved):
:tada:

Issue #1617, opened 2022-11-29
change linked channel in an existing run


trilopin:
Trigger in run list at channels RHS

![Image](https://user-images.githubusercontent.com/4096774/204826186-b711a432-92b5-48dc-ae58-b49a40a99c8f.png)

[Discussion at community](https://community.mattermost.com/core/pl/uokc1tjiipnx3xcpkzq6fmuc4w) with the following decision:
- we keep RHS home to users without playbooks
- we show zero cases when no run is in progress and/or finished in the current channel
- we keep the user in the current channel after linking a run to a different existing channel

[Figma](https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=1296%3A246033&t=UJdE2vtmVR1nEVg9-4)

PR #1678, opened 2022-12-13
1617 change linked channel

trilopin:
## Summary
Two main new features from the run card:
- link tun to a different channel
- rename run

Both are managed with a modal triggered from kebab menu.

https://user-images.githubusercontent.com/4096774/207548878-aec8e5fd-30b7-4852-aaf5-36bc12fb25a8.mp4

### Zero cases in RHS Run list
- If the user can create runs and there are 0 runs, a primary button is shown.
- If there are run in the "other" state (in progress vs finished) a link/button will be shown

Screenshot 2022-12-13 at 20 05 22

### RHS Home changes
The RHS Home won't be shown when no runs anymore, only when no playbooks.
The code that showed playbooks inside the RHS Home has been removed, that goal is now covered by the start-a-run modal.

### Honour permissions
RHS start-run ctas now are disabled when user has no permissions to do such action. It's delayed to the modal since the permission is playbook based.
  
## UPDATE
From @abhijit-singh suggestions:
- fixed modal style: border, input height, padding etc...
- prevent user from navigating to RHS details after moving a run to another channel (triggered before if only 1 run)
- animation when we move a run to another channel

2/5 on the current approach, it seems a bit complicated due to the distance between where action triggers and where animation shows... Errors are very unlikely in the updateRun but will hide the card from the list even if the remote op fails. I've checked other mutations and we are currently just assuming they don't fail 😱.

https://user-images.githubusercontent.com/4096774/208472746-289855e4-a81c-4bd4-b267-792e47718545.mp4

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1617
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1679

## Checklist
- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


trilopin:
Code is not fully ready, but I'd love to have an early ux review.

abhijit-singh (changes requested):
Looks good for the most part @trilopin! A few observations:
1. In the case when there are 2 runs in a channel and then the user links one of them to a different channel, which leaves just one run in the channel, we automatically navigate to the run detail view in the RHS. This could be jarring to the user because they might not have context on what that run is and why they're looking at it suddenly. It will be ideal if we can keep them on the runs list page even with a single run as an exception here.
2. Non blocking but the link to a different channel selector modal has some dissimilarities from the [design](https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=1296%3A246033&t=7ZP7YO2BT8eKnOHm-4). Adding the run name, removing the modal header divider, and making the input box larger would fix it.
image
3. Also non blocking and definitely a nice to have but it'll be cool to have some kind of transition for the run card that was linked to a different channel. Like in the task inbox or the one I had in the design — [prototype link](https://www.figma.com/proto/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?page-id=321%3A141102&node-id=786%3A222641&viewport=-5557%2C12731%2C1&scaling=min-zoom&starting-point-node-id=786%3A222641)

trilopin:
> 1. In the case when there are 2 runs in a channel and then the user links one of them to a different channel, which leaves just one run in the channel, we automatically navigate to the run detail view in the RHS. This could be jarring to the user because they might not have context on what that run is and why they're looking at it suddenly. It will be ideal if we can keep them on the runs list page even with a single run as an exception here.

Ok, will try to avoid the expected automated flow in this case.

> 2. Non blocking but the link to a different channel selector modal has some dissimilarities from the [design](https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=1296%3A246033&t=7ZP7YO2BT8eKnOHm-4). Adding the run name, removing the modal header divider, and making the input box larger would fix it.

Done


> 3. Also non blocking and definitely a nice to have but it'll be cool to have some kind of transition for the run card that was linked to a different channel. Like in the task inbox or the one I had in the design — [prototype link](https://www.figma.com/proto/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?page-id=321%3A141102&node-id=786%3A222641&viewport=-5557%2C12731%2C1&scaling=min-zoom&starting-point-node-id=786%3A222641)

will open a separate task for this one

webapp/src/components/modals/run_update.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const onSubmit = () => {
+        onSubmitStart?.();
+        if (field === 'name') {
+            updateRun({name});
+            onSubmitEnd?.('');
+        }
+
+        if (field === 'channel_id') {
+            // delay is to allow animation to shine
+            setTimeout(() => {
```
lieut-data, line 70: nit: Is there a way to handle this on the render side instead? In principle, it seems like this animation might be allowed to fire regardless of what triggers the change (e.g. if someone else finishes the run, and I happen to be looking at the list).

webapp/src/components/modals/run_update.tsx
```diff
@@ -0,0 +1,197 @@
...
+            }, 700);
+        }
+    };
+
+    const isFormValid = () => {
+        if (field === 'name') {
+            return name !== '' && name !== run?.name;
+        }
+        return channelId !== '' && channelId !== run?.channel_id;
+    };
```
lieut-data, line 82: I'm wondering if there's an opportunity to simplify and just track the "value" of this modal, vs. having separate fields tracking separate values depending on how the field is configured? Then this block might look like:

```ts
    const isFormValid = () => {
        return value !== '' && value !== run?[field];
    };
```

Possibly other blocks might be clarified too. 0/5!

lieut-data (approved):
A huge effort, @trilopin! Nothing blocking, just a few thoughts -- thank you :)

webapp/src/components/modals/run_update.tsx
trilopin: Not sure if I understand your point. Name behavior is very different from channel behavior. One is plain text and the other is a special resource with its own selector ChannelSelector. I tried to re-use the modal envelope but I'm not that confident about making it more generic given the different nature of the fields.

trilopin: yeah, this is a bit on the line. I'm more inclined toward small duplication than a premature forced abstraction.

trilopin:
Changed animation/modal/handlers approach:
- update hooks now are very close to where animation happens
- simple handlers passed to modals
- modals split in two simpler components
- use onAnimationEnd event instead of ugly setTimeout

webapp/src/components/rhs/rhs_run_list.tsx
```diff
@@ -399,16 +408,37 @@ const BlueCheckmark = styled(CheckIcon)`
...
     const participatIDsWithoutOwner = props.participantIDs.filter((id) => id !== props.ownerUserID);
+    const [movedChannel, setMovedChannel] = useState({channelId: '', channelName: ''});
+    const updateRun = useUpdateRun(props.id);
 
     return (
-        <CardWrapper progress={props.progress * 100}>
+        <CardWrapper
+            progress={props.progress * 100}
+            className={removed ? 'removed' : ''}
+            onAnimationEnd={() => {
```
lieut-data, line 430: TIL! No idea `onAnimationEnd` was built-in :tada:

lieut-data (approved):

m-ripper (approved):
Well done!

abhijit-singh (approved):
Perfect 💯 

Thank you @trilopin!

PR #1682, opened 2022-12-15
MM-48917 Make interactive playbook run dialogs translatable

m-ripper:
## Summary
Changed all hard-coded strings of the interactive dialogs related to playbook runs in the backend to use the i18n translated string instead.
This means, that the following dialogs can now be translated to the user's locale:
-  AddChecklistItemDialog - `/playbook checkadd 0 0`
- FinishPlaybookRunDialog - `/playbook finish 0`
- PlaybookRunDialog - `/playbook run`
- UpdatePlaybookRunDialog - `/playbook update 0`
- AddToTimelineDialog

## Ticket Link

[MM-48917](https://mattermost.atlassian.net/browse/MM-48917)

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -2443,13 +2443,7 @@ func (s *PlaybookRunServiceImpl) newFinishPlaybookRunDialog(playbookRun *Playboo
 		"RunName": playbookRun.Name,
 		"Count":   outstanding,
 	}
-
-	message := T("app.user.run.confirm_finish.question", data)
-	if outstanding == 1 {
-		message = T("app.user.run.confirm_finish.one_outstanding") + " " + message
-	} else if outstanding > 1 {
-		message = T("app.user.run.confirm_finish.num_outstanding", data) + " " + message
-	}
+	message := T("app.user.run.confirm_finish.num_outstanding", data)
 
 	return &model.Dialog{
 		Title:            T("app.user.run.confirm_finish.title"),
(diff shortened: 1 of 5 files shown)


server/app/playbook_run_service.go
```diff
@@ -2476,41 +2505,43 @@ func (s *PlaybookRunServiceImpl) newPlaybookRunDialog(teamID, ownerID, postID, c
...
+	introductionText := T("app.user.run.update_status.intro") + " "
 
-		introductionText += fmt.Sprintf(" This post will be broadcasted to %d channel%s.", broadcastChannelNum, plural)
+	if broadcastChannelNum == 1 {
+		introductionText += T("app.user.run.update_status.one_channel")
+	} else {
+		data := map[string]interface{}{
+			"Count": broadcastChannelNum,
+		}
+		introductionText += T("app.user.run.update_status.num_channel", data)
```
lieut-data, line 2544: Not necessary for this PR, but I recall there is if/else logic available in translations, and we might consider folding this into a single translation string, simply passing `Count == 1` to handle the other case inline.

lieut-data (approved):
Awesome work, @m-ripper :)

server/app/playbook_run_service.go
calebroseland: Server-side I don't think conditionals are supported beyond simple plural rules, but indeed since this is only branching on the count/plural, this would ideally just be one translation. For our translators, it would be better to not introduce a new translation if we're going to remove it right away.

calebroseland (approved):
LGTM, just the one thing below. Thanks, @m-ripper!

4/5 that `app.user.run.update_status.one_channel` and `app.user.run.confirm_finish.one_outstanding` should be removed/unified before this goes in, but I'd be okay with deferring that for a follow up if necessary..

Issue #1676, opened 2022-12-13
double RHS when creating a Run


trilopin:
#bug When I ran a playbook for the first time on a new installation, I landed on this screen (screenshot attached) after clicking on `Start run` in the modal.
Steps to reproduce: Create a new channel, then in the Channels RHS, click on `Use` for a playbook template > Create playbook > Click on `Run` on the playbook page > Choose `Link to existing channel` (specify channel just created) > Start run.

https://user-images.githubusercontent.com/4096774/207314753-9dde9fb6-ff5f-4f7c-a884-932a9233b9b0.mp4

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/6toj19hact8ainqenqimfjddpo) *by @abhijit-singh*._

PR #1685, opened 2022-12-15
 GH-1676: only open rhs when in a channel

calebroseland:
# Summary
Adds a check to ensure we're viewing a channel before opening RHS. `currentChannelId` can sometimes have a value even though you're not in a channel. 

# Ticket Link
- Fixes: #1676 

## Checklist

- ~[ ] Telemetry updated~
- ~[ ] Gated by experimental feature flag~
- [x] Unit tests updated

### DIFF
--- a/webapp/src/rhs_opener.ts
+++ b/webapp/src/rhs_opener.ts
@@ -7,6 +7,8 @@ import {getCurrentChannel} from 'mattermost-redux/selectors/entities/channels';
 import {getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
 import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
 
+import {matchPath} from 'react-router-dom';
+
 import {fetchPlaybookRuns, telemetryEventForPlaybookRun} from 'src/client';
 import {currentPlaybookRun, isPlaybookRunRHSOpen, inPlaybookRunChannel} from 'src/selectors';
 import {PlaybookRunStatus} from 'src/types/playbook_run';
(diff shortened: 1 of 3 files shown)


trilopin (approved):
Thanks 🎉

webapp/src/rhs_opener.ts
```diff
@@ -24,12 +26,14 @@ export function makeRHSOpener(store: Store<GlobalState>): () => Promise<void> {
         const currentChannel = getCurrentChannel(state);
         const currentTeam = getCurrentTeam(state);
         const playbookRun = currentPlaybookRun(state);
+        const url = new URL(window.location.href);
+        const isInChannel = matchPath(url.pathname, {path: '/:team/:path(channels|messages)/:identifier/:postid?'});
```
lieut-data, line 30: Do these parameters implicitly have to match whatever was registered by the webapp? (i.e. are these brittle)

Wondering if it makes sense to just verify that we're in the Channels product instead?

lieut-data (approved):

lieut-data:
Also: any hope of an E2E test?

calebroseland:
> Also: any hope of an E2E test?

Yup, was going to try to get back to this for that today

webapp/src/rhs_opener.ts
calebroseland: This is indeed brittle, but unavoidable to my understanding. All of the `currentProduct` and like selectors are all exclusively in mm-webapp, and it'd be just as brittle IMO to raw-select the products and manually recreate `useCurrentProductId`. I'm only 1/5 on the current path-check; happy to switch it over.

PR #1686, opened 2022-12-16
Enable sorted imports in ESLint

m-ripper:
## Summary

Some IDEs, like IntelliJ, by default will try to sort the imports. 
This can cause PRs to be cluttered with changes that are not necessarily related to it.

In order to avoid this in the future. The `sort-imports` rule of ESLint is now enabled and configured.

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


calebroseland (approved):
Looks good! I like things that reduce noise.

trilopin (approved):
nice Max, thank you!

PR #1687, opened 2022-12-16
MM-48607 Fix bug that caused the upgrade modal to not be shown when requesting an update

m-ripper:
## Summary

Fix a bug that caused the upgrade modal to not be shown when requesting an update. 
The bug was caused because the click event that caused the modal to be shown was propagated and immediately triggered `onHide` of the upgrade modal which closed the modal. 

## Ticket Link

[MM-48607](https://mattermost.atlassian.net/browse/MM-48607)

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
@@ -433,7 +433,10 @@ const RequestUpdateButton = ({type, onClick, disabled = false}: {disabled: boole
             {type === 'dotmenu' ? (
                 <DotMenuItem
                     disabled={disabled}
-                    onClick={() => setShowUpgradeModal(true)}
+                    onClick={(e) => {
+                        e.stopPropagation();
+                        setShowUpgradeModal(true);
+                    }}
                 >
                     {formatMessage({defaultMessage: 'Request update...'})}
                     <KeyVariantCircleIcon


webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
```diff
@@ -433,7 +433,10 @@ const RequestUpdateButton = ({type, onClick, disabled = false}: {disabled: boole
             {type === 'dotmenu' ? (
                 <DotMenuItem
                     disabled={disabled}
-                    onClick={() => setShowUpgradeModal(true)}
+                    onClick={(e) => {
+                        e.stopPropagation();
```
lieut-data, line 437: Ooh, curious. What ends up happening without this directive? Should `DotMenuItem` itself handle this instead?

lieut-data (approved):

webapp/src/components/backstage/playbook_runs/playbook_run/status_update.tsx
m-ripper: Without the directive, the click on the button is also registered by the modal but outside of the modal, and therefore `onHide` is immediately called, which closes the modal

`DotMenuItem` is a styled DOM element, otherwise I would agree that it's a good idea to make the propagation stop there.

trilopin (changes requested):
I'm not sure how that change affected bit now the modal appears but doesn't properly work.

In the following video, I show first how it works in the case I'm a viewer, and then how it fails when I'm a participant. The modal opens correctly but it closes no matter where I click. Here are a few more details:
- It closes immediately, without the animation
- It closes even if I click on the start trial button
- it closes even if I click on the modal but not on the buttons

https://user-images.githubusercontent.com/4096774/208389190-0935ba0c-b5a3-400f-9745-04a520bb5662.mp4

m-ripper:
> I'm not sure how that change affected to it. Now the modal appears but doesn't properly work.
> 
> In the following video, I show first how it works in the case I'm a viewer, and then how it fails when I'm a participant. The modal opens correctly but it closes no matter where I click. Here are a few more details:
> 
>     * It closes immediately, without the animation
> 
>     * It closes even if I click on the start trial button
> 
>     * it closes even if I click on the modal but not on the buttons
> 
> 
> CleanShot.2022-12-19.at.10.08.27.mp4

Thanks for pointing out! I found out that the cause of the issues for participants is, that the `RequestUpdateButton` is in a `Dropdown`. This has the side-effect that closing the dropdown will cause the modal to be removed from the document and this caused the issues.

I've fixed this by moving the modal outside the dropdown.

lieut-data (approved):
Thanks for digging into this deeper @trilopin & @m-ripper!

trilopin (approved):
Cool 🎉

PR #1688, opened 2022-12-19
MM-46983 Visual feedback for invalid run names

m-ripper:
## Summary

Implement visual feedback when a run name is invalid. A detailed error message will be shown in that case. 
Currently, it will only be checked if the run name exceeds 64 characters.

[Screencast from 21.12.2022 15:18:00.webm](https://user-images.githubusercontent.com/8669935/208926553-743590ef-30c8-485b-9d40-07fa01476cb1.webm)

## Ticket Link

[MM-46983](https://mattermost.atlassian.net/browse/MM-46983)

## Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/playbooks/start_run_spec.js
+++ b/tests-e2e/cypress/integration/playbooks/start_run_spec.js
@@ -147,37 +147,6 @@ describe('playbooks > start a run', () => {
     });
 
     describe('from playbook editor', () => {
-        it('error is shown when maximum length of run name is exceeded', () => {
-            // # Visit the selected playbook
-            cy.visit(`/playbooks/playbooks/${testPlaybook.id}/outline`);
-
-            // # Click start a run button
-            cy.findByTestId('run-playbook').click();
-
-            cy.get('#root-portal.modal-open').within(() => {
-                // # Wait the modal to render
-                cy.wait(500);
-
-                // * Assert template name is empty
-                cy.findByTestId('run-name-input').should('have.value', '');
-
-                // # Type run name that exceeds maximum length
-                cy.findByTestId('run-name-input').type('a'.repeat(RUN_NAME_MAX_LENGTH + 1));
-
-                // # Click start button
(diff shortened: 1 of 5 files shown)


calebroseland (dismissed):
> but there’s no user facing error to explain

When the run name is empty, its somewhat self-explanatory why Start Run is disabled, but in the event that it's too long, it may not be obvious. 

I'd suggest either an explicit error message (describing the 64 char limit in a friendly tone) just below the run name input. 

There's anther pattern of truncating pastes/rejecting keystrokes after the limit at the input level, and that could be considered, but it's not my favorite. (Though I'd say slightly better than no error message)

cc @abhijit-singh

lieut-data (approved):
Agreed with @calebroseland's feedback (but letting his review block). IMO, we should always leave the buttons clickable and instead show an error on such an attempt.

m-ripper:
I've added a small error message now but I am 0/5 in terms of the design.
[Screencast from 20.12.2022 17:50:00.webm](https://user-images.githubusercontent.com/8669935/208721427-f2fab729-5260-4e4e-afd5-d21fb3602b55.webm)

abhijit-singh:
This is great @m-ripper. The flow looks good.  I'd request matching the styling of the error state to this design — https://www.figma.com/file/eq9Nwyot47KZ9gb4pf5C2L/%E2%96%B6%EF%B8%8F-Multiple-runs-in-a-channel?node-id=1649%3A256088&t=1J7Z2BFAbNoCXlEU-4

image

trilopin (approved):
It's cool 🎉.  

However, I would want to leave some suggestions/questions for the future:
- I'd convert BaseInput in a component with the styling + a maxLength optional prop and handle inside the state.
- would make sense to have a component with label+field+error?
- would make sense to stop writing on the input once the limit has been reached? 
- would make sense to trigger the check on keydown event instead of onsubmit?

abhijit-singh (approved):

PR #1689, opened 2022-12-19
Lint: named imports single/multiline

calebroseland:
## Summary
Lints named-imports single/multiline behavior. Max 3 named-imports go to single line. (Excludes default import, so `import Thing, {a, b, c} from 'thing'` is possible.) >= 4 named-imports and it will switch to multiline. 

Also adds `make fix-style`


trilopin:
100% with the motivation.

My only question is, should we do one only PR for @m-ripper PR and this one?

trilopin (approved):

calebroseland:
@trilopin 0/5  they can go in together or separately (after #1686 is merged this will auto-rebase to `master`). I just wanted the diff to be clear in this as it would also touch a bunch of the same lines.

m-ripper (approved):
Very nice!

PR #1691, opened 2022-12-20
Upgrade to Cypress v12

lieut-data:
## Summary
Upgrade to Cypress 12, with requisite e2e test updates to respect test isolation: no longer is the browser state preserved between tests, so logging in again is explicitly required, and any UI must be setup afresh.


tests-e2e/cypress/integration/channels/general_actions_spec.js
```diff
@@ -14,15 +14,27 @@ describe('channels > general actions', () => {
...
             testSysadmin = user;
 
             cy.apiCreateUser().then((resp) => {
                 testUser = resp.user;
                 cy.apiAddUserToTeam(team.id, resp.user.id);
+
+                cy.apiLogin(testUser);
+
+                // TODO: Make this work with CRT enabled.
+                cy.apiSaveCRTPreference(testUser.id, 'off');
```
lieut-data, line 30: This isn't strictly specific to these changes, just that I had had `always_on` configured on my server and ran into this, so fixed it for at least when the server doesn't require it.

trilopin:
> Upgrade to Cypress 12, with requisite e2e test updates to respect test isolation: no longer is the browser state preserved between tests, so logging in again is explicitly required, and any UI must be setup afresh.

Is the isolation something we desire or just a requirement from cypress12? 

Have you tried to play with [cy.session](https://docs.cypress.io/guides/references/migration-guide#Command-Cypress-API-Changes) to avoid changing that many tests?

tests-e2e/cypress/integration/playbooks/edit_metrics_spec.js
```diff
@@ -528,7 +531,7 @@ const addMetric = (type, title, target, description) => {
 
 const verifyViewMetric = (index, title, target, description) => {
     cy.getStyledComponent('ViewContainer').should('have.length.of.at.least', index + 1).eq(index).within(() => {
-        cy.getStyledComponent('Title').should('have.text', title);
+        cy.getStyledComponent('Title-').should('have.text', title);
```
trilopin, line 534: why is the dash now needed?

tests-e2e/cypress/integration/runs/rdp_main_retrospective_spec.js
```diff
@@ -190,22 +191,26 @@ describe('runs > run details page', () => {
...
+
             beforeEach(() => {
-                cy.apiLogin(testViewerUser).then(() => {
-                    cy.visit(`/playbooks/runs/${testRun.id}`);
-                });
+                // Login as the test viewer
+                cy.apiLogin(testViewerUser);
+
+                // # Visit the playbook run
+                cy.visit(`/playbooks/runs/${testRun.id}`);
```
trilopin, line 213: Don't we need to chain the visit to the login? Aren't you exposed to visiting the URL before finishing the login?

trilopin (approved):
I'm happy you did this, cypress12 is expected to be a bit faster than 10 🎉 

I left some questions about changes I don't understand.

tests-e2e/cypress/integration/playbooks/edit_metrics_spec.js
lieut-data: /This/ one appears to be spurious, at least after poking at the spec again. But in the case of `rdp_rhs_runinfo_spec.js`, it boils down to matching both `ParticipantsContainer` and `Participants`, which ultimately traces back to this change in [Cypress 12](https://docs.cypress.io/guides/references/changelog#12-0-0):

> The [.within()](https://docs.cypress.io/api/commands/within) command now requires a single subject and throws an error if given more than one subject. This change adds consistency around how .within() behaves across commands. Previously some commands inside a .within() callback would silently select the first element, while others would use all of the previously yielded subjects, and others would throw an error.

Given all styled components look like `<Name>-`, I'm going to test changing the `getStyledComponent` helper to add this dash all the time as a way of sidestepping the problem going forward. (Unless we reuse a name ;P)

m-ripper (approved):
Nice! Let's hope the performance is noticeably better.

tests-e2e/cypress/integration/runs/rdp_main_retrospective_spec.js
lieut-data: I /think/ the way Cypress's promise model works internally amounts to something like:
```
await cy.apiLogin(...)
await cy.visit(...)
```

And everything works out in the end. The only time `.then` is needed is if the results of one operation need to be referenced explicitly in another, e.g. the creation of a playbook run and then using that playbook run's ID.

lieut-data:
> Is the isolation something we desire or just a requirement from cypress12?

@trilopin, you're right that we could revert to the previous behaviour. I was actually surprised by how few tests required changes (e.g. < 20 out of 500+ tests). My inclination is to double down on test isolation for hygiene -- something I've heard from SDETs in general as being good practice. Thoughts?

trilopin:
> @trilopin, you're right that we could revert to the previous behaviour. I was actually surprised by how few tests required changes (e.g. < 20 out of 500+ tests). My inclination is to double down on test isolation for hygiene -- something I've heard from SDETs in general as being good practice. Thoughts?

I was very surprised too. I have nothing against isolation, just was curious.

PR #1692, opened 2022-12-20
1680 Fix playbook create permissions

### LINKED ISSUE #1680: create playbook cta active even without permissions
#bug primary button to create playbook is active even without permissions (team override scheme).

Api returns a 403 correctly, but button should be disabled and a tooltip shown.

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/hgupcr8hujghdx1jpwzghzbjgo) *by @trilopin*._

m-ripper:
## Summary

Fix check if playbook can be created when assigned to multiple teams.
Instead of checking if the user has permissions in any team he's member of, it will now only be checked for the currently selected team.

## Ticket Link

Fixes #1680 

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


trilopin (approved):
Thanks @m-ripper 

please do not merge this pr before we merge the linked-channel one

trilopin:
I guess `useCanCreatePlaybooksOnAnyTeam` is here since the no-team-sidebar ages. Do you think we can remove that function? @m-ripper @lieut-data

webapp/src/components/backstage/runs_page_no_content.tsx
```diff
@@ -91,37 +87,33 @@ const NoContentPlaybookRunSvgContainer = styled.div`
...
-    const teams = useSelector<GlobalState, Team[]>(getMyTeams);
-    const goToMattermost = () => {
-        navigateToUrl('');
-    };
-    const [playbookExist, setplaybookExist] = useState(false);
+    const teamId = useSelector(getCurrentTeamId);
+    const [playbookExist, setPlaybookExist] = useState(false);
+    const [,,
+        {setSelectedPlaybook},
+    ] = usePlaybooksCrud({team_id: '', per_page: BACKSTAGE_LIST_PER_PAGE});
```
lieut-data, line 94: nit: I wonder if `usePlaybooksCrud` should eventually be migrated to always returning an object, like `usePlaybooksRouting`?

webapp/src/components/rhs/rhs_home.tsx
```diff
@@ -109,10 +109,7 @@ const ListSection = styled.div`
 const RHSHome = () => {
     const dispatch = useDispatch();
     const currentTeam = useSelector(getCurrentTeam);
```
lieut-data, line 111: Should we simplify to just `getCurrentTeamId` if we only use the id?

lieut-data (approved):
Thanks, @m-ripper! Can we look to add e2e coverage for this change?

webapp/src/components/backstage/runs_page_no_content.tsx
m-ripper: That's a good point. I haven't looked into `usePlaybooksCrud` in-depth but judging by the signature there, there should be a better way and I'll gladly take a look if that can be migrated.

Issue #1618, opened 2022-11-29
telemetry playbookrun_create


trilopin:
Move telemetry event playbookrun_create to frontend.

Add the following new information to the event:
- place: playbook_editor | slash_command | channels_rhs | channel_keyword
- channel_created: true | false
- channel_id: channel-id used 
- channel_creation_method: 'create' | 'use' 
- channel_id_orig: channel-id default when creating (the one configured in PBE)

The questions we can answer with this:
- how many runs the users started
- where are users starting a run? split by place
- are users using the defaults? how many users change from creating to link-to-a-existing? how many users changed the channel that was default?

The event should be moved from the server to the webapp (release must be coordinated with Sri/Jose).

To discuss: 
- are there more ways to create a run?
- anyone is concerned about moving this important event?

Note that names may need a bit more thinking, I just used ones to explain the concept.

Pending Tao's feeback

PR #1694, opened 2022-12-20
1618 telemetry multiple runs

trilopin:
## Summary
Telemetry for run creation.

New modal trigger an event `playbookrun_create`with the following data:
- playbookId: playbook used
- channelMode: 'create_new_channel' | 'link_existing_channel'
- public: true if run if created channel is public
- hasPlaybookChanged: true if playbook has changed or filled in the modal process
- hasNameChanged: true if user changed default name
- hasSummaryChanged: true if user changed default summary
- hasChannelModeChanged: true if user changed the channel mode of the playbook
- hasChannelIdChanged: true if the user changed the linked channel in the playbook
- hasPublicChanged: true if the user changed the visibility of the channel that is configured in the playbook
- place: 'channels_rhs_home' | 'channels_rhs_runlist' | 'backstage_playbook_list' | 'backstage_playbook_editor'

Questions we could answer with these changes:
- how many runs are created?
- how many runs are created split by the place where the modal is triggered?
- how many runs are created using default vs any change?
- what props are changed the most when creating runs? does it happen more when they are started from some places?

E2E checking telemetry:
- [X] start run from RHS
- [x] start run from backstage

This pr don't cover (followup pr coming):
- migrate old server event 
- track run creation through dialog server side with this same event / properties
- track page views for rhs_home and rhs_runlist

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1618

## Checklist
- [X] Telemetry updated
- [X] Unit tests updated


m-ripper (approved):
Looks good to me, thank a lot.

gbochora (approved):
Great work! 🎉

PR #1695, opened 2022-12-21
page tracking for all RHS variants +  e2e test

trilopin:
## Summary
Follow up PR for telemetry data

Added pageview tracking for all three variants of RHS.

Migrated all places with standard interception style to use the new ui_command that is more flexible. 

ui_command changes:
- Renamed ui_command from expectTelemetryToBe to expectTelemetryToContain, since it's not required to add every event captured (we should check just the relevant ones for the test)
- add optional arg to tell the command how many telemetry call should expect (default to auto)

## Ticket Link
No

## Checklist
- [x] Telemetry updated
- [x] Unit tests updated


tests-e2e/cypress/support/plugin/ui_commands.js
```diff
@@ -311,12 +311,7 @@ Cypress.Commands.add('expectTelemetryToContain', (items) => {
     cy.get('@telemetry.all').then((xhrs) => {
         items.forEach((item) => {
             const matches = xhrs.filter((xhr) => xhr.request.body.name === item.name && xhr.request.body.type === item.type);
-            if (matches.length === 0) {
-                expect.fail(`No telemetry event found for ${item.name} of type ${item.type}`);
-            }
-            if (matches.length > 1) {
-                expect.fail(`Unsupported multiple telemetry event found for ${item.name} of type ${item.type}`);
-            }
+            expect(matches).to.have.lengthOf(1, `Unsupported number of matches (${matches.length} != 1) telemetry event found for ${item.name} of type ${item.type}`);
```
gbochora, line 314: Very compact.

gbochora (approved):
Nice work!

trilopin:
@lieut-data when merging master I noticed that you did a similar (but working better) code in the ui_command to avoid adding all events to the expectTelemetry command. 

Thanks a lot!

tests-e2e/cypress/support/plugin/ui_commands.js
```diff
@@ -308,11 +308,23 @@ Cypress.Commands.add('interceptTelemetry', () => {
...
+
+// cy.expectTelemetryToContain expects to find the given telemetry events in the order given among the
 // recorded telemetry. It doesn't fail if other telemetry events happen to occur in between.
-Cypress.Commands.add('expectTelemetryToBe', (items) => {
-    // Wait for at least as many telemetry events as requested.
-    items.forEach(() => cy.wait('@telemetry'));
+Cypress.Commands.add('expectTelemetryToContain', (items, opts) => {
+    const options = {...defaultExpectTelemetryToContainOptions, ...opts};
+
+    // Wait for at least as many telemetry events as requested if auto, or explicit number is passed.
```
lieut-data, line 320:
```suggestion
// Wait for at least as many telemetry events as requested if auto, or explicit number if passed.
```

Would a test that passes a number here likely be brittle in the face of unexpected code changes?

lieut-data (approved):
Wow! A New Years gift :)

tests-e2e/cypress/support/plugin/ui_commands.js
trilopin: The passed number is the number of guaranteed requests to be waited for.

If I want to test follow&unfollow events but the pageview is additionally tracked, I should wait for 3 events to ensure my desired events are captured. If I stopped tracking the pageview the test will fail because we can not satisfy the third wait. 

I could add a comment to the passed param about which events are waited but not asserted, in order to solve a potential error in an easier way. But that is only a comment...


What I would eventually want, but I still didn't completely solve:
- [X] validate just the properties passed 
- [X] check that some events triggered
- [ ] wait automatically for all events
- [ ] ignore the order of the events

trilopin: As we add more and more events, it's not very nice to assert all events that trigger every time we want to ensure that one event is triggered.

Issue #1658, opened 2022-12-07
fix list, info and update commands


gbochora:
Add multiple runs compatibility for commands:
- /playbook list
- /playbook info
- /playbook update

No playbooks code is executed for these commands, they are handled on the mm server side.

PR #1696, opened 2022-12-21
Fix update command

gbochora:
## Summary
This PR: 
- fixes the `\playbook update` command; now, it supports multiple runs in the channel.
- removes the `\playbook list` command because the start-a-run modal now covers that goal.

https://user-images.githubusercontent.com/4368372/208988099-8fd4da55-ccda-411b-b8c8-0366c4c667dc.mp4

## Ticket Link
#1658

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


server/api/playbook_runs.go
```diff
@@ -1156,6 +1158,30 @@ func (h *PlaybookRunHandler) getChannelRunsAutocomplete(c *Context, w http.Respo
...
+	vars := mux.Vars(r)
+	channelID := vars["channel_id"]
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	playbookRuns, err := h.playbookRunService.GetPlaybookRunsForChannelByUser(channelID, userID)
+	if err != nil {
+		h.HandleErrorWithCode(w, c.logger, http.StatusInternalServerError,
+			fmt.Sprintf("unable to retrieve runs for channel id %s", channelID), err)
+	}
+	if len(playbookRuns) == 0 {
```
m-ripper, line 1171: Does it make sense to return an empty array instead of an error?

m-ripper (approved):
Nice Work! 🎉 
The comment I left is only something to think about but does not necessarily need a change.

trilopin (approved):

PR #1697, opened 2022-12-22
MM-49060: status update post for non-participants

lieut-data:
## Summary
Non-participants won't necessarily have the playbook run loaded into Redux, so don't skip rendering if we can't find it: instead just render the original post message as created by the server. The additional interaction is only relevant for participants anyway.

Before:
CleanShot 2022-12-22 at 14 41 36@2x

After:
CleanShot 2022-12-22 at 14 41 51@2x

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-49060


iomodo (approved):
Nice coverage!

gbochora (approved):
Nice!

Issue #1693, opened 2022-12-20
Runs list CTA to run playbook shown even if disabled in permissions


m-ripper:
When the user has no runs but at least one playbook, the runs overview page shows a button to start a new playbook run, even if it is disabled in the permissions (no matter if team or global).

When the button is clicked, the user is redirected to Channels and a modal is opened. 
Instead, the user should not be redirected and the `new_run_playbook_modal.tsx` should be shown. This modal takes permissions into account. 

![Screenshot from 2022-12-20 15-09-39](https://user-images.githubusercontent.com/8669935/208686300-ddc119d0-cf46-4d62-b3c9-c391a9ba4a76.png)

[Screencast from 20.12.2022 15:44:37.webm](https://user-images.githubusercontent.com/8669935/208695056-964ce21e-8177-4fb2-87be-8d39595b611a.webm)

lieut-data:
I'm going to remove this from the v7.7 milestone simply because it requires advanced permissions to disable running playbooks + no existing runs, which is relatively rare.

PR #1698, opened 2022-12-22
Use new playbook run modal when no runs

lieut-data:
## Summary
The new playbook run modal supports checking permissions to handle the case when a playbook cannot be run (including all playbooks). But tweak the styling when a playbook is disabled, otherwise it blocks the action but doesn't explain why.

CleanShot 2022-12-22 at 16 39 25

## Ticket Link
Fixes: https://github.com/mattermost/mattermost-plugin-playbooks/issues/1693


lieut-data:
Oops, I just realized this overlaps significantly with https://github.com/mattermost/mattermost-plugin-playbooks/pull/1692/files. I'll rebase on that one.

lieut-data:
Fixed!

webapp/src/components/rhs/rhs_welcome_view.tsx
```diff
@@ -1,117 +0,0 @@
-// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
-// See LICENSE.txt for license information.
-
-import React, {useEffect, useState} from 'react';
-import {FormattedMessage} from 'react-intl';
```
trilopin, line 5: nice removal

webapp/src/components/backstage/runs_page_no_content.tsx
```diff
@@ -87,11 +66,12 @@ const NoContentPlaybookRunSvgContainer = styled.div`
 
 const NoContentPage = () => {
     const dispatch = useDispatch();
-    const teamId = useSelector(getCurrentTeamId);
+    const teamId = useSelector<GlobalState, string>(getCurrentTeamId);
```
trilopin, line 69: aren't those types already inferred?

webapp/src/components/backstage/runs_page_no_content.tsx
```diff
@@ -103,30 +83,34 @@ const NoContentPage = () => {
...
     const handleClick = () => {
         if (playbookExist) {
-            goToMattermost();
-            dispatch(startPlaybookRun(teamId));
+            dispatch(openPlaybookRunNewModal({
+                teamId,
+                onRunCreated: (runId: string, channelId: string, statsData: object) => {
+                    navigateToPluginUrl(`/runs/${runId}?from=run_modal`);
+                    refreshLHS();
+                    telemetryEvent(PlaybookRunEventTarget.Create, {...statsData, place: 'backstage_runs_no_content'});
```
trilopin, line 93: it would be great to have an e2e just with one case (not all permutations we already have in start-run-xxx.spec) checking that run is created, the navigation is done to RDP and the telemetry sets the right place.

trilopin (approved):
Thanks for your help finishing these details 🎉

m-ripper (approved):
Well done! I have nothing to comment besides what @trilopin already commented on.

webapp/src/components/backstage/runs_page_no_content.tsx
lieut-data: Hah, you're seeing through my TypeScript weaknesses. I actually only added them because I saw them somewhere else.

webapp/src/components/backstage/runs_page_no_content.tsx
lieut-data: Absolutely, yes! I was hoping to get to that before the PR, but would like to return to this.

PR #1703, opened 2022-12-30
Safer json marshal

gbochora:
## Summary
Changed `MarshalJSON` receiver type from pointer to value. Pointer receiver is a little risky because if you call `json.Marshal` and pass a value type, the custom marshal function will not be called. I did it a few weeks ago and spent hours trying to find out what the issue was.

Here is an example by @trilopin: https://go.dev/play/p/wPlv8mKp0jX if you run it as is, it works for both pointer and value types. However, if you change the receiver type to a pointer(`(a *A)`), it will work only when passing a pointer. 

## Ticket Link
NONE

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


crspeller (approved):

iomodo (approved):
nice

PR #1707, opened 2023-01-04
MM-48600 Fix run actions modal not updating state when changing run

m-ripper:
## Summary

The run actions modal did not show the correct data when changing runs.

This was caused because `useState` does not use the provided initial state if the component has mounted with a state already which is the case here, when the first run is visited. 

To avoid this issue, all states will be synced when the run changes.

[Screencast from 04.01.2023 15:49:45.webm](https://user-images.githubusercontent.com/8669935/210581554-ca316de4-49b3-4e6e-bc15-c6f7e697b954.webm)

## Ticket Link

[MM-48600](https://mattermost.atlassian.net/browse/MM-48600)

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

[MM-48600]: https://mattermost.atlassian.net/browse/MM-48600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webapp/src/components/run_actions_modal.tsx
+++ b/webapp/src/components/run_actions_modal.tsx
@@ -1,7 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useEffect, useState} from 'react';
+import React, {useState} from 'react';
+import {useUpdateEffect} from 'react-use';
 import {useDispatch, useSelector} from 'react-redux';
 import {useIntl} from 'react-intl';
 import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


calebroseland (changes requested):
Batching everything into one effect here would yield some interesting side effects.


CleanShot 2023-01-04 at 13 50 50

If anything changes on the run while someone is modifying Run Actions (let's say someone else joins the run while they're entering a complex webhook)—any pending changes would be reset without any warning.

To avoid that it would be better to keep them tied to only their individual dependencies like here: 
https://github.com/mattermost/mattermost-plugin-playbooks/blob/69714edbdc02fc23a2a9db280ecf877b9ea021fc/webapp/src/components/checklist_item/checklist_item.tsx#L102-L124

(see also: https://github.com/mattermost/mattermost-plugin-playbooks/pull/1556#pullrequestreview-1174625602 re: `useProxyState`)

gbochora (approved):
Nice!

PR #1708, opened 2023-01-04
MM-48599 Improve toast for participating in a run and toasts in general

m-ripper:
## Summary

Increased the duration the toast for participating in a run is shown from 3 to 4.5 seconds. 

Additionally, toasts will not be hidden while hovered. 

[Screencast from 04.01.2023 17:54:43.webm](https://user-images.githubusercontent.com/8669935/210608030-39b7800e-5c77-49de-91b0-652c250313be.webm)

## Ticket Link

[MM-48599](https://mattermost.atlassian.net/browse/MM-48599)

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

[MM-48599]: https://mattermost.atlassian.net/browse/MM-48599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webapp/src/components/backstage/toast.tsx
+++ b/webapp/src/components/backstage/toast.tsx
@@ -14,6 +14,9 @@ export interface ToastProps {
     buttonName?: string;
     buttonCallback?: () => void;
     closeCallback?: () => void;
+    interactive?: boolean;
+    onMouseEnter?: () => void;
+    onMouseLeave?: () => void;
 }
 
 export const Toast = (props: ToastProps) => {
@@ -30,9 +33,12 @@ export const Toast = (props: ToastProps) => {
             iconName = 'information-outline';
         }
     }
+    const interactive = props.interactive ?? Boolean(props.buttonCallback);
     return (
         <StyledToast
             toastStyle={props.toastStyle ?? ToastStyle.Success}
+            onMouseEnter={interactive ? props.onMouseEnter : undefined}
+            onMouseLeave={interactive ? props.onMouseLeave : undefined}
         >
             <StyledIcon className={`icon icon-${iconName}`}/>
             <StyledText>{props.content}</StyledText>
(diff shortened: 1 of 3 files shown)


webapp/src/components/backstage/toast_banner.tsx
```diff
@@ -8,6 +8,7 @@ const Ctx = React.createContext({} as ToastFuncs);
 
 const DEFAULT_DURATION = 3000;
 let toastCount = 0;
+const toastTimeoutMap :Record<number, number> = {};
```
calebroseland, line 11:
```suggestion
const toastTimeoutMap: Record<number, number> = {};
```

calebroseland (approved):
LGTM! Just a super tiny whitespace nit

webapp/src/components/backstage/toast.tsx
```diff
@@ -14,6 +14,9 @@ export interface ToastProps {
     buttonName?: string;
     buttonCallback?: () => void;
     closeCallback?: () => void;
+    interactive?: boolean;
```
crspeller, line 17: Why the interactive flag? Seems fine to hold the toast open for non-interactive toasts if the mouse is on it. We can remove the `?` in this case. We also don't seem to set the flag anywhere?

crspeller (changes requested):
In addition I am wondering what happens if you spam the join button a bunch of times. Does this cause any strangeness? Was having an issue with that.

abhijit-singh (approved):
Looks good to me, thanks @m-ripper!

m-ripper:
> In addition I am wondering what happens if you spam the join button a bunch of times. Does this cause any strangeness? Was having an issue with that.

@crspeller I figured out the strangeness was caused because `lodash.debounce` was used which resets the timer with every call, therefore a second toast could only be added if the user waited 2 seconds before triggering the toast again.
I've changed this to use `lodash.throttle` instead which makes sure that the user can trigger the toast every 4.5 seconds which means that there should only be a single toast at all times if the user doesn't hover over them.

crspeller (approved):

PR #1712, opened 2023-01-09
MM-48582 Playbook import via drag and drop

m-ripper:
## Summary

Playbooks can be imported by dropping a file onto the import buttons.

Additionally, the LHS is now refreshed when a playbook is imported.

[Screencast from 12.01.2023 16:54:38.webm](https://user-images.githubusercontent.com/8669935/212115666-a5733ea7-eb80-493c-815a-ba9f1cb82f18.webm)

## Ticket Link

[MM-48582](https://mattermost.atlassian.net/browse/MM-48582)
[MM-48591](https://mattermost.atlassian.net/browse/MM-48591)

## Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

[MM-48582]: https://mattermost.atlassian.net/browse/MM-48582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests-e2e/cypress/integration/playbooks/list_spec.js
```diff
@@ -107,6 +107,67 @@ describe('playbooks > list', () => {
...
+            // # Open the product
+            cy.visit('/playbooks');
+
+            // # Switch to Playbooks
+            cy.findByTestId('playbooksLHSButton').click();
+
+            cy.findByTestId('titlePlaybook').within(() => {
+                // # Drop loaded fixture onto import button
+                cy.findByText('Import').selectFile('@playbookExport', {
+                    action: 'drag-drop'
```
trilopin, line 143: TIL 🎉

webapp/src/components/backstage/import_playbook.tsx
```diff
@@ -8,37 +8,56 @@ import {importFile} from 'src/client';
...
+        importFile(data, teamId)
+            .then(({id}) => cb(id))
+            .catch(errorHandler);
+    };
+
+    const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        if (!e.target.files?.[0]) {
+            return;
         }
+        readFile(e.target.files[0])
```
trilopin, line 40: would make sense to limit the file size before reading the file and handle it as a user-readable error?

webapp/src/components/backstage/import_playbook.tsx
```diff
@@ -8,37 +8,56 @@ import {importFile} from 'src/client';
...
+    });
+
+    const uploadFile = (data: FileData) => {
+        importFile(data, teamId)
+            .then(({id}) => cb(id))
+            .catch(errorHandler);
+    };
+
+    const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        if (!e.target.files?.[0]) {
```
trilopin, line 37: what happens if the user drop more than one file?

trilopin: I tested, the first is imported, rest are ignored. Could we fail with an explicit error when more than 1 file is uploaded?

trilopin (changes requested):
Good job Max,

How hard would be to do the style change to visually tell the user that the import button is the drop area when drag start?

Please check the tiny details I wrote.

webapp/src/components/backstage/import_playbook.tsx
m-ripper: That's probably a good idea. What file size were you thinking about if you had any in mind so far?

m-ripper:
I've added a visual indicator for the dropzone and updated the video above. I'm 2/5 in regards of the UX that the dropzone gets shown when the user drags a file to any part of the page. In Channels, it is only shown when dragged into the dropzone.

webapp/src/components/assets/files_overlay_image.tsx
```diff
@@ -0,0 +1,11 @@
...
...ECMQAAwBVYgXm+psR1CmQorE337hq3PXDXK5Mm7rFsx33l+UEYxUmaZntpmsRRGPi5+4ZjW9v/X6+v67XPLeh2QPd3Bto9MLHsJ0ZRVlEUwTjNED5CWRrDfVBVFob9zJqAXvvfDMy+rCNeD+h+2h8MDcB2cmdFKOMQwnhf+ggLqZTSSkmJ0b4W4n0EZ5SsJve2YNBq/PsBYrEr3gog4waj8EQTIkiAy6+Q2DgxKAww0NCAwVxRc7UqFFSLAKXQATgAN5ABAzAUtllhLQAUANKkbXNpu+7u2raFvT95N+bt/E8t9oGqC9/3v+97iS8St6Y4iBZ0aqzDCLorq2sTHJ+cnh0ewrnDwQP3YlBG7QpROSiyoxRGAbOAH5yeHK/9xupKlz2A54kiAqSgU+B10Fle6kqcz1ycScOkwDyX1KO27/bD0AGEYd/121E0GBQwjjSVpjq7mDnvVlha7gTqT2hoJUdBCzoV1iEGQdC5/LF5lR4dZZK5PPXEBepOr2Xatqhg22av54AMN6mmUeQ52Cq92vx52ekEAe/8sZIjS1RV0CkkzH4fjsZZVsChl/LQ45q5EIal67qGgD8tA4SYLdARu76UATogIaOdrVdwEDy8+Hslxy7oCD583L7OkXsiuXstUwjL0jX66aP8j6ZblhBmy5uoKOUwbhZuN1SfIU2v5JgFHcXct/m7bELej/uSvC0MSp6KMIQtRfRjP8FR5Nfbi4wveb64ygjgI7WioHu6u3MP5Os1I7Pq9ZC8RslTERqK6HmYbhCRF9nD/L5i/b1RPEewCrqXB8OFG5nWQVR7HsibwkDfPGEB3YQi4smKKvIxRII8yJBXewT8gu7tItv0FOpISBEkEuTVHgGzoJvbl6YH9mj6kJheAWUk3KTK9dRIfMVKTh0CUtDNbu2MxtWWR9+g6S216fkiMBL1JPIcI6Gu5JQF3catND3eXXHRM03/6EiUZRWJe4wEebWn8pDj7/5r0z8+EsODOhLvFM8RpKB7tj//QJcl0/QNRGIPX+3xPPT502hc5H8sS5vtmwYiUdztvWc5CBep58oLSzQ93zfNRSJqx57gCtAMM3T9RPGE0CBoJHw/dky4DTMvt4QJ1zls0zcvwnE8U1jsW70mB9i86fmRELYko3GJ1NKbNz0/EjqSYV+E+N/k1XR+AWSp9BXRFCxOAAAAAElFTkSuQmCC';
```
m-ripper, line 3: I don't like this solution, but didn't find a better one at the moment. I want to use [this image](https://github.com/mattermost/mattermost-webapp/blob/master/images/filesOverlay.png) that is already included in the webapp but due to hashes I cannot use the already existing one. 

The Plugin API doesn't seem to support adding static files from a plugin, that's why I am using a base64 version of the the image, but maybe I am missing something. If there's a SVG version of the image, I will use it instead.

calebroseland: I pushed a small fix to the webpack config that should support PNGs now—same outcome as what you have, just avoids having to do it manually. I think they have to be inlined because I think the current mattermost-server plugin build/flow that playbooks is in doesn't support URL resources (or other webpack chunks).

webapp/src/components/assets/files_overlay_image.tsx
```diff
@@ -0,0 +1,14 @@
...
```
calebroseland, line 6: e.g. 
```suggestion
import source from './files_overlay.png';
```

webapp/src/components/backstage/import_playbook.tsx
```diff
@@ -8,37 +8,76 @@ import {importFile} from 'src/client';
...
+
+    const readFile = (file: File) => new Promise<FileData>((resolve, reject) => {
+        if (file.size > fileSizeLimit) {
+            addToast({
+                content: formatMessage({defaultMessage: 'The file size exceeds the limit of 5MB.'}),
+                toastStyle: ToastStyle.Failure,
+            });
+            reject(new Error('File size limit exceeded'));
+            return;
+        }
```
calebroseland, line 34:
```suggestion
        }
        if (file.type !== 'application/json') {
            addToast({
                content: formatMessage({defaultMessage: 'The file must be a valid JSON playbook template.'}),
                toastStyle: ToastStyle.Failure,
            });
            reject(new Error('File must be a JSON file'));
            return;
        }
```

calebroseland (approved):
Awesome work @m-ripper! Thanks for the expanded indicators.

One thought on a helpful error message if you try to upload an unsupported file type, but not essential so feel free to proceed without.

abhijit-singh (approved):
Love it @m-ripper!

trilopin (approved):
Good job @m-ripper 

Non blocking suggestion regarding messages accuracy

PR #1715, opened 2023-01-10
useMemo with debounce

lieut-data:
## Summary
`debounce` requires memoization, otherwise every component re-render recreates the debouncing function. We change some existing, memoized instances of `debounce` using `useCallback` to use `useMemo` primarily to address (es)linting issues.

While I'm in here, fix an instance of `should('not.be.visible')` to be `should('not.exist')` that failed a recent CI run -- see https://docs.cypress.io/guides/references/migration-guide#Non-existent-element-assertions.

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-48509

### DIFF
--- a/webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
+++ b/webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
@@ -1,4 +1,4 @@
-import React, {useRef, useState} from 'react';
+import React, {useMemo, useRef, useState} from 'react';
 import styled from 'styled-components';
 import {useIntl} from 'react-intl';
 import debounce from 'debounce';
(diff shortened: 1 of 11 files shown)


webapp/src/hooks/crud.ts
```diff
@@ -155,11 +160,14 @@ export function usePlaybooksCrud(
...
+    const setSearchTerm = useCallback((term: string) => {
         setLoading(true);
         setParams({search_term: term, page: 0});
-    };
-    const setSearchTermDebounced = debounce(setSearchTerm, searchDebounceDelayMilliseconds);
+    }, [setLoading, setParams]);
+    const setSearchTermDebounced = useMemo(
+        () => debounce(setSearchTerm, searchDebounceDelayMilliseconds),
+        [setSearchTerm],
+    );
```
lieut-data, line 170: I probably could have inlined `setSearchTerm`, but opted to keep the diff smaller this time.

webapp/src/components/datetime_input.tsx
```diff
@@ -101,7 +101,7 @@ const DateTimeInput = ({
         const datetimes = parseDateTimes(locale, query)?.map(({start}) => DateTime.fromJSDate(start.date()));
         const duration = parse(locale, query, Mode.DurationValue);
         setOptions(makeOptions(query, datetimes, duration ? [duration] : [], mode) || null);
-    }, 150), [locale, setOptions, makeOptions]);
+    }, 150), [locale, setOptions, makeOptions, mode]);
```
lieut-data, line 104: A few missing dependencies reported by the linter.

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective.tsx
```diff
@@ -113,13 +126,6 @@ const Retrospective = ({
         );
     };
 
-    const onMetricsChange = debounce((metrics_data: RunMetricData[]) => {
-        updateRetrospective(playbookRun.id, playbookRun.retrospective, metrics_data);
-    }, DEBOUNCE_2_SECS);
-    const onReportChange = debounce((retrospective: string) => {
-        updateRetrospective(playbookRun.id, retrospective, playbookRun.metrics_data);
-    }, DEBOUNCE_2_SECS);
-
```
lieut-data, line 122: These are pulled up higher to avoid conditional hook declaration.

m-ripper (approved):
Nice work!

calebroseland (approved):
Sweet, now we have a definitive pattern!

PR #1716, opened 2023-01-11
Live reload for server code

trilopin:
## Summary
Proposal with live reload at server side.

Based on [modd](https://github.com/cortesi/modd) tool, the same that focalboard uses.

The tool is automatically installed and started when you do `make watch`:
- the first deploy is complete, not triggered by modd
- webapp part is still handled by webpack (run in daemon section at modd.conf)
- go files (excluding tests) are watched and, when they change, the server is compiled and the plugin is uploaded again

Some extra experiments are done with graphql files, to trigger `make graphql` but I'm still 1/5 about if that's needed.

Tested on:
- [X] MacOS 
- [ ] Windows
- [x] Linux

~~Still getting sometimes `Unable to move plugin from temporary directory to final destination. Another plugin may be using the same directory name., destination already exists`, not sure why just fail sometimes.~~

## Ticket Link
Nope

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


lieut-data:
Throwing @cpoile on this in the hopes that it might help with Calls development too :)

trilopin:
kind reminder to get a review @cpoile @crspeller

cpoile (approved):
Nice @trilopin!

crspeller (approved):

PR #1732, opened 2023-01-13
Clean unused code

trilopin:
## Summary

**key points**
- Removed promptForFeedback in favor of the current google form. It was triggered by surveybot before. I've removed it from client.ts, bot API and bot service.
- ~~RemoveChecklist and RemoveItem are not available in the UI in the context of a run. I've removed the API and the client.ts funcs. For playbook, it's allowed but handled through graphql Update mnutation.~~
- One Makefile command to install all go tools (gotestsum, golangci-lint and mockgen)
- Old favorite endpoints that weren't used, have been removed (addFavorite deleteFavorite) in favor of the currently used Graphql mutations.
- Run_modal has been renamed to remove the new prefix (added when old and new existed)
- RHSState redux & types has been removed  (it's currently handled with local state, not redux)
- A bunch of unused functions removed from client.ts
- A bunch of unused svg assets removed
- Removed unused hooks: useProfilesInCurrentChannel, useCanCreatePlaybooks, useProfilesInChannel and related tests
- Unused backstage/playbook_edit/automation/categorize_playbook_run.tsx has been removed
- Unused backstage/playbook_preview_cards.tsx has been removed
- Unused components/multi_select.tsx has been removed
- Some unused styled removed from add_participant_modal and become_participant_modal
- RHSHomePlaybook at rhs/rhs_home_item has been removed
- Unused rhs/rhs_timeline_placeholder has been removed
- Old stats fetchStats and related types removed

**ts-prune report after these changes** 
```
> ts-prune

src/hooks/index.ts:61 - useKeyPress
src/hooks/index.ts:183 - clearLocks
src/hooks/index.ts:225 - useCanRestrictPlaybookCreation
src/hooks/index.ts:243 - useExperimentalFeaturesEnabled
src/hooks/index.ts:257 - useThing
src/hooks/index.ts:336 - useDropdownPosition
src/hooks/index.ts:419 - useFormattedUsername
src/hooks/index.ts:53 - FetchMetadata
src/hooks/index.ts:560 - usePrevious
src/hooks/index.ts:28 - getPlaybookOrFetch
src/hooks/index.ts:204 - useCategories
src/hooks/index.ts:22 - PlaybookCreateQueryParameters
src/hooks/index.ts:24 - useHasChannelPermission
src/hooks/index.ts:28 - useHasPlaybookPermissionById
src/hooks/index.ts:14 - usePlaybookViewTelemetry
src/hooks/index.ts:22 - usePlaybookRunViewTelemetry
src/types/permissions.ts:6 - Permission
src/components/rhs/rhs_run_details_tour_dialog.tsx:146 - default
src/components/tutorial/tours.ts:9 - AutoTourStatus
src/components/tutorial/tours.ts:55 - TTCategoriesMapToAutoTourStatusKey
src/components/tutorial/tutorial_tour_tip/index.ts:92 - useElementAvailable
```

Tours are kept since we plan to do one for checklists.

**Open questions**
- some make commands are held in custom.mk instead of Makefile. we don't clearly separate what we should place where. I'd suggest separating them clearly or merging them.
- I'd remove unused hooks. We can annotate with ts-prune-ignore the ones that we can keep even unused. any thoughts?

Feel free to raise your hand to keep some removed items.

## Ticket Link
No

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


server/api/categories.go
```diff
@@ -35,8 +35,6 @@ func NewCategoryHandler(router *mux.Router, api *pluginapi.Client, categoryServi
 	categoriesRouter := router.PathPrefix("/my_categories").Subrouter()
 	categoriesRouter.HandleFunc("", withContext(handler.getMyCategories)).Methods(http.MethodGet)
 	categoriesRouter.HandleFunc("", withContext(handler.createMyCategory)).Methods(http.MethodPost)
-	categoriesRouter.HandleFunc("/favorites", withContext(handler.addFavorite)).Methods(http.MethodPost)
-	categoriesRouter.HandleFunc("/favorites", withContext(handler.deleteFavorite)).Methods(http.MethodDelete)
```
trilopin, line 39: these are done through `setRunFavorite`in the case of runs and `updatePlaybook` in the case of playbooks

webapp/src/components/backstage/playbook_editor/outline/section_retrospective.tsx
```diff
@@ -120,4 +115,21 @@ const RetrospectiveTextContainer = styled.div`
     padding: 0 8px;
 `;
 
+const Card = styled.div`
```
trilopin, line 118: moved here from a removed file `src/components/backstage/playbook_preview_cards'` where the only used thing was this styled

Makefile
```diff
@@ -75,17 +85,6 @@ endif
 	cd tests-e2e && npm run fix
 
 
-## Check & install golangci version
-# Pinned version to commit until 10.2 is released since some macos version have problems
-# with the native diff tools used by typecheck (gometalinter problem)
-.PHONY: check-golangci
-check-golangci:
```
lieut-data, line 82: nit: Is this a target that's "familiar" to other repositories like the server or other plugins? Wondering if it makes sense to keep the interface consistent, even if only as a legacy entry that simply calls the new one.

webapp/src/client.ts
```diff
@@ -180,6 +173,7 @@ export async function fetchCheckAndSendMessageOnJoin(channelId: string) {
     return Boolean(data.viewed);
 }
 
+// ts-prune-ignore-next It's very likely to need this one to avoid fetching each channelswitch the channel actions
```
lieut-data, line 176: nit: comment doesn't parse for me?

lieut-data (approved):
💙

webapp/src/client.ts
trilopin: What do you mean by `doesn't parse`? is it being reported as unused? is your IDE complaining about sth?

crspeller (approved):
Love it! :D

Makefile
trilopin: I have no problem keeping that but I can not find those repositories.

mattermost-server is installing it direcly on the [command](https://github.com/mattermost/mattermost-server/blob/master/Makefile#L338) for running golangci

focalboard just check but don't install through [makefile](https://github.com/mattermost/focalboard/blob/main/Makefile#L112)

For mattermost-plugin-template it is embedded inside check-style (not sure how important since we are moving from plugin to product)

crspeller:
>  some make commands are held in custom.mk instead of Makefile. we don't clearly separate what we should place where. I'd suggest separating them clearly or merging them.

That stuff I think is the shared stuff with other plugins. We can fix it when we go to MPA or monorepo. 

> I'd remove unused hooks. We can annotate with ts-prune-ignore the ones that we can keep even unused. any thoughts?

I am totally fine with removing anything unused. It's always in the history.

webapp/src/components/datetime_input.tsx
```diff
@@ -181,5 +181,4 @@ export const useDateTimeInput = ({defaultValue, ...props}: Partial<Exclude<Props
     return {input, value};
 };
 
-export default DateTimeInput;
```
calebroseland, line 184: 1/5 We should probably keep this default export (even though current usage is through `useDateTimeInput`)

calebroseland (approved):
Sweet sweet red lines

webapp/src/client.ts
trilopin: ok, comment makes no sense if we move forward with the current overfetch solution.

Will change it

trilopin: oh :)

That ts-prune-ignore marks a function that fetch all channels with runs which I considered as a possible different way of solving the overfetch problem. Then, I solved through the extra data in the WS event and realize that channel actions are valid for channels without runs.

So, everything wrong 😂. Regardless of the final solution for overfetch, I'll remove the comment in this pr and re-evaluate the removal in a follow-up PR.

Makefile
trilopin: I kept the change in the merge, but still open to recover if you think it's important

PR #1734, opened 2023-01-13
MM-48590 Add indicator for adding participants when at least one member alrady exists

m-ripper:
## Summary

At the moment, it is not obvious how to add participants to a run in the RHS if there are members already. 
This PR adds a link for that case, similar to the link when there are no participants.

[Screencast from 13.01.2023 16:47:20.webm](https://user-images.githubusercontent.com/8669935/212362464-c0091bb9-f3da-4c2d-b6df-cea8974deb1c.webm)

## Ticket Link

[MM-48590](https://mattermost.atlassian.net/browse/MM-48590)

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/rhs/rhs_participants.tsx
+++ b/webapp/src/components/rhs/rhs_participants.tsx
@@ -26,6 +26,16 @@ const RHSParticipants = (props: Props) => {
         props.setShowParticipants(true);
     };
 
+    const addParticipant = (
+        <LinkAddParticipants
+            to={'#'}
+            onClick={showParticipants}
+        >
+            {formatMessage({defaultMessage: 'Add participant'})}
+            <OpenInNewIcon size={11}/>
+        </LinkAddParticipants>
+    );
+
     const becomeParticipant = (
         <Tooltip
             id={'rhs-participate'}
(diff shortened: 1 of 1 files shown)


lieut-data (approved):
I think this makes sense. How does it render when there are lots of participants? cc @abhijit-singh for design review as well.

crspeller (approved):

abhijit-singh (changes requested):
This is great @m-ripper! A couple of points:
1. Would it be possible to have an icon-button instead of the `Add participant` text in this case? (`icon-account-multiple-plus-outline` would be ideal)
2. Non-blocking but can we trigger the add participant modal directly from here instead of first taking the user to the participants list?

Attaching a screencap showing the proposed behaviour and here's the [figma link](https://www.figma.com/proto/rU4vGGtxk3lBIghH4o0yj6/Checklists?page-id=112%3A113286&node-id=601%3A225994&viewport=-4599%2C-8111%2C0.74&scaling=min-zoom&starting-point-node-id=601%3A225994).

https://user-images.githubusercontent.com/14369607/213241235-02194648-09a4-42c4-bc71-7870d80b452c.mov

m-ripper:
Here's the latest style:
[Screencast from 18.01.2023 19:07:02.webm](https://user-images.githubusercontent.com/8669935/213260247-2b8d6eb8-237f-45cc-87cb-977018224085.webm)

Sadly it is not that easy at the moment to show the modal for searching users immediately, but this is good for the future.

PR #1735, opened 2023-01-13
MM-48089 Fix URLs for email notifications

m-ripper:
## Summary
The URLs in the digest messages were relative which meant that they would work when opened from the app but not if opened from an email client for example.

Now, these URLs are absolute which fixes the issue.

## Ticket Link

[MM-48089](https://mattermost.atlassian.net/browse/MM-48089)

## Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

[MM-48089]: https://mattermost.atlassian.net/browse/MM-48089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -128,15 +128,15 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRuns(requesterInfo RequesterInfo, op
 	return s.store.GetPlaybookRuns(requesterInfo, options)
 }
 
-func (s *PlaybookRunServiceImpl) buildPlaybookRunCreationMessageTemplate(playbookTitle, playbookID string, playbookRun *PlaybookRun, reporter *model.User) (string, error) {
+func (s *PlaybookRunServiceImpl) buildPlaybookRunCreationMessageTemplate(playbookTitle, playbookID, siteURL string, playbookRun *PlaybookRun, reporter *model.User) (string, error) {
 	return fmt.Sprintf(
 		"##### [%s](%s%s)\n@%s ran the [%s](%s) playbook.",
 		playbookRun.Name,
-		GetRunDetailsRelativeURL(playbookRun.ID),
+		getRunDetailsURL(siteURL, playbookRun.ID),
 		"%s", // for the telemetry data injection
 		reporter.Username,
 		playbookTitle,
-		GetPlaybookDetailsRelativeURL(playbookID),
+		getPlaybookDetailsURL(siteURL, playbookID),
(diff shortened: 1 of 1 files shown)


server/app/playbook_run_service.go
```diff
@@ -2217,6 +2217,12 @@ func (s *PlaybookRunServiceImpl) buildTodoDigestMessage(userID string, force boo
 		return nil, err
 	}
 
+	siteURL := s.pluginAPI.Configuration.GetConfig().ServiceSettings.SiteURL
+	if siteURL == nil {
+		logrus.Error("cannot send webhook on creation, please set siteURL")
```
trilopin, line 2222: could this message be copied from another place? I don't see the relation with webhooks here

m-ripper: Good catch, thanks! 😅

trilopin (changes requested):
Thanks @m-ripper 

There are more GetRunDetailsRelativeURL related to digest messages which could be exposed to the same problem. I would check any message that the playbooksbot is posting..

m-ripper:
> Aha, that explains the e-mail oddity. But I wonder, is this a general class of problem where if I use a relative URL in a post that triggers an e-mail, it too will be broken?
> 
> One benefit of the relative URL at post time is that it allows the SiteURL to change. (E-mails can't do that, of course, but also I find they don't get referenced "historically" like posts do.)
> 
> I wonder if we might consider doing something in the e-mail template on the server instead? (Blocking review to discuss.)

I like the idea and it would probably make it a lot cleaner and easier! I'll take a look at this approach. 👍🏻

m-ripper:
> > Aha, that explains the e-mail oddity. But I wonder, is this a general class of problem where if I use a relative URL in a post that triggers an e-mail, it too will be broken?
> > One benefit of the relative URL at post time is that it allows the SiteURL to change. (E-mails can't do that, of course, but also I find they don't get referenced "historically" like posts do.)
> > I wonder if we might consider doing something in the e-mail template on the server instead? (Blocking review to discuss.)
> 
> I like the idea and it would probably make it a lot cleaner and easier! I'll take a look at this approach. 👍🏻

I've found a better solution in the server, thank you for the hint!
I am closing this PR in favor of [#22101](https://github.com/mattermost/mattermost-server/pull/22101) in the server repository.

PR #1736, opened 2023-01-13
connect apollo dev tools

trilopin:
## Summary
Oneliner to connect dev tools.

Not sure yet if this should be only done in development (instrospection is already disabled server side for production)

It needs EnableTesting and EnableDeveloper as true in the mattermost-server.

## Ticket Link
Nope

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


crspeller (approved):
Nice!

PR #1744, opened 2023-01-17
MM-48580: avoid error loading without teams

lieut-data:
## Summary
If a user doesn't have any teams, redirect them to the `/select_team` handler.

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-48580

### DIFF
--- a/webapp/src/components/backstage/main_body.tsx
+++ b/webapp/src/components/backstage/main_body.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import {
     Redirect,
     Route,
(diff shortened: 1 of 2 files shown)


calebroseland (approved):
Nice and straightforward 🙂

webapp/src/components/backstage/main_body.tsx
```diff
@@ -36,9 +36,17 @@ const useInitTeamRoutingLogic = () => {
     const currentTeamId = useSelector(getCurrentTeamId);
     const currentUserId = useSelector(getCurrentUserId);
 
+    const history = useHistory();
+
+    useEffect(() => {
+        if (currentUserId?.length > 0 && teams.length === 0) {
```
crspeller, line 42: Is it possible for us to reach this point before teams are loaded?  That would cause undesirable redirection here.

lieut-data: Indeed, I was worried about that too, and figured the `currentUserId` check would at least guarantee we aren't doing this before "most" of the loading happens -- CI tests helped build a bit of confidence that this worked (as it definitely didn't before). I'd be happy to use something even more reliable though if we know of a "bit" to follow.

Thoughts, @crspeller?

crspeller: I think the only way to make sure this is reliable is to handle it webapp side. We don't have any control over the fetching of teams.

lieut-data: Indeed. Given it's isolated to rendering a Playbooks component, I'm inclined to think this won't make the problem worse where we fail altogether today. Is that reasonable?

crspeller: It could make it worse if we falsely redirect users when we have not fetched teams yet. I am just not sure if that is possible.

lieut-data:
I'm going to close this out without merging for now, as it's a low-frequency edge case, and I'm not yet sure how to achieve this in a bulletproof way.

Issue #1731, opened 2023-01-12
channel actions fetched twice on channel switch


trilopin:
when we switch channels we are fetching some actions-related data:
- GET /plugin/playbooks/api/v0/actions/channels/<channel_id>
- GET /plugin/playbooks/api/v0/actions/channels/<channel_id>?trigger_type=new_member_joins

Eventually, we'll stop fetching always, but we should check if both GET calls are needed in the meantime.

PR #1748, opened 2023-01-19
fix overfetch channel actions

trilopin:
## Summary
Two different actions on the channelactions management to reduce the number of calls on every channel switch. There are two kind of fetch:

**/plugins/playbooks/api/v0/actions/channels/<Channelid>**

Make channel actions modal fetch action ONLY when is shown, not when it's rendered but hidden. Before this change, channel actions were fetched on every channel switch.

**/plugins/playbooks/api/v0/actions/channels/<Channelid>?trigger_type=new_member_joins**

Several actions done here:
 - kept hasviewedchannel redux reducer/actions/seelctors to guarantee backwards compatibility with old servers
 - added prev_viewed_at to websocket data at mattermost-server
 - use prev_viewed_at to trigger just the first time

Before 
https://user-images.githubusercontent.com/4096774/213535360-fda5437a-67b0-416d-bbfc-9f216c37fd81.mp4

After
https://user-images.githubusercontent.com/4096774/213535329-e6caa3d5-71d9-42dd-a531-17601581a656.mp4

This PR relies on[ some small change](https://github.com/mattermost/mattermost-server/pull/22111) on mattermost-server. If the mattermost-server don't send that new field, we will work as we previously did (just prevent fetch after viewing channels once per session)

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1731

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/websocket_events.ts
+++ b/webapp/src/websocket_events.ts
@@ -203,11 +203,11 @@ export const handleWebsocketChannelUpdated = (getState: GetStateFunc, dispatch: 
 export const handleWebsocketChannelViewed = (getState: GetStateFunc, dispatch: Dispatch) => {
     return async (msg: WebSocketMessage<{ channel_id: string, prev_viewed_at?: number}>) => {
         const channelId = msg.data.channel_id;
-        const isAlreadyViewed = msg.data.prev_viewed_at === 0;
+        const isAlreadyViewed = Boolean(msg.data.prev_viewed_at);
 
         // if the user has already viewed the channel, there's no need to fetch actions again
         // prev_viewed_at is not coming for server 7.7 and below, so we additionally check redux
-        if (!isAlreadyViewed || hasViewedByChannelID(getState())[channelId]) {
+        if (isAlreadyViewed || hasViewedByChannelID(getState())[channelId]) {
             return;
         }
 
--- a/webapp/src/websocket_events.ts
+++ b/webapp/src/websocket_events.ts
@@ -201,12 +201,13 @@ export const handleWebsocketChannelUpdated = (getState: GetStateFunc, dispatch: 
 };
 
 export const handleWebsocketChannelViewed = (getState: GetStateFunc, dispatch: Dispatch) => {
-    return async (msg: WebSocketMessage<{ channel_id: string }>) => {
+    return async (msg: WebSocketMessage<{ channel_id: string, prev_viewed_at?: number}>) => {
         const channelId = msg.data.channel_id;
+        const isAlreadyViewed = msg.data.prev_viewed_at === 0;
 
-        // if the user has already viewed the channel,
-        // there's no need to fetch actions again
-        if (hasViewedByChannelID(getState())[channelId]) {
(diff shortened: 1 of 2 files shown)


webapp/src/websocket_events.ts
```diff
@@ -201,12 +201,13 @@ export const handleWebsocketChannelUpdated = (getState: GetStateFunc, dispatch:
 };
 
 export const handleWebsocketChannelViewed = (getState: GetStateFunc, dispatch: Dispatch) => {
-    return async (msg: WebSocketMessage<{ channel_id: string }>) => {
+    return async (msg: WebSocketMessage<{ channel_id: string, prev_viewed_at?: number}>) => {
         const channelId = msg.data.channel_id;
+        const isAlreadyViewed = Boolean(msg.data.prev_viewed_at);
```
lieut-data, line 206: nit: Could we test `msg.data.prev_viewed_at > 0` instead?

webapp/src/websocket_events.ts
```diff
@@ -201,12 +201,13 @@ export const handleWebsocketChannelUpdated = (getState: GetStateFunc, dispatch:
...
 
-        // if the user has already viewed the channel,
-        // there's no need to fetch actions again
-        if (hasViewedByChannelID(getState())[channelId]) {
+        // if the user has already viewed the channel, there's no need to fetch actions again
+        // prev_viewed_at is not coming for server 7.7 and below, so we additionally check redux
+        if (isAlreadyViewed || hasViewedByChannelID(getState())[channelId]) {
```
lieut-data, line 210: nit: if we have this information in Redux, why do we need the extra metadata in the websocket event? Is there a chance the websocket data is "stale" relative to Redux?

lieut-data (approved):

webapp/src/websocket_events.ts
trilopin: could be undefined if the server is not updated, 

if it's undefined, we'll still use the old flow.

I can do `msg.data.prev_viewed_at !== undefined && msg.data.prev_viewed_at > 0`, if it feels more explicit

crspeller (approved):

webapp/src/websocket_events.ts
lieut-data: This all makes sense! Excited about the future simplification once we can rely on the presence of this field.

trilopin:
I realized that the welcome message won't be sent if the first view of a channel is done on the mobile app (and the user is not connected at the time in desktop/browser).

In the previous approach, mobile was also ignored, but at least the message was sent in the first desktop view.

🤔

trilopin:
Closed.

The modal one was already extracted to a different (merged) pr.

I wrote a [proposal](https://docs.google.com/document/d/1lk83O-7dENMdP_24LfNgPR6eHfQGK3XNU_LGhpgJyBE/edit?usp=sharing) with different alternatives, and after some discussions I believe that would make sense to solve it 100% on server-side (hook or mpa).

PR #1753, opened 2023-01-23
fix several styles at  channel selector

trilopin:
## Summary
Several styles improved at channel selector:
- font-size is always 12px, no only when unknown channel
- channel icon is 12px, clear icon is 14 px 
- title/subtitle is 6px (title line height was larger before)
- channel icon margin is 7px

|Before|After|
|---|---|
||| 

## Ticket Link
Nope

## Checklist
- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


iomodo (approved):

abhijit-singh (approved):
Thanks @trilopin! I could spot a few other minor things in this modal. They're non-blocking so approving the PR but listing them out below::
1. The modal subtitle seems to have the wrong font. It should be `Open sans` and not `Metropolis`. `Metropolis` is supposed to be used only for headings. 
2. There should be ~14px more space between the subtitle and the divider between the modal header and body content.
image

3. Channel selector > Selected channel chip > The padding on the left of the public/private icon can be reduced to 6px instead of the current 8px. 
image

4. The space between the icon and the text can be reduced to 4px instead of the current 7px.
image

lieut-data:
@calebroseland, can you take a look at this one today as we try to get this in for the bug bash tomorrow?

calebroseland (approved):

trilopin:
@abhijit-singh I missed completely the point 2, will include it in a different PR

PR #1759, opened 2023-01-24
Avoid spurious session errors

lieut-data:
## Summary
The impetus for this change was to find and eliminate the source of the following ERROR logs in production:

> GetSessionByld: We encountered an error finding the session.

and logging a blank `sessionID`. The code in question was our `MessageHasBeenPosted` hook handler, which assumed that all messages posted had come from an active session. But this isn't the case for bots, to which the actions service intentionally wanted to respond.

Upon examining how the fetched session was used, I realized it was only to suppress one line in the `/playbook run` modal that linked to the Playbooks product as a way of explaining how to create a playbook:

CleanShot 2023-01-24 at 16 28 40@2x

As this functionality is not available on mobile, the message made no sense, so we attempted to suppress it in that case:

IMG_4C97897D5C17-1

Unfortunately, the suppression is driven by the session of the user that created the post, not the user consuming it. In some cases, that might be the same person, achieving the desired outcome, but in all other cases, it's more than likely to be wrong for one of the users.

The good news is that none of this is really necessary anymore, as our primary use case for launching a playbook no longer relies on this interactive dialog in the first place. So just rip out all this handling, sidestepping the potential for the ERROR log altogether.

## Ticket Link
None.

### DIFF
--- a/server/api/signal.go
+++ b/server/api/signal.go
@@ -60,12 +60,6 @@ func (h *SignalHandler) playbookRun(c *Context, w http.ResponseWriter, r *http.R
 		return
 	}
 
-	isMobile, err := getBoolField("isMobile", req.Context, w)
-	if err != nil {
-		h.returnError(publicErrorMessage, err, c.logger, w)
-		return
-	}
-
 	postID, err := getStringField("postID", req.Context, w)
 	if err != nil {
 		h.returnError(publicErrorMessage, err, c.logger, w)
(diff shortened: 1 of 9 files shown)


lieut-data:
cc @angeloskyratzakos, this should squash one of the ERROR logs we discovered together this morning.

crspeller (approved):

PR #1780, opened 2023-01-26
When focus is lost on LHS stop polling

crspeller:
## Summary
The desktop app loads playbooks in the background on startup and has the LHS visible. This will cause the polling for LHS updates to run in the background which is not desired. 

This patch attempts to mitigate this by only polling when the LHS is actually in focus.

### DIFF
--- a/webapp/src/components/sidebar/playbooks_sidebar.tsx
+++ b/webapp/src/components/sidebar/playbooks_sidebar.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import styled from 'styled-components';
 import {useSelector} from 'react-redux';
 import {getCurrentTeamId} from 'mattermost-redux/selectors/entities/teams';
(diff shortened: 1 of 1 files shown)


webapp/src/components/sidebar/playbooks_sidebar.tsx
```diff
@@ -47,18 +47,35 @@ export const playbookLHSQueryDocument = graphql(/* GraphQL */`
...
     const normalizeCategoryName = useReservedCategoryTitleMapper();
-    const {data, error} = useQuery(playbookLHSQueryDocument, {
+    const {data, error, startPolling, stopPolling} = useQuery(playbookLHSQueryDocument, {
         variables: {
             userID: 'me',
             teamID,
             types: [PlaybookRunType.Playbook],
         },
         fetchPolicy: 'cache-and-network',
-        pollInterval: 60000, // Poll every minute for updates
```
crspeller, line 59: No polling by default since the desktop app loads this in the background.

webapp/src/components/sidebar/playbooks_sidebar.tsx
```diff
@@ -47,18 +47,35 @@ export const playbookLHSQueryDocument = graphql(/* GraphQL */`
...
     });
 
+    useEffect(() => {
+        const focus = () => {
+            startPolling(pollInterval);
```
crspeller, line 65: Note this won't actually fire immediately on browser. But as soon as the user clicks anything this event fires. This is probably an acceptable issue given we hope to replace this with subscriptions soon.

lieut-data (approved):
Nice!! This seems perfect, honestly :)

trilopin (approved):
Love it, simple and effective solution

Issue #1741, opened 2023-01-16
set reminder for status update (snooze) is not working when multiple runs


trilopin:
#bug Snooze Status update is not fully working with multiple-runs in a channel. 

HTTP Status returned is 204 but UI is not updated. From the user's perspective seems to do nothing.

https://user-images.githubusercontent.com/4096774/212743226-58b0a7d4-5f77-4da7-b311-b2b903f42dda.mp4

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/5d3no8doribo7cc1s6bixg748w) *by @trilopin*._

trilopin:
This problem is bigger than it seemed.

reminder is extracting the run affected through the channeled, but if we have multiple runs with multiple reminders everything gets wrong.

I'm currently adding playbookrunid to post metadata and wiring everything again from there

PR #1781, opened 2023-01-27
Fix status update reminders a multiple run in a channelcontext

trilopin:
## Summary
Makes status update reminders work with multiple runs in a channel.

**Problem**
We were assuming one run per channel and we navigated channel->run at redux selector expecting one only result. 
That broke UI by showing same run name for multiple status-update-request in the same channel when multiple were asked. Additionally, when you wanted to snooze the status update, the post wasn't deleted.

**Approach**
Inject playbookRunId in the post.props when is type=custom_update_status:
- for new ones through code changes
- for old ones with a SQL migration

In the webapp component, several changes have been done:
- Stop relying on redux data, removed selector
- Exrtact playbookRunId from post.props and use an inline GraphQL query to get the relevant data

- [X] server e2e added
- [ ] cypress e2e added
- [ ] MySQL migration

## Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-playbooks/issues/1741

## Checklist
- [ ] ~~Telemetry updated~~
- [X] Unit tests updated


trilopin:
@agarciamontoro I've added you as a reviewer to increase confidence & awareness with the database migration (posts table).

server/sqlstore/migrations.go
```diff
@@ -2471,4 +2471,24 @@ var migrations = []Migration{
...
+			if e.DriverName() == model.DatabaseDriverMysql {
+
+			} else {
+				if _, err := e.Exec(`
+					UPDATE posts p
+					SET props=jsonb_insert(props,'{playbookRunId}', CONCAT('"',i.id,'"')::jsonb)
+					FROM ir_incident i
+					WHERE i.reminderpostid = p.id AND p.deleteat = 0
+					AND p.type = 'custom_update_status'	AND p.props -> 'playbookRunId' IS NULL
+				`); err != nil {
```
lieut-data, line 2487: I'm inclined to skip this migration, and instead pick the first playbook run when there are multiple and we haven't previously recorded a post prop. Until v7.7, there could only be one run per channel, with channels being created explicitly for the purpose of that playbook run vs. being long-lived entities likely to inherit additional runs later on.

If we did decide this migration was necessary, we'd need to route this through server team to evaluate the potential performance impact, measure on a large server, etc.

lieut-data (changes requested):
Nice work, @trilopin! Just one thought below.

agarciamontoro:
Was my review removed by on purpose?

trilopin:
> Was my review removed on purpose?

Yes, based on the Ci failures and @lieut-data feedback

agarciamontoro:
Ok! Feel free to re-assign me if needed :)

tests-e2e/cypress/integration/channels/update_request_post_spec.js
```diff
@@ -38,7 +39,7 @@ describe('channels > update request post', () => {
             // # Enable threads view
             cy.apiSaveCRTPreference(testParticipant.id, 'on');
 
-            // # Create a public playbook
+            // # Create a public playbook with 2 runs in the same channel
```
trilopin, line 42: multiple runs should be the new normal for tests, I'd rather go this way instead of making a path for single runs and another for multiple (same testing, double maintenance)

webapp/src/components/update_request_post.tsx
```diff
@@ -29,19 +30,84 @@ import {
...
+            channelID: $channelID,
+            statuses: ["InProgress"],
+            first: 1,
+        ) {
+            edges {
+                node {
+                    id
+                    name
+                    previousReminder
+                    reminderTimerDefaultSeconds
```
trilopin, line 63: I've tried to use a fragment here to reuse the list of fields but I couldn't make the types work.

modd.conf
```diff
@@ -7,6 +7,10 @@
     prep +onchange: make graphql && make bundle && make upload-to-server
 }
 
+**/*.tsx {
+    prep +onchange: make graphql
```
trilopin, line 11: we should keep an eye on this, it's for inline queries but not 100% sure it's needed

modd.conf
```diff
@@ -7,6 +7,10 @@
     prep +onchange: make graphql && make bundle && make upload-to-server
```
lieut-data, line 7: TIL about https://github.com/cortesi/modd. Should we add i18n stuff somewhere here?

lieut-data (approved):
Nice!

modd.conf
trilopin: This is the live reload I added for server. 

re: i18n I'm not sure if I would add it here since:
- does not help to the development process 
- can be run in a pre-commit hook

I always forget to run i18n-extract so I added to my pre-commit hook:
```
#!/bin/sh
make fix-style
make i18n-extract
```

trilopin: And I run a `make check-style` in the pre-push hook

modd.conf
crspeller: Needs it for ts files too as you could have a query or fragment in a TS file if it is related to a custom hook. 

Also the graphql generator stuff has its own watcher is we wanted to use that instead.

crspeller (approved):

PR #1786, opened 2023-01-30
GH-1726 Update run details view style

### LINKED ISSUE #1726: RHS: Run view
See [Figma file](https://www.figma.com/file/rU4vGGtxk3lBIghH4o0yj6/Checklists?node-id=609%3A244876&t=tdE7PsxEVGqYZ733-0) for reference

Image

m-ripper:
## Summary

Update style of run details view: Remove hard-coded height and implement smaller tweaks.

## Ticket Link

Fixes #1726 

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~


webapp/src/components/assets/icons/external_link.tsx
```diff
@@ -23,7 +23,6 @@ const ExternalLink = (props : {className?: string}) => (
         <path
             d='M8.494 0.250048V1.74405H11.194L3.814 9.12405L4.876 10.186L12.256 2.80605V5.50605H13.75V0.250048H8.494ZM12.256 12.256H1.744V1.74405H7V0.250048H1.744C1.336 0.250048 0.982 0.394048 0.682 0.682048C0.394 0.970048 0.25 1.32405 0.25 1.74405V12.256C0.25 12.664 0.394 13.012 0.682 13.3C0.982 13.6 1.336 13.75 1.744 13.75H12.256C12.664 13.75 13.012 13.6 13.3 13.3C13.6 13.012 13.75 12.664 13.75 12.256V7.00005H12.256V12.256Z'
             fill='currentColor'
-            fillOpacity='0.48'
```
iomodo, line 26: why remove this?

iomodo (changes requested):
Looks good @m-ripper 
We should change the header to show the title right?

webapp/src/components/assets/icons/external_link.tsx
m-ripper: I've talked about the header with @abhijit-singh and we came to the conclusion that the title in the header is not really useful because the title can be quite long and the header space is quite narrow.  Also, we didn't have a nice solution for renaming. 

About the removed fillOpacity: This caused the icon to always seem lighter than it actually is. This way it's easier to replicate the color that is used in the designs.

lieut-data:
Skipping this in line with the shifting priorities on Playbooks features.

PR #1789, opened 2023-01-31
Migrate run details page to GraphQL.

crspeller:
## Summary
Migrating the run details page to GraphQL


lieut-data:
Wow, this was an incredible change @gbochora! Unfortunately, we're looking at dramatically reducing the scope of the Playbooks featureset, and I don't think we'll need these changes right now. Still, appreciate the effort, and sad that we didn't get to ship this together :/

PR #1794, opened 2023-02-01
GH-1773 Fix potential duplicate runs

### LINKED ISSUE #1773: Double clicks can start duplicate runs
Clicking too fast when starting a run can cause multiple runs to be started.

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/xxno8tska7fuf8as7fniud1sya) *by @m-ripper*._

m-ripper:
## Summary

When the internet of the user or the server is slow, it is possible to start multiple runs at once by clicking the "Start run" button multiple times. The "Start run" button is now disabled while the client is waiting for the response from the server.

[Screencast from 01.02.2023 16:06:50.webm](https://user-images.githubusercontent.com/8669935/216080548-7966b752-3c38-49b2-ae8d-b2c19d471e47.webm)

## Ticket Link

Fixes #1773 

## Checklist

- [ ] ~~Telemetry updated~~
- [ ] ~~Unit tests updated~~

### DIFF
--- a/webapp/src/components/modals/run_playbook_modal.tsx
+++ b/webapp/src/components/modals/run_playbook_modal.tsx
@@ -63,6 +63,7 @@ const RunPlaybookModal = ({
     const [createPublicRun, setCreatePublicRun] = useState(false);
     const [searchTerm, setSearchTerm] = useState('');
     const [showsearch, setShowsearch] = useState(true);
+    const [isSubmitting, setIsSubmitting] = useState(false);
     const canCreatePlaybooks = useCanCreatePlaybooksInTeam(teamId || '');
 
     let userId = useSelector(getCurrentUserId);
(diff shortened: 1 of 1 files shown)


webapp/src/components/modals/run_playbook_modal.tsx
```diff
@@ -138,7 +140,7 @@ const RunPlaybookModal = ({
                 onRunCreated(newPlaybookRun.id, newPlaybookRun.channel_id, statsData);
             }).catch(() => {
             // show error
-            });
+            }).finally(() => setIsSubmitting(false));
```
trilopin, line 143: wouldn't this be the case for using debounce?

m-ripper: Debounce could work as well, but would be more complicated to implement for no gain in usage, because the time frame would need to be hard-coded and cancel/flush would have to be used. This has the same result but is a lot simpler, what do you think?

trilopin (approved):

PR #1795, opened 2023-02-01
GH-1767 Default to current channel when linking a new playbook run

### LINKED ISSUE #1767: Default to the current channel when choosing to link a new playbook run
If a playbook defaults to "Create a run channel", and I choose to "Link run to an existing channel" instead, let's default to the currently selected channel (if there is one), instead of forcing the user to start from scratch as present:

image

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/dohj39bx4pgc5gnm9g1o99bjer) *by @lieut-data*._

m-ripper:
## Summary

If a playbook defaults to "Create a run channel" and "Link run to an existing channel" is chosen, the default channel will be the current one instead of none.

## Ticket Link

Fixes #1767 

## Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated

### DIFF
--- a/tests-e2e/cypress/integration/channels/rhs/start_run_rhs_spec.js
+++ b/tests-e2e/cypress/integration/channels/rhs/start_run_rhs_spec.js
@@ -240,8 +240,8 @@ describe('channels rhs > start a run', () => {
                         // # Fill run name
                         cy.findByTestId('run-name-input').clear().type('Test Run Name');
 
-                        // # Fill Town square as the channel to be linked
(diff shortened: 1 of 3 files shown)


gbochora (approved):
Nice!

trilopin (approved):
The approach is perfect, nicely done!

It would be even greater if you added a couple of e2e tests testing explicitly this new behavior:
- when the user click link-to-existent-channel but the playbook already had one
- when the user click link-to-existent-channel but we are in the backstage (no currentChannelId)
- when all conditions are met to show current channel

It's also fine if we already have these cases and we add additional assertions.

PR #1802, opened 2023-02-02
GH-1764 Prevent unnamed playbooks

### LINKED ISSUE #1764: Prevent unnamed playbooks
We currently allow playbooks to be saved with an empty title, either by directly editing and simply deleting the name, or by importing a playbook without a name:

image

Let's prevent this in two ways:
* If there was a previous name, say because the user just deleted the name, then restore that and ignore the attempt to clear the name. Do this after taking into account whitespaces, etc.
* If there was no previous name, generate one similar to the name we use when create a playbook from a template: "@username's Playbook"

No need to fix existing playbooks without names, as these can just be fixed by the user as needed.

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/i64rnimzrt8m7kzqjcnjo91xbo) *by @lieut-data*._

m-ripper:
## Summary

Make sure playbooks always have a title:
- When importing or creating a playbook with an empty title, the title will be generated based on the username. 
- When updating a playbook with an empty title, the current title will be used and the new one omitted.

## Ticket Link

Fixes #1764 

## Checklist

- [ ] ~~Telemetry updated~~
- [x] Unit tests updated


lieut-data:
Thanks again, @m-ripper! I'm going to close this out, despite it being a solid change, as we figure out next steps for the Playbooks featureset.

PR #1805, opened 2023-02-04
Morph v0.30.0 v0.36.0

gbochora:
## Summary
- add morph migration files for v0.30.0-v0.36.0. 
- add constraints validation for migration tests
- move schema, index, and constraint info extraction in one function

## Ticket Link
No ticket

## Checklist

~~- [ ] Telemetry updated~~
- [x] Unit tests updated


lieut-data (approved):
We will probably need to revisit where we land with Playbooks after MPA, but I'm going to accept this change as-is. Love the unit tested migrations, @gbochora -- thank you, again.

PR #1808, opened 2023-02-13
Gh 1727 runmodal templates

trilopin:
## Summary
Add templates to "start a run" modal.

Key points:
- that will always appear as the last group of "playbooks"
- when selected, run will be created without playbook, directly from the template
- template data can be changed in the second step of the modal (channel-mode, public/private, channel_name, run sumary)

## Checklist
- [X] Telemetry updated
- [ ] Unit tests updated


server/api/playbook_runs.go
```diff
@@ -152,31 +153,59 @@ func (h *PlaybookRunHandler) checkEditPermissions(next http.Handler) http.Handle
 	})
 }
 
+type playbookRunCreateOptions struct {
```
trilopin, line 156: a type from the client package was used and it seemed non-consistent with the rest of the client package approach

Aside from creating type here, the main change is to add Playbook as optional field

lieut-data:
@trilopin, I'm going to close this one out for now as we move to the monorepo, but feel free to re-open if you're inclined to keep the momentum on this effort :)

PR #1821, opened 2023-02-24
MM-50738 Remove keyword prompt post instead of editing it.

crspeller:
## Summary
Removes the editing of the keyword prompt action post and instead deletes it, leaving the notifications of run creation to the regular channels. 

## Ticket Link
https://mattermost.atlassian.net/browse/MM-50738

### DIFF
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -266,6 +266,7 @@ func (h *PlaybookRunHandler) createPlaybookRunFromDialog(c *Context, w http.Resp
 			TeamID:      request.TeamId,
 			ChannelID:   playbook.GetRunChannelID(),
 			Name:        name,
+			PostID:      state.PostID,
 			PlaybookID:  playbookID,
 			Type:        app.RunTypePlaybook,
 		},
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -266,7 +266,6 @@ func (h *PlaybookRunHandler) createPlaybookRunFromDialog(c *Context, w http.Resp
 			TeamID:      request.TeamId,
 			ChannelID:   playbook.GetRunChannelID(),
 			Name:        name,
-			PostID:      state.PostID,
 			PlaybookID:  playbookID,
 			Type:        app.RunTypePlaybook,
 		},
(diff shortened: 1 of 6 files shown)


server/api/signal.go
```diff
@@ -60,30 +60,22 @@ func (h *SignalHandler) playbookRun(c *Context, w http.ResponseWriter, r *http.R
...
 	}
 
-	if err := h.playbookRunService.OpenCreatePlaybookRunDialog(req.TeamId, req.UserId, req.TriggerId, postID, "", []app.Playbook{pbook}, post.Id); err != nil {
+	if err := h.playbookRunService.OpenCreatePlaybookRunDialog(req.TeamId, req.UserId, req.TriggerId, "", "", []app.Playbook{pbook}); err != nil {
 		h.returnError("can't open dialog", errors.Wrap(err, "can't open a dialog"), c.logger, w)
 		return
 	}
 
 	ReturnJSON(w, &model.PostActionIntegrationResponse{}, http.StatusOK)
+	if err := h.api.Post.DeletePost(req.PostId); err != nil {
```
trilopin, line 75: you call `returnJSON` before DeletePost, is that intended?

server/api/playbook_runs.go
```diff
@@ -342,33 +334,6 @@ func (h *PlaybookRunHandler) createPlaybookRunFromDialog(c *Context, w http.Resp
...
-		return errors.Wrapf(err, "unable to retrieve the playbook with ID %q", playbookID)
-	}
-
-	user, err := h.pluginAPI.User.Get(userID)
-	if err != nil {
-		return errors.Wrapf(err, "unable to retrieve the user with ID %q", userID)
-	}
-
-	// Update the message and remove the attachments; i.e., the buttons
-	post.Message = fmt.Sprintf("@%s started the run [%s](%s) using the [%s](%s) playbook.",
```
trilopin, line 362: 1/5 wouldn't be useful to insert the content of the post not as edit but as a new post to still communicate the trigger-result in the channel?

server/api/signal.go
trilopin: silly question, do DeletePost handle permissions, what if this method is called with an arbitrary req.postId, would it be deleted? should we check that the author is playbooksBot?

trilopin (changes requested):
Thanks @crspeller, I left a couple of comments and questions.

I've tested manually and it works perfectly 🎉.

A server test api_signal_test to check that the post is correctly deleted would be highly appreciated.

server/api/signal.go
crspeller: Just following the form of the `ignoreKeywords` handler that does this. I don't think it matters. 

No that's a good question. You can't call this endpoint without a valid triggerid it will fail before it gets to the deletion portion.

server/api/playbook_runs.go
crspeller: I think there are already plenty of notification mechanisms for run start. (it does post an ephemeral message)

trilopin (approved):
unblocking here

server test still desired ❤️

calebroseland (approved):

jupenur (approved):

PR #1826, opened 2023-02-27
MM-50818 Fix module loading issues caused by shared modules

hmhealey:
The errors that we were seeing where either Boards or Playbooks were failing to load when they were both loaded at the same time seems like it was happening due to an issue with the resolution of shared modules by module federation. Specifically, it seems to be due to how it was trying to choose which version of those shared modules to use between the ones bundled with the web app, Boards, and Playbooks.

Since I've been trying to make the web app the authority for these dependencies, I just changed it so that the Playbooks build will never include "fallback" versions of React and such since we know they'll always exist in the web app.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50818

#### Related Pull Requests
https://github.com/mattermost/focalboard/pull/4605
https://github.com/mattermost/mattermost-webapp/pull/12282

### DIFF
--- a/webapp/webpack.config.js
+++ b/webapp/webpack.config.js
@@ -130,7 +130,7 @@ if (TARGET_IS_PRODUCT) {
 
                 // Set these to false so that any version provided by the web app will be accepted
                 requiredVersion: false,
-                version: false
+                version: false,
             };
         }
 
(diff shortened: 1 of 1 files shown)


lieut-data (approved):

hmhealey:
@lieut-data I need a minute. I forgot one part of this

hmhealey:
Okay, now it's good to go

lieut-data (approved):

lieut-data:
Yep, CI on this branch isn't working yet -- feel free to merge @hmhealey :)

PR #1827, opened 2023-02-28
Update Webpack dev server configuration in product mode

hmhealey:
Both of these changes are to fix some issues with the Webpack dev server since I run my local instance on https after having had to set it up for something else.


lieut-data (approved):
Looks good, but maybe we just need a file-level ignore for using `process` here?

trilopin (approved):
🎉

lieut-data:
Thanks! I'll merge this, as we're still sorting out the CI steps for this particular branch.

PR #1849, opened 2023-06-02
Backport: MM-52476 Fix guest users access to playbooks. (#23279)

crspeller:
## Summary

Backport of https://github.com/mattermost/mattermost/pull/23279

### DIFF
--- a/server/app/permissions_service.go
+++ b/server/app/permissions_service.go
@@ -67,7 +67,7 @@ func (p *PermissionsService) getPlaybookRole(userID string, playbook Playbook) [
 	// Public playbooks
 	if playbook.Public {
 		// Public playbooks are public to those who can list channels on a team. (Not guests)
-		if p.api.HasPermissionToTeam(userID, playbook.TeamID, model.PermissionListTeamChannels) {
+		if p.pluginAPI.User.HasPermissionToTeam(userID, playbook.TeamID, model.PermissionListTeamChannels) {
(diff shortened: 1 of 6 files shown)


server/app/permissions_service.go
```diff
@@ -66,10 +66,13 @@ func (p *PermissionsService) getPlaybookRole(userID string, playbook Playbook) [
 
 	// Public playbooks
 	if playbook.Public {
-		if playbook.DefaultPlaybookMemberRole == "" {
-			return []string{playbook.DefaultPlaybookMemberRole}
+		// Public playbooks are public to those who can list channels on a team. (Not guests)
+		if p.pluginAPI.User.HasPermissionToTeam(userID, playbook.TeamID, model.PermissionListTeamChannels) {
```
lieut-data, line 70: Why not another `guest` check vs. inferring something from a semi-related permission?

crspeller: Because hard guest checks are cheating vs using the permissions system properly.

lieut-data: In general, I agree, but using it properly would be to have a `model.PermissionViewPublicPlaybooks` or some such. This just ties together two things that happen to be related, but who knows how the product will change in the future? At least the hard-coded guest check is unlikely to regress.

crspeller: That's true. We should add the `model.PermissionListPublicPlaybooks` or some such. We can fix this on master. Not going to fix here for the backport.

PR #1856, opened 2023-06-27
[fix] CD revisiting naming for artifacts generated

phoinixgrr:
## Summary

We are currently using the scheme: `${GITHUB_REPOSITORY#*/}-${GITHUB_REF_NAME}.tar.gz` to name the playbook-plugin artifact, in the CD process. 

CircleCI pipelines used the following naming scheme: `${GITHUB_REF_NAME}.tar.gz`
Previous CircleCI release example: https://github.com/mattermost/mattermost-plugin-playbooks/releases/tag/v1.36.2

We are aligning with the previous naming scheme.

## Ticket Link

Ticket: https://mattermost.atlassian.net/browse/CLD-5870

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -109,7 +109,7 @@ jobs:
 
       - name: cd/rename-artifacts
         run: |
-          mv dist/*.tar.gz dist/${GITHUB_REPOSITORY#*/}-${GITHUB_REF_NAME}.tar.gz
+          mv dist/*.tar.gz dist/playbooks-${GITHUB_REF_NAME}.tar.gz
 
       - name: cd/rename-artifacts
         uses: aws-actions/configure-aws-credentials@67fbcbb121271f7775d2e7715933280b06314838 # v1.7.0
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yaml
```diff
@@ -109,7 +109,7 @@ jobs:
 
       - name: cd/rename-artifacts
         run: |
-          mv dist/*.tar.gz dist/${GITHUB_REPOSITORY#*/}-${GITHUB_REF_NAME}.tar.gz
+          mv dist/*.tar.gz dist/${GITHUB_REF_NAME}.tar.gz
```
mvitale1989, line 112:
```suggestion
          mv dist/*.tar.gz dist/playbooks-${GITHUB_REF_NAME}.tar.gz
```
Shouldn't it be like this to match the old format? Afaiu the `GITHUB_REF_NAME` variable is only the name of the given git tag

phoinixgrr: Aha! Good catch Mario. rebasing and pushing the fix.

phoinixgrr: @mvitale1989  Done with https://github.com/mattermost/mattermost-plugin-playbooks/compare/35594279667c2d936a85114f516b78eca9db7c16..93c90347386be52e60a5ab5996be876ba1bb9a8f

toninis (approved):

PR #1860, opened 2023-07-21
Add docs link in playbooks intro

esethna:
Added link to docs in the playbooks intro screen


esethna:
@sbishel does this PR need to be moved over to the monorepo?

sbishel:
> does this PR need to be moved over to the monorepo?

I don't think so, playbooks is not part of the mono-repo anymore.

lieut-data:
@esethna, feel free to reopen if needed!

PR #1866, opened 2023-08-17
restore the missing Plugin telemetry key

lieut-data:
## Summary
For now, we're just hard-coding this while still honouring the server configuration as to whether or not to emit telemetry in the first place.

In the future, we should integrate knowledge of the service environment and use that to key off which key (if any) to use.

## Ticket Link
None.


crspeller (approved):

lieut-data:
I'll look into the e2e test issues when we pre-package and run the suite there.

PR #1869, opened 2023-08-25
MM-53753 Fix error messages for too long json

crspeller:
Turns out that if you wrap a nil error it returns nil instead of a new error. We were using Wrapf when we meant Errorf leading to a function returning (`nil`, `nil`) instead of (`nil`, `error`) as intended.


sbishel (approved):

esarafianou (approved):
LGTM

calebroseland (approved):

crspeller:
e2e failures are unrelated and will be fixed separately. See: https://github.com/mattermost/mattermost-plugin-playbooks/pull/1866

PR #1870, opened 2023-09-06
Updated legacy links

cwarnermm:
## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


lieut-data (approved):
Thank you, @cwarnermm!

crspeller (approved):

cwarnermm:
@lieut-data @crspeller - I see there are some checks that aren't successful on this PR and blocking me from merging this change. Is there something I can do to make these checks successful?

lieut-data:
@cwarnermm, it's not your changes. I think something has regressed in the e2e runs when we execute them from this side of the equation, vs. in the monorepo. I'll take a look separately, and just get your changes merged.

PR #1872, opened 2023-09-11
Drop cloud e2e tests & sync with monorepo

lieut-data:
## Summary
The Playbooks e2e tests have been failing for a while. Let's fix them!
* Remove the deprecated `.circleci/` configuration. (I wasted some time "fixing" these before realizing it's not run anymore!)
* Update our version of the e2e tests with the copy in `monorepo`. I'll push these changes there after merge. Long term, it would be great to avoid duplicating this in the first place...
* We don't need to run `master` and `cloud` e2e tests anymore, since we do these e2e tests in the monorepo already. So just test against latest when making direct plugin changes.
* Address a handful of "e2e flakiness"
* Transition to the new `DisableAppBar` configuration option
* Use Redux store subscriptions to trigger welcome messages.

This last change deserves some additional discussion. https://github.com/mattermost/mattermost/pull/24003 introduced a breaking change in that the server stopped emitting a `channel_viewed` websocket message in favour of the new `multiple_channels_viewed` message, but also changed the semantics to only emit if there was something to "mark as read". Meaning if there were no new posts, the server doesn't bother notifying the client at all.

We used to depend on this websocket event to trigger the "welcome message action", showing an optional ephemeral message the first time a channel was loaded. Now, we trigger the same heuristic by subscribing to the Redux store and watching the current channel change.

## Ticket Link
None.


cpoile (approved):
Nice, I like the new welcome message logic. But I'm always a bit worried about adding listeners to the store (we already do so much). But I guess it can't be helped. :)

cpoile:
Heads up that Alejandro is out this week, maybe Michael?

calebroseland (approved):
Nice! Overall I don't think the additional listener will add too much overhead, but long-term, it might be nice to write additional plugin api hooks for common patterns like this (channel viewed, X route accessed, etc.)

PR #1877, opened 2023-10-24
Move RHS opener to GraphQL

crspeller:
Migrates the RHS opener logic to GraphQL to reduce the overhead of fetching all the runs on the team. We where fetching all the data on each run when all we really needed was if there was an active run in the channel.

Status posts, timeline events, and metrics where set behind an option that GraphQL excludes and fetches separately itself if requested. 

This came up because a customer was encountering the `getTimelineEventsForPlaybookRun` subquery executing with near 1000 runs. This appears to be the only situation outside of having 1000 runs in a channel that this could happen.

### DIFF
--- a/server/api/graphql.go
+++ b/server/api/graphql.go
@@ -25,6 +25,7 @@ type GraphQLHandler struct {
 	config             config.Service
 	permissions        *app.PermissionsService
 	playbookStore      app.PlaybookStore
+	runStore           app.PlaybookRunStore
 	licenceChecker     app.LicenseChecker
 
 	schema *graphql.Schema
@@ -42,6 +43,7 @@ func NewGraphQLHandler(
 	configService config.Service,
 	permissions *app.PermissionsService,
 	playbookStore app.PlaybookStore,
+	runStore app.PlaybookRunStore,
 	licenceChecker app.LicenseChecker,
 ) *GraphQLHandler {
 	handler := &GraphQLHandler{
@@ -53,6 +55,7 @@ func NewGraphQLHandler(
 		config:             configService,
 		permissions:        permissions,
 		playbookStore:      playbookStore,
+		runStore:           runStore,
 		licenceChecker:     licenceChecker,
 	}
 
@@ -85,18 +88,22 @@ func NewGraphQLHandler(
 type ctxKey struct{}
 
 type GraphQLContext struct {
-	r                  *http.Request
-	playbookService    app.PlaybookService
-	playbookRunService app.PlaybookRunService
-	playbookStore      app.PlaybookStore
-	categoryService    app.CategoryService
-	pluginAPI          *pluginapi.Client
-	logger             logrus.FieldLogger
-	config             config.Service
-	permissions        *app.PermissionsService
-	licenceChecker     app.LicenseChecker
-	favoritesLoader    *dataloader.Loader[favoriteInfo, bool]
-	playbooksLoader    *dataloader.Loader[playbookInfo, *app.Playbook]
+	r                    *http.Request
+	playbookService      app.PlaybookService
+	playbookRunService   app.PlaybookRunService
+	playbookStore        app.PlaybookStore
+	runStore             app.PlaybookRunStore
+	categoryService      app.CategoryService
+	pluginAPI            *pluginapi.Client
+	logger               logrus.FieldLogger
+	config               config.Service
+	permissions          *app.PermissionsService
+	licenceChecker       app.LicenseChecker
+	favoritesLoader      *dataloader.Loader[favoriteInfo, bool]
+	playbooksLoader      *dataloader.Loader[playbookInfo, *app.Playbook]
+	statusPostsLoader    *dataloader.Loader[string, []app.StatusPost]
+	timelineEventsLoader *dataloader.Loader[string, []app.TimelineEvent]
+	runMetricsLoader     *dataloader.Loader[string, []app.RunMetricData]
 }
 
 // When moving over to the multi-product architecture this should be handled by the server.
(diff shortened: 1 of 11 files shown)


lieut-data:
It's not immediately obvious to me what switching to GraphQL solves in this case: can you elaborate?

Looks like a number of new e2e tests failing (there was just one failing on `master` recently with the i18n changes). Is there an expected semantic difference expected with this change, @crspeller?

server/api/graphql_loader_run.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		result[i] = &dataloader.Result[V]{
+			Data: V(timelineEvents),
+		}
+	}
+
+	return result
+}
+
+func graphQLRunMetricsLoader[V []app.RunMetricData](ctx context.Context, keys []string) []*dataloader.Result[V] {
```
crspeller, line 75: Only after writing the metrics loaded did I notice it isn't supported by GraphQL. Leaving in for potential future support.

server/api/graphql.go
```diff
@@ -124,20 +131,27 @@ func (h *GraphQLHandler) graphQL(c *Context, w http.ResponseWriter, r *http.Requ
 	// dataloaders
 	favoritesLoader := dataloader.NewBatchedLoader(graphQLFavoritesLoader[bool], dataloader.WithBatchCapacity[favoriteInfo, bool](loaderBatchCapacity))
 	playbooksLoader := dataloader.NewBatchedLoader(graphQLPlaybooksLoader[*app.Playbook], dataloader.WithBatchCapacity[playbookInfo, *app.Playbook](loaderBatchCapacity))
+	statusPostsLoader := dataloader.NewBatchedLoader(graphQLStatusPostsLoader[[]app.StatusPost], dataloader.WithBatchCapacity[string, []app.StatusPost](loaderBatchCapacity))
```
crspeller, line 134: All new capabilities have dataloaders for efficiency.

webapp/src/rhs_opener.ts
```diff
@@ -5,18 +5,40 @@ import {Store} from 'redux';
...
+        $status: String!,
+    ) {
+        runs(
+            teamID: $teamID,
+            statuses: [$status],
+            participantOrFollowerID: $participant,
+        ) {
+            edges {
+                node {
+                    channel_id: channelID
```
crspeller, line 33: Only the channel and team ID are needed to be able to open the RHS. We always re-fetch when the user views a run anyway.

crspeller:
@lieut-data Reviews where requested before completion. Now it should be clear we are not making the slow query. 

The tests I think are fixed now. (accidentally hit rerun all)

server/api/graphql_loader_run.go
```diff
@@ -0,0 +1,103 @@
+package api
+
+import (
+	"context"
+
+	"github.com/graph-gophers/dataloader/v7"
+	"github.com/mattermost/mattermost-plugin-playbooks/server/app"
+)
+
+func graphQLStatusPostsLoader[V []app.StatusPost](ctx context.Context, keys []string) []*dataloader.Result[V] {
```
lieut-data, line 10: nit: could `keys` be called `playbookRunIDs` for clarity?

server/app/playbook_run.go
```diff
@@ -929,6 +929,15 @@ type PlaybookRunStore interface {
 
 	// GetStatusAsTopicMetadataByIDs gets PlaybookRunIDs and TeamIDs from runs by statusIDs
 	GetStatusAsTopicMetadataByIDs(statusIDs []string) ([]TopicMetadata, error)
+
+	// GetStatsPostsByIDs gets the status posts for playbook runs
+	GetStatusPostsByIDs(playbookRunID []string) (map[string][]StatusPost, error)
```
lieut-data, line 934: nit: it's ambiguous if `ByIDs` here refers to playbook runs or posts (or timeline events or metrics below).

server/api/graphql_loader_run.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return populateResultWithError(err, result)
+	}
+
+	timelineEvents, err := c.runStore.GetTimelineEventsByIDs(keys)
+	if err != nil {
+		return populateResultWithError(err, result)
+	}
+
+	timelineEventsByRunID := make(map[string][]app.TimelineEvent)
```
lieut-data, line 56: nit: would it be consistent to use the generic type here?
```suggestion
	timelineEventsByRunID := make(map[string]V)
```

server/sqlstore/playbook_run.go
```diff
@@ -776,6 +778,51 @@ func (s *playbookRunStore) GetPlaybookRun(playbookRunID string) (*app.PlaybookRu
...
+	statusPostsByRunID := make(map[string][]app.StatusPost)
+	for _, statusPost := range statusPosts {
+		statusPostsByRunID[statusPost.PlaybookRunID] = append(statusPostsByRunID[statusPost.PlaybookRunID], statusPost.StatusPost)
+	}
+
+	return statusPostsByRunID, nil
+}
+
+func (s *playbookRunStore) GetTimelineEventsByIDs(playbookRunIDs []string) ([]app.TimelineEvent, error) {
+	return s.getTimelineEventsForPlaybookRun(s.store.db, playbookRunIDs)
```
lieut-data, line 796: Should we return a map like the method above?

lieut-data (approved):
A bunch of nits, but nothing blocking. Nice work!

wiggin77 (approved):
LGTM :+1:   Nice work.

server/sqlstore/playbook_run.go
crspeller: The only reason the other one returns a map is because they don't include the runID in the struct so we have to give that info to the user somehow. In this case it does include it so we leave the decision to create a map to the caller.

server/app/playbook_run.go
crspeller: The parameter documents what you need to pass.

crspeller:
@lindy65 All I need here is some general testing around the RHS opening functionality. Also just a general test if playbooks are loading correctly everywhere.

server/app/playbook_run.go
lieut-data: Does it, though?
```go
store.GetStatusPostsByIDs([]string{"abcaakaa..."})
```

And what if we ever did want to support more than one filter parameter (e.g. `PostIDs`, or `TeamIDs`).

I'm only 0/5 -- it's a nit at best :)

lindy65:
Thanks @crspeller - is there a Jira ticket for this PR so I can track post merge?

lindy65 (approved):
RHS and Playbooks look good to me @crspeller 👍 

Let me know if I need to create a Jira ticket :)

PR #1878, opened 2023-10-27
MM-54989 Fix team run team interaction.

crspeller:
#### Summary

Adding checks for team permissions when accessing runs. 

#### Ticket

https://mattermost.atlassian.net/browse/MM-54989

### DIFF
--- a/server/api_runs_test.go
+++ b/server/api_runs_test.go
@@ -515,6 +515,55 @@ func TestRunRetrieval(t *testing.T) {
 	})
 }
 
+func TestRunPostStatusUpdateDialog(t *testing.T) {
+	e := Setup(t)
+	e.CreateBasic()
+
+	t.Run("post an update", func(t *testing.T) {
+		dialogRequest := model.SubmitDialogRequest{
+			TeamId: e.BasicTeam.Id,
+			UserId: e.RegularUser.Id,
+			State:  "{}",
+			Submission: map[string]interface{}{
+				app.DialogFieldMessageKey:           "someupdate",
+				app.DialogFieldReminderInSecondsKey: "100000",
+				app.DialogFieldFinishRun:            false,
+			},
+		}
+		dialogRequestBytes, err := json.Marshal(dialogRequest)
+		require.NoError(t, err)
+
+		result, err := e.ServerClient.DoAPIRequestBytes("POST", e.ServerClient.URL+"/plugins/"+manifest.Id+"/api/v0/runs/"+e.BasicRun.ID+"/update-status-dialog", dialogRequestBytes, "")
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, result.StatusCode)
+	})
+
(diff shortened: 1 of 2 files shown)


lieut-data (approved):

server/api_runs_test.go
```diff
@@ -548,6 +548,15 @@ func TestRunPostStatusUpdate(t *testing.T) {
 		assert.Error(t, err)
 	})
 
+	t.Run("no permissions to run", func(t *testing.T) {
+		_, err := e.ServerAdminClient.RemoveTeamMember(e.BasicRun.TeamID, e.RegularUser.Id)
+		require.NoError(t, err)
+		err = e.PlaybooksClient.PlaybookRuns.UpdateStatus(context.Background(), e.BasicRun.ID, "update", 600)
```
esarafianou, line 554: UpdateStatus doesn't call the `POST /plugins/playbooks/api/v0/runs/<RUN_ID>/update-status-dialog` endpoint. It calls the `runs/<RUN_ID>/status` endpoint which is not what we'd like to test: https://github.com/mattermost/mattermost-plugin-playbooks/blob/a10bb42418667c0f5337f44b39e582f7a8fc81cf/client/playbook_runs.go#L141C27-L141C43.

server/app/permissions_service.go
```diff
@@ -410,6 +410,10 @@ func (p *PermissionsService) RunManageProperties(userID, runID string) error {
 }
 
 func (p *PermissionsService) runManagePropertiesWithPlaybookRun(userID string, run *PlaybookRun) error {
```
esarafianou, line 412: The `POST /plugins/playbooks/api/v0/runs/<RUN_ID>/update-status-dialog` doesn't seem to ever call the `runManagePropertiesWithPlaybookRun` function so this fix won't get applied in that endpoint: https://github.com/mattermost/mattermost-plugin-playbooks/blob/a10bb42418667c0f5337f44b39e582f7a8fc81cf/server/api/playbook_runs.go#L922-L971

esarafianou (changes requested):

server/app/permissions_service.go
crspeller: It is a little confusing if you don't know what is going on. There is actually some registered middleware that is taking care of it. You can see it here: https://github.com/mattermost/mattermost-plugin-playbooks/blob/d03dfd8d4e200e8db5c5f64ae05a7ba4b68032fc/server/api/playbook_runs.go#L79

server/api_runs_test.go
crspeller: The code paths for checking permissions are identical. The only difference is the input.
However I added a test for it just to be safe.

sbishel (approved):

server/app/permissions_service.go
esarafianou: Aah, that makes sense. Thanks for pointing to that! I missed it.

esarafianou (approved):
LGTM!

PR #1880, opened 2023-10-31
Prevent extra JSON from being stored in actions.

crspeller:
## Summary
Prevents extra JSON from being stored in the DB for channel actions.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-54130

### DIFF
--- a/server/safemapstructure/safemapstructure.go
+++ b/server/safemapstructure/safemapstructure.go
@@ -1,0 +1,20 @@
+package safemapstructure
+
+import (
+	"github.com/mitchellh/mapstructure"
+)
+
+func Decode(input interface{}, output interface{}) error {
+	config := &mapstructure.DecoderConfig{
+		Metadata:  nil,
+		Result:    output,
+		MatchName: func(a string, b string) bool { return a == b }, // Only match exactly
+	}
+
+	decoder, err := mapstructure.NewDecoder(config)
+	if err != nil {
+		return err
+	}
+
+	return decoder.Decode(input)
+}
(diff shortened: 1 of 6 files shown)


server/safemapstructure/safemapstructure_test.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	input := map[string]interface{}{
+		"tEst":       "incorrect",
+		"Test":       "incorrect",
+		"Other_test": "incorrect",
+		"other_tEst": "incorrect",
+	}
+
+	// Do it a bunch of times since map order is randomized
```
jupenur, line 22: We can just do this once because the input contains no zero values.

server/safemapstructure/safemapstructure.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/mitchellh/mapstructure"
+)
+
+func Decode(input interface{}, output interface{}) error {
+	config := &mapstructure.DecoderConfig{
+		Metadata:  nil,
+		Result:    output,
+		MatchName: func(a string, b string) bool { return a == b }, // Only match exactly
```
jupenur, line 11: Exact matches are handled in a different code path so `MatchName` never gets called with equal values of `a` and `b`. Which means this could also be written as:

```suggestion
		MatchName: func(string, string) bool { return false }, // Only match exactly
```

0/5 on whether that's better or worse than what you've got here.

jupenur (approved):

server/safemapstructure/safemapstructure.go
crspeller: I did notice that. I just thought that this was more clear.

sbishel (approved):

lieut-data (approved):

PR #1895, opened 2024-03-06
MM-56402: Close the store on Deactivate

agnivade:
Not closing the store would keep DB connections open,
allowing them to remain in the server connection pool.

https://mattermost.atlassian.net/browse/MM-56402

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -506,6 +506,11 @@ func (p *Plugin) GetTopicMetadataByIds(c *plugin.Context, topicType string, topi
 	return topicsMetadata, nil
 }
 
+func (p *Plugin) OnDeactivate() error {
+	logrus.Info("Shutting down store..")
+	return p.pluginAPI.Store.Close()
+}
+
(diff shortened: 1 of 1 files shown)


agnivade:
While I look for a server side solution. There's no harm in doing the due diligence from plugin side :smile: 

@lieut-data - AFAIK, only playbooks, focalboard were the early plugins using the store API. The new ones are probably msteams and the metrics plugin.

@isacikgoz - Can we ensure to add the same code in the PR for the metrics plugin as well? Thanks.

jespino (approved):
LGTM

lieut-data (approved):

agnivade:
@lieut-data - How do you want to test this? I tested locally of course, but I don't have a reader-writer setup.

lieut-data:
@agnivade, from a Playbooks perspective, I think the e2e test coverage ought to be good enough. And I think @jespino has confirmed the semantically same change for MS Teams resolves the issue there too.

PR #1900, opened 2024-03-26
Fixing slash commands on runs

crspeller:
## Summary
Fixing slash commands on runs. Note this doesn't fix the UI around this feature. Followup ticket: https://mattermost.atlassian.net/browse/MM-57509

## Ticket Link
https://mattermost.atlassian.net/browse/MM-56343


lieut-data (approved):

esarafianou (approved):
looks good! Let's only add a test for a non channel-member running a slash command

PR #1908, opened 2024-05-29
[MM-47629] Don't crash when a run channel has been deleted

JulienTant:
## Summary
This PR allows for a run page to be visited even if the associated channel has been removed.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-47629

## Checklist
~~- [ ] Telemetry updated~~
- [x] e2e tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -1266,11 +1266,12 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
 	// Get main channel details
 	channel, err := s.pluginAPI.Channel.Get(playbookRun.ChannelID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve channel id '%s'", playbookRun.ChannelID)
+		s.pluginAPI.Log.Warn("failed to retrieve channel id", "channel_id", playbookRun.ChannelID)
+		channel = &model.Channel{}
 	}
-	team, err := s.pluginAPI.Team.Get(channel.TeamId)
+	team, err := s.pluginAPI.Team.Get(playbookRun.TeamID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve team id '%s'", channel.TeamId)
+		return nil, errors.Wrapf(err, "failed to retrieve team id '%s'", playbookRun.TeamID)
 	}
 
 	numParticipants, err := s.store.GetHistoricalPlaybookRunParticipantsCount(playbookRun.ChannelID)
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -1267,7 +1267,6 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
 	channel, err := s.pluginAPI.Channel.Get(playbookRun.ChannelID)
 	if err != nil {
 		s.pluginAPI.Log.Warn("failed to retrieve channel id", "channel_id", playbookRun.ChannelID)
-		channel = &model.Channel{}
 	}
 	team, err := s.pluginAPI.Team.Get(playbookRun.TeamID)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
```diff
@@ -144,6 +144,8 @@ const PlaybookRunDetails = () => {
         }
     }, [urlHash]);
 
+    const channelDeleted = Boolean(channel?.delete_at) || channelResult.isErrorCode(404);
```
JulienTant, line 147: covers both soft and hard delete.

server/app/playbook_run_service.go
```diff
@@ -1266,11 +1266,12 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
 	// Get main channel details
 	channel, err := s.pluginAPI.Channel.Get(playbookRun.ChannelID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve channel id '%s'", playbookRun.ChannelID)
+		s.pluginAPI.Log.Warn("failed to retrieve channel id", "channel_id", playbookRun.ChannelID)
+		channel = &model.Channel{}
 	}
-	team, err := s.pluginAPI.Team.Get(channel.TeamId)
+	team, err := s.pluginAPI.Team.Get(playbookRun.TeamID)
```
JulienTant, line 1272: use the run teamID because the channel might not exist

server/app/playbook_run_service.go
```diff
@@ -1266,11 +1266,12 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
 	// Get main channel details
 	channel, err := s.pluginAPI.Channel.Get(playbookRun.ChannelID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve channel id '%s'", playbookRun.ChannelID)
+		s.pluginAPI.Log.Warn("failed to retrieve channel id", "channel_id", playbookRun.ChannelID)
```
JulienTant, line 1269: I guess this is the main core change: considering that a missing channel is not a big deal and we can continue the process without it.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_info_overview.tsx
```diff
@@ -197,24 +198,13 @@ const RHSInfoOverview = ({run, role, channel, runMetadata, followState, editable
...
-                            size={14}
-                            color={'var(--button-bg)'}
-                        />
-                    </ItemLink>
-                </> : <ItemDisabledContent>
-                    {role === Role.Participant ? <RequestJoinButton onClick={showRequestJoinConfirm}>{formatMessage({defaultMessage: 'Request to Join'})}</RequestJoinButton> : null}
-                    <LockOutlineIcon size={20}/> {formatMessage({defaultMessage: 'Private'})}
-                </ItemDisabledContent>
-                }
+                <ChannelRow
```
JulienTant, line 201: I moved the channel row to it's own component because we started to have a lot of conditions.

server/app/playbook_run_service.go
```diff
@@ -1266,11 +1266,12 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
...
+		channel = &model.Channel{}
```
lieut-data, line 1270: If possible, I'd propose we leave `channel` to be `nil` -- this is more reflective of the underlying state, and might even help us track down other regressions.

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
lieut-data: Might we infer that the channel doesn't exist by virtue of `channel` being `undefined`?

lieut-data (approved):
Looks good! Just two thoughts, but not blocking.

server/app/playbook_run_service.go
JulienTant: yes!

JulienTant:
> Might we infer that the channel doesn't exist by virtue of channel being undefined?

@lieut-data Unless it's been hard deleted, the channel is returned with `deleted_at` in this case. Also if I test for the channel neing undefefined to detect hard deletes, I would probably have to add `and the request is not pending`

webapp/src/components/backstage/playbook_runs/playbook_run/playbook_run.tsx
lieut-data: Ah, good point! Forgot how this worked ;p

calebroseland (approved):
LGTM 👍

PR #1918, opened 2024-06-13
[MM-58377] Make sure we have an enterprise license when activating the plugin

JulienTant:
## Summary
Add license check when starting playbooks if the version is 2.0+

Also see the Mattermost PR: https://github.com/mattermost/mattermost/pull/27335

Note: there is another ticket for cleaning up all the license checks for pro and enterprise:  https://mattermost.atlassian.net/browse/MM-58700

## Ticket Link
https://mattermost.atlassian.net/browse/MM-58377

### DIFF
--- a/server/api_playbooks_test.go
+++ b/server/api_playbooks_test.go
@@ -23,17 +23,7 @@ func TestPlaybooks(t *testing.T) {
 	e.CreateClients()
 	e.CreateBasicServer()
 
-	t.Run("unlicenced servers can't create a private playbook", func(t *testing.T) {
-		id, err := e.PlaybooksClient.Playbooks.Create(context.Background(), client.PlaybookCreateOptions{
-			Title:  "test1",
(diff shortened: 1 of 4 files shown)


server/api_playbooks_test.go
```diff
@@ -98,6 +65,7 @@ func TestPlaybooks(t *testing.T) {
 			Title:  "test6 - to be archived",
 			TeamID: e.BasicTeam.Id,
 		})
+		t.Log("err is", err)
```
mgdelacroix, line 68: nit: we can probably remove this
```suggestion
```

mgdelacroix (approved):

server/api_playbooks_test.go
JulienTant: Ooopsie! Thanks :D

calebroseland (approved):

PR #1939, opened 2024-09-05
MM-60420: treat bots as participants

lieut-data:
## Summary
In support of the MS Teams Tab App, and to unlock other security workflows involving bot participation in runs, let's allow bots to be participants of playbook runs.

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-60420

### DIFF
--- a/server/sqlstore/playbook_run.go
+++ b/server/sqlstore/playbook_run.go
@@ -156,13 +156,15 @@ func NewPlaybookRunStore(pluginAPI PluginAPIClient, sqlStore *SQLStore) app.Play
 	// the user is not a member of the channel they won't have permissions to get the user list
 	participantsCol := `
         COALESCE(
-			(SELECT string_agg(rp.UserId, ',')
+			(SELECT string_agg(x.UserId, ',') FROM (
+				SELECT rp.UserId
 				FROM IR_Incident as i2
 					JOIN IR_Run_Participants as rp on rp.IncidentID = i2.ID
+					LEFT JOIN Bots b ON (b.UserId = rp.UserId)
 				WHERE i2.Id = i.Id
 				AND rp.IsParticipant = true
-				AND rp.UserId NOT IN (SELECT UserId FROM Bots)
-			), ''
+				ORDER BY (CASE WHEN b.UserId IS NULL THEN 0 ELSE 1 END), rp.UserId
+			) x), ''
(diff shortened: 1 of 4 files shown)


server/sqlstore/playbook_run.go
```diff
@@ -156,23 +156,26 @@ func NewPlaybookRunStore(pluginAPI PluginAPIClient, sqlStore *SQLStore) app.Play
...
+			(SELECT string_agg(x.UserId, ',') FROM (
+				SELECT rp.UserId
 				FROM IR_Incident as i2
 					JOIN IR_Run_Participants as rp on rp.IncidentID = i2.ID
+					LEFT JOIN Bots b ON (b.UserId = rp.UserId)
 				WHERE i2.Id = i.Id
 				AND rp.IsParticipant = true
-				AND rp.UserId NOT IN (SELECT UserId FROM Bots)
-			), ''
+				ORDER BY (CASE WHEN b.UserId IS NULL THEN 0 ELSE 1 END), rp.UserId
```
lieut-data, line 166: We still JOIN so we can order bots after human participants.

server/sqlstore/playbook_run.go
```diff
@@ -891,13 +894,12 @@ func (s *playbookRunStore) GetPlaybookRunIDsForChannel(channelID string) ([]stri
 }
 
 // GetHistoricalPlaybookRunParticipantsCount returns the count of all members of a playbook run's channel
-// since the beginning of the playbook run, excluding bots.
+// since the beginning of the playbook run.
```
lieut-data, line 897: Note that this is effectively "retroactive."

sbishel (approved):

esethna (approved):

JulienTant (approved):
LGTM @lieut-data; I was wondering what was the reasoning behind preventing bots to become participants in the first place?

lieut-data:
> LGTM @lieut-data; I was wondering what was the reasoning behind preventing bots to become participants in the first place?

The driving impetus in the past was to prevent the `@playbooks` bot from always being a participant in the run because the bot would join the channel and channel membership == run participation. But when we split those apart, we never revisited this decision -- new runs since that change (circa late 2022) won't have `@playbooks` as a participant despite the bot joining the channel.

PR #1942, opened 2024-09-13
Support the MS Teams Tab App

lieut-data:
## Summary
This is the "backend" for https://github.com/mattermost/mattermost-teams-tab, allowing colleagues in Teams to connect to a Mattermost server, proving their membership in a configured tenant via the [Microsoft ID token](https://learn.microsoft.com/en-us/entra/identity-platform/id-tokens), and getting read-only access to playbook runs that have explicitly invited the `@msteams` bot.

## Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-60362
Fixes: https://mattermost.atlassian.net/browse/MM-60363
Fixes: https://mattermost.atlassian.net/browse/MM-60366
Fixes: https://mattermost.atlassian.net/browse/MM-60367
Fixes: https://mattermost.atlassian.net/browse/MM-60368

## Checklist
- [x] Unit tests updated


client/actions.go
```diff
@@ -18,7 +18,7 @@ type ActionsService struct {
 // Create an action. Returns the id of the newly created action.
 func (s *ActionsService) Create(ctx context.Context, channelID string, opts ChannelActionCreateOptions) (string, error) {
 	actionURL := fmt.Sprintf("actions/channels/%s", channelID)
-	req, err := s.client.newRequest(http.MethodPost, actionURL, opts)
+	req, err := s.client.newAPIRequest(http.MethodPost, actionURL, opts)
```
lieut-data, line 21: This is the first in a series of `s/newRequest/newAPIRequest/` to allow the `client` package to be used with non-`/api` endpoints.

client/tabapp.go
```diff
@@ -0,0 +1,65 @@
...
+	client *Client
+}
+
+type TabAppGetRunsOptions struct {
+	Page    int `json:"page"`
+	PerPage int `json:"per_page"`
+}
+
+func (s *TabAppService) GetRuns(ctx context.Context, token string, options TabAppGetRunsOptions) (*TabAppResults, error) {
+	url := fmt.Sprintf("plugins/%s/tabapp/runs", manifestID)
```
lieut-data, line 43: We don't use the `/api` endpoint directly here.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/mattermost/mattermost-plugin-playbooks
 
-go 1.19
+go 1.21
```
lieut-data, line 3: This came for free -- trusting to e2e tests for coverage here.

plugin.json
```diff
@@ -1,33 +1,47 @@
...
-        "footer": "",
-        "settings": [
-        {
-            "key": "EnableExperimentalFeatures",
-            "type": "bool",
-            "display_name": "Enable Experimental Features:",
-            "help_text": "Enable experimental features that come with in-progress UI, bugs, and cool stuff."
-        }
-        ]
+  "id": "playbooks",
```
lieut-data, line 2: Noisy because of some whitespace cleanup.

server/api/api.go
```diff
@@ -41,8 +41,8 @@ func NewHandler(pluginAPI *pluginapi.Client, config config.Service) *Handler {
 	}
 
 	root := mux.NewRouter()
+	root.Use(LogRequest)
```
lieut-data, line 44: Moving this to the root to log all requests, not just those on `api/`.

server/api/permutation_test.go
```diff
@@ -0,0 +1,91 @@
+package api
```
lieut-data, line 1: Copied verbatim from the MS Teams plugin to help with testing permutations -- will lift this to a common place soonish.

server/api/tabapp.go
```diff
@@ -0,0 +1,426 @@
...
+	pluginapi "github.com/mattermost/mattermost-plugin-api"
+
+	"github.com/mattermost/mattermost-plugin-playbooks/server/app"
+	"github.com/mattermost/mattermost-plugin-playbooks/server/config"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	MicrosoftTeamsAppDomain = "https://playbooks.integrations.mattermost.com"
+	ExpectedAudience        = "api://playbooks.integrations.mattermost.com/8f7d5beb-ed24-4d95-aa31-c26298d5a982"
```
lieut-data, line 21: Both hard-coded to match our production deploy.

server/api/tabapp_test.go
```diff
@@ -0,0 +1,502 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	keyID           = "my-key-id"
+	keyWithoutAlgID = "SjE4tvzAwAoo6GB32-g1QAdgIck"
+)
+
+// TestValidateToken was inspired by https://github.com/MicahParks/keyfunc/blob/main/keyfunc_test.go.
+func TestValidateToken(t *testing.T) {
```
lieut-data, line 27: These unit tests focus on the validation of the token -- I've added top-level tests that focus on the tabapp results themselves.

server/api/tabapp_test.go
```diff
@@ -0,0 +1,502 @@
...
+			_, priv, jwtKeyFunc := makeKeySet(t)
+			r := makeRequest(t, newToken(t, priv, nil))
+			expectedTenantIDs := []string{}
+
+			validationErr := validateToken(jwtKeyFunc, r, expectedTenantIDs, params.EnableDeveloper)
+			require.NotNil(t, validationErr)
+			assert.Equal(t, http.StatusUnauthorized, validationErr.StatusCode)
+		})
+
+		t.Run("hmac key pretending to be rsa", func(t *testing.T) {
```
lieut-data, line 183: cc @esarafianou

server/api_tabapp_test.go
```diff
@@ -0,0 +1,234 @@
...
+			"Origin": "example.com",
+		})
+		require.NoError(t, err)
+		require.Equal(t, http.StatusNoContent, response.StatusCode)
+		assertCORS(t, api.MicrosoftTeamsAppDomain, response)
+	})
+
+	t.Run("CORS headers, mis-matched Origin header, developer mode", func(t *testing.T) {
+		setTabApp(t, true)
+		setDeveloperMode(t, true)
```
lieut-data, line 113: We basically only test with developer mode, as we can't spoof a token for the production configuration.

lieut-data:
Swapping @sbishel for @crspeller given the dual overlap of Playbook and the "Fast Futures" nature of this initiative.

.github/workflows/ci.yaml
```diff
@@ -11,7 +11,7 @@ defaults:
 
 env:
   TERM: xterm
-  GO_VERSION: 1.19.5
+  GO_VERSION: 1.21.13
```
esarafianou, line 14: Since we're updating it here, can we have it match the `mattermost` golang version: https://github.com/mattermost/mattermost/blob/master/server/go.mod#L3-L5 ?

go.mod
esarafianou: Same comment as above, but only if no additional changes are needed. otherwise bumping to 1.22 can happen independently.

esarafianou (commented):
Thank you Jesse! One comment around Full name and two optional comments

esarafianou (changes requested):
Marking this as requested changes for the Full name one

server/api/tabapp.go
```diff
@@ -0,0 +1,426 @@
...
+		// instead.
+		user, err := h.pluginAPI.User.Get(statusPost.UserID)
+		if err != nil {
+			logrus.WithField("user_id", statusPost.UserID).WithError(err).Warn("Failed to get status post user")
+			continue
+		}
+		users[statusPost.UserID] = limitedUser{
+			UserID:    user.Id,
+			FirstName: user.FirstName,
+			LastName:  user.LastName,
```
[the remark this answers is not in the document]
lieut-data: Gah, thank you @esarafianou -- this was the last change I addressed and I should have given this a bit more "brain soak" time. Will address!

esarafianou (approved):

client/tabapp.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	req.Header.Add("Authorization", token)
+
+	tabAppResults := new(TabAppResults)
+	resp, err := s.client.do(ctx, req, tabAppResults)
+	if err != nil {
+		return nil, err
+	}
+	resp.Body.Close()
```
JulienTant, line 62: nit because unsure: errcheck?

server/tabapp.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	// Create the bot, if needed. This will allow the MS Teams plugin to use the
+	// bot normally as well.
+	if botUserID == "" {
+		bot := &model.Bot{
+			Username:    "msteams",
+			DisplayName: "MS Teams",
+			OwnerId:     "playbooks",
+		}
```
JulienTant, line 91: If the MSTeams plugin is not installed, and as a consequence we don't have the MSTeams bot, what's the advantage of creating a MSTeams bot here instead of using the Playbooks one?

I'm just thinking if the name of the MSTeams bot has to be changed for some reason or we want to update the avatar for the bot, we'll have to remember to do it everywhere 😅

JulienTant (approved):
Sorry it took me a minute to review. Swapping my review into "Approved" as I don't have anything that is a blocker to move forward.

crspeller (changes requested):
I don't see any current security problems, however it seems unlikely that we found them all in this unconventional flow. 
I suspect there are some flows around adding users to public channels that would cause some issues here. Adding users to public channels doesn't usually get that much scrutiny since from as security perspective it doesn't change data access. However adding this magic msteams user is now a dangerous action since in a large company msteams access and MM access might be wildly different.

The feature is also really unintuitive. Why would adding some random MS teams user suddenly publish my run publicly to the whole company?

lieut-data:
> I don't see any current security problems, however it seems unlikely that we found them all in this unconventional flow. 

Thanks, @crspeller! @esarafianou, I know you took a look at the security surface originally: is there anything else we should prioritize given some of this additional context?

> I suspect there are some flows around adding users to public channels that would cause some issues here. Adding users to public channels doesn't usually get that much scrutiny since from as security perspective it doesn't change data access. However adding this magic msteams user is now a dangerous action since in a large company msteams access and MM access might be wildly different.

At the moment the user has to be added as a run participant, and not just a channel member. We do have automation to make run participants into channel members, but I don't believe it happens in the reverse anymore. Can you confirm?

> The feature is also really unintuitive. Why would adding some random MS teams user suddenly publish my run publicly to the whole company?

I agree re: unintuitive. That said, this is much more of a "fast futures" experiment. Given it's behind an explicit plugin configuration as such, can do we anything else to build confidence towards shipping this experiment? I absolutely see us firming up the sharing flow if there's adoption.

crspeller (approved):
> At the moment the user has to be added as a run participant, and not just a channel member.

Right! That makes sense.

Alright let's consider this to be a FF style experiment and see if we get any signal.

PR #1973, opened 2024-12-18
Validate props in retrospective post (aider assisted)

esarafianou:
## Summary
Validate props in the retrospective post

## Ticket Link
https://mattermost.atlassian.net/browse/MM-61164

## Checklist

- ~~[ ] Telemetry updated~~
- ~~[ ] Unit tests updated~~

### DIFF
--- a/webapp/src/components/retrospective_post.tsx
+++ b/webapp/src/components/retrospective_post.tsx
@@ -10,6 +10,13 @@ import {Metric, MetricType} from 'src/types/playbook';
 
 import {RunMetricData} from 'src/types/playbook_run';
 
+import {
+    isArrayOf,
+    isMetric,
+    isMetricData,
+    safeJSONParse,
+} from 'src/utils';
+
 import {ClockOutline, DollarSign, PoundSign} from './backstage/playbook_edit/styles';
 import {metricToString} from './backstage/playbook_edit/metrics/shared';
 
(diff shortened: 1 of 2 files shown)


jespino (approved):
LGTM

webapp/src/components/retrospective_post.tsx
```diff
@@ -33,8 +33,8 @@ export const RetrospectivePost = (props: Props) => {
 
     const mdText = (text: string) => messageHtmlToComponent(formatText(text, markdownOptions), true, messageHtmlToComponentOptions);
 
-    const metricsConfigs: Array<Metric> = JSON.parse(props.post.props.metricsConfigs);
-    const metricsData: Array<RunMetricData> = JSON.parse(props.post.props.metricsData);
+    const metricsConfigs: Array<Metric> = props.post.props?.metricsConfigs ? JSON.parse(props.post.props.metricsConfigs) : [];
+    const metricsData: Array<RunMetricData> = props.post.props?.metricsData ? JSON.parse(props.post.props.metricsData) : [];
```
larkox, line 37: This validation is probably not enough. The following inputs should break this:
- A non JSON parseable string value like `"{"`
- A non string value like `{}`
- A non list JSON string like = `"{}"`

If we want to be 100% safe, we would need to add some try catch (or create a "safeJSONParse" function that does the try catch for us), and some type guard.

```
const parsedMetricsConfig = safeJSONParse(props.post.props?.metricsConfigs);
const parsedMetricsData = safeJSONParse(props.post.props?.metricsData);

const metricsConfigs = isArrayOf(parsedMetricsConfig, isMetric) ? parsedMetricsConfig : [];
const metricsData = isArrayOf(parsedMetricsData, metricsData) ? parsedMetricsData : [];
```

larkox (commented):
Sorry by the drive by review, but since I handled something similar not long ago, I thought I would jump in.

webapp/src/components/retrospective_post.tsx
jespino: good catch

webapp/src/utils.ts
```diff
@@ -132,3 +132,40 @@ export function runCallsSlashCommand(command: string, channelId: string, teamId:
         },
     }, window.origin);
 }
+
+export function safeJSONParse<T>(value: unknown): T | null {
```
larkox, line 136: Nitpick: I like the semantics of `undefined` better than `null`, but it is just about taste.

webapp/src/utils.ts
```diff
@@ -132,3 +132,40 @@ export function runCallsSlashCommand(command: string, channelId: string, teamId:
...
+    if (!value || typeof value !== 'string') {
```
larkox, line 137: I think this would be enough:

```suggestion
    if (typeof value !== 'string') {
```

We can let the case of empty string being handled by `JSON.parse`.

larkox (approved):
LGTM, just a couple of nitpicks you can ignore.

jespino (approved):
LGTM

crspeller (approved):

PR #1984, opened 2025-01-27
Update to use the new version of mattermost-redux

hmhealey:
## Summary
As part of trying to start publishing mattermost-redux again, I've used Playbooks as a test bed to ensure that the published package actually works since it's the first time that we've published it from a monorepo and because I'm trying to do things "the right way" versus the old mattermost-redux package which was rather clunky and required a lot of workarounds.

The commits for this can be roughly broken down into a few categories:
1. Replacing the `mattermost-webapp` dependency with `mattermost-redux` or `@mattermost/types`
2. Updating the import paths to `mattermost-redux` and removing the overrides to make the previous ones work (except for Jest which I fought with for hours and gave up on) 
3. Updating some code to add null checks or rename things to support the latest version of `mattermost-redux`
4. Adding some dependencies to Playbooks which were implicitly available as part of the web app
5. Finally, updating Node and Babel to make the package work and because it was on an old version and I accidentally did a lot of this work on Node 20 before realizing that Playbooks used an older one

This is a draft for now since it's using a prerelease version of `mattermost-redux` and the `@mattermost` packages.

Web app PR: https://github.com/mattermost/mattermost/pull/30020


NOTICE.txt
lieut-data, line 1: Propose either omitting the changes to this file and letting the automation submit a PR to do so, or explicitly running https://github.com/mattermost/notice-file-generator to effect all required changes. (If the files don't perfectly match that output, we'll get an automated PR clobbering these changes anyway.)

lieut-data (approved):
This looks great, @hmhealey! Thank you for the heavy lifting :)

cc @esarafianou, as this will resolve a number of the warning re: the old webapp dependency.

esarafianou:
> cc.  @esarafianou, as this will resolve a number of the warning re: the old webapp dependency.

This is awesome! 😍

webapp/package.json
```diff
@@ -1,31 +1,34 @@
 {
   "private": true,
   "engines": {
-    "npm": ">=7.24.0",
-    "node": ">=16.10.0"
+    "node": ">=18.10.0",
+    "npm": "^9.0.0 || ^10.0.0"
```
calebroseland, line 5: We update `.nvm` as well

calebroseland (approved):
Awesome! A few lines possibly slipped find-replace, but aside from that this is looking great!

webapp/src/components/channel_actions_modal.tsx
```diff
@@ -5,7 +5,7 @@ import React, {useCallback, useEffect, useState} from 'react';
 import {useDispatch, useSelector} from 'react-redux';
 import {useIntl} from 'react-intl';
 
-import {getCurrentChannelId} from 'mattermost-webapp/packages/mattermost-redux/src/selectors/entities/common';
+import {getCurrentChannelId} from 'mattermost-redux/selectors/entities/common';
```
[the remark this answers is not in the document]
hmhealey: Those should be fixed now. They're from a PR which came after I started working on this, and I missed them when rebasing on top of that

hmhealey:
@calebroseland FYI, I shipped the 10.6.0 version of mattermost-redux yesterday if you want to upgrade to that.

I'd do that myself but I don't want to interfere with what you're trying

.github/actions/e2e-test/action.yaml
```diff
@@ -112,8 +112,8 @@ runs:
       if: always()
       uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
       with:
-        name: cypress-test-results
+        name: cypress-test-results-${{ matrix.runId }}
         path: |
           e2e-tests/results/junit
+          e2e-tests/tests/screenshots
```
calebroseland, line 118: unique artifacts for e2e results and include screenshots

.github/workflows/ci.yaml
```diff
@@ -257,7 +257,7 @@ jobs:
         ports:
           - 9200:9200
       mattermost-server:
-        image: mattermost/mattermost-enterprise-edition:master
+        image: mattermostdevelopment/mattermost-enterprise-edition:master
```
calebroseland, line 260: Upgrade mattermost to current/master, thanks @lieut-data! (https://github.com/mattermost/mattermost-plugin-playbooks/pull/1985/commits/512149fe9b21f7022aebd23b6fa273c822695f26)

lieut-data: Sorry I didn't get that PR merged with these changes earlier!!

PR #1994, opened 2025-03-08
MM-62690: Channel Membership gating in Playbooks 

catalintomai:
## Summary
See MM-59287, MM-61195, MM-62691 - all in one PR since they overlap/are inter-dependent.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-59287
https://mattermost.atlassian.net/browse/MM-61195
https://mattermost.atlassian.net/browse/MM-62691


server/app/playbook_run_service.go
```diff
@@ -1287,10 +1282,25 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
...
-		NumParticipants: numParticipants,
-		Followers:       followers,
+		TeamName:  team.Name,
+		Followers: followers,
 	}
+
+	// Return early if user doesn't have channel access
+	if !hasChannelAccess {
+		return metadata, nil
+	}
```
sbishel, line 1292: Shouldn't this be checked right at the top of the function. Do we need to return anything?

webapp/i18n/nb_NO.json
```diff
@@ -144,5 +144,93 @@
...
+    "zWgbGg": "I dag",
+    "yqpcOa": "Bruk",
+    "z3B83t": "Søk etter en Playbook",
+    "zELxbG": "Lagrede meldinger",
+    "zINlao": "Eier",
+    "zSOvI0": "Filtre",
+    "ypIsVG": "Gjenopprett oppgave",
+    "zz6ObK": "Gjenopprett",
+    "zxj2Gh": "Sist oppdatert {time}",
+    "zx0myy": "Deltakere"
```
sbishel, line 235: Weird...is this supposed to be in here?

sbishel (approved):

lieut-data (changes requested):
Thanks, Catalin! There's a few cases here where I'm wondering if we can relax the checks, and also a general request to fail more aggressively vs logging a warning and just dropping expected semantics.

server/app/playbook_run_service.go
catalintomai: In the https://mattermost.atlassian.net/browse/MM-61195 ticket, they only complain about divulging "`channel_name`  , `channel_display_name`  
`num_participants`", so I kept these as what we return "by default".
It's up to us to draw the line between what needs permissions and what doesn't. I drew it according to the ticket right now =).

server/app/playbook_run_service.go
```diff
@@ -1287,10 +1282,25 @@ func (s *PlaybookRunServiceImpl) GetPlaybookRunMetadata(playbookRunID string) (*
...
+
+	// Only set NumParticipants if user has channel access
+	metadata.NumParticipants = numParticipants
```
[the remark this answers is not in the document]
catalintomai: see comment above - https://github.com/mattermost/mattermost-plugin-playbooks/pull/1994/files#r2003367744. Re-reading the ticket, there might be a misunderstanding - they consider num_participants "Private channel data". If it is a run property I'm ok to add it to the existing set that we return permission-independently: team name and followers.

server/api/graphql_run.go
```diff
@@ -181,7 +182,11 @@ func (r *RunResolver) Followers(ctx context.Context) ([]string, error) {
 		return nil, err
 	}
 
-	metadata, err := c.playbookRunService.GetPlaybookRunMetadata(r.ID)
+	userID := c.r.Header.Get("Mattermost-User-ID")
+
+	// Check if user has permission to view the channel
+	hasChannelAccess := c.pluginAPI.User.HasPermissionToChannel(userID, r.ChannelID, model.PermissionReadChannel)
+	metadata, err := c.playbookRunService.GetPlaybookRunMetadata(r.ID, hasChannelAccess)
```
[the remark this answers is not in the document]
catalintomai: In GetPlaybookRunMetadata we are returning the followers, it is not gated. The check is for other properties.

webapp/i18n/nb_NO.json
catalintomai: yeah, don't know where it comes from...Hope it disappears by the time this gets checked-in.. = ).

server/app/playbook_run_service.go
```diff
@@ -3051,6 +3060,29 @@ func (s *PlaybookRunServiceImpl) leaveActions(playbookRun *PlaybookRun, userID s
...
+		permission = model.PermissionManagePublicChannelMembers
+	}
+
+	if !s.pluginAPI.User.HasPermissionToChannel(requesterID, channel.Id, permission) {
+		logrus.WithFields(logrus.Fields{
+			"user_id":    requesterID,
+			"channel_id": channel.Id,
+		}).Warn("leaveActions: user does not have permission to manage channel members")
+		return
+	}
```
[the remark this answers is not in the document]
catalintomai: as an observation, the "leaveActions" change was not called out explicitly in the security tickets, it just made sense to me, if we do the checks for the "participateActions" one, so if we make a change here it might make sense to do it for "participateActions" as well.

server/app/playbook_run_service.go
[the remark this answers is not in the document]
lieut-data: I'm inclined to treat this as a property of the run, with access gated by access to the run alone. cc @DSchalla for any "PM" perspective :)

server/app/playbook_run_service.go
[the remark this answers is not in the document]
catalintomai: @lieut-data , I double-check the existing codebase and couldn't find a precedent for the whole request failure off a permission check. I hear you about the "subtle breakage", however, this looks to be a bit of a convoluted scenario, off the frequent/critical path so maybe the blast radius is small where we don't have to depart from the curent approach?

server/app/playbook_run_service.go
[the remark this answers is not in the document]
lieut-data: Sounds good! Thanks :)

webapp/i18n/nb_NO.json
catalintomai: Healed itself..

lieut-data (approved):

lindy65 (approved):
Thanks @catalintomai - I'll QA-defer on this PR as it's not possible to test Playbooks (plugins) on spinwicks

enzowritescode (approved):

.github/actions/test-with-db/action.yaml
```diff
@@ -58,7 +58,7 @@ runs:
       if: always()
       uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
       with:
-        name: test-results-${{ github.workflow }}
+        name: test-results-${{ github.workflow }}-${{ github.job }}
```
catalintomai, line 61: @toninis - does this seem ok to you? Thanks.

calebroseland: I've got a similar fix along with other CI fixes on this PR https://github.com/mattermost/mattermost-plugin-playbooks/pull/1984/files#diff-61b091761616d618147c87a7ab7d425bb6d42a91a2f132177050dc1c2404de58

There's also other CI image bumps that address the 14+ `CreateCacheEntry` warnings and other log clutter.

It also addresses critical e2e issues—was running on a Dec 2023 version of mattermost, many broken specs were updated, save for a few [flaky ones](https://mattermost.atlassian.net/browse/MM-63692).

e2e-tests/tests/integration/playbooks/channels/post_type_components_spec.js
```diff
@@ -52,76 +52,88 @@ describe('channels > post type components', {testIsolation: true}, () => {
 
     describe('update post (custom_run_update)', () => {
         it('displays in run channel', () => {
-            // # Go to the playbook run channel
-            cy.visit(`/${testTeam.name}/channels/test-run`);
+            // # Check if we have access to the channel before proceeding
```
calebroseland, line 55: Related ticket https://mattermost.atlassian.net/browse/MM-63645. 

I've skipped one of these in https://github.com/mattermost/mattermost-plugin-playbooks/pull/1984—if this spec still works afterwards, let's keep your version and unskip)

PR #1999, opened 2025-03-25
MM-63410- Check all requestor information for playbook retrieval

sbishel:
## Summary
Retrieves requesterInfo from the playbooks handler, rather than just setting the UserId Property. This will ensure that isGuest gets set and handled properly. 
## Ticket Link

Fixes: https://mattermost.atlassian.net/browse/MM-63410

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


server/api/playbook_runs.go
```diff
@@ -603,12 +603,23 @@ func (h *PlaybookRunHandler) getPlaybookRunByChannel(c *Context, w http.Response
 	channelID := vars["channel_id"]
 	userID := r.Header.Get("Mattermost-User-ID")
 
+	// Check if user has permission to view the channel
+	hasChannelAccess := h.pluginAPI.User.HasPermissionToChannel(userID, channelID, model.PermissionReadChannel)
```
sbishel, line 607: This check is not required specifically for this fix. However, my AI generated unit tests did flush this out as an issue.

enzowritescode (approved):

lieut-data (approved):

server/api/playbook_runs.go
```diff
@@ -603,12 +603,23 @@ func (h *PlaybookRunHandler) getPlaybookRunByChannel(c *Context, w http.Response
...
+	if !hasChannelAccess {
+		h.HandleErrorWithCode(w, c.logger, http.StatusNotFound, "Not found",
+			errors.Errorf("playbook run for channel id %s not found", channelID))
```
catalintomai, line 610: should the message be about the permissions instead (also the "NotFound" messge above)?

server/api/playbook_runs.go
catalintomai: I would remove this, if it is not needed for this particular ticket (and maybe open a different ticket, after checking with security that this indeed an issue). I know security signed off, but maybe they didn't check the scenario that this gating cover. cc: @enzowritescode

lieut-data (approved):

catalintomai (approved):

server/api/playbook_runs.go
enzowritescode: I had reviewed it but not tested it (since I haven't tested anything for this PR yet). Just created SEC-7791

enzowritescode (approved):

lindy65 (approved):
Thanks @sbishel - I've QA-deferred to test post merge

PR #2004, opened 2025-04-07
MM-63668: [Playbooks/Mobile] Add support for filtering playbook runs by the the end_at field (retrieve only active runs)

catalintomai:
## Summary
 Implemented include_ended query parameter (defaults to true for backward compatibility)  
  - When set to false, only active runs (with EndAt = 0) are returned  
  - All API, service layer, and database components updated to support the new filter
 
## Ticket Link
https://mattermost.atlassian.net/browse/MM-63668


server/api/graphql_root_run.go
```diff
@@ -49,6 +49,8 @@ func (r *RunRootResolver) Runs(ctx context.Context, args struct {
 	First                   *int32
 	After                   *string
 	Types                   []string
+	// Default true will be applied by the schema
```
larkox, line 52: I am not a big fan of "default to true". In general I prefer to "negate the variable name" (in this case `omitDeleted`), so the default is false. But 0/5 on this.

webapp/src/components/update_post.tsx
```diff
@@ -57,11 +57,15 @@ export const UpdatePost = (props: Props) => {
...
-    });
+
+    // Send telemetry when this component is rendered
+    useEffect(() => {
+        telemetryView(PlaybookRunViewTarget.StatusUpdate, {
+            post_id: props.post.id,
+            channel_type: channel?.type || '', // not always available
+            playbook_run_id: props.post.props.playbookRunId || '',
+        });
+    }, [props.post.id]);
```
larkox, line 68: Again this seemingly unrelated change 🤔

larkox (commented):
This looks great. Thanks!

server/api/graphql_root_run.go
catalintomai: @larkox - fair enough, will try changing it.

larkox (approved):

esarafianou (approved):
LGTM!
Just one comment: We need to update the server/api/api.yaml OpenAPi spec with the new query param.

PR #2006, opened 2025-04-07
MM-Mobile support/ticket 2 since param

catalintomai:
## Summary
1. Added ActivitySince field to PlaybookRunListOptions in client/playbook_run.go for filtering runs updated after a given timestamp  
3. Added parsing of the since parameter in the parsePlaybookRunsFilterOptions function in server/api/playbook_runs.go  
4. Updated API documentation in server/api/api.yaml to include the new parameter  
5. Modified response structure to include a new FinishedIDs field containing minimal information about runs that finished since the provided timestamp  

## Ticket Link
https://mattermost.atlassian.net/browse/MM-63670


server/app/playbook_run.go
```diff
@@ -463,11 +466,18 @@ type TimelineEvent struct {
...
-	PageCount  int           `json:"page_count"`
-	PerPage    int           `json:"per_page"`
-	HasMore    bool          `json:"has_more"`
-	Items      []PlaybookRun `json:"items"`
+	TotalCount  int           `json:"total_count"`
+	PageCount   int           `json:"page_count"`
+	PerPage     int           `json:"per_page"`
+	HasMore     bool          `json:"has_more"`
+	Items       []PlaybookRun `json:"items"`
+	FinishedIDs []FinishedRun `json:"finished_ids,omitempty"` // Only populated when since parameter is used
```
larkox, line 474: I have mixed feelings with this approach of the finished ids. My gut feeling is that we want to keep at least a few finished runs in the local memory (we may want to check while offline how that run in that channel finished).

If we only get the ids, yes, we could just move the local copy to finished, but we may be missing the "latest updates before the end". In general with this pattern what we would do is completely delete it from the local storage.

Getting only the ids would be for the "delete scenario" (which apparently never happens), not so much the run ended scenario. We could consider deleting the ended playbooks, and you can only check the ended playbooks while online, but I expect that to bite us sooner rather than later.

server/sqlstore/playbook_run.go
```diff
@@ -359,6 +359,12 @@ func (s *playbookRunStore) GetPlaybookRuns(requesterInfo app.RequesterInfo, opti
 	queryForResults = queryStartedBetweenTimes(queryForResults, options.StartedGTE, options.StartedLT)
 	queryForTotal = queryStartedBetweenTimes(queryForTotal, options.StartedGTE, options.StartedLT)
 
+	// Filter by UpdateAt for the activity since parameter
+	if options.ActivitySince > 0 {
+		queryForResults = queryForResults.Where(sq.GtOrEq{"i.UpdateAt": options.ActivitySince})
+		queryForTotal = queryForTotal.Where(sq.GtOrEq{"i.UpdateAt": options.ActivitySince})
+	}
```
larkox, line 366: I wonder if this is the only check, we have to make sure the update at of the run gets properly updated whenever the checklist items are updated too. When we show the list of playbooks we have this "progress bar" that tell us the proportion of checks that has been checked. So we need the checklist information up to date when we fetch the list of runs.

larkox (commented):
I am not a big fan right now of the "FinishedRuns" separation, just because we probably should keep them in the local database.

Also a comment about what we consider an updated playbook.

server/app/playbook_run.go
catalintomai: @larkox
"we want to keep at least a few finished runs in the local memory" - makes sense to prob. keeping the "recent" ended ones only on the client. 

"but we may be missing the "latest updates before the end"". - are the "change" structures in https://github.com/mattermost/mattermost-plugin-playbooks/pull/2008/files that capture the info before the actual deletion be enough for this?
"Getting only the ids would be for the "delete scenario" (which apparently never happens), not so much the run ended scenario" - not sure I get that, you can use the ids to show/tag the ended ones and use the deltas/updates received so far to display what changed about them till the end.
Thoughts?

larkox: If I understood correctly, the delta updates are sent through websocket, right? So, if the app is closed, a run is updated, then ended, and we open the app, we will not get that incremental change, right? Since the app is closed and the websocket not connected.

catalintomai: "a run is updated, then ended, and we open the app.." - that is correct. What can we do is maybe get the incremental changes even if the run has ended, and not through sockets, but through API calls - since the last time the run has been viewed on the client.

server/sqlstore/playbook_run.go
catalintomai: As mentioned before, in https://github.com/mattermost/mattermost-plugin-playbooks/pull/2008/files we detect if the checklist/checklist item changes and will update the playbook run's UpdateAt (currently the PlaybookRunUpdate's no the PlaybookRun's). Once we do that, this should be fixed.

catalintomai: The PlaybookRun's UpdateAt field does get updated when changes are made to checklists or checklist items, but this should happen indirectly through the call to UpdatePlaybookRun, which is called from various places.

server/app/playbook_run.go
larkox: Let's try making a scenario to clarify.

- I have a run with name "Run number 1" with update at X
- I see it in my phone
- I close the app
- On a different client I update the name to "Final Run"
- On that different client, I end the run
- I open the app, and check for the changes since X. The server returns an object like: `{finishedRuns: ['runid']}`
- The app marks that run as finished, but on my phone it is still called "Run number 1", because I never received information about whether the run changed names.

The quick solution for this is to just return the whole entity, doesn't matter if it is finished or not.

larkox: Since the full item will be visible in the interface, let's get the full item. Nevertheless, I wonder if @enahum has any other thought on this.

catalintomai: @larkox, as discussed, I removed the FinishedRun/FinishedIds support.

server/api/api.yaml
```diff
@@ -120,7 +128,9 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/PlaybookRunList"
+                oneOf:
+                  - $ref: "#/components/schemas/PlaybookRunList"
+                  - $ref: "#/components/schemas/PlaybookRunListWithFinished"
```
enahum, line 133: I'm confused about this, is it a list of runs that finished or just the ids ?

catalintomai: @enahum , see previous discussion with DanielE. Initially we wanted to provide the client the list of ids for the finished runs **only**, and then we backed from that and we return the runs for all cases. This PlaybookRunListWithFinished should be prob. removed now since we don't return a mixture of runs and ids anymore and the description section updated to reflect the latest.

calebroseland (approved):

PR #2008, opened 2025-04-07
MM-63674: [Playbooks/Mobile] Add support for playbook run incremental update events

catalintomai:
## Summary
Core Infrastructure
  Configuration-Driven Feature Toggle:
  - New EnableIncrementalUpdates configuration flag with backward compatibility
  - Defaults to false for safe rollout, with automatic fallback to full updates when disabled
  - Server-side feature detection in IsIncrementalUpdatesEnabled()
  
  Server-Side Change Detection:
  - DetectChangedFields() function performs comprehensive state comparison between previous and current playbook runs
  - Granular field-level diff generation across scalar fields, string arrays, and complex nested objects
  - Specialized handlers for status posts, timeline events, metrics data, and checklist changes
  - Deletion tracking via checklist_deletes, timeline_event_deletes, status_post_deletes

  WebSocket Event System
  New Incremental Event Type:
  - WEBSOCKET_PLAYBOOK_RUN_UPDATED_INCREMENTAL event for targeted updates
  - PlaybookRunUpdate data structure with timestamp-based conflict resolution
  - Preserves existing WEBSOCKET_PLAYBOOK_RUN_UPDATED events for backward compatibility
  Enhanced Payload Structure:
  - changed_fields: Only modified fields instead of complete objects
  - ChecklistUpdate and ChecklistItemUpdate interfaces for granular checklist changes
  - Timeline event and status post incremental updates with Map-based merging
  - Item reordering support via items_order field tracking

  Frontend Integration
  Update Application Logic:
  - New applyIncrementalUpdate() utility function (341 lines) with idempotent operations
  - Timestamp-based update ordering prevents out-of-order race conditions
  - Map-based merging for timeline events and status posts handles concurrent updates
  - Graceful fallback to fetchPlaybookRun() when local state is missing
  
  Redux Integration:
  - Updated reducers handle WEBSOCKET_PLAYBOOK_RUN_INCREMENTAL_UPDATE_RECEIVED action
  - Dual state synchronization across playbook_runs and myPlaybookRunsByTeam reducers
  - Defensive programming against missing state scenarios with automatic full fetch
  
Note: this includes some of larkox's changes from https://github.com/mattermost/mattermost-plugin-playbooks/pull/2041/files

## Ticket Link
https://mattermost.atlassian.net/browse/MM-63674


larkox (approved):
This looks great!

server/app/playbook_run.go
```diff
@@ -199,6 +199,428 @@ type PlaybookRun struct {
...
+	Fields map[string]interface{} `json:"fields,omitempty"`
+
+	// ItemUpdates contains changes to existing checklist items
+	ItemUpdates []ChecklistItemUpdate `json:"item_updates,omitempty"`
+
+	// ItemDeletes contains IDs of deleted checklist items
+	ItemDeletes []string `json:"item_deletes,omitempty"`
+
+	// ItemInserts contains new checklist items
+	ItemInserts []ChecklistItem `json:"item_inserts,omitempty"`
```
enahum, line 235: these are great, hopefully we can find a way to do this also with the REST API and not only with the WS events

server/config/configuration.go
```diff
@@ -21,6 +21,12 @@ type Configuration struct {
 	EnableTeamsTabApp    bool   `json:"enableteamstabapp"`
 	TeamsTabAppTenantIDs string `json:"teamstabapptenantids"`
 	TeamsTabAppBotUserID string
+
+	// EnableIncrementalUpdates controls whether the server sends incremental WebSocket updates
+	// instead of full playbook run objects. When enabled, the server compares previous and current
+	// states to determine what fields changed and only sends those changes.
+	// This is set to false by default for backward compatibility.
+	EnableIncrementalUpdates bool `json:"enableincrementalupdates"`
```
enahum, line 29: any reason why this is an opt-in setting ? is this setting exposed to the client? (I see the code below for serialize, but not sure if that gets to the client)

my questions are cause it seems that we would need to keep code in the client to handle both scenarios, setting on or off

enahum (commented):
some questions

server/config/configuration.go
catalintomai: ""any reason why this is an opt-in setting ? is this setting exposed to the client?" - @enahum , that is the reason I had a question some time before about the standard way of enabling this for mobile. This is a regular plugin setting, we can do it the way you recommend, or you had it done previously (if any). Need guidance here.

server/app/playbook_run.go
catalintomai: how would that work? We don't keep the history of changes on the server, the mobile client has the previous state, so not sure we can build these for an arbitrary API call. For the WS events we have the run's "before" and immediate "after".

larkox (approved):

catalintomai:
@calebroseland , please review when you get the chance. Thanks.

larkox:
I feared something like that happened :( I was planning to add some logic to check what users added "extra" were not among the channel members, and send the websocket events to those too. But thought it was overkill. It seems that it is not.

I can take a look and apply the changes here.

esarafianou:
Skipping review. I removed myself from the reviewers.

larkox (approved):

lindy65 (approved):
Thanks @catalintomai - tested with the tar.gz you supplied on a server I spun up and didn't find any issues with what we discussed via DM

webapp/src/utils/playbook_run_updates.ts
```diff
@@ -0,0 +1,379 @@
...
+
+    // Handle timeline events specially by merging them with existing events
+    if (changedFields.timeline_events) {
+        updatedRun = applyTimelineUpdates(updatedRun, changedFields.timeline_events);
+    }
+
+    // Apply checklist updates if provided by the server
+    if (changedFields.checklists) {
+        updatedRun = applyChecklistUpdates(updatedRun, changedFields.checklists);
+    }
```
calebroseland, line 66: Jumping over to array apis isn't necessary IMO; I think a destructuring and a rest+spread approach would be simpler.

```suggestion
    const {timeline_events, checklists, ...basicFields} = changedFields;

    // Start with a shallow copy of the run with basic fields updated
    let updatedRun = {...run, ...basicFields};

    // Handle timeline events specially by merging them with existing events
    if (timeline_events) {
        updatedRun = applyTimelineUpdates(updatedRun, timeline_events);
    }

    // Apply checklist updates if provided by the server
    if (checklists) {
        updatedRun = applyChecklistUpdates(updatedRun, checklists);
    }
```

JulienTant (approved):
LGTM Overall, great work!

My only point of concern is the mix of playbooks/runs taks lists with and without ID in the DB

PR #2041, opened 2025-06-26
Add sort order to playbooks and checklists

larkox:
## Summary
I made some changes to add sort order also to the playbooks, to better handle a couple of things related to incremental updates and order in general.

## Ticket Link
NONE

### DIFF
--- a/server/app/playbook.go
+++ b/server/app/playbook.go
@@ -221,6 +221,9 @@ type Checklist struct {
 
 	// UpdateAt is when this checklist was last modified
 	UpdateAt int64 `json:"update_at" export:"-"`
+
+	// Sort order of the checklist items
+	SortOrder []string `json:"sort_order"`
 }
 
 func (c Checklist) GetItems() []ChecklistItemCommon {
@@ -229,6 +232,14 @@ func (c Checklist) GetItems() []ChecklistItemCommon {
 		items[i] = &c.Items[i]
 	}
 	return items
+}
+
+func (c Checklist) GetSortOrder() []string {
+	sortOrder := make([]string, len(c.Items))
+	for i, item := range c.Items {
+		sortOrder[i] = item.ID
+	}
+	return sortOrder
 }
 
 func (c Checklist) Clone() Checklist {
--- a/server/app/playbook_run.go
+++ b/server/app/playbook_run.go
@@ -200,6 +200,17 @@ type PlaybookRun struct {
 	// Type determines a type of a run.
 	// It can be RunTypePlaybook ("playbook") or RunTypeChannelChecklist ("channel")
 	Type string `json:"type"`
+
+	// Sort order of the checklists
+	SortOrder []string `json:"sort_order"`
+}
+
+func (p PlaybookRun) GetSortOrder() []string {
+	sortOrder := make([]string, len(p.Checklists))
+	for i, checklist := range p.Checklists {
+		sortOrder[i] = checklist.ID
+	}
+	return sortOrder
 }
 
 // PlaybookRunUpdate represents an incremental update to a playbook run
@@ -251,6 +262,18 @@ type ChecklistItemUpdate struct {
 
 	// Fields contains the changed fields of the checklist item
 	Fields map[string]interface{} `json:"fields"`
+}
+
+func compareSortOrder(prevSortOrder, currSortOrder []string) bool {
+	if len(prevSortOrder) != len(currSortOrder) {
+		return false
+	}
+	for i, id := range prevSortOrder {
+		if id != currSortOrder[i] {
+			return false
+		}
+	}
+	return true
 }
 
 // DetectChangedFields compares two playbook runs and returns a map of changed fields
@@ -348,6 +371,10 @@ func DetectChangedFields(previous, current *PlaybookRun) map[string]interface{} 
 	}
 	if previous.Type != current.Type {
 		changes["type"] = current.Type
+	}
+
+	if !compareSortOrder(previous.SortOrder, current.SortOrder) {
+		changes["sort_order"] = current.SortOrder
 	}
 
 	// Check array fields where order doesn't matter (unordered sets)
@@ -473,6 +500,11 @@ func GetChecklistUpdates(previous, current []Checklist) []ChecklistUpdate {
 			if prev.Title != checklist.Title {
 				fields["title"] = checklist.Title
 			}
+
+			if !compareSortOrder(prev.SortOrder, checklist.SortOrder) {
+				fields["sort_order"] = checklist.SortOrder
+			}
+
 			update.Fields = fields
 
 			// Get item updates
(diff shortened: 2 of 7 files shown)


server/app/playbook.go
```diff
@@ -221,6 +221,9 @@ type Checklist struct {
 
 	// UpdateAt is when this checklist was last modified
 	UpdateAt int64 `json:"update_at" export:"-"`
+
+	// Sort order of the checklist items
+	SortOrder []string `json:"sort_order"`
```
catalintomai, line 226: Is the "`SortOrder`" naming a bit misleading, since ids have no real ordering meaning? Maybe `OrderByIds` or smth. else?

catalintomai: Also, we seem to propose a lightweight approach to the GetChecklistItemUpdates, which shoudl report specific position changes as opoosed to the binary changed vs. unchanged. Should we remove then the position-tracking approach then?

server/app/playbook_run.go
```diff
@@ -455,24 +486,31 @@ func GetChecklistUpdates(previous, current []Checklist) []ChecklistUpdate {
...
-			UpdatedAt: now,
+			UpdatedAt: checklist.UpdateAt,
 		}
 
 		// Check if checklist exists in previous state
 		if prev, exists := prevMap[checklist.ID]; exists {
 			// Compare fields
 			fields := make(map[string]interface{})
+
+			if prev.UpdateAt != checklist.UpdateAt {
```
catalintomai, line 502: I think ChecklistUpdate.UpdateAt was intended to capture "when this update event was created/processed". If we want to capture the "when the checklist entity was last modified" (which is indeed more useful for clients handling incremental updates.), it'd might make sense to use a different field or handle it differently - or maybe remove ChecklistUpdate.UpdatedAt if we don't make use of it and add a separate field ChecklistUpdate.ChecklistUpdatedAt for the things that 
I^m not necessarily fond of having the "fields["update_at"]" added , just because I see fields as capturing the user-visible content changes and not internal metadata timestamps.

server/app/playbook_run_service.go
```diff
@@ -1132,11 +1113,6 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
 	// Remove pending reminder (if any), even if current reminder was set to "none" (0 minutes)
 	s.RemoveReminder(playbookRunID)
 
-	err = s.resetReminderTimer(playbookRunID)
```
catalintomai, line 1135: This is part of the existing codebase, so we'd need to review with the previous developers.

server/sqlstore/playbook_run.go
```diff
@@ -1538,9 +1544,22 @@ func (s *playbookRunStore) updateParticipating(playbookRunID string, userIDs []s
...
 	}
 
+	if err = s.touchPlaybookRun(playbookRunID); err != nil {
+		return errors.Wrapf(err, "failed to touch playbook run '%s'", playbookRunID)
+	}
+
 	return nil
 }
 
+func (s *playbookRunStore) touchPlaybookRun(playbookRunID string) error {
```
catalintomai, line 1554: Thanks for fixing this bug = )!

server/app/playbook_run.go
larkox: I updated the updateAt to the one from the entity so I can update the updateAt on the client with the "real value". Sure, the difference is going to be probably milliseconds, but still.

Regarding this particular change to add the `updateAt` to the list of changes, I have to admit I don't remember the exact reason why I added this particular change. I think I originally added it to playbooks, and then just kept moving the change through the three entities. For some reason something was not updating, and I thought sending the update at would solve it, but it does not make sense to me right now 🤔 That being said, I don't have a strong feeling about having it here or having it in the update object, so it is not duplicated.

Regarding having sort_order in the changes, it is important (the client needs to know the sort order) and simplifies the logic. Update at has to arrive to the client one way or another, since the client needs to know when it was updated to understand if there are changes between what it retrieves and what it has. As I said in the previous paragraph, 0/5 on whether to have it in the fields/changes or in the update object.

server/app/playbook_run_service.go
larkox: This was sending a websocket event with the playbook, but whatever changed when resetting the reminder, was not changing anything in the playbook itself, but in the metadata, and the websocket event did not have any extra information from the one that is sent a couple of lines after this. So the websocket event was duplicated, creating other problems on the client.

Nevertheless, agree it would be great to have feedback from @calebroseland or @lieut-data .

server/app/playbook.go
catalintomai: Ok, so I will add `ItemsOrder` (instead of `SortOrder`) to the https://github.com/mattermost/mattermost-plugin-playbooks/pull/2008/files for both checklist and playbook run and remove the array position tracking approach.

server/api/playbook_runs.go
```diff
@@ -196,9 +196,9 @@ func (h *PlaybookRunHandler) createPlaybookRunFromPost(c *Context, w http.Respon
 		return
 	}
 
-	h.poster.PublishWebsocketEventToUser(app.PlaybookRunCreatedWSEvent, map[string]interface{}{
+	h.poster.PublishWebsocketEventToChannel(app.PlaybookRunCreatedWSEvent, map[string]interface{}{
```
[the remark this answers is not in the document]
lieut-data: At the time this was introduced, playbook runs always created a new channel, and only the creating user needed to know about it and redirect to the new channel in question. I suspect this was never revisited when we supported creating runs in an existing channel.

server/app/playbook_run_service.go
lieut-data: IIRC, resetting the timer prevents spurious reminders from showing up after the run was already finished.

> So the websocket event was duplicated, creating other problems on the client.

This worries me: why would a duplicate event cause issues?

larkox: > This worries me: why would a duplicate event cause issues?

The issue is mostly related to incremental updates. Also, related in how it is calculated.

Before this is called, some changes has already been done to the entity. So when this was called, it took a new copy from the database (already updated), set the reminder, and trigger the event, making the comparison with the already update entity.

Then, we create a new event, but this new event, the update at is the same as the previous. So if we want to decide whether there are changes based on the update_at, both update_at are the same. So the client would think "oh, I have already done this" and not do it. So changes are lost. Does it make sense?

lieut-data: > So the client would think "oh, I have already done this" and not do it. So changes are lost. Does it make sense?

It parses, but seems surprisingly brittle. How does the client normally recover from these types of lost websocket events?

lieut-data: > Then, we create a new event, but this new event, the update at is the same as the previous. 

Ah, I missed this -- thank you for clarifying, @larkox! Can we touch the `UpdateAt` here instead?

larkox:
Close as this is no longer needed.

PR #2049, opened 2025-07-18
Add FIPS support to build process and update .gitignore

stafot:
## Summary
- Introduced FIPS compliance build options in the Makefile, allowing for the creation of FIPS-compliant plugins using Docker.
- Added new targets for building and bundling FIPS plugins.
- Updated the manifest generation to support outputting to a FIPS-specific directory.
- Modified .gitignore to include FIPS-related directories and build cache.

## Ticket Link
https://mattermost.atlassian.net/browse/CLD-9414

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


.github/workflows/ci.yaml
```diff
@@ -106,10 +107,56 @@ jobs:
         uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0
         with:
           go-version: "${{ env.GO_VERSION }}"
-          cache: false
+          cache: true
 
-      - name: ci/ensure-build-on-all-platforms
-        run: make dist
+      - name: ci/setup-node
```
hmhealey, line 112: Why do we need to set up the web app's dependencies now when we didn't have to before? I'm actually kind of surprised that that wasn't part of this pipeline before

Makefile
```diff
@@ -285,12 +337,63 @@ else
...
+ifneq ($(HAS_SERVER),)
+	mkdir -p dist-fips/$(PLUGIN_ID)/server
+	cp -r server/dist-fips dist-fips/$(PLUGIN_ID)/server/dist
+endif
+ifneq ($(HAS_WEBAPP),)
+	if [ -d webapp/dist ]; then \
+		mkdir -p dist-fips/$(PLUGIN_ID)/webapp && \
+		cp -r webapp/dist dist-fips/$(PLUGIN_ID)/webapp/; \
+	else \
+		echo "Warning: webapp/dist not found, skipping webapp in FIPS bundle"; \
```
hmhealey, line 369: I feel like this should error out since, if the plugin has `HAS_WEBAPP` set but the web app is missing, it's going to result in a broken build

Makefile
```diff
@@ -227,6 +241,44 @@ else
...
+	# Login to Chainguard registry if credentials are available
+	@if [ -n "$(CHAINGUARD_DEV_USERNAME)" ] && [ -n "$(CHAINGUARD_DEV_TOKEN)" ]; then \
+		echo "Logging into Chainguard registry..."; \
+		echo "$(CHAINGUARD_DEV_TOKEN)" | docker login cgr.dev --username "$(CHAINGUARD_DEV_USERNAME)" --password-stdin; \
+	else \
+		echo "Warning: CHAINGUARD_DEV_USERNAME and CHAINGUARD_DEV_TOKEN not set. Using public image if available."; \
+	fi
+	
+	# Build directly in the FIPS container without external script
+	# Create local cache directory for CI/ACT compatibility
```
hmhealey, line 261: For my own understanding, the FIPS build requires us to build the server code in the provided Docker image without running any external scripts, is that right?

.github/actions/e2e-test/action.yaml
```diff
@@ -16,7 +16,7 @@ runs:
       uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0
       with:
         go-version: "${{ env.GO_VERSION }}"
-        cache: false
+        cache: true
```
hmhealey, line 19: Why change this caching behaviour?

Makefile
hmhealey, line 1: Should any of these changes make it back into the starter template so that any new plugins we make will have FIPS available? I get that we probably don't want this turned on for community plugins and such, but it seems like it would help share it among our plugin builds

hmhealey (commented):
I've got a few questions, mostly for my own understanding since I know very little about FIPS compliance

Makefile
```diff
@@ -285,12 +337,63 @@ else
...
 
+## Builds and bundles the FIPS plugin.
+.PHONY: dist-fips
+dist-fips: apply server-fips webapp bundle-fips
+
+## Builds both normal and FIPS distributions.
+.PHONY: dist-all
+dist-all: clean
+	@echo "==> Building both normal and FIPS distributions in parallel..."
+	$(MAKE) dist & $(MAKE) dist-fips & wait
```
agnivade, line 392: Building in parallel will interleave logs from both builds, and make debugging difficult. I'd suggest to make it serial to help debug faster in case something goes wrong.

Makefile
```diff
@@ -11,11 +11,25 @@ DLV_DEBUG_PORT := 2346
...
+
 export GO111MODULE=on
 
 # We need to export GOBIN to allow it to be set
 # for processes spawned from the Makefile
-export GOBIN ?= $(PWD)/bin
+ifeq ($(FIPS_ENABLED),true)
+	export GOBIN ?= /go/bin
+	GO_FIPS ?= docker run --rm -v $(PWD):/plugin -v $(HOME)/.cache:/root/.cache -w /plugin -e GOFLAGS -e GO111MODULE $(FIPS_IMAGE) go
+	GO_BUILD_TAGS_FIPS = fips
```
agnivade, line 27: Where are these GO_FIPS and GO_BUILD_TAGS_FIPS variables used?

Makefile
```diff
@@ -227,6 +241,44 @@ else
...
+		--entrypoint="" \
+		--user root \
+		-v $(PWD):/plugin \
+		-v $(PWD)/.build-cache:/root/.cache \
+		-w /plugin \
+		-e GO_VERSION \
+		$(FIPS_IMAGE) \
+		/bin/sh -c "\
+			echo 'Go already available: ' && go version && \
+			export GO111MODULE=on && \
```
agnivade, line 273: We can remove this line. It's not needed anymore.

Makefile
```diff
@@ -227,6 +241,44 @@ else
...
+			export CGO_ENABLED=1 && \
+			cd /plugin/server && \
+			env GOOS=linux GOARCH=amd64 go build -tags fips -trimpath -buildvcs=false -o dist-fips/plugin-linux-amd64-fips && \
```
agnivade, line 276: What is the purpose of passing `-tags fips`? I don't see it being used in the codebase anywhere.

Makefile
```diff
@@ -285,12 +337,63 @@ else
...
+ifneq ($(wildcard $(ASSETS_DIR)/.),)
+	cp -r $(ASSETS_DIR) dist-fips/$(PLUGIN_ID)/
+endif
+ifneq ($(HAS_PUBLIC),)
+	cp -r public dist-fips/$(PLUGIN_ID)/public/
+endif
+ifneq ($(HAS_SERVER),)
+	mkdir -p dist-fips/$(PLUGIN_ID)/server
+	cp -r server/dist-fips dist-fips/$(PLUGIN_ID)/server/dist
+endif
```
agnivade, line 363: A lot of these lines are duplicated from the `bundle` target. Could we refactor them?

build/manifest/main.go
```diff
@@ -86,10 +86,15 @@ func main() {
...
 	case "dist":
-		if err := distManifest(manifest); err != nil {
+		if err := distManifest(manifest, "dist"); err != nil {
 			panic("failed to write manifest to dist directory: " + err.Error())
 		}
 
+	case "dist-fips":
+		if err := distManifest(manifest, "dist-fips"); err != nil {
+			panic("failed to write manifest to dist-fips directory: " + err.Error())
+		}
```
agnivade, line 96: We could potentially optimize this to `case "dist", "dist-fips":`, and then just use `distManifest(manifest, cmd)`. But just a minor thing. Feel free to ignore.

agnivade (changes requested):

.github/workflows/ci.yaml
hmhealey: I see that up there, but that's in the `lint` job instead of the `dist` one. Since this one sets up its own workspace, it looks like we've previously just linted the web app and that we haven't been compiling it.

From what I can tell, this container must already come with Node by default, so we haven't needed to set up a specific version of Node before, and we haven't had to install the NPM dependencies manually because the Makefile is set up to do that if they're missing

Makefile
```diff
@@ -11,11 +11,25 @@ DLV_DEBUG_PORT := 2346
 DEFAULT_GOOS ?= $(shell go env GOOS)
 DEFAULT_GOARCH ?= $(shell go env GOARCH)
 
+# FIPS Support - similar to mattermost server
+# To build FIPS-compliant plugin: make dist-fips
+# Requires Docker to be installed and running
+FIPS_ENABLED ?= false
+FIPS_IMAGE ?= cgr.dev/mattermost.com/go-msft-fips:1.24.4@sha256:8ab847d56930279a3ea36763277080106406354ec31c5f57f9d7fa787ecadcb2
```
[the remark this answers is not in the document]
esarafianou: @stafot ignore this comment. We discussed we prefer pinning and updated on every release.

Makefile
stafot: feel free to inherit them there!

hmhealey (approved):

agarciamontoro:
I just cherry-picked all commits in [the v2.4.1...v2.4.2 diff](https://github.com/mattermost/mattermost-plugin-playbooks/compare/v2.4.1...v2.4.2) except for [a commit](https://github.com/mattermost/mattermost-plugin-playbooks/commit/97b419be1ce42f35efb42070ec7ef7114bd7f569) that modified the CI pipeline and had conflict with the changes in this PR.

agarciamontoro:
It worked! Thank you, @stafot <

lieut-data:
Superseded by https://github.com/mattermost/mattermost-plugin-playbooks/pull/2139

PR #2054, opened 2025-08-06
MM-64738: fix incorrect requesterid when adding invitees to created run

calebroseland:
## Summary
Fixes an issue where the playbooks bot was the actor adding invited users, rather than the correct user starting the run.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-64738

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated


catalintomai (approved):
@calebroseland , Awesome surgical fix. Would be great if we could add some test coverage as well, invitation during run creation (`TestGraphQLChangeRunParticipants`?)  seems to be an untested code path.

PR #2055, opened 2025-08-07
Implement condition system for property fields

JulienTant:
## Summary

Implements a comprehensive condition system for evaluating property field values in both Go server-side and TypeScript client-side codebases.

### Key Features

- **Cross-language compatibility**: Identical behavior between Go and TypeScript implementations
- **Type-safe evaluation**: Supports text (case-insensitive), select (exact match), and multiselect (any-of logic) field types
- **Logical operators**: AND/OR operations with proper nesting support
- **Comprehensive validation**: Input validation with proper error handling
- **Shared test data**: JSON-driven test suite ensuring both implementations handle identical scenarios

### Implementation Details

**Server-side (Go):**
- `server/app/condition.go`: Core condition evaluation logic
- `server/app/condition_test.go`: Traditional unit tests
- `server/app/condition_json_test.go`: JSON-driven tests

**Client-side (TypeScript):**  
- `webapp/src/types/conditions.ts`: Type definitions and execution logic
- `webapp/src/types/conditions.test.ts`: Traditional unit tests
- `webapp/src/types/conditions-json.test.ts`: JSON-driven tests

**Shared test data:**
- `testdata/condition-test-cases.json`: 24 comprehensive test cases used by both implementations

### Test Coverage

- ✅ Text field case-insensitive matching
- ✅ Select field exact matching
- ✅ Multiselect "any of" logic
- ✅ Logical operators (AND/OR)
- ✅ Nested conditions
- ✅ Edge cases and error handling
- ✅ Type safety validation

Both Go and TypeScript test suites pass all 24 test cases, ensuring behavioral parity.

## Ticket Link

Part of property field conditions feature development.

## Checklist
- [x] Unit tests updated (both Go and TypeScript)
- [ ] Telemetry updated

### DIFF
--- a/server/app/condition.go
+++ b/server/app/condition.go
@@ -1,3 +1,6 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
 package app
 
 import (
@@ -218,26 +221,26 @@ func (cc *ComparisonCondition) validateValueForFieldType(field PropertyField) er
 			return errors.New("text field condition value must be a string")
 		}
 		return nil
-		
+
 	case model.PropertyFieldTypeSelect:
 		// Select fields must have string values (single option ID)
 		var stringValue string
 		if err := json.Unmarshal(cc.Value, &stringValue); err != nil {
 			return errors.New("select field condition value must be a string")
 		}
-		
+
 		// Validate that the value is a valid option ID
 		if len(field.Attrs.Options) == 0 {
 			return errors.New("condition value does not match any valid option for select field")
 		}
-		
+
 		for _, option := range field.Attrs.Options {
 			if option.GetID() == stringValue {
 				return nil
 			}
 		}
 		return errors.New("condition value does not match any valid option for select field")
-		
+
 	case model.PropertyFieldTypeMultiselect:
 		// Multiselect fields must have array values
 		var arrayValue []string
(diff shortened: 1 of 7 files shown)


server/app/condition.go
```diff
@@ -0,0 +1,340 @@
...
+
+		value, valueExists := valueMap[c.IsNot.FieldID]
+		if !valueExists {
+			return true
+		}
+
+		return isNot(field, value, c.IsNot.Value)
+	}
+
+	return true
```
jgheithcock, line 124: I would have thought that an undefined (presumably future) Condition type would default to false for Evaluate, but I'm don't have a strong opinion - I'd really say this was a program error and should return an error as `validate()` does below.

server/app/condition.go
```diff
@@ -0,0 +1,340 @@
...
+func (c *Condition) validate(currentDepth int, propertyFields []PropertyField) error {
+	conditionCount := 0
+
+	if c.And != nil {
+		conditionCount++
+		if len(c.And) == 0 {
+			return errors.New("and condition must have at least one nested condition")
+		}
+		if currentDepth >= MaxConditionDepth {
+			return errors.New("condition nesting depth exceeds maximum allowed")
```
jgheithcock, line 136: Seems always good to put in the error how deep they can go. If we have no real plans to ever make nested conditions, you could even say "conditions cannot be nested". As above, this is more of a nit than a request.

server/app/condition.go
```diff
@@ -0,0 +1,340 @@
...
+		if err := json.Unmarshal(cc.Value, &stringValue); err != nil {
+			return errors.New("text field condition value must be a string")
+		}
+		return nil
+
+	case model.PropertyFieldTypeSelect:
+		var arrayValue []string
+		if err := json.Unmarshal(cc.Value, &arrayValue); err != nil {
+			return errors.New("select field condition value must be an array")
+		}
```
jgheithcock, line 224: This seems odd. Both `PropertyFieldTypeSelect` and `PropertyFieldTypeMultiselect` have an array as a value? The code is the same, except for the error message. Seems that `PropertyFieldTypeSelect` should not be an array, or if it is, it should only have one value in it? From your comment below for `is`, seems this is 'as expected' but I'd love to understand this better.

jgheithcock (approved):

calebroseland (approved):

JulienTant:
@esarafianou sorry for the last minute tag 😅 I'll go ahead and merge this as I believe the risk is low but feel free to leave a comment if you find anything and I'll create a PR to deal with it in high priority.

The main risk I identified is that I built this feature in a recursive way, so I added a MaxDepth constant to limit how deep we can go (only one AND/OR, which matches with the design spec).

PR #2056, opened 2025-08-18
Fix templates adding temporary ids

larkox:
## Summary
Revert changes related to adding ids to elements. These ids were making the templates to use always the same ids in all the runs created from the template, messing how we store the data on mobile.

## Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-65081

### DIFF
--- a/e2e-tests/tests/integration/playbooks/runs/rdp_rhs_runinfo_spec.js
+++ b/e2e-tests/tests/integration/playbooks/runs/rdp_rhs_runinfo_spec.js
@@ -198,9 +198,8 @@ describe('runs > run details page > run info', {testIsolation: true}, () => {
                 cy.apiDeleteChannel(testRun.channel_id).then(() => {
                     cy.visit(`/playbooks/runs/${testRun.id}`);
 
-                    // https://mattermost.atlassian.net/browse/MM-63682
-                    // * Assert channel status private
-                    getOverviewEntry('channel').contains('Private');
+                    // * Assert channel status shows deleted
+                    getOverviewEntry('channel').contains('Channel deleted');
                 });
             });
         });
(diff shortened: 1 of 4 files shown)


catalintomai:
> > @catalintomai I played with the plugin for a while, and wasn't able to reproduce any weird behavior. Not sure if this is no longer needed, or I didn't look in the right place, but I was not able to reproduce any issue just by deleting these lines. Let me know if you find any.
> 
> @larkox, did you try the webapp flow turning on the incremental update flag? If you have some server provisioned with the current, you could also shoot a DM to Lindy, she tested this feature before and she might know better which scenarios were failing before. Thanks.

I tried the fix on local server and didn't see anything suspicious. @lindy65 might also want to run her checks. I don't know if the scenario in the ticket has been validated for the original PR.

larkox:
@isacikgoz After discussing with @yasserfaraazkhan , it seems the errors also happen on master, so they should not be related with this PR. They will probably be handled in a different PR.

larkox:
@lindy65 @yasserfaraazkhan I add you both mainly so Lindy can take a look to the playbooks web side, and Yasser can take a look to the mobile issue it is trying to solve. If you guys want to organize differently, feel free 😄

lindy65:
> @lindy65 @yasserfaraazkhan I add you both mainly so Lindy can take a look to the playbooks web side (that nothing is broken), and Yasser can take a look to the mobile issue it is trying to solve. If you guys want to organize differently, feel free 😄

Hey @larkox - are you open to building me a tar.gz that will work on my Windows machine (not the entire tar.gz as that's too large to upload to a `/cloud` server for testing)?

larkox:
@lindy65 That sounds fine. I will dm it to you the moment I have it (hopefully today).

lindy65 (approved):
Hey @larkox - tested on a server that Yasser shared with me using the playbooks build from this PR and I didn't find any issues on the webapp side with both `Enable Incremental Updates` set to `true` and `false`

larkox:
@yasserfaraazkhan That shouldn't be related at all with this PR. Are you ok if we merge this with that failing? (it seems it would be no problem to do so)

yasserfaraazkhan:
@larkox this should be good to merge

PR #2059, opened 2025-08-21
MM-65008: migrate post menu to registerPostDropdownMenuAction

jgheithcock:
## Summary

- Replace deprecated `registerPostDropdownMenuComponent` with `registerPostDropdownMenuAction`
- Move text labels and action/filter into webapp/src/components/post_menu.tsx
- Add `makeStartPlaybookRunAction` and `makeAttachToPlaybookRunAction`
- DRY filter with `shouldShowPostMenuForPost` returning strict boolean
- Simplify webapp/src/index.tsx to always register actions (min server bumped)
- Remove deprecated component-based post menu items
- Cleanup unused imports

**Note* this also bumps `min_server_version` to 7.10 to ensure we have `registerPostDropdownMenuAction`

## Ticket Link

https://mattermost.atlassian.net/browse/MM-65008

### DIFF
--- a/webapp/src/components/post_menu.tsx
+++ b/webapp/src/components/post_menu.tsx
@@ -22,14 +22,10 @@ import {addToTimeline, showPostMenuModal, startPlaybookRun} from 'src/actions';
 import {useAllowAddMessageToTimelineInCurrentTeam} from 'src/hooks';
 import {isProfessionalLicensedOrDevelopment} from 'src/license';
 
-interface Props {
-    postId: string;
-}
-
 function shouldShowPostMenuForPost(store: Store, postId: string) {
     const state = store.getState() as GlobalState;
     const post = getPost(state, postId);
-    return !!post && !isSystemMessage(post);
+    return Boolean(post) && !isSystemMessage(post);
 }
 
 export const StartPlaybookRunPostMenuText = () => {
--- a/webapp/src/components/post_menu.tsx
+++ b/webapp/src/components/post_menu.tsx
@@ -2,16 +2,14 @@
 // See LICENSE.txt for license information.
 
 import React from 'react';
-import {useDispatch, useSelector} from 'react-redux';
+import {Store} from 'redux';
 
 import styled from 'styled-components';
 
 import {GlobalState} from '@mattermost/types/store';
-import {Post} from '@mattermost/types/posts';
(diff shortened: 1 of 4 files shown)


webapp/src/components/post_menu.tsx
```diff
@@ -107,3 +42,56 @@ const PositionedKeyVariantCircleIcon = styled(KeyVariantCircleIcon)`
...
+
+export function makeStartPlaybookRunAction(store: Store) {
+    return {
+        text: StartPlaybookRunPostMenuText,
+        action: (postId: string) => {
+            const state = store.getState() as GlobalState;
+            const post = getPost(state, postId);
+            if (!post || isSystemMessage(post)) {
+                return;
+            }
```
JulienTant, line 65: It does look like we are reapplying the filter here. If it is indeed necessary, maybe we can reuse `shouldShowPostMenuForPost`?

webapp/src/components/post_menu.tsx
```diff
@@ -107,3 +42,56 @@ const PositionedKeyVariantCircleIcon = styled(KeyVariantCircleIcon)`
...
+
+export function makeAttachToPlaybookRunAction(store: Store) {
+    return {
+        text: AttachToPlaybookRunPostMenuText,
+        action: (postId: string) => {
+            const state = store.getState() as GlobalState;
+            const post = getPost(state, postId);
+            if (!post || isSystemMessage(post)) {
+                return;
+            }
```
JulienTant, line 86: RE: It does look like we are reapplying the filter here. If it is indeed necessary, maybe we can reuse `shouldShowPostMenuForPost`?

JulienTant (approved):
LGTM, I left one nitpick.

webapp/src/components/post_menu.tsx
hmhealey: You could change it so that `shouldShowPostMenuForPost` returns the post if it's valid and then use it in both places with `Boolean` wrapping it in the filter. You'd probably want a different name for it though, and you'd still need to null check `Post` though

hmhealey (approved):
Unless there's any more major changes, looks great!

lindy65 (approved):
Thanks @jgheithcock - looks good to me 👍

PR #2060, opened 2025-08-21
MM-64486: remove telemetry

lieut-data:
## Summary
Telemetry is no more with v11. Clean things up!

## Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-64486

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -107,8 +107,20 @@ jobs:
         with:
           go-version: "${{ env.GO_VERSION }}"
           cache: false
-
-      - name: ci/ensure-build-on-all-platforms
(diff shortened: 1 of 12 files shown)


JulienTant:
@lieut-data I was just thinking about that, we should probably also change the PR template to remove `Telemetry updated` from the checklist

catalintomai (approved):

lieut-data:
@lindy65, not sure if you figured out a workflow with Workflows for getting the plugin tested, but I noted that we do build the tarball in the artifacts https://github.com/mattermost/mattermost-plugin-playbooks/actions/runs/17472703479?pr=2060

CleanShot 2025-09-05 at 11 48 00@2x

PR #2063, opened 2025-08-27
Fix double websocket when changing owner or assignee

larkox:
## Summary
When we add a new participant because we added a owner or assignee from outside of the run, we were sending two different websocket events with the same update at:
- One for adding the participant
- One for setting the assignee

This makes sure only one websocket is sent in this scenario.

## Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-65176

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -457,7 +457,7 @@ func (s *PlaybookRunServiceImpl) CreatePlaybookRun(playbookRun *PlaybookRun, pb 
 		}
 	}
 
-	err = s.AddParticipants(playbookRun.ID, invitedUserIDs, playbookRun.ReporterUserID, false)
+	err = s.AddParticipants(playbookRun.ID, invitedUserIDs, playbookRun.ReporterUserID, false, false)
 	if err != nil {
 		logrus.WithError(err).WithFields(map[string]any{
 			"playbookRunId":  playbookRun.ID,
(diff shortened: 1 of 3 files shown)


server/app/playbook_run_service.go
```diff
@@ -3536,7 +3536,7 @@ func (s *PlaybookRunServiceImpl) AddParticipants(playbookRunID string, userIDs [
 	}
 
 	// ws send run
-	if len(usersToInvite) > 0 {
+	if len(usersToInvite) > 0 && !omitWebsocket {
```
catalintomai, line 3539: nit: use `sendWebsocket` instead, `!omitWebsocket `is a bit harder to read.

larkox:
@lindy65 @yasserfaraazkhan again, I add you to this PR so one can verify the web part and the other the mobile part, unless you want to do it differently.

The idea here is to tests all the scenarios where a user is added to a run, and verify I didn't break anything. If I am not mistaken, the cases are the following:
- Actively adding a user to a run
- Changing the owner to someone outside of the run (that will add the user)
- Changing the assignee of a task to someone outside of the run (that will also add the user)
- Creating a run from a playbook that already has some "default" participants

lindy65:
Hi @larkox :)
Sure, I can test the web side of things by spinning up a `/cloud` server but I'll need a linux-amd64 tar.gz to upload please. Can you DM it to me?

yasserfaraazkhan (approved):

PR #2064, opened 2025-08-28
Playbooks CI upload playbooks bundle

calebroseland:
## Summary
Upload plugin bundle when build is successful

## Ticket Link
- 

## Checklist

- [ ] Telemetry updated
- [ ] Unit tests updated

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -110,6 +110,14 @@ jobs:
 
       - name: ci/ensure-build-on-all-platforms
         run: make dist
+      - name: ci/upload-build-artifact
+        if: success()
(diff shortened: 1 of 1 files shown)


JulienTant (approved):
LGTM! In the future we should consider using https://github.com/mattermost/actions-workflows if it makes sense (we can look at how calls does it!)

larkox (approved):
The code looks good to me. Only two concerns:

Where is the artifact right now? I cannot find it :sweat_smile:

And... is the full artifact, or artifacts separated by platform? The current "full package" is too big to upload, and we need them separate.

lindy65 (approved):
Thanks @calebroseland :+1:

PR #2068, opened 2025-09-03
Deprecated MySQL for v11

JulienTant:
## Summary
Deprecated MySQL for v11

## Ticket Link
N/A

## Checklist
- [x] Unit tests updated


server/app/playbook.go
```diff
@@ -756,7 +756,7 @@ func (cpm *ChannelPlaybookMode) UnmarshalText(text []byte) error {
 
 // Scan parses a ChannelPlaybookMode back from the DB
 func (cpm *ChannelPlaybookMode) Scan(src interface{}) error {
-	txt, ok := src.([]byte) // mysql
+	txt, ok := src.([]byte)
 	if !ok {
 		txt, ok := src.(string) //postgres
 		if !ok {
```
lieut-data, line 762: Should we just do the postgres way all the time instead of just removing the comment?

server/sqlstore/migrations_utils.go
```diff
@@ -161,19 +109,17 @@ var dropColumnPG = func(e sqlx.Ext, tableName, colName string) error {
 }
 
 func addPrimaryKey(e sqlx.Ext, sqlStore *SQLStore, tableName, primaryKey string) error {
-	hasPK := 0
-
-	dbSelectionLine := "AND tco.table_schema = (SELECT DATABASE())"
-	if e.DriverName() == model.DatabaseDriverPostgres {
-		dbSelectionLine = "AND tco.table_catalog = (SELECT current_database())"
+	if e.DriverName() != model.DatabaseDriverPostgres {
```
lieut-data, line 112: nit: probably don't need to check this ever again after initialization?

lieut-data (commented):
:tada: -- two thoughts below.

server/app/playbook.go
JulienTant: Fixed with 5d0f4743

lieut-data (approved):

jgheithcock (approved):
Other than my question, looks good

PR #2071, opened 2025-09-09
Add experimental features toggle with real-time updates

JulienTant:
## Summary

Add experimental features configuration setting that controls visibility of properties/attributes UI with real-time websocket updates.

## Ticket Link

MM-65161

## Checklist
- [ ] Unit tests updated

### DIFF
--- a/client/settings.go
+++ b/client/settings.go
@@ -9,6 +9,7 @@ import (
 )
 
 type GlobalSettings struct {
+	EnableExperimentalFeatures bool `json:"enable_experimental_features"`
 }
 
 // SettingsService handles communication with the settings related methods.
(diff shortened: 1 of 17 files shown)


jgheithcock (changes requested):
This seems a very generic vague tag for a specific feature. Should we not call this `enableplaybooksexperiments` or something more specific? In particular, if this is showing up in the main Mattermost app, we should give the user some clue as to what it does. Is there a standard we should be following for this?

jgheithcock (approved):
Having talked this over with Julien, I see now that this is an existing setting and the user will see it is only for Plugins so I am okay (also, we hope to remove this flag sooooon.)

Willyfrog (approved):

Willyfrog:
A nitpick afterthought (we don't need to implement it): should the state of this feature also depend on having the properties feature enabled in the server?
it doesn't have to go in this PR, but enabling the feature if the server doesn't allow properties might have a very negative experience

PR #2082, opened 2025-09-18
MM-65685: fix undefined pollution in myPlaybookRuns

calebroseland:
## Summary
Fixes an error where the `myPlaybookRuns` reducer store was polluted with an `undefined` entry, which would cause HTTP-Fetch errors. `ReceivedTeamPlaybookRunConnections` is added so `myPlaybookRuns` doesn't hook into it. 

## Ticket Link
https://mattermost.atlassian.net/browse/MM-65685

## Checklist

- [ ] Unit tests updated


lindy65 (approved):
Thanks @calebroseland - tested with the tar.gz from this PR on a `/cloud` server and didn't see any  `...runs/undefined` in the url of HTTP network requests

PR #2091, opened 2025-10-01
Don't delete fields or options used in conditions

JulienTant:
## Summary

This PR prevents the deletion of playbook attribute fields and attribute options that are being used in conditions. When a user attempts to delete a attribute field or option that is referenced in any condition, the system now returns an error with a clear message indicating that the field/option is in use.
It also prevents the user from changing the attribute type if it's being used in a condition.

## Ticket Link

- https://mattermost.atlassian.net/browse/MM-65154
- https://mattermost.atlassian.net/browse/MM-65155
- https://mattermost.atlassian.net/browse/MM-66144

## Checklist
- [x] Unit tests updated


jgheithcock (approved):
LGTM

calebroseland (changes requested):
Overall LGTM! Just one request for the refetching on-error bit

PR #2094, opened 2025-10-03
Add e2e tests for playbook and run attributes

JulienTant:
## Summary

This PR adds comprehensive e2e tests for playbook and run attributes functionality, along with a production bug fix for multiselect property clearing.

### Changes:
1. **E2E Tests for Playbook Attributes** (`playbook_attributes_spec.js`)
   - Tests for creating, editing, and deleting attributes (text, select, multiselect)
   - Tests for attribute limits and validation
   - Tests for attribute visibility and ordering

2. **E2E Tests for Run Attributes** (`run_attributes_spec.js`)
   - Tests for attribute inheritance from playbook to run
   - Tests for editing attribute values from run details page and channel RHS
   - Tests for clearing attributes (text, select, multiselect)
   - Tests for attribute independence (run attributes remain unchanged when playbook attributes change)
   - Tests for channel message posting when attributes are modified

3. **Production Bug Fix**
   - Fixed multiselect property clear functionality in `property_multiselect.tsx`
   - The issue was that `handleStopEdit` only saved values when `tempValue !== null`, preventing clearing
   - Now properly saves `null` values to clear the field

## Ticket Link

https://mattermost.atlassian.net/browse/MM-66085

## Checklist
- [x] Unit tests updated
- [x] E2E tests added


webapp/src/components/backstage/playbook_properties/playbook_properties.tsx
```diff
@@ -116,8 +116,17 @@ const PlaybookProperties = ({playbookID}: Props) => {
             return;
         }
 
+        // Find the highest number in existing "Attribute X" names
+        const attributeNumbers = properties
```
JulienTant, line 120: I ran into an issue by adding attributes (Attribute 1-5), removing the first one, then trying to add a new one, we would try to create (Attribute `properties.length + 1`)... which was a duplicate of Attribute 5 again.

webapp/src/components/rhs/properties/property_multiselect.tsx
```diff
@@ -37,28 +37,30 @@ const MultiselectProperty = (props: Props) => {
...
+        props.onValueChange(tempValue ?? null);
+        setTempValue(undefined);
     };
 
     const selectOptions = props.field.attrs?.options?.map((option) => ({
         value: option.id,
         label: option.name,
     })) || [];
 
+    // initialValue is displayValue IF tempValue does not have a value yet (undefined when not set), but as
```
JulienTant, line 55: It's not pretty to have `string[] | null | undefined` but I needed to know what was the current state:

- `undefined` = we never set any value here
- `null` = the selector is empty (the lib does not return an empty array but null)
- `string[]` = the list of selected IDs

lindy65 (approved):
Thanks @JulienTant - makes my heart so happy to see these :)

PR #2114, opened 2025-10-16
Update dependencies for React 18

hmhealey:
## Summary
Along with updating the version of React, we also have to update a bunch of libraries that depend on it because the versions we previously used only work on React 18. For the ones that we upgraded in the web app repo, I updated them to match the web app, but for some other ones, I just overrode the dependencies of those dependencies because that seemed to be all that prevented them from working with React 18.

Along with that, there's some minor changes to some types and some small fixes which I'll note below.

## Checklist

- [x] Unit tests updated


webapp/package.json
```diff
@@ -24,22 +24,23 @@
     "js-trim-multiline-string": "^1.0.8",
     "lodash": "4.17.21",
     "luxon": "3.6.1",
-    "mattermost-redux": "10.9.0",
+    "mattermost-redux": "11.1.0-2",
```
hmhealey: I updated mattermost-redux and the client/type packages because I forgot that mattermost-redux had a dependency on a specific major version of Redux. I cut these adhoc releases off of the latest 11.1 branch of the monorepo, and we can replace them with proper ones after 11.1.0 releases

webapp/package.json
```diff
@@ -102,14 +102,14 @@
     "jest-canvas-mock": "2.3.1",
     "jest-environment-jsdom": "29.7.0",
     "jest-junit": "13.0.0",
+    "patch-package": "8.0.1",
```
hmhealey: This is temporary until I either merge https://github.com/mattermost/react-bootstrap/pull/6 or figure out a better way to modify its dependency without it

webapp/package.json
```diff
@@ -102,14 +102,14 @@
...
     "postcss-styled-syntax": "0.7.1",
     "process": "0.11.10",
-    "react-beautiful-dnd": "13.1.0",
-    "react-bootstrap": "1.6.1",
+    "react-bootstrap": "github:mattermost/react-bootstrap#05559f4c61c5a314783c390d2d82906ee8c7e558",
```
hmhealey: As mentioned elsewhere, this is only used during development, but it'll now match the web app's version instead of accidentally depending on a newer version of React Bootstrap

webapp/package.json
```diff
@@ -122,9 +122,25 @@
...
-      "react-dom": "17.0.2"
+      "react": "$react",
+      "react-dom": "$react-dom"
+    },
+    "react-input-autosize": {
+      "react": "$react"
+    },
+    "react-select": {
+      "react": "$react",
+      "react-dom": "$react-dom"
```
hmhealey: The web app uses a newer version of React Select (5.9.0) which supports React 18 and has some accessibility improvements, but upgrading requires some breaking changes

webapp/src/types/actions.ts
```diff
@@ -58,163 +58,163 @@ export const WEBSOCKET_PLAYBOOK_RUN_INCREMENTAL_UPDATE_RECEIVED = manifest.id +
 // so we respect their naming convention (all caps)
 export const PUBLISH_TEMPLATES = (manifest.id + '_PUBLISH_TEMPLATES').toUpperCase();
 
-export interface ReceivedToggleRHSAction {
+export type ReceivedToggleRHSAction = {
```
hmhealey: I've got absolutely no idea why, but these types weren't able to be passed into `dispatch` previously because of some weird behaviour where interfaces aren't compatible with `UnknownAction` unless those interfaces supported arbitrary fields

PR #2221, opened 2026-03-25
Adding bulk editing capability to Playbooks > Outline > Tasks

asaadmahmood:
## Summary
Add ability to enter a bulk edit mode in Playbooks editor
https://www.loom.com/share/7be4007b5bb3485ca26dde38dbe1a945

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68227

## Checklist

- [ ] Unit tests updated


jgheithcock:
Cool to see but don't let my review hold back merging. Thanks @asaadmahmood !

lindy65:
Hi @asaadmahmood - is there a Jira ticket for this feature?  

Please also remember to add E2E test/s so I can review the PR :)

asaadmahmood:
@lindy65 no there isn't one yet.

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -347,6 +395,198 @@ const ChecklistList = ({
...
+                    return cl;
+                }
+                return {
+                    ...cl,
+                    items: cl.items.filter((_, itemIdx) => !deletedIndices.has(itemIdx)),
+                };
+            });
+
+            if (playbookRun) {
+                dispatch(playbookRunUpdated({...playbookRun, checklists: newChecklists}));
```
calebroseland: Bug: this specific logic branch (playbook-runs) does not actually delete the selected tasks—they're only removed from redux. Tasks reappear on refresh.

lindy65 (approved):
Thanks @asaadmahmood - feature and tests look good 👍

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -347,6 +395,198 @@ const ChecklistList = ({
...
+    const handleBulkDelete = async () => {
+        const selectedByChecklist = new Map<number, Set<number>>();
+        selectedItems.forEach(({checklistIndex, itemIndex}: {checklistIndex: number; itemIndex: number; item: ChecklistItem}) => {
+            if (!selectedByChecklist.has(checklistIndex)) {
+                selectedByChecklist.set(checklistIndex, new Set());
+            }
+            selectedByChecklist.get(checklistIndex)!.add(itemIndex);
+        });
+
+        if (playbookRun && playbookRun.type === PlaybookRunType.ChannelChecklist) {
```
calebroseland:
```suggestion
        if (playbookRun) {
```

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -347,6 +395,198 @@ const ChecklistList = ({
...
+                    ...cl,
+                    items: cl.items.filter((_, itemIdx) => !deletedIndices.has(itemIdx)),
+                };
+            });
+
+            if (playbookRun) {
+                dispatch(playbookRunUpdated({...playbookRun, checklists: newChecklists}));
+            } else {
+                setChecklistsForPlaybook(newChecklists);
+            }
```
calebroseland:
```suggestion
            setChecklistsForPlaybook(newChecklists);
```

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -347,6 +395,198 @@ const ChecklistList = ({
...
+                const sortedIndices = [...itemIndices].sort((a, b) => b - a);
+                for (const idx of sortedIndices) {
+                    deleteTotal++;
+                    try {
+                        await clientDeleteChecklistItem(playbookRun.id, checklistIndex, idx); // eslint-disable-line no-await-in-loop
+                    } catch {
+                        deleteFailures++;
+                    }
+                }
+            }
```
calebroseland: This nested loop should not have an await in it. On large bulk-deletes, this will have a significant series of delays by doing it sequentially. 

If possible, could we use the `await Promise.allSettled(...)` pattern like `handleBulkAssign`? I also like the return-error handling patterns used there better. But understanding the index-based approach, if it's strictly necessary—it'd be fine as-is.

webapp/src/components/checklist/checklist_list.tsx
```diff
@@ -347,6 +395,198 @@ const ChecklistList = ({
         setChecklistsForPlaybook(newChecklists);
     };
 
+    const isSelectedIndex = (clIdx: number, itemIdx: number) => {
+        return selectedIndices.has(`${clIdx}-${itemIdx}`);
+    };
+
+    const handleBulkAssign = async (userId: string) => {
```
calebroseland: Could we consider moving the bulk-action handlers and maybe some of the selection logic over to `multi_select_action_bar.tsx` as custom hook like `useBulkActions`?

Maybe something like 
```tsx
const {handleBulkAssign, handleBulkDueDate, handleBulkDelete} = useBulkActions({setPlaybook})
```

calebroseland (dismissed):
I like the pattern overall! Bulk-edit will help with a good number of flows, and potential future enhancements for other actions and bulk-drag might be nice, too.

The change request is mainly just for the playbook run task deletion bug, and potentially some other refactoring (e.g. `useBulkActions`), but most other comments are for discussion sake.

I do have some concerns on the UX: 
1. I find the dual-checkmarks to be a bit confusing
2. having the bulk-edit button in the playbook editor UI but not the run/checklist UI is a minor discrepancy

Image

calebroseland (changes requested):
Thanks for the updates, Asaadhttps://github.com/mattermost/mattermost-plugin-playbooks/actions/runs/24734175620#summary-72356467032 came back with some suggestions to fill a few testing gaps -- if any of those are false positives, I'm not sure how to dismiss.

Thanks for adding to RHS, but there's a couple minor UI things there: 

Image





Also, these were a few findings from a local agent review: 

> 1. Fix silent bulk assign/due-date failures — In webapp/src/hooks/bulk_actions.ts:79-106 and :121-148, change the filter from r.status ===
   'rejected' to also catch r.status === 'fulfilled' && r.value?.error, and only dispatch playbookRunUpdated for items whose individual call
   succeeded. Root cause: setAssignee and setDueDate in webapp/src/client.ts:311-327 swallow errors and return {error} instead of rejecting,
   so Promise.allSettled always reports success.
>  2. Reposition the bulk action bar for RHS — In webapp/src/components/checklist/multi_select_action_bar.tsx:190-201, replace the
  viewport-relative position: fixed; left: 50%; transform: translateX(-50%) with a portal into an RHS-local positioned ancestor (or switch
  to position: absolute inside a positioned parent). Currently the bar centers on the viewport, straddling the RHS/channel divider when bulk
   edit is used from rhs_checklist_list.tsx.
>  3. Prevent bulk-select from trapping users in edit mode — In webapp/src/components/checklist_item/checklist_item.tsx:401-404, guard the
  bulkEditMode onClick so it's a no-op when the item is currently editing: onClick={props.bulkEditMode && props.onItemSelect && !isEditing ?
   ... : undefined}. Without this guard, a user who enters bulk mode while an item is open for editing can't reach Cancel or Save — every
  click selects instead.

asaadmahmood:
@calebroseland apologies, my agent might have added it to the RHS, isn't meant to be added to the checklist side of things only playbooks.

calebroseland (approved):
Approving to unblock, all looks good to me, just an E2E case needs fixing.

PR #2229, opened 2026-04-02
[MM-68190] Support export and import of attributes and conditionals

jgheithcock:
## Summary

Extends playbook export/import to include **property field definitions** (attributes) and **conditions** (conditional tasks). Previously, exporting a playbook to JSON and reimporting it would lose these definitions.

- On **export**: properties and conditions are fetched separately and serialized as additive `"properties"` and `"conditions"` JSON keys. The export version is intentionally kept at 1 — Go's `json.Decoder` silently ignores unknown keys, so old servers importing new exports gracefully degrade (playbook imports without attributes/conditionals) rather than hard-rejecting with "Unsupported version".
- On **import**: property fields are recreated with new IDs, an `old_id → new_id` mapping is built (same pattern as `CopyPlaybookPropertiesToRun`), condition expressions are remapped via `SwapPropertyIDs`, conditions are created via the new `CreateConditionsFromExport` method on `ConditionService`, and checklist item `condition_id` references are updated on the playbook.

### Key changes

| File | Changes |
|------|---------|
| `server/app/export.go` | New `ExportPropertyField`/`ExportCondition` types, custom `UnmarshalJSON` for versioned condition deserialization, `generatePropertiesExport`/`generateConditionsExport` helpers, extended `GeneratePlaybookExport` signature |
| `server/app/playbook.go` | Changed `ConditionID`/`ConditionAction` export tags from `"-"` to key names; added `ImportWithProperties` and `GetPlaybookConditionsForExport` to `PlaybookService` interface |
| `server/app/playbook_service.go` | `ImportWithProperties` implementation with full audit trail, `GetPlaybookConditionsForExport` delegation |
| `server/app/condition.go` | Added `CreateConditionsFromExport` and `GetPlaybookConditionsForExport` to `ConditionService` interface |
| `server/app/condition_service.go` | Implementation of the two new methods (parallels `CopyPlaybookConditionsToPlaybook` pattern) |
| `server/api/playbooks.go` | Export handler fetches properties/conditions; import handler decodes and passes them through |
| `server/app/export_test.go` | Unit tests for export with properties, condition round-trip, unsupported version, null expression |
| `server/app/playbook_service_test.go` | Unit tests for `ImportWithProperties` ID remapping, early return, and error handling |
| `server/api_playbooks_test.go` | Integration tests for export→import with properties and conditions, verifying remapped IDs |

## Ticket Link

https://mattermost.atlassian.net/browse/MM-68190

## Checklist
- [x] ~~Gated by experimental feature flag~~ (additive change — purely new JSON keys, no feature flag needed)
- [x] Unit tests updated

### DIFF
--- a/server/api/playbooks.go
+++ b/server/api/playbooks.go
@@ -609,7 +609,21 @@ func (h *PlaybookHandler) exportPlaybook(c *Context, w http.ResponseWriter, r *h
 		return
 	}
 
-	export, err := app.GeneratePlaybookExport(playbook)
+	// Fetch properties and conditions for the playbook
+	properties, err := h.propertyService.GetPropertyFields(playbookID)
+	if err != nil {
+		h.HandleError(w, c.logger, err)
+		return
+	}
+
+	// Fetch conditions through the playbook service which manages the export
+	conditions, err := h.playbookService.GetPlaybookConditionsForExport(playbookID)
+	if err != nil {
+		h.HandleError(w, c.logger, err)
+		return
+	}
+
+	export, err := app.GeneratePlaybookExport(playbook, properties, conditions)
 	if err != nil {
 		h.HandleError(w, c.logger, err)
 		return
@@ -659,7 +673,9 @@ func (h *PlaybookHandler) importPlaybook(c *Context, w http.ResponseWriter, r *h
 	userID := r.Header.Get("Mattermost-User-ID")
 	var importBlock struct {
 		app.Playbook
-		Version int `json:"version"`
+		Version    int                    `json:"version"`
+		Properties []app.ExportPropertyField `json:"properties,omitempty"`
+		Conditions []app.ExportCondition   `json:"conditions,omitempty"`
 	}
 	if err := json.NewDecoder(r.Body).Decode(&importBlock); err != nil {
 		h.HandleErrorWithCode(w, c.logger, http.StatusBadRequest, "unable to decode playbook import", err)
@@ -700,7 +716,7 @@ func (h *PlaybookHandler) importPlaybook(c *Context, w http.ResponseWriter, r *h
 		return
 	}
 
-	id, err := h.playbookService.Import(playbook, userID)
+	id, err := h.playbookService.ImportWithProperties(playbook, userID, importBlock.Properties, importBlock.Conditions)
 	if err != nil {
 		h.HandleError(w, c.logger, err)
 		return
--- a/server/api/playbooks.go
+++ b/server/api/playbooks.go
@@ -609,14 +609,12 @@ func (h *PlaybookHandler) exportPlaybook(c *Context, w http.ResponseWriter, r *h
 		return
 	}
 
-	// Fetch properties and conditions for the playbook
 	properties, err := h.propertyService.GetPropertyFields(playbookID)
 	if err != nil {
 		h.HandleError(w, c.logger, err)
 		return
 	}
 
-	// Fetch conditions through the playbook service which manages the export
 	conditions, err := h.playbookService.GetPlaybookConditionsForExport(playbookID)
 	if err != nil {
 		h.HandleError(w, c.logger, err)
@@ -716,7 +714,12 @@ func (h *PlaybookHandler) importPlaybook(c *Context, w http.ResponseWriter, r *h
 		return
 	}
 
-	id, err := h.playbookService.ImportWithProperties(playbook, userID, importBlock.Properties, importBlock.Conditions)
+	id, err := h.playbookService.Import(app.PlaybookImportData{
+		Playbook:   playbook,
+		Version:    importBlock.Version,
+		Properties: importBlock.Properties,
+		Conditions: importBlock.Conditions,
+	}, userID)
 	if err != nil {
 		h.HandleError(w, c.logger, err)
 		return
--- a/server/api/playbooks.go
+++ b/server/api/playbooks.go
@@ -673,9 +673,9 @@ func (h *PlaybookHandler) importPlaybook(c *Context, w http.ResponseWriter, r *h
 	userID := r.Header.Get("Mattermost-User-ID")
 	var importBlock struct {
 		app.Playbook
-		Version    int                    `json:"version"`
+		Version    int                       `json:"version"`
 		Properties []app.ExportPropertyField `json:"properties,omitempty"`
-		Conditions []app.ExportCondition   `json:"conditions,omitempty"`
+		Conditions []app.ExportCondition     `json:"conditions,omitempty"`
 	}
 	if err := json.NewDecoder(r.Body).Decode(&importBlock); err != nil {
 		h.HandleErrorWithCode(w, c.logger, http.StatusBadRequest, "unable to decode playbook import", err)
--- a/server/app/condition_service.go
+++ b/server/app/condition_service.go
@@ -232,11 +232,6 @@ func (s *conditionService) CreateConditionsFromExport(playbookID string, exportC
 	return conditionMapping, nil
 }
 
-// GetPlaybookConditionsForExport retrieves all conditions for a playbook without auth overhead.
-func (s *conditionService) GetPlaybookConditionsForExport(playbookID string) ([]Condition, error) {
-	return s.store.GetPlaybookConditions(playbookID, 0, MaxConditionsPerPlaybook)
-}
-
 // CreatePlaybookCondition creates a new stored condition for a playbook
 func (s *conditionService) CreatePlaybookCondition(userID string, condition Condition, teamID string) (*Condition, error) {
 	auditRec := s.auditor.MakeAuditRecord("createCondition", model.AuditStatusFail)
--- a/server/app/condition_service.go
+++ b/server/app/condition_service.go
@@ -141,6 +141,84 @@ func (s *conditionService) CopyPlaybookConditionsToPlaybook(sourcePlaybookID, ta
 	return conditionMapping, nil
 }
 
+// CreateConditionsFromExport creates conditions from exported data with property ID remapping.
+// This parallels CopyPlaybookConditionsToPlaybook but starts from deserialized export data
+// rather than from conditions already in the database.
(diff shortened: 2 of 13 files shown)


server/app/condition_service.go
```diff
@@ -141,6 +141,84 @@ func (s *conditionService) CopyPlaybookConditionsToPlaybook(sourcePlaybookID, ta
...
+	conditionMapping := make(map[string]*Condition)
+	if len(exportConditions) == 0 {
+		return conditionMapping, nil
+	}
+
+	newProperties, err := s.propertyService.GetPropertyFields(playbookID)
+	if err != nil {
+		logrus.WithError(err).WithField("playbook_id", playbookID).Warn("failed to fetch property fields for validation during import")
+		newProperties = []PropertyField{}
+	}
```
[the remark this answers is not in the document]
jgheithcock: This is the deliberate best-effort pattern used throughout the codebase. If GetPropertyFields fails, newProperties is empty, so every condition will fail IsValid and be individually skipped with a log. The import still succeeds with whatever it can create. Failing hard here would prevent the entire import for a transient error.

server/app/condition_service.go
```diff
@@ -141,6 +141,84 @@ func (s *conditionService) CopyPlaybookConditionsToPlaybook(sourcePlaybookID, ta
...
+		if err != nil {
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"playbook_id":  playbookID,
+				"condition_id": exportCondition.ID,
+				"version":      exportCondition.Version,
+			}).Warn("failed to create condition during import, skipping")
+			continue
+		}
+
+		conditionMapping[exportCondition.ID] = createdCondition
```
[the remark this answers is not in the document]
jgheithcock: Re the license gate on import: No change is needed for license checks — import the data regardless of license state, this matches how Duplicate works. The data is inert without the license. The properties/conditions are stored but have no effect unless the license activates them.

server/app/playbook_service.go
```diff
@@ -104,6 +104,160 @@ func (s *playbookService) Import(playbook Playbook, userID string) (string, erro
...
+				auditRec.AddErrorDesc("failed to get playbook for condition remapping: " + err.Error())
+				logrus.WithError(err).WithField("playbook_id", newPlaybookID).Warn("failed to get playbook for condition remapping")
+			} else {
+				// Write back only the condition IDs that were actually recreated,
+				// using the saved exported references (the stored items have empty IDs).
+				for _, ref := range exportedCondRefs {
+					if newConditionID, ok := oldConditionIDToNewConditionID[ref.conditionID]; ok {
+						updatedPlaybook.Checklists[ref.checklistIdx].Items[ref.itemIdx].ConditionID = newConditionID
+						updatedPlaybook.Checklists[ref.checklistIdx].Items[ref.itemIdx].ConditionAction = ref.conditionAction
+					}
```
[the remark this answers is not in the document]
jgheithcock: Good point, added

server/app/playbook_service_test.go
```diff
@@ -449,3 +450,271 @@ func TestPlaybookService_Duplicate(t *testing.T) {
...
+			CreateConditionsFromExport(newPlaybookID, exportConditions, gomock.Any()).
+			DoAndReturn(func(pbID string, conds []app.ExportCondition, mappings *app.PropertyCopyResult) (map[string]*app.Condition, error) {
+				assert.Contains(t, mappings.FieldMappings, oldFieldID, "field mappings should contain old field ID")
+				assert.Equal(t, newFieldID, mappings.FieldMappings[oldFieldID], "field mapping should be old->new")
+				assert.Contains(t, mappings.OptionMappings, oldOptionID, "option mappings should contain old option ID")
+				assert.Equal(t, newOptionID, mappings.OptionMappings[oldOptionID], "option mapping should be old->new")
+
+				return map[string]*app.Condition{
+					oldConditionID: {ID: newConditionID},
+				}, nil
```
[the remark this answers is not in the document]
jgheithcock: Go allows composite literals without the & inside map literals when the value type is a pointer; the compiler infers it. So this isn't actually a bug but it is cleaner to be explicit so that has been changed.

server/app/condition_service.go
```diff
@@ -141,6 +141,102 @@ func (s *conditionService) CopyPlaybookConditionsToPlaybook(sourcePlaybookID, ta
...
+			continue
+		}
+
+		// Remap field and option IDs in-place
+		if err := exportCondition.ConditionExpr.SwapPropertyIDs(propertyMappings); err != nil {
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"playbook_id":  playbookID,
+				"condition_id": exportCondition.ID,
+				"version":      exportCondition.Version,
+			}).Warn("failed to remap property IDs in condition during import, skipping")
```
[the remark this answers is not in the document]
jgheithcock: This is a false positive. Looking at the caller in playbook_service.go (line 209-212), propertyMappings is constructed with CopiedFields: copiedFields, which is built from successfully created fields. SwapPropertyIDs uses propertyMappings.CopiedFields (passed by the caller), not newProperties (which is only used for IsValid). The data is already there.

server/api/playbooks.go
```diff
@@ -700,7 +716,7 @@ func (h *PlaybookHandler) importPlaybook(c *Context, w http.ResponseWriter, r *h
 		return
 	}
 
-	id, err := h.playbookService.Import(playbook, userID)
+	id, err := h.playbookService.ImportWithProperties(playbook, userID, importBlock.Properties, importBlock.Conditions)
```
JulienTant: `Import` was enough, we don't want to add a `with` everytime we add something new :P 

Also maybe we should pass the whole importBlock to have access to the version etc

server/app/export.go
```diff
@@ -58,14 +60,117 @@ func generateMetricsExport(metrics []PlaybookMetricConfig) []interface{} {
...
+		exported = append(exported, exportCondition)
+	}
+	return exported
+}
+
 // GeneratePlaybookExport returns a playbook in export format.
 // Fields marked with the stuct tag "export" are included using the given string.
-func GeneratePlaybookExport(playbook Playbook) ([]byte, error) {
+// If properties and conditions are provided, they are also included in the export.
+// Note: CurrentPlaybookExportVersion is kept at 1 to maintain backward compatibility.
```
JulienTant: I strongly disagree with this. If we are adding fields, we should bump the version number. 

Also this comment belongs to a spec more than a comment, this is a log of a design decision not a comment on the code itself :P

server/app/playbook_service.go
```diff
@@ -104,6 +104,170 @@ func (s *playbookService) Import(playbook Playbook, userID string) (string, erro
 	return newID, nil
 }
 
+// ImportWithProperties imports a playbook with its properties and conditions.
+// This calls Create directly (rather than Import) so a single audit record
+// covers the full operation including property/condition setup.
+func (s *playbookService) ImportWithProperties(playbook Playbook, userID string, properties []ExportPropertyField, conditions []ExportCondition) (string, error) {
```
JulienTant: If it's an option,I would argue that we deserve an import service, or at least create some smaller functions with easy to test input>output to increase coverage and readability

Also RE:why not just reuse import

server/app/export.go
```diff
@@ -58,14 +60,117 @@ func generateMetricsExport(metrics []PlaybookMetricConfig) []interface{} {
...
+// NewExportCondition creates an ExportCondition from a Condition
+func NewExportCondition(c Condition) ExportCondition {
+	return ExportCondition{
+		ID:            c.ID,
+		ConditionExpr: c.ConditionExpr,
+		Version:       c.Version,
+	}
+}
+
+func generatePropertiesExport(properties []PropertyField) []interface{} {
```
JulienTant: Why do we return `[]interface{}` instead of `[]ExportPropertyField`?

server/app/export.go
```diff
@@ -58,14 +60,117 @@ func generateMetricsExport(metrics []PlaybookMetricConfig) []interface{} {
...
+	exported := make([]interface{}, 0, len(properties))
+	for _, property := range properties {
+		exportProperty := NewExportPropertyField(property)
+		exported = append(exported, exportProperty)
+	}
+	return exported
+}
+
+func generateConditionsExport(conditions []Condition) []interface{} {
```
JulienTant: Why do we return `[]interface{}` instead of `[]ExportCondition`?

server/app/playbook_service.go
```diff
@@ -104,6 +104,170 @@ func (s *playbookService) Import(playbook Playbook, userID string) (string, erro
...
+	auditRec.AddEventResultState(playbook)
+
+	return newPlaybookID, nil
+}
+
+// GetPlaybookConditionsForExport delegates to the condition service to fetch conditions
+// for export without authentication overhead.
+func (s *playbookService) GetPlaybookConditionsForExport(playbookID string) ([]Condition, error) {
+	return s.conditionService.GetPlaybookConditionsForExport(playbookID)
+}
```
JulienTant: This could be 

```suggestion
func (s *playbookService) GetPlaybookConditionsForExport(playbookID string) ([]Condition, error) {
     result, err := s.conditionService.GetPlaybookConditions("", playbookID, 0, MaxConditionsPerPlaybook)
     if err != nil {
             return nil, err
     }
     return result.Items, nil
}
```

And we would not need `conditionService.GetPlaybookConditionsForExport` at all. 

Note that the comment "without authentication overhead" is scary for no valid reason haha.

JulienTant (changes requested):
Thanks @jgheithcock ! That's a lot haha

Also something else is the lack of e2e tests :P

lindy65 (approved):
Thanks @jgheithcock - tested export/import of a playbook with existing attributes and all looks good.

Will you be adding E2Es to this PR or via a separate PR?

lindy65 (approved):
Thanks @jgheithcock - tests look good 👍

JulienTant (approved):
Thanks @jgheithcock

jgheithcock:
Noting that the CI errors in FIPS run are not connected to our code. "Unrecognized config permissions tag value" warnings with "tag_value":"sysconsole_write_*_read" indicate that the Cypress test harness is sending a config payload that the latest FIPS server image now rejects. (@saturninoabril fyi)

PR #2236, opened 2026-04-17
MM-68333 Use WithTooltip shared from web app

hmhealey:
## Summary
This a POC for starting to use the shared package in Playbooks. It should be good to merge except for the fact that it requires a server that has the upcoming version of the shared package to run, so it's not suitable for merge until either we update the minimum server version to 11.8.0+ or we figure out how to get newly added shared components to work without the accompanying server version. This is also why the E2E tests are failing.

The new components are part of `@mattermost/shared`, but Webpack is configured to use them as external dependencies. This means that they won't be bundled into Playbooks, and they'll be loaded from the web app at runtime.

Related web app PR: https://github.com/mattermost/mattermost/pull/36037

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68333

## Screenshots
The only notable UX difference with this is that, instead of forcing the tooltips to always be below the attached button, they use the same positioning logic as the web app which prioritizes tooltip position based on what has space available. It prioritizes above the button > below the button > to the left/right of the button. In the third case below, it puts the tooltip on the left side instead of having it off centre above or below it.
Before|After
-|-
0 before|0b after
1 before|1b after
2 before|2b after

## Checklist

- [x] Unit tests updated


webapp/src/components/backstage/playbook_editor/controls.tsx
```diff
@@ -209,21 +207,14 @@ export const AutoFollowToggle = ({playbook}: ControlProps) => {
...
-        </Tooltip>
-    );
-
     return (
         <SecondaryButtonLargerCheckbox
             checked={isFollowing}
             disabled={archived}
         >
-            <OverlayTrigger
-                placement={'bottom'}
```
hmhealey, line 224: As mentioned in the description, this is a functional change. There's currently a `forcedPlacement` prop on `WithTooltip`, but I don't think we need to use it

webapp/src/components/backstage/playbook_editor/playbook_editor.tsx
```diff
@@ -90,23 +91,21 @@ const PlaybookEditor = () => {
     const archived = playbook.delete_at !== 0;
 
     const archivedTooltip = archived && (
-        <Tooltip
-            delay={{show: 0, hide: 1000}}
```
hmhealey, line 105: These are no longer customizable. If we want to be able to customize them, we'll have add that to the `WithTooltip` component

webapp/src/components/backstage/playbook_runs/playbook_run/header_button.tsx
```diff
@@ -28,11 +23,10 @@ interface HeaderButtonProps {
 
 const HeaderButton = ({tooltipId, tooltipMessage, Icon, onClick, isActive, clicked, size, iconSize, 'aria-label': ariaLabel, 'data-testid': dataTestId}: HeaderButtonProps) => {
     return (
-        <Tooltip
+        <WithTooltip
             id={tooltipId}
-            placement={'bottom'}
-            shouldUpdatePosition={true}
```
hmhealey, line 34: This is handled automatically

webapp/src/components/sidebar/create_playbook_dropdown.tsx
```diff
@@ -116,21 +109,20 @@ const CreatePlaybookDropdown = (props: CreatePlaybookDropdownProps) => {
...
                         aria-label={formatMessage({defaultMessage: 'Create Playbook Dropdown'})}
                         data-testid='create-playbook-dropdown-toggle'
                     >
                         <PlusIcon size={18}/>
                     </Button>
-                    {inputImportPlaybook}
-                </>
-            </OverlayTrigger>
+                </WithTooltip>
+                {inputImportPlaybook}
```
hmhealey, line 124: This change is because `WithTooltip` adds the required attributes to the child component instead of adding an extra div, so it's not possible to pass a fragment into it. This element has `display: none` though, so it didn't need a tooltip to begin with

webapp/src/components/widgets/copy_link.tsx
```diff
@@ -40,21 +39,17 @@ const CopyLink = ({
     };
 
     return (
-        <Tooltip
+        <WithTooltip
             id={id}
-            placement='bottom'
-            delay={OVERLAY_DELAY}
-            onExited={() => setWasCopied(false)}
```
hmhealey, line 47: This getting removed is a bug. It's not super noticeable, but I'll fix that. It might require a new version of the shared package though, so that might not be fixed right away

webapp/update-shared-packages-dev.sh
hmhealey, line 1: These scripts might be useful for others if they're trying to make changes to using the shared package while working on Playbooks, but I don't expect that to happen too often. I'll probably remove them before merge

webapp/update-shared-packages-npm.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+VERSION=${1:-11.7.0-0}
+
+npm add @mattermost/types@${VERSION} @mattermost/client@${VERSION} mattermost-redux@${VERSION} @mattermost/shared@${VERSION}
```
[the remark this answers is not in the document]
hmhealey: That's not an issue. This script isn't intended to be merged

jgheithcock (approved):
I (and Opus) didn't find anything you haven't already noted. I look forward to a fully shared library. Thanks Harrison!

JulienTant:
@hmhealey passing along to Caleb, he's more likely to provide valuable feedback on this matter!

webapp/src/components/backstage/playbook_runs/playbook_run/retrospective/timeline_event_item.tsx
```diff
@@ -398,7 +399,7 @@ const TimelineEventItem = (props: Props) => {
                         )}
                     >
                         <ClockOutlineIcon size={12}/>
```
calebroseland: Compass icons don't forward refs, so this tooltip is broken. Maybe wrap in a div/span that the tooltip can attach to? Otherwise, perhaps we should wrap the whole timestamp as the icon is a pretty small hitbox.

webapp/src/components/backstage/playbook_runs/playbook_run/rhs_participants.tsx
```diff
@@ -191,16 +192,15 @@ const ParticipantRow = ({id, teamName, isRunOwner, manageMode, removeFromRun, ch
...
             return (
                 <HoverButtonContainer>
-                    <Tooltip
+                    <WithTooltip
                         id={`${id}-tooltip`}
-                        shouldUpdatePosition={true}
-                        content={formatMessage({defaultMessage: 'Send message'})}
+                        title={formatMessage({defaultMessage: 'Send message'})}
                     >
                         <SendMessageButton
```
calebroseland, line 199: `SendMessageButton` is similarly a plain FC / no forwardRef

webapp/package.json
```diff
@@ -24,7 +25,7 @@
     "js-trim-multiline-string": "^1.0.8",
     "lodash": "4.17.21",
     "luxon": "3.6.1",
-    "mattermost-redux": "11.5.0",
+    "mattermost-redux": "11.7.0-0",
```
calebroseland: `11.7.0`

calebroseland (changes requested):
Overall looks great! Thanks for adding this-- also looking forward to a potential GenericModal. 

I see you commented a few additional changes you'll be making before merge.--I've added a couple of additional ones that should be addressed.



I've attached my agent's report in case you want to skim through any of the other points, but most are nits/nice-to-haves. 
[review-report.md](https://github.com/user-attachments/files/27602732/review-report.md)

webapp/src/components/widgets/copy_link.tsx
hmhealey: This is fixed now

webapp/update-shared-packages-dev.sh
hmhealey: Removed

webapp/package.json
hmhealey: This now requires technically 11.8.0, but I can't merge this until 11.8.0 is actually released, so it'll have to stay as a prerelease version for now

calebroseland (approved):

PR #2243, opened 2026-04-21
ci: strip debug symbols from release binaries

JulienTant:
## Summary

Pass `-ldflags="-s -w"` via `GO_BUILD_FLAGS` in the `dist` and `dist-fips` CI jobs so release builds omit the Go symbol table (`-s`) and DWARF debug info (`-w`). Local developer builds (`make dist` run outside CI) are unaffected. Go's `BuildInfo` section is preserved, so `go version -m` still works on the stripped binaries.

### Binary size impact

Measured locally by building each target with and without `-ldflags="-s -w"`:

| Target | Unstripped | Stripped | Saved | Reduction |
|---|---:|---:|---:|---:|
| linux/amd64   | 40.2 MiB | 28.2 MiB | 12.0 MiB | 29.8% |
| linux/arm64   | 38.0 MiB | 26.6 MiB | 11.5 MiB | 30.1% |
| darwin/amd64  | 41.6 MiB | 29.1 MiB | 12.5 MiB | 30.0% |
| darwin/arm64  | 39.7 MiB | 26.9 MiB | 12.8 MiB | 32.3% |
| windows/amd64 | 40.9 MiB | 28.9 MiB | 12.0 MiB | 29.2% |

Consistent ~30% reduction per binary; the bundle ships all five, so ~60 MiB off the uncompressed tarball.

## Ticket Link
N/A

## Checklist
- [ ] ~~Unit tests updated~~

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -106,7 +106,7 @@ jobs:
     container:
       image: mattermostdevelopment/mattermost-build-server:${{ needs.go.outputs.version }}
     env:
-      GO_BUILD_FLAGS: '-ldflags=-s -w'
+      GO_BUILD_FLAGS: '-ldflags="-s -w"'
     steps:
       - name: ci/checkout-repo
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -143,7 +143,7 @@ jobs:
         password: ${{ secrets.DOCKERHUB_DEV_TOKEN }}
     env:
       FIPS_ENABLED: true
-      GO_BUILD_FLAGS: '-ldflags=-s -w'
+      GO_BUILD_FLAGS: '-ldflags="-s -w"'
     steps:
       - name: ci/checkout-repo
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,8 +105,6 @@ jobs:
       - lint
     container:
       image: mattermostdevelopment/mattermost-build-server:${{ needs.go.outputs.version }}
-    env:
-      GO_BUILD_FLAGS: '-ldflags="-s -w"'
     steps:
       - name: ci/checkout-repo
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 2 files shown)


JulienTant:
@lindy65 this PR changes the binary generation - no code change so nothing to test

pavelzeman (approved):
Approved. 

It doesn't look like we're doing anything FIPS specific to plugins, but it'd be good to have Eva confirm.

JulienTant:
> Approved.
> 
> It doesn't look like we're doing anything FIPS specific to plugins, but it'd be good to have Eva confirm.

Right I'll add Eva as reviewer and Alejandro, as he's the one usually creating the FIPS build for playbooks

esarafianou (changes requested):
@JulienTant shouldn't we change the Makefile server build step instead so that the debug symbols are stripped not only from the dev binaries created during CI but also from the final release artifact?

Plugins are released from our internal centralized delivery repo that triggers the plugin's makefile.

agarciamontoro (changes requested):
Do we know how this affects the generation of stacktraces when we receive panic reports from Sentry?

pavelzeman:
I think we'll want to co-ordinate this PR with [the change in mattermost/mattermost](https://github.com/mattermost/mattermost/pull/36181), specifically when it comes to striping the debug info and retaining it for released builds (just in case it's ever needed).

JulienTant:
> Plugins are released from our internal centralized delivery repo that triggers the plugin's makefile.

Completely missed that! Thanks. Fixed with b2d0fc3

I also made a chance to make sure that when "MM_DEBUG" is set, we don't strip the debug symbols. Curious if it's something we might be interested for the main repo @pavelzeman?


@agarciamontoro Not sure what is used for Sentry specifically, but stacktraces are complete and unaffected. I ran a small test locally to made sure of it. If Sentry reuse the panic stacktrace, it should work fine/

agarciamontoro (approved):
Thanks, @JulienTant! I also verified the stacktraces are intact, so this looks safe to me :)

JulienTant:
Ping @esarafianou

pavelzeman:
Please don't merge quite yet - we need to decide if we are shipping two sets of bits (with and without debug symbols) or if we strip the symbols and ship a separate .debug info package (that could be used with a debugger).

this applies here, in monorepo and anywhere else where we'd want to revisit the size of shipping binaries

JulienTant:
> Please don't merge quite yet - we need to decide if we are shipping two sets of bits (with and without debug symbols) or if we strip the symbols and ship a separate .debug info package (that could be used with a debugger).
> 
> this applies here, in monorepo and anywhere else where we'd want to revisit the size of shipping binaries

Copy!

PR #2245, opened 2026-04-22
feat(metrics): serve metrics via Mattermost plugin endpoint

lieut-data:
## Summary

Replace the standalone Prometheus HTTP server (previously running on port 9093) with a `ServeMetrics` plugin method that integrates with the Mattermost server's built-in plugin metrics endpoint. This removes the need for a separate `service.go` and eliminates the dependency on `MetricsSettings.Enable` gating — metrics tracking is now always active, and exposure is controlled by the Mattermost server.

This helps support https://github.com/mattermost/mattermost/pull/34678, which needs to see and interact with the hook in order to surface plugin metrics automatically.

Relates-to: https://mattermost.atlassian.net/browse/MM-68436


lieut-data:
/test-analysis-override neither AI nor the Calls plugin have test coverage here, this is overkill

calebroseland (approved):

PR #2246, opened 2026-04-22
[MM-68372] Scope metric updates to playbook; strip import metric IDs

jgheithcock:
## Summary

This change ensures playbook metric configuration updates only apply to rows for the same playbook, and that imported playbooks do not carry forward metric database IDs (consistent with export).

- **Store:** When updating an existing `IR_MetricConfig` row by ID, the query also requires the row's `PlaybookID` to match the playbook being saved.
- **Service:** `Import` clears `Metrics[].ID` before `Create` so new playbooks always get new metric rows.
- **Tests:** Store and app unit tests for the above. Regression test for the store path was confirmed to fail when the extra `PlaybookID` condition is removed.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-68372

## Checklist
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_service.go
+++ b/server/app/playbook_service.go
@@ -92,6 +92,11 @@ func (s *playbookService) Import(data PlaybookImportData, userID string) (string
 	model.AddEventParameterToAuditRec(auditRec, "numConditions", len(data.Conditions))
 	model.AddEventParameterAuditableToAuditRec(auditRec, "playbook", playbook)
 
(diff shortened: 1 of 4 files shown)


JulienTant (approved):
Approving but I'm curious: if export omits them, why do we have to strip them?

jgheithcock:
> Approving but I'm curious: if export omits them, why do we have to strip them?

As they could be injected.

hmohammed-prodsec (approved):

PR #2247, opened 2026-04-22
CF-14: Ability to automatically archive Channel after resolution

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68480

## Checklist

Summary

  Adds an opt-in Auto-archive channel setting to playbooks. When enabled, finishing a run automatically archives its channel; restoring the run un-archives it. The setting is snapshotted onto the run at creation, so toggling it on the playbook later does not retroactively change behaviour for in-flight runs. Linked channels (existing channels the run did not create) are never auto-archived, regardless of the toggle.

  Changes

  Backend

  - New AutoArchiveChannel field on the playbook model, persisted via migration.
  - Three new fields on the run: a snapshot of the playbook's setting at run creation, a marker for whether the channel was created by the run, and a marker for whether the channel was actually auto-archived (used by restore).
  - Run-finish flow archives the channel (only when the run created it), records a channel_archived timeline event, and persists the archived marker before the archive call so a partial failure remains recoverable.
  - Run-restore flow un-archives the channel when the run-archived marker is set, clears the marker, and records a channel_unarchived timeline event. The marker is always cleared on restore — even if the channel was manually un-archived in the meantime — so a subsequent finish starts from a clean state.
  - Restore posts the status message after un-archiving so it lands in an active channel.

  Frontend

  - New When a run finishes section in the playbook editor outline with the Auto-archive channel toggle and a confirmation banner shown when enabled.
  - Toggle is disabled with an explanatory tooltip when the playbook is configured to link an existing channel.
  - The RHS "Time to wrap up?" Finish button now goes through the shared finish-run flow (modal + checks) instead of the legacy direct-dispatch action, so the auto-archive snapshot path is consistent everywhere a run can be finished.

  How to test

  1. Toggle visibility & default — Open a playbook with Create new channel mode. Confirm the toggle appears under When a run finishes and is off by default.
  2. Persistence — Enable the toggle, reload the editor, confirm it remains on and the confirmation banner is shown. Toggle off and confirm the banner disappears.
  3. Linked-channel disable — Switch the playbook to Link existing channel. The toggle should be disabled and hovering it should show a tooltip explaining why.
  4. Auto-archive on finish — With the toggle on, start a run, then click Finish run from the RHS Time to wrap up? footer. The run channel should become archived and the timeline should show a channel archived event.
  5. Un-archive on restore — From the run details page, restore the finished run. The channel should be un-archived, the status update message should appear in it, and the timeline should show a channel unarchived event.
  6. Toggle off / default — With the toggle off, finish a run. The channel must remain active.
  7. Manual un-archive before restore — With the toggle on, finish a run, manually un-archive the channel from the channel header, then restore the run. The restore should succeed and not error; future finishes of the same run should still auto-archive correctly.

  Security

  - Playbook editor toggle — Same persona as other playbook edits: users with edit permission on the playbook (typically members/admins of the playbook). Viewers cannot change it.
  - Run finish (triggers auto-archive) — Same persona as the existing finish-run permission: the run owner, run participants, and channel/team admins per existing checks. No new privilege is granted; auto-archive simply piggybacks on an action the user is already authorized to perform.
  - Run restore (triggers un-archive) — Same persona as the existing restore-run permission. The un-archive uses the plugin's channel-update capability and does not bypass channel permissions.
  - Linked channels are never auto-archived even if the flag is true on the run, preventing a playbook author from archiving a shared channel they don't own by linking it into a playbook.
  - Setting is snapshotted at run creation, so a malicious or mistaken edit to the playbook after a run starts cannot retroactively cause an unexpected archive on finish.
  - No new API surface is exposed beyond the existing playbook update / run finish / run restore endpoints; their authorization checks are unchanged.


catalintomai:
@coderabbitai review

catalintomai:
@coderabbitai review

catalintomai:
@coderabbitai review

catalintomai:
closed for https://github.com/mattermost/mattermost-plugin-playbooks/pull/2254

PR #2254, opened 2026-05-04
CF-14: Ability to automatically archive Channel after resolution

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68480

## Checklist

Summary

  Adds an opt-in Auto-archive channel setting to playbooks. When enabled, finishing a run automatically archives its channel; restoring the run un-archives it. The setting is snapshotted onto the run at creation, so toggling it on the playbook later does not retroactively change behaviour for in-flight runs. Linked channels (existing channels the run did not create) are never auto-archived, regardless of the toggle.

  Changes

  Backend

  - New AutoArchiveChannel field on the playbook model, persisted via migration.
  - Three new fields on the run: a snapshot of the playbook's setting at run creation, a marker for whether the channel was created by the run, and a marker for whether the channel was actually auto-archived (used by restore).
  - Run-finish flow archives the channel (only when the run created it), records a channel_archived timeline event, and persists the archived marker before the archive call so a partial failure remains recoverable.
  - Run-restore flow un-archives the channel when the run-archived marker is set, clears the marker, and records a channel_unarchived timeline event. The marker is always cleared on restore — even if the channel was manually un-archived in the meantime — so a subsequent finish starts from a clean state.
  - Restore posts the status message after un-archiving so it lands in an active channel.

  Frontend

  - New When a run finishes section in the playbook editor outline with the Auto-archive channel toggle and a confirmation banner shown when enabled.
  - Toggle is disabled with an explanatory tooltip when the playbook is configured to link an existing channel.
  - The RHS "Time to wrap up?" Finish button now goes through the shared finish-run flow (modal + checks) instead of the legacy direct-dispatch action, so the auto-archive snapshot path is consistent everywhere a run can be finished.

  How to test

  1. Toggle visibility & default — Open a playbook with Create new channel mode. Confirm the toggle appears under When a run finishes and is off by default.
  2. Persistence — Enable the toggle, reload the editor, confirm it remains on and the confirmation banner is shown. Toggle off and confirm the banner disappears.
  3. Linked-channel disable — Switch the playbook to Link existing channel. The toggle should be disabled and hovering it should show a tooltip explaining why.
  4. Auto-archive on finish — With the toggle on, start a run, then click Finish run from the RHS Time to wrap up? footer. The run channel should become archived and the timeline should show a channel archived event.
  5. Un-archive on restore — From the run details page, restore the finished run. The channel should be un-archived, the status update message should appear in it, and the timeline should show a channel unarchived event.
  6. Toggle off / default — With the toggle off, finish a run. The channel must remain active.
  7. Manual un-archive before restore — With the toggle on, finish a run, manually un-archive the channel from the channel header, then restore the run. The restore should succeed and not error; future finishes of the same run should still auto-archive correctly.

  Security

  - Playbook editor toggle — Same persona as other playbook edits: users with edit permission on the playbook (typically members/admins of the playbook). Viewers cannot change it.
  - Run finish (triggers auto-archive) — Same persona as the existing finish-run permission: the run owner, run participants, and channel/team admins per existing checks. No new privilege is granted; auto-archive simply piggybacks on an action the user is already authorized to perform.
  - Run restore (triggers un-archive) — Same persona as the existing restore-run permission. The un-archive uses the plugin's channel-update capability and does not bypass channel permissions.
  - Linked channels are never auto-archived even if the flag is true on the run, preventing a playbook author from archiving a shared channel they don't own by linking it into a playbook.
  - Setting is snapshotted at run creation, so a malicious or mistaken edit to the playbook after a run starts cannot retroactively cause an unexpected archive on finish.
  - No new API surface is exposed beyond the existing playbook update / run finish / run restore endpoints; their authorization checks are unchanged.


server/app/playbook_run_service.go
```diff
@@ -1299,6 +1303,45 @@ func (s *PlaybookRunServiceImpl) FinishPlaybookRun(playbookRunID, userID string)
 	}
 
 	s.metricsService.IncrementRunsFinishedCount(1)
+
+	if s.shouldAutoArchiveChannel(playbookRunToModify) {
```
JulienTant, line 1322: nit: It is so annoying not being able to use TX in plugins 😅

server/app/playbook_run_service.go
```diff
@@ -1522,6 +1590,35 @@ func (s *PlaybookRunServiceImpl) RestorePlaybookRun(playbookRunID, userID string
...
+func (s *PlaybookRunServiceImpl) shouldAutoArchiveChannel(run *PlaybookRun) bool {
+	return run.AutoArchiveChannel && run.ChannelCreatedByRun && run.ChannelID != ""
+}
+
+// restoreChannelByID clears DeleteAt on the given channel (un-archives it).
+// NOTE: Get→mutate→Update is not atomic. Channel.Update sends the full channel object, so any
+// concurrent channel property edit (display name, purpose) made between Get and Update will be
+// silently overwritten. This is acceptable as a best-effort operation until the plugin API
+// exposes a dedicated RestoreChannel method.
+func (s *PlaybookRunServiceImpl) restoreChannelByID(channelID string, logger *logrus.Entry) bool {
```
JulienTant, line 1596: nit: Code is very similar to the code we have to delete a channel, should we also have a helper for "deleteChannelByID"? that would declutter the code above for certain!

server/sqlstore/migrations.go
```diff
@@ -1739,6 +1739,26 @@ var migrations = []Migration{
 				return errors.Wrapf(err, "failed creating index IR_Condition_RunID_DeleteAt")
 			}
 
+			return nil
```
JulienTant, line 1742: Not saying to change this because I don't know which is the "right" way, but we also have migrations in `sqlstore/migrations/postgresql`. Advantage to the other one is support for rollbacks

JulienTant (approved):
Looks good @catalintomai ! Just left a few nits

server/sqlstore/migrations.go
catalintomai: I believe the other one says `// WARNING: Disable morph migrations until proper testing
	// if err := sqlStore.runMigrationsWithMorph(); err != nil {
	// 	return fmt.Errorf("failed to complete migrations (with morph): %w", err)
	// }`

server/app/playbook_run_service.go
catalintomai: added one.

jgheithcock (approved):
Nothing to add. LGTM

lindy65:
Thanks @catalintomai 

According to "_Frontend changes above: Toggle is disabled with an explanatory tooltip when the playbook is configured to link an existing channel._" I should not be able to enable the toggle switch to auto archive the playbook if I'm linking it to an existing channel but this doesn't seem to be enforced at the moment:
image

I ran the above with the playbook linked to Town Square (existing channel) and didn't receive any notification that Town Square couldn't be archived (this was kind of expected but I was just wondering what would happen as the playbook was set to auto-archive the channel even although it was an existing channel). There were some js.console errors though - not sure if relevant?
image

However, if I go back to the playbook in the backstage (after running it for the first time), I now see the tooltip over the toggle button BUT the toggle is in the enabled state. (Probably better to fix the above case where toggle can be enabled even although Town Square was chosen as existing channel?)
image

Adding also that if I've already run an existing playbook that created a new channel and then go to the backstage and change the setting in that playbook to link to an existing channel, the toggle to auto-archive the channel remains enabled. (After a hard refresh the toggle is still enabled but does show the tooltip on hover. I feel it should automatically toggle to the off position and show the tooltip on hover as soon as the setting to link to an existing channel is selected?)
image



In this flow: Finish run > restart run from archived channel 
  - there is an active "retrospective post" in the channel from the previously "finished" run
      - This doesn't make sense to me because I "restarted" the run
         - Should this post be immediately greyed out (as previous ones are if I keep "finishing" and "restarting" the run in the channel)
image

image

This is just a "nit" but, could this text be on one line and also left-aligned with the toggle and section header above?
image

catalintomai:
Hi @lindy65 , good catch, I believe I fixed the issues, please give it a try. re: "there is an active "retrospective post" in the channel from the previously "finished" run" - this should not be specific to this PR's work, so I suggest opening a ticket to address it separately. Thanks.

lindy65:
> "there is an active "retrospective post" in the channel from the previously "finished" run" - this should not be specific to this PR's work

Hey @catalintomai - thanks, the issues with the "auto-archive channel" toggle are fixed 👍 

Sorry to come back to the issue of the active retrospective post in channel that posts after "finishing" / auto-archiving the channel.

I know you said it's not related to the code in this PR and I realise that the actual post is part of a playbook's process when finishing a run.

However, now that we've added the auto-archive feature, it doesn't make sense to post the retrospective post AFTER the channel has been archived.

I tested a flow of clicking on the "Yes, start retrospective" from the finished run/auto-archived channel and...
- I am taken to the retrospective in the backstage
- I type something in the retrospective and "Publish" (screenshot 1)
- I navigate back to "Channels" from the backstage and am viewing the archived channel. 
    - A retrospective has been posted by the system but, the text I typed into the retrospective in the backstage is not visible (channel deleted = can't make a post?) (screenshot 3)

So, although I agree that the actual "Retrospective" post made in channel by the system (if retrospective is selected in the backstage before the run) doesn't have anything to do with your PR, the flow that happens by auto-archiving a channel after a run just doesn't make sense as far as the "Retrospective" post goes.

Does doing something like this make sense in this PR or is it a bigger change that needs to be done in Playbooks code itself? (screenshot 4)

image

image

image

image

catalintomai:
Hi @lindy65 , re: "... it doesn't make sense to post the retrospective post AFTER the channel has been archived." I agree. However, re: flow : user finishes run → channel auto-archives → user goes back later and clicks the retrospective button from the archived channel - same sequence exists without auto-archive if someone manually archives the channel, so the root cause is entirely in PublishRetrospective and predates this PR - looks like a pre-existing Playbooks bug that this feature exposed. That is why I recommend opening a separate ticket. Thanks.

lindy65 (approved):
Thanks @catalintomai - I've created https://mattermost.atlassian.net/browse/MM-68815 to track the active Retrospective post in an archived playbook channel

catalintomai:
@coderabbitai review

server/app/playbook_run_service.go
```diff
@@ -1530,6 +1565,87 @@ func (s *PlaybookRunServiceImpl) RestorePlaybookRun(playbookRunID, userID string
 	return nil
 }
 
+// shouldAutoArchiveChannel returns true if the run's channel should be auto-archived on finish:
+// AutoArchiveChannel was snapshotted from the playbook at run creation and the channel was
+// created by this run (not linked).
+func (s *PlaybookRunServiceImpl) shouldAutoArchiveChannel(run *PlaybookRun) bool {
```
edgarbellot, line 1578: This guard relies on `ChannelCreatedByRun` and `AutoArchiveChannel` being immutable (correctly enforced in [playbook_run.go#L614-L615](https://github.com/mattermost/mattermost-plugin-playbooks/blob/5f5b8ebdd178731c4d273e2340a754869aa940c2/server/sqlstore/playbook_run.go#L614-L615)), but `ChannelID` is mutable via `UpdateRun` in [graphql_root_run.go#L200-L237](https://github.com/mattermost/mattermost-plugin-playbooks/blob/5f5b8ebdd178731c4d273e2340a754869aa940c2/server/api/graphql_root_run.go#L200-L237), which only requires `ManageChannelProperties` on the target.

Chain: create a run with auto-archive (so `ChannelCreatedByRun=true`) → `UpdateRun` to swap `ChannelID` to a victim channel in the same team → finish the run → `pluginAPI.Channel.Delete` archives the victim under the plugin's elevated identity. Net effect: `ManageChannelProperties` escalates to `DeleteChannel`.

Would it make sense to either snapshot the original channel (e.g. `OriginalChannelID`) and compare here, or reset `ChannelCreatedByRun` whenever `ChannelID` changes via `UpdateRun`?

server/app/playbook_run_service.go
```diff
@@ -1530,6 +1565,87 @@ func (s *PlaybookRunServiceImpl) RestorePlaybookRun(playbookRunID, userID string
...
+	return run.AutoArchiveChannel && run.ChannelCreatedByRun && run.ChannelID != ""
+}
+
+// deleteChannelByID archives (soft-deletes) the given channel. Returns true on success.
+func (s *PlaybookRunServiceImpl) deleteChannelByID(channelID string, logger *logrus.Entry) bool {
```
edgarbellot, line 1583: Since `pluginAPI.Channel.Delete` runs under the plugin's elevated identity, this archives the channel regardless of whether the user who triggered `FinishPlaybookRun` actually has `DeletePublicChannel`/`DeletePrivateChannel`. `RunManageProperties` (see [permissions_service.go#L459](https://github.com/mattermost/mattermost-plugin-playbooks/blob/5f5b8ebdd178731c4d273e2340a754869aa940c2/server/app/permissions_service.go#L459)) accepts any run participant, so a participant who normally cannot delete the channel ends up able to archive it implicitly by finishing the run.

Is this working as expected? If not, would it be worth adding a user-side permission check before invoking `Channel.Delete` here? something like:

```go
perm := model.PermissionDeletePublicChannel
if channel.Type == model.ChannelTypePrivate {
    perm = model.PermissionDeletePrivateChannel
}
if !s.pluginAPI.User.HasPermissionToChannel(userID, channelID, perm) {
    logger.Info("skipping auto-archive: user lacks delete permission")
    return false
}
```

That would keep the destructive action aligned with what the invoking user could do directly.

server/app/playbook_run_service.go
```diff
@@ -1530,6 +1565,87 @@ func (s *PlaybookRunServiceImpl) RestorePlaybookRun(playbookRunID, userID string
...
+	}
+	return true
+}
+
+// restoreChannelByID clears DeleteAt on the given channel (un-archives it).
+// NOTE: Get→mutate→Update is not atomic. Channel.Update sends the full channel object, so any
+// concurrent channel property edit (display name, purpose) made between Get and Update will be
+// silently overwritten. This is acceptable as a best-effort operation until the plugin API
+// exposes a dedicated RestoreChannel method.
+func (s *PlaybookRunServiceImpl) restoreChannelByID(channelID string, logger *logrus.Entry) bool {
```
edgarbellot, line 1596: This is the symmetric concern to [deleteChannelByID](https://github.com/mattermost/mattermost-plugin-playbooks/blob/5f5b8ebdd178731c4d273e2340a754869aa940c2/server/app/playbook_run_service.go#L1576-L1582): `pluginAPI.Channel.Update` runs under the plugin's identity, so any run participant with `RunManageProperties` can trigger an un-archive on the linked channel even if they could not unarchive it directly.

In isolation the impact is limited because `AutoArchivedChannel` is only set by the finish flow, but combined with `ChannelID` being mutable via `UpdateRun` ([graphql_root_run.go#L200-L237](https://github.com/mattermost/mattermost-plugin-playbooks/blob/5f5b8ebdd178731c4d273e2340a754869aa940c2/server/api/graphql_root_run.go#L200-L237)), it broadens the surface.

If this is not expected, would it be reasonable to also gate this on the user having `ManagePublic/PrivateChannelProperties` (or an equivalent) before clearing `DeleteAt`, so the restore mirrors what the user could do without the bot?

edgarbellot (changes requested):
@catalintomai great work! left some comments/questions

server/app/playbook_run_service.go
catalintomai: @edgarbellot - When a run's channel is swapped via UpdateRun, the "created by this run" flag is now cleared, so auto-archive should not fire on the new channel. A test covering this path was added as well.

server/app/playbook_run_service.go
catalintomai: @edgarbellot  - the way I see it - the authorization chain is: the playbook admin enables AutoArchiveChannel=true → the bot executes that policy when the configured event fires → the user finishing/restoring the run is the trigger, not the authorizer. This is the same pattern used throughout the codebase: channel creation at run start, member management, status posts, and webhooks all run under the bot's elevated identity. In short , it is by design.

server/app/playbook_run_service.go
catalintomai: @edgarbellot - same reasoning as above.

server/app/playbook_run_service.go
edgarbellot: Thank you! 🚀

server/app/playbook_run_service.go
edgarbellot: got it, thanks for clarifying! Approving the PR :)

edgarbellot (approved):

asaadmahmood (approved):
LGTM

PR #2255, opened 2026-05-04
CF-2: Unique Identifier for Incidents

catalintomai:
# Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68468

## Checklist

Summary

Adds sequential run identifiers to Playbooks. Each playbook can be configured with an optional prefix (e.g. `INC`) that is combined with a per-playbook auto-incrementing counter to produce identifiers like `INC-00042`. The identifier is assigned atomically at run creation time, displayed in the run RHS and run list, and resolved as `{SEQ}` in run/channel name templates.

---

Changes

Backend

- New database columns: `RunNumberPrefix`, `NextRunNumber`, and `ChannelNameTemplate` on `playbooks`; `RunNumber` and `SequentialID` on `playbook_runs`. A partial unique index enforces prefix uniqueness per team (case-insensitive, active playbooks only).
- `ResolveRunCreationParams` is the recommended pre-creation step that validates the template and resolves the owner/creator without consuming a run number. `CreatePlaybookRun` always performs the sequential-ID allocation itself via `resolveAndAllocate` — callers cannot inject a pre-set `RunNumber` or `SequentialID`. Calling `ResolveRunCreationParams` first surfaces template errors before a counter is consumed.
- New template engine (`template_engine.go`) shared by channel-name and run-name templates. Handles property-value formatting, date values, user display names, and placeholder clean-up (orphaned separators, multi-space collapse).
- New `PATCH /playbooks/{id}` endpoint for partial update of `run_number_prefix` and `channel_name_template`. Uses `DisallowUnknownFields`; whitespace-only templates and templates that reference nonexistent property fields are rejected at save time.
- Input validation: prefix max 32 chars, alphanumeric + hyphens; channel name template max 1024 chars; templates are validated against the playbook's current property fields at save time. `SEQ`, `OWNER`, `CREATOR`, and `PROPERTY_USER` are reserved as property field names (case-insensitive) so user-defined fields cannot shadow the built-in tokens.
- `ReminderMessageTemplate` (status updates) resolves placeholders at send time. Unknown tokens pass through unchanged.
- Slash-command surfaces (`/playbook run`, status update commands) honor the prefix and template resolution.

Frontend

- New `TemplateInput` component: text input with `{token}` autocomplete triggered by typing `{` or clicking the **Insert variable** button. Suggests system tokens (`SEQ`, `OWNER`, `CREATOR`) and custom property fields. Shows a live preview resolving tokens against current playbook state.
- Run name field in the **Start a run** modal is optional when the playbook has a template that is dialog-safe (no property placeholders). The modal pre-fills the name from the template and shows a live preview. When the resolved name exceeds 64 characters, an inline error is shown and the Start button is disabled.
- Sequential ID badge displayed in the run RHS overview, run RHS title chip, and run list row when a run has an assigned sequential ID. Every run receives a sequential ID — a bare zero-padded number when no prefix is set (e.g. `00001`), or a prefixed ID when a prefix is configured (e.g. `INC-00001`).
- Playbook editor exposes the run number prefix field and channel-name-template field; both saved via debounced REST PATCH. The form rolls back optimistically and surfaces a toast on 400/409/other failures.
- Playbook and run TypeScript types extended with `run_number_prefix`, `next_run_number`, `run_number`, and `sequential_id`.

---

How to Test

1. **Configure a prefix** — Open a playbook, go to **Actions**, set **Run number prefix** to `INC`. Save.
2. **Start a run** — Start a new run from that playbook. Confirm the run name pre-fills from the template (if set) and is optional. After the run is created, verify the RHS overview shows `INC-00001`.
3. **Sequential increment** — Start a second run. Verify its sequential ID is `INC-00002`.
4. **Template with `{SEQ}`** — Edit the playbook's run name template to include `{SEQ}` (e.g. `Incident {SEQ}`). Start a third run and verify the channel is named `Incident INC-00003` (because prefix `INC` is set, `{SEQ}` resolves to `INC-00003`).
5. **Template with `{OWNER}` / `{CREATOR}`** — Set the template to `Run by {OWNER}`. Verify the run name resolves to the owner's display name at creation time.
6. **Run list** — Open the runs list. Verify runs from the prefixed playbook display their sequential ID.
7. **Prefix uniqueness** — Try to configure two active playbooks in the same team with the same prefix. Verify the second save is rejected with an error toast.
8. **No prefix** — Create a playbook without a prefix. Start a run. Verify the sequential ID badge still appears in the RHS, but shows a bare counter number (e.g. `00001`) with no prefix label.
9. **Template autocomplete** — In the playbook editor run name template field, type `{` or click the **Insert variable** button. Verify the dropdown shows `SEQ`, `OWNER`, `CREATOR`, and any custom property fields.
10. **Template too long** — Set a run name template whose resolved value exceeds 64 characters. Open the **Start a run** modal and verify an inline error is shown ("The resolved run name exceeds the 64-character limit…") and the Start button is disabled.

---

Security

- The PATCH endpoint requires the **manage-properties** permission on the playbook. Only users with that permission may change the prefix or channel-name template; non-members of a private playbook receive 403.
- Run creators (any member with playbook run-start permission) trigger the sequential ID increment at creation time. The counter is incremented atomically in the database to prevent duplicate IDs even under concurrent requests.
- Prefix uniqueness is enforced at the database level (partial unique index) so two concurrent editors cannot assign the same prefix to different playbooks in the same team.
- Input validation is performed server-side: prefix format (alphanumeric + hyphens, max 32 chars) and template length (max 1024 chars) are validated before any write. The `NextRunNumber` counter cannot be set via the REST API — it is server-managed only.
- Property values in templates are resolved server-side at run creation; client-supplied initial property values are validated against the playbook's field definitions before being applied, preventing injection of arbitrary values into channel names.
- The sequential ID is **frozen at creation**: changing a playbook's prefix afterwards does not retroactively rewrite existing runs' identifiers — only new runs see the new prefix.


server/app/playbook_service.go
```diff
@@ -621,3 +638,49 @@ func (s *playbookService) ReorderPropertyFields(playbookID, fieldID string, targ
...
+		return 0, errors.Wrap(err, "playbook_service.IncrementRunNumber")
+	}
+	return n, nil
+}
+
+func (s *playbookService) UpdateChannelNameTemplateIfUnchanged(playbookID, oldTemplate, newTemplate string) (bool, error) {
+	return s.store.UpdateChannelNameTemplateIfUnchanged(playbookID, oldTemplate, newTemplate)
+}
+
+func (s *playbookService) UpdateRunNumberPrefix(playbookID, prefix, userID string) error {
```
edgarbellot, line 715: Every other state-changing method in this file (`createPlaybook`, `updatePlaybook`, `archivePlaybook`, etc.) creates an audit record via `MakeAuditRecord`/`LogAuditRec`. Changing the run number prefix affects the identity of all future run identifiers, so it seems like it should be audited too.

Would it make sense to add an audit record here similar to the other methods?

And one nit: the `userID` parameter is accepted here but never used, is this intentional?

server/app/errors.go
```diff
@@ -34,3 +34,12 @@ var ErrPropertyOptionsInUse = errors.New("property options are in use")
...
 var ErrPropertyFieldTypeChangeNotAllowed = errors.New("property field type change not allowed")
+
+// ErrReservedPropertyFieldName occurs when trying to create or update a property field with a reserved name.
+var ErrReservedPropertyFieldName = errors.New("reserved property field name")
+
+// ErrInternalPrecondition occurs when a caller skipped a mandatory setup step; maps to HTTP 500.
+var ErrInternalPrecondition = errors.New("internal precondition violated")
+
+// ErrRunNumberPrefixImmutable occurs when a caller tries to change the run number prefix after runs exist.
+var ErrRunNumberPrefixImmutable = errors.New("run number prefix cannot be changed after runs have been created")
```
edgarbellot: `ErrRunNumberPrefixImmutable` is declared here and the [HTTP handler](https://github.com/mattermost/mattermost-plugin-playbooks/blob/f7fad63e8fcf4561f92079ea6db465fcfe02489e/server/api/playbooks.go#L871) is already wired up to return a 409 when it occurs. However, [UpdateRunNumberPrefix](https://github.com/mattermost/mattermost-plugin-playbooks/blob/f7fad63e8fcf4561f92079ea6db465fcfe02489e/server/app/playbook_service.go#L670) never actually returns it - it only checks for uniqueness, not whether runs already exist.

Should the service check `NextRunNumber > 1` and return this error when runs exist?

server/api/playbooks.go
```diff
@@ -751,6 +862,36 @@ func (h *PlaybookHandler) validateMetrics(pb app.Playbook) error {
...
+	if errors.Is(err, app.ErrReservedPropertyFieldName) {
+		h.HandleErrorWithCode(w, logger, http.StatusBadRequest, app.ErrReservedPropertyFieldName.Error(), err)
+		return true
+	}
+	return false
+}
+
+// validateTemplateWithFields validates a channel name template against a known field list and prefix.
+// Empty template is always valid. Pass nil fields to allow system tokens only (new-playbook case).
+func (h *PlaybookHandler) validateTemplateWithFields(w http.ResponseWriter, logger logrus.FieldLogger, template, prefix string, fields []app.PropertyField) bool {
```
edgarbellot: The GraphQL `UpdatePlaybook` mutation sends `ChannelNameTemplate` [directly to the store](https://github.com/mattermost/mattermost-plugin-playbooks/blob/f7fad63e8fcf4561f92079ea6db465fcfe02489e/server/api/graphql_root_playbook.go#L299) without going through the same validation - and since the frontend uses `@UpdatePlaybook` (GraphQL) to save the template, this is the primary code path in production.

Note: `graphql_root_playbook.go` wasn't modified in this PR, but the gap becomes more significant now that the validation logic exists and is applied elsewhere. Would it make sense to apply the same validation in the GraphQL handler as a follow-up, or is there a reason GraphQL intentionally skips it?

server/app/playbook_run_service.go
```diff
@@ -1061,6 +1098,8 @@ func (s *PlaybookRunServiceImpl) UpdateStatus(playbookRunID, userID string, opti
 		originalRun = playbookRunToModify.Clone()
 	}
 
+	options.Message = s.resolveMessageForRun(options.Message, playbookRunToModify)
```
edgarbellot, line 1171: This line resolves `{FieldName}` placeholders in the user's status update message before it's posted and [broadcast to external channels](https://github.com/mattermost/mattermost-plugin-playbooks/blob/f7fad63e8fcf4561f92079ea6db465fcfe02489e/server/app/playbook_run_service.go#L1130). That means the actual property values (e.g. `Classification: CONFIDENTIAL`) end up in channels that may have members who don't have access to the run itself.

A couple of questions: if a user writes `{SomeField}` without intending to use template expansion (just using curly braces for formatting), they'll get their message silently modified before it's sent - is that the intended UX? 

And for the broadcast case specifically, should there be a mechanism to warn the user that property values will be visible to a wider audience?

edgarbellot (changes requested):
@catalintomai great work! Just left a few clarifying questions, but overall looks good

server/app/playbook_service.go
catalintomai: I added the audit record to `UpdateRunNumberPrefix`. The `userID` parameter is used exclusively for the audit record (like in `Update`, `Archive`, and `Restore`)

server/app/errors.go
catalintomai: `ErrRunNumberPrefixImmutable` is a leftover from an earlier design where I considered making the prefix immutable once runs existed. The final design intentionally allows the prefix to change at any time — a playbook can start runs as INC-0001…INC-0005 and then switch to CHANGES-0006 onward. I removed both the sentinel and the handler.

server/api/playbooks.go
catalintomai: I moved the `ChannelNameTemplate` save path to REST, given that GraphQL is being deprecated and we should not add more to its codebase (is still used for other playbook fields, so we're not removing it yet fwik).

server/app/playbook_run_service.go
catalintomai: re: "This line resolves {FieldName} placeholders..." - The user sees the fully resolved message in the status update dialog before submitting, so they can edit or remove values before posting - user reviews the expanded message and explicitly submits it.
re: "if a user writes {SomeField}..." - a user writing {SomeField} for formatting purposes should see it unchanged in the posted message, as long as no property field happens to be named SomeField. 
re: "And for the broadcast case specifically," - agreed this is the broadcast exposure is a gap, but this predates this PR and needs further design (like warning in the dialog when the message contains resolved property values and broadcast is enabled??).

edgarbellot: The broadcast channel is pre-existing, but the automatic expansion of `{FieldName}` against `run.PropertyValues` is new in this PR ([playbook\_run\_service.go:1101](https://github.com/mattermost/mattermost-plugin-playbooks/blob/672c1b086fba09c0fc5e22bbb5d544b6897e09f3/server/app/playbook_run_service.go#L1101)). So the conduit is old, the sensitive payload flowing through it is new.

Picture a security incident playbook with broadcast enabled to `#incidents-announce` (200+ members, possibly contractors). An incident lead writes `"Investigating {AffectedSystem}, classification {Classification}, cause: {RootCause}"` thinking about their private channel. The resolved version (`"...payments-service-prod, Restricted, SQL injection in /api/payments"`) lands in the broadcast channel automatically: unpatched vuln, customer impact, and a Restricted field leaked. The compose-time preview only helps if the user is actively thinking about broadcast scope, which the UI doesn't currently surface.

My proposal, either works:

1.  Minimal guardrail in this PR: when `StatusUpdateBroadcastChannelsEnabled` is true and the resolved message differs from the raw template, show a one-line notice in the dialog (*This message contains run property values and will be posted to N broadcast channel(s)*). Extends your point nº1 to the broadcast case.
2.  Defer with a tracking issue filed and linked before merge so it doesn't get lost.

What do you think?

catalintomai: re: "which the UI doesn't currently surface." - while existing dialog already shows 'This update for the run X will be broadcasted to channel such and such' - they can get the actual channel names on hover - so the user should be somewhat informed of the scope, more warnings are welcome I guess. Either approach is fine, for 1 - @asaadmahmood , wdyt?

asaadmahmood: By warning do we mean showing a modal or something after the try to send the status update?

asaadmahmood: @catalintomai @edgarbellot ?

jgheithcock (approved):
I do not have any concerns that have not already been brought up.

server/app/playbook_run_service.go
catalintomai: @asaadmahmood re: "By warning do we mean showing a modal or something after the try to send the status update?" - not sure. From what I gather @edgarbellot suggests to augment the existing 'This update for the run X will be broadcasted to channel such and such' with a warning (see above).

The current UI looks this: 
Screenshot 2026-05-12 at 14 19 47

server/sqlstore/migrations.go
```diff
@@ -1739,6 +1739,42 @@ var migrations = []Migration{
...
+			if err := addColumnToPGTable(e, "IR_Incident", "RunNumber", "BIGINT NOT NULL DEFAULT 0"); err != nil {
+				return errors.Wrapf(err, "failed adding RunNumber to IR_Incident")
+			}
+			if err := addColumnToPGTable(e, "IR_Incident", "SequentialID", "VARCHAR(64) NOT NULL DEFAULT ''"); err != nil {
+				return errors.Wrapf(err, "failed adding SequentialID to IR_Incident")
+			}
+
+			// Partial functional index: case-insensitive prefix uniqueness per team,
+			// excluding archived playbooks (DeleteAt=0) and prefix-less ones (prefix<>'').
+			if _, err := e.Exec(createPartialUniquePGIndex(
```
JulienTant, line 1769: Should we consider a unique index on RunNumberPrefix+Sequence number?

JulienTant (approved):

lindy65:
Thanks @catalintomai 

1. Is the icon outlined in the screenshot supposed to be a dropdown arrow?
image

2. I'm testing _7. Prefix uniqueness — Try to configure two active playbooks in the same team with the same prefix. Verify the second save is rejected._ and am not able to insert the same prefix in the highlighted field in the screenshot below. I type it and it just disappears.

If this is what is supposed to be happening because 2 active playbooks in the same team with the same prefix cannot be created?  I think we need some form of error message or notification/indicator to the user that there is another playbook in the same team with the same prefix
image

I'm still in the midst of testing but want to confirm the behavior in 2 above before continuing as there are a lot of js.console errors.

Thanks!

catalintomai:
@lindy65 re: "Is the icon outlined in the screenshot supposed to be a dropdown arrow?" - it's a "Insert variable" button (not a dropdown indicator) - it should "say"/express:  "insert a template token" ({SEQ}, {OWNER}, {CREATOR}, and any custom property field names).

catalintomai:
@lindy65  - re. "If this is what is supposed to be happening because 2 active playbooks in the same team with the same prefix cannot be created?" - yes, this is by design. Agree on the silent failure, will add a visible error message.

lindy65:
> @lindy65 - re. "If this is what is supposed to be happening because 2 active playbooks in the same team with the same prefix cannot be created?" - yes, this is by design. Agree on the silent failure, will add a visible error message.

Hey @catalintomai 

A. Thanks for adding the error message.
1. Is it supposed to flash briefly on the screen? Or is it suppossed to stay visible and be dismissable?
    - Currently it is flashing briefly and is difficult to read in entirety in the short time it's visible
2. Also, I've created a new team on the spinwick which had no playbooks yet. I created the first playbook and tried the `Run number prefix` of `NO.` and the error message fired (which shouldn't happen in a new team with the first playbook being created
   - There are many errors in the js.console
   
image

B. Following on from A above, I tried to use a completely unique prefix that hadn't been used on either team and, at first, I got the same red error flash on the screen. The second time I added exactly the same prefix it stayed in the text box with no error flashing but, when I went to "run name template" and tried to add `{SEQ}` to the channel name, the prefix disappeared from the "Run number prefix" text box and I got the error outlined in the screenshot below. There are also errors in the js.console
image



> @lindy65 re: "Is the icon outlined in the screenshot supposed to be a dropdown arrow?" - it's a "Insert variable" button (not a dropdown indicator) - it should "say"/express: "insert a template token" ({SEQ}, {OWNER}, {CREATOR}, and any custom property field names).

C. Thanks - the hover-over text says `Insert variable` - is that correct?



I'm going to pause testing again while you take a look 

Please let me know when I can resume testing :) Thanks!

catalintomai:
Hi @lindy65 :

  A1 — toast flashes too briefly: I am reusing existing logic — toast duration is shared across the plugin. Happy to file a follow-up to make failure toasts persistent + dismissable if you'd like that scoped separately.
  A2 — NO. error on a fresh team: the prefix was being rejected because it contains a . (only letters, numbers, and hyphens are allowed), but the toast incorrectly said "already uses that prefix." Toasts should be now accurate:
  - duplicate prefix → "Another active playbook in this team already uses that prefix."
  - invalid characters → "Invalid prefix: must start and end with a letter or number and contain only letters, numbers, and hyphens."
 
  B — prefix disappears after editing the run name template: fixed. The prefix now stays put when you edit the template afterward. Added a regression test.
  C — "Insert variable" tooltip: correct — that button inserts a template token ({SEQ}, {OWNER}, {CREATOR}, custom property fields).

also, the .js AbortError flood is real noise but not breaking anything - they come from typing fast in the template field, each new request cancels the previous one.
 
Thanks!

lindy65:
Thanks for all that @catalintomai 

1. I tested running a playbook with 4 Attributes added and then added all the attributes to the "Run Name Template" field. (I'm not sure whether customers are going to do this but.... they might)
  - I hit an issue where, in the next screen after you click "Run" from the backstage, the long run name now exceeds the character limit of 64 characters.
  - Also, the bottom of the modal breaks when all the Attributes are added and the "Cancel" button does nothing. (I can close the modal using the "X" though).
     - Checking that it's correct that I cannot edit the run name from this screen because it's pulled from the "Run Name Template" in the backstage?
  image

2. Came across something that *might/could* happen from a user perspective. I had inserted a custom Attribute in the "Run Name Template" field in the backstage. When I went to edit that field due to the error in 1) above, I deleted one of the brackets on the variable in the field and, when I ran the playbook, the next screen didn't add the Attribute to be filled in by the user because the variable was not recognized due to the missing bracket
  - I know it might seem like a silly thing but... it happened to me 😆 
  - Is there any kind of check we can add to the "Run Name Template" field that will ensure all brackets are in place / variable names are correct?
     - I'm only bringing this to attention because that field is editable and mistakes happen when editing

image

3. Yes please, I think those error messages need to stay visible so the user is aware of what the problem is
> Happy to file a follow-up to make failure toasts persistent + dismissable if you'd like that scoped separately.

server/app/playbook_run_service.go
edgarbellot: Yep, my suggestion is to be more explicit about which channels the update will be broadcast to, so users are aware before posting and can avoid accidentally sharing sensitive information in unintended channels.

Approving the PR to avoid blocking.

edgarbellot (approved):

catalintomai:
Hi @lindy65 

1a. Run name too long — the modal now shows an inline error: "The resolved run name exceeds the 64-character limit. Edit the playbook template to use fewer or shorter fields."
1b. Cancel button / modal breaking — should be fixed - the modal body should scroll and the Cancel button stay pinned at the bottom regardless of how many attribute fields are shown.
1c. Can't edit run name — correct, by design. When a template is set the name is fully derived from {SEQ} + the attribute values, so the free-text field is intentionally read-only there.
2. Malformed brackets — already handled in the template editor with two inline warnings: one for unclosed braces ("Template has an unclosed variable…") and one for unrecognised field names ("{FieldName} will appear as literal text…"). The error surfaces before the user ever hits Run.

Thanks!

lindy65 (approved):
> 1a. Run name too long — the modal now shows an inline error: "The resolved run name exceeds the 64-character limit. Edit the playbook template to use fewer or shorter fields." 1b. Cancel button / modal breaking — should be fixed - the modal body should scroll and the Cancel button stay pinned at the bottom regardless of how many attribute fields are shown. 1c. Can't edit run name — correct, by design. When a template is set the name is fully derived from {SEQ} + the attribute values, so the free-text field is intentionally read-only there. 2. Malformed brackets — already handled in the template editor with two inline warnings: one for unclosed braces ("Template has an unclosed variable…") and one for unrecognised field names ("{FieldName} will appear as literal text…"). The error surfaces before the user ever hits Run.

Thanks for all the above @catalintomai - retested and did not find any other issues 👍

asaadmahmood (approved):
LGTM

catalintomai:
@coderabbitai - are all comments you generated for this PR been solved?

catalintomai:
@coderabbitai can you check if the issues in  https://github.com/mattermost/mattermost-plugin-playbooks/pull/2255#issuecomment-4548816121 are fixed or not?

catalintomai:
@coderabbitai, since "Would you like me to open a GitHub issue to track item #4 as a follow-up?" is outside, should you approve this PR?

PR #2256, opened 2026-05-04
CF-5: Create Placeholder for Task Assignment

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68471

## Checklist

Summary

  Adds role-based ("placeholder") task assignment for playbook checklist items. Instead of hard-coding a specific user at playbook authoring time, a task can be assigned to a role — Run Owner, Run Creator, or a User-type property field — which is resolved to a concrete user when the run starts and re-resolved when the underlying role changes.

  This eliminates the need to edit playbook templates whenever team membership changes and lets the same template route work to whoever holds a given role on each run.

  Changes

  Backend

  - Checklist items gain an assignment type (owner, creator, property_user) and an optional property-field reference, alongside the existing direct user assignment.
  - On run creation, role placeholders are resolved: owner → run owner, creator → run reporter, property_user → user resolved from the run-level property value.
  - When the run owner changes, owner tasks are re-resolved automatically; creator and manually assigned tasks are preserved.
  - Editing a single task no longer wipes the role-based metadata of sibling tasks.
  - Property-field references are remapped from playbook-level to run-level IDs when copying property fields onto the run.
  - assignee_id, assignee_type, and assignee_property_field_id are mutually exclusive on the assignee endpoint; switching modes clears the others.
  - Bug fix (pre-existing): `OwnerChanged` timeline events now set `SubjectUserID` to the new owner and `CreatorUserID` to the actor, matching the semantic contract for that event type.

  Frontend

  - New Assignee dropdown in the checklist item editor with three modes: specific user, role (Run Owner / Run Creator), or User-type property field.
  - The dropdown is available both in the playbook editor (template authoring) and inside an active run.
  - In a run, role-based tasks display the resolved user; switching to a specific user clears the role.

  How to Test

  1. Create a playbook with checklist tasks and assign them to: a specific user, "Run Owner", "Run Creator", and a User-type property field.
  2. Start a run and confirm each task shows the correct resolved user (owner → starter/owner, creator → starter, property → user from that property's value).
  3. Change the run owner from the RHS and verify Owner-typed tasks update to the new owner, while Creator-typed and manually assigned tasks stay unchanged.
  4. Edit one task's assignee in the run and confirm sibling tasks keep their role assignment.
  5. Switch a role-based task to a specific user (in both playbook editor and run) and confirm the role indicator is cleared.
  6. Change the value of the User-type property on the run and verify property-bound tasks reflect the new user.
  7. Archived playbook — confirm starting a run from it still fails cleanly.

  Security

  Authorization reuses the existing run/playbook permission model with one new access-control constraint added for the property_user path.

  - Playbook authors / members with edit access on a playbook: can set or change the assignment type (specific user, role, or property field) on template tasks.
  - Run participants and run admins: can change a run task's assignee, including switching between specific user, role, and property-field modes. The same handler that previously gated direct assignment now gates all three modes.
  - Run owner changes: only users authorized to change the owner can trigger automatic re-resolution of Owner-typed tasks; the system uses the new owner's identity, not the requester's.
  - Property-field assignment: the referenced field must exist on the run; requests targeting fields not bound to the run are rejected with 400. Field IDs are validated for format before lookup.
  - Mode exclusivity: the API rejects requests that set more than one of assignee_id / assignee_type / assignee_property_field_id, preventing ambiguous state.
  - Resolution uses stored identities: owner, creator, and property-user resolution read server-side run state, so a caller cannot impersonate another user by manipulating the assignment payload.
  - Auto-participant side effect (property_user): when a user-type property field is set to a user who is not yet a run participant, that user is automatically added to the run and notified. Because any run participant can write property values, this could otherwise allow a regular participant to invite arbitrary team members. The auto-add is therefore gated: only the run owner or a system admin triggers membership expansion; other participants can still set the property value but the auto-add is skipped for them.


catalintomai:
@coderabbitai review

catalintomai:
@coderabbitai review

server/app/playbook_run_service.go
```diff
@@ -2141,7 +2166,203 @@ func (s *PlaybookRunServiceImpl) SetAssignee(playbookRunID, userID, assigneeID s
...
+	if resolvedUserID == "" {
+		s.pluginAPI.Log.Debug("property field has no user value set; task will be unassigned", "field_id", runFieldID, "run_id", playbookRunID)
+	}
+
+	var originalRun *PlaybookRun
+	if s.configService.IsIncrementalUpdatesEnabled() {
+		originalRun = playbookRun.Clone()
+	}
+
+	noChangeNeeded := applyPropertyUserAssigneeUpdate(itemToCheck, propertyFieldID, resolvedUserID)
```
JulienTant: We should use the `runFieldID` here so we never have to match against the parent id.

server/api_playbooks_test.go
```diff
@@ -1616,6 +1616,77 @@ func TestPlaybooksDuplicate(t *testing.T) {
...
+
+		duplicated, err := e.PlaybooksClient.Playbooks.Get(context.Background(), newID)
+		require.NoError(t, err)
+		require.Len(t, duplicated.Checklists, 1)
+		require.Len(t, duplicated.Checklists[0].Items, 4)
+		assert.Equal(t, "", duplicated.Checklists[0].Items[0].AssigneeType)
+		assert.Equal(t, "owner", duplicated.Checklists[0].Items[1].AssigneeType)
+		assert.Equal(t, "creator", duplicated.Checklists[0].Items[2].AssigneeType)
+		assert.Equal(t, "property_user", duplicated.Checklists[0].Items[3].AssigneeType)
+		assert.Equal(t, fakeFieldID, duplicated.Checklists[0].Items[3].AssigneePropertyFieldID)
```
JulienTant: That's a mistake (in the prod code) - The new playbook must have a dedicated field ID - we must create of copy of the fields so each playbook (and run) is independant.

We have something similar going on with conditions

server/api/playbook_runs.go
```diff
@@ -477,11 +476,7 @@ func (h *PlaybookRunHandler) createPlaybookRun(playbookRun app.PlaybookRun, user
 		return nil, errors.Wrap(app.ErrMalformedPlaybookRun, "must provide team or channel to create playbook run")
 	}
 
-	if playbookRun.OwnerUserID == "" {
```
JulienTant, line 485: I think that's a collateral damage?

server/app/playbook_run_service.go
```diff
@@ -1854,22 +1885,36 @@ func (s *PlaybookRunServiceImpl) ChangeOwner(playbookRunID, userID, ownerID stri
...
+		updatedRun = playbookRunToModify
+	}
+	s.sendPlaybookRunObjectUpdatedWS(playbookRunID, originalRun, updatedRun)
 
 	// Mark success and add result state for audit
 	auditRec.Success()
-	model.AddEventParameterToAuditRec(auditRec, "oldOwnerId", oldOwner.Id)
-	model.AddEventParameterToAuditRec(auditRec, "newOwnerId", newOwner.Id)
+	if oldOwner != nil {
+		model.AddEventParameterToAuditRec(auditRec, "oldOwnerID", oldOwner.Id)
```
JulienTant: Not sure if it matters but `oldOwnerId`>`oldOwnerID` might have an impact

JulienTant (changes requested):

server/api/playbook_runs.go
catalintomai: Intentional, not collateral damage. The hardcoded `OwnerUserID`: `request.UserId` was removed from `createPlaybookRunFromDialog` in server/api/playbook_runs.go — the dialog submission handler — flagged by coderabbitai in r3182511175. That value was bypassing the new API-layer resolution logic (explicit value → `DefaultOwnerID` → current user), which only fires when `OwnerUserID` is empty. Once removed, the early rejection of an empty `OwnerUserID` had to go with it, otherwise every dialog-created run would be rejected before reaching the resolution logic.
 
The connection to this feature: `AssigneeType: "owner" `tasks resolve at run creation time to `run.OwnerUserID`. If the dialog always forced the submitter as owner, then owner-type tasks in a playbook with a configured `DefaultOwnerID` would resolve to the wrong person. The fix ensures owner resolution is consistent across the REST and dialog creation paths, which is necessary for the owner placeholder to behave correctly regardless of how the run was started.

server/app/playbook_run_service.go
catalintomai: Reverted "oldOwnerID" back to "oldOwnerId" and "newOwnerID" back to "newOwnerId" to preserve the existing audit log keys.

server/api_playbooks_test.go
catalintomai: Should be fixed — Duplicate now calls `remapAssigneePropertyFieldIDs` with the field mappings returned by `CopyPlaybookPropertiesToPlaybook`, in the same update pass that already remaps condition IDs. The test was also corrected: it previously used a fake hardcoded "field-123" ID which had no mapping and just round-tripped through unchanged. It now creates a real property field, uses its actual ID, and asserts the duplicated item references a new distinct field ID belonging to the new playbook.

server/app/playbook_run_service.go
catalintomai: Should be fixed — `applyPropertyUserAssigneeUpdate` now receives `runFieldID` instead of `propertyFieldID`, so `item.AssigneePropertyFieldID` always stores the run-level field ID. Added a regression test as well.

jgheithcock (approved):
I've nothing else to add not already covered.

jgheithcock (approved):

JulienTant (approved):

lindy65:
Hey @catalintomai,

I'm not sure whether I'm not understanding how this feature should work or if something is missing...

In the backstage, I have set up `user`-type attributes
image

I have added various `user`-type attributes to tasks on the playbook template in the backstage
image

When I run the playbook, the `user`-type attributes are listed in the RHS as well as on the tasks. There is no text field to the right of the `user`-type attribute to add a team mate name though? In the test steps above, step 6 says "Change the value of the User-type property on the run and verify property-bound tasks reflect the new user."

I'm not seeing a way to do this?

image

I can re-assign the task to an actual user in this modal but that removes the `user`-type attribute which seem counter-intuitive if we're wanting all tasks with the attribute to change to the team member who is, for example, the QA person?
image

Please let me know how this should be working because I'm a bit confused! Thanks :)

lindy65 (approved):
Thanks @catalintomai - that fixed it! Looks good to me now :)

catalintomai:
Hi @lindy65 , it might be buried in the description/test steps, but there is more to it. The user attribute could be set as the assignee when creating/editing a task for a playbook/playbook run as well (not only in RHS). 
Screenshot 2026-05-13 at 22 08 39

lindy65:
> Hi @lindy65 , it might be buried in the description/test steps, but there is more to it. The user attribute could be set as the assignee when creating/editing a task for a playbook/playbook run as well (not only in RHS). Screenshot 2026-05-13 at 22 08 39

Thanks @catalintomai - I did test this during my review as well 👍

catalintomai:
@dryrunsecurity nit drs_719a4034 This is by design and consistent with the existing pattern throughout the plugin. All service methods that mutate a run use the same read-modify-write pattern via UpdatePlaybookRun — there is no optimistic concurrency control anywhere in the store layer for playbook runs. The race window for task assignee updates is narrow and the last-write-wins outcome is acceptable for this use case. Addressing this across the board would require a schema change and is out of scope for this PR.

catalintomai:
@coderabbitai review

catalintomai:
@coderabbitai Thanks for the finding. After investigation, this is not a bug in the current implementation.

`CopyPlaybookPropertiesToRun` only copies property **field definitions** from the playbook to the run — it does not copy property **values**, because playbooks don't have property values (values only exist at the run level). At run creation time, `GetRunPropertyValues` would always return empty, so calling `resolvePropertyUserAssignmentsFromRun` there would be a no-op.

The correct behavior is exactly what `TestRunCreationFromPlaybook_PropertyUserAssigneeType` documents: `AssigneeID` starts empty at creation and is resolved later when someone calls `SetRunPropertyValue` on the run.

edgarbellot (approved):
LGTM, thank you!

asaadmahmood:
@catalintomai minor feedback on this, but I'll be pushing a fix for this that you may just need to double check.
https://www.loom.com/share/c89d5ce9b28441ed96a759dc5ca66e24

asaadmahmood (approved):

catalintomai:
@coderabbitai review

webapp/src/components/rhs/properties/property_user.tsx
```diff
@@ -0,0 +1,265 @@
...
+                            <Profile
+                                userId={uid}
+                                withoutName={true}
+                            />
+                            <RemoveButton
+                                aria-label={formatMessage({id: 'playbooks.property_user.remove_user', defaultMessage: 'Remove user'})}
+                                onClick={() => handleRemoveUser(uid)}
+                            >
+                                <CloseIcon size={12}/>
+                            </RemoveButton>
```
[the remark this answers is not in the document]
catalintomai: @coderabbitai I believe this was fixed. can you check?

catalintomai:
@coderabbitai re: "Requested changes" - are the more changes? Or are all addressed?

catalintomai:
@coderabbitai - can you check if the issues in https://github.com/mattermost/mattermost-plugin-playbooks/pull/2256#issuecomment-4547611622 are fixed or not?

catalintomai:
@coderabbitai re: "No remaining open issues from the previous CHANGES_REQUESTED review. The PR looks good to proceed." - then why not approve the PR?

PR #2257, opened 2026-05-04
CF-9: Restrict Playbook Edits to Playbook Admin

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68475

## Checklist

Restrict Playbook Edits to Playbook Admins

  Summary

  Adds an opt-in "admin-only edit" mode for playbooks. When enabled, only playbook admins and system admins can modify a playbook's configuration — regular members get read-only access. The toggle is surfaced in the playbook editor's Settings section and is visible only to users who have permission to change it.

  ---
  Changes

  Backend

  - Added AdminOnlyEdit boolean field to the playbook model, stored via a new DB migration.
  - Introduced a PlaybookEdit permission check that gates all write operations (update, duplicate, delete, archive, property management, property field reorder, conditions) behind the new flag. When AdminOnlyEdit is false, the existing PlaybookManageProperties logic applies unchanged.
  - Added IsPlaybookAdmin helper that resolves the admin role from DefaultPlaybookAdminRole or falls back to the built-in playbook admin role.
  - Toggling the AdminOnlyEdit flag itself is additionally restricted to playbook admins and system admins, preventing privilege escalation in both directions.

  Frontend

  - Playbook editor derives a canEdit flag from the playbook's admin_only_edit field and the current user's role. Non-admin members see the title, description, and all sections as read-only when the lock is active.
  - New AdminOnlyEditToggle component renders a toggle with help text inside a "Settings" section; the section is only shown to playbook admins and system admins.
  - Optimistic UI update with rollback on save failure.

  ---
  How to Test

  1. Enable the lock as a playbook admin
    - Open a playbook as a playbook admin → go to the Outline tab → locate the Settings section → enable "Only admins can edit this playbook" → confirm the toggle saves.
  2. Verify read-only for regular members
    - Log in as a member who is not a playbook admin → open the same playbook → confirm the title, description, and all sections are not editable.
  3. Playbook admin retains edit access
    - Log in as a playbook admin → open the locked playbook → confirm edits to title, description, and checklist items are still possible.
  4. System admin always has edit access
    - Log in as a system admin → open the locked playbook → confirm all fields are editable and the Settings section with the toggle is visible.
  5. Disable the lock
    - As a playbook admin or system admin, disable the toggle → confirm regular members can edit again.
  6. Non-admin cannot toggle the lock
    - Log in as a regular member → open the locked playbook → confirm the Settings section is not visible (toggle is hidden from non-admins).
  7. Verify archive is blocked for regular members
    - With the lock enabled, log in as a regular member → attempt to archive the playbook (via the playbook list dot-menu) → confirm the archive action is denied.
  8. Verify conditions are blocked for regular members (API)
    - With the lock enabled, as a regular member send `POST /plugins/playbooks/api/v0/playbooks/{id}/conditions` → confirm a 403 response. As a playbook admin the same request must succeed with 201.

  ---
  Security

  - System admins always retain full edit access, regardless of the AdminOnlyEdit flag.
  - Playbook admins (members with the admin role on the playbook) can edit the playbook and toggle the lock on or off.
  - Regular playbook members become read-only when the lock is active. They cannot enable or disable the lock.
  - Unauthenticated / non-member users are unaffected — existing access controls still gate who can view a playbook at all.
  - The permission to flip the AdminOnlyEdit flag is evaluated against the old state of the playbook, preventing a regular member from enabling the lock and then being unable to disable it due to a race on their own permissions.


client/playbook.go
```diff
@@ -117,6 +118,7 @@ type PlaybookCreateOptions struct {
 	RemoveChannelMemberOnRemovedParticipant bool                   `json:"remove_channel_member_on_removed_participant"`
 	ChannelID                               string                 `json:"channel_id" export:"channel_id"`
 	ChannelMode                             ChannelPlaybookMode    `json:"channel_mode" export:"channel_mode"`
+	AdminOnlyEdit                           bool                   `json:"admin_only_edit"`
```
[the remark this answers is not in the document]
catalintomai: Not addressing this intentionally. The permission model defines editors as `playbook_admin members ∪ system admins` — when the former set is empty, system admins can still edit and can promote a member to playbook admin at any time. "Only system admins can edit" is a valid, intentional configuration (e.g. a tightly governed playbook locked down at creation), not a stranded state. Rejecting it at the API layer would add a constraint that doesn't exist in the underlying permission model and would block legitimate use cases.

catalintomai:
Thanks for the analysis, DryRun. This is working as intended — here's the reasoning.

**The design intent of `AdminOnlyEdit`**

`PlaybookManageProperties` grants edit access to two distinct groups:
1. Users whose **playbook-level role** has `PermissionPrivatePlaybookManageProperties` granted (`RolesGrantPermission`)
2. Users with **team-level** `PermissionPrivatePlaybookManageProperties` (`HasPermissionToTeam`)

`PlaybookEdit` with `AdminOnlyEdit=true` is **deliberately more restrictive**, not less. It intentionally bypasses group 2 (team-level RBAC holders) and limits edits exclusively to system admins and playbook admins. A team admin who would normally pass `PlaybookManageProperties` via `HasPermissionToTeam` is *denied* by the new code path.

**The edge case the scanner found**

The only scenario where `AdminOnlyEdit=true` could be less restrictive than `PlaybookManageProperties` is if:
- The `PlaybookRoleAdmin` role has had `PermissionPrivatePlaybookManageProperties` explicitly revoked (non-default configuration), AND
- The playbook admin has no team-level `PermissionPrivatePlaybookManageProperties` either

In that corner case, `PlaybookManageProperties` would deny the playbook admin, but `PlaybookEdit` would still allow them. This is intentional: `AdminOnlyEdit` means "**playbook admins own the edit rights to this playbook**." The feature defines edit access at the playbook-admin level, independent of how manage-properties permissions are wired in the broader RBAC system. Revoking ManageProperties from the admin role in RBAC while keeping the user as a playbook admin would be a contradictory configuration.

**Summary**: The bypass is by design. `AdminOnlyEdit=true` creates a stricter, playbook-scoped authorization policy that supersedes (rather than extends) the team-level RBAC fallback. Marking this as false positive for `drs_6c0751d7`.

server/api/playbooks.go
```diff
@@ -651,6 +651,10 @@ func (h *PlaybookHandler) duplicatePlaybook(c *Context, w http.ResponseWriter, r
 		return
 	}
 
+	if !h.PermissionsCheck(w, c.logger, h.permissions.PlaybookEdit(userID, playbook)) {
```
JulienTant, line 676: Just making sure this is desired - we will only allow admins to duplicate a playbook?

JulienTant (changes requested):
Request change as my question introduce a breaking change; If I'm not missing something, with this change only admins will be able to duplicate a playbook - no matter if said playbook is locked or unlocked.

server/api/playbooks.go
catalintomai: @JulienTant , as said in the description "that gates all write operations (update, duplicate, delete, property management, conditions) behind the new flag." I guess it depends if the definition of edit covers these actions as well. One could say that duplicate is more of a create operation than an edit one, so I see your point. cc: @DSchalla - any input re: customer needs?

DSchalla: Duplication is OK - Key need is to prevent operators (playbook users) modifying the process defined by officers and zone leads (playbook admins). I assume duplication can also be prevented by taking away the create playbook permission, correct?

Thanks folks.

jgheithcock (approved):
I don't have anything to add to Julien's and CodeRabbit's comments.

JulienTant (approved):

lindy65:
Thanks @catalintomai 

SCENARIOS 1-8 ALL DISCOVERED *BEFORE* PLAYBOOK WAS RUN FROM THE BACKSTAGE (when the "only admins can edit this playbook" toggle is set to true/enabled). If you'd like to take a look at these issues to confirm which are valid, once fixed, I'll re-test as well as test access to editing by an ordinary member of the playbook during a run.

1. When the "only admins can edit this playbook" toggle is set to true/enabled, and I view the playbook in the backstage as an ordinary member of the playbook:
    - I see Attributes section of the playbook is still clickable - should it be greyed out at this level?
    - If I try to add an attribute, I get a very long error message at the bottom of the screen that isn't very helpful (see below)

image


2. When the "only admins can edit this playbook" toggle is set to true/enabled, as an ordinary member of the playbook, I can edit the description(?) field below the playbook's name in the backstage - should this be blocked - no "edit pencil" visible on the right?

image

image

3. When the "only admins can edit this playbook" toggle is set to true/enabled, should an ordinary member of the playbook be able to edit where the run starts? In the screenshot below, the playbook admin had set it to create a new channel. The ordinary member changed it to run linked to Town Square

image

4. As an ordinary member of the playbook, I can bulk edit tasks when the "only admins can edit this playbook" toggle is set to true/enabled - which means I can edit tasks, assign them to people, delete them, etc. (2nd screenshot below, I've deleted tasks that were visible in 1st screenshot below)

image

image

5. As an ordinary member of the playbook, I can edit the Retrospective template and it is saved, when the "only admins can edit this playbook" toggle is set to true/enabled

image

image 

6. As an ordinary member of the playbook I can configure metrics in the retrospective from the "Reports" tab of the playbook when the "only admins can edit this playbook" toggle is set to true/enabled. When clicking on "Save" it is not saved and also not added to the playbook section in the backstage (correct) but... I should not be able to *edit* this section. There are some js.console errors

image

image

image

7. Correctly, when the "only admins can edit this playbook" toggle is set to true/enabled, I cannot duplicate a playbook as an ordinary member. However, I **can** export the playbook. Should this be allowed?

image

8. As I understand it, an ordinary member *should be able to* **import** a playbook and this **is** working. However, when I view the imported playbook, I should not be able to see the "Settings" section showing the field "Only admins can edit this playbook" right? (before I exported the playbook, this setting was set to true from the playbook admin's account).
    - Or is this field now accessible because I **imported** a playbook and am now "owner" of it?

image

catalintomai:
Hi @lindy65 ,
  1. Attributes section — All attribute interactions (adding, editing, reordering) should be now disabled for non-admin members. 
  2. Description field — The description field should be now read-only for non-admin members, consistent with the title and other fields.
  3. "Where the run starts" — All channel-mode controls should be now read-only for non-admin members.
  4. Bulk edit tasks — The bulk-edit button should be now hidden for non-admin members, and the checklist itself should be read-only.
  5. Retrospective template — The retrospective template field should be now read-only for non-admin members.
  6. Metrics in retrospective — All metric editing controls should be now disabled for non-admin members.
  7 (export): Export is intentionally allowed. The admin_only setting restricts editing the playbook definition, not reading it. Members can already view the full playbook in the editor, so downloading that same data as JSON is consistent with their read access.
  8. Correct behaviour. When a user imports a playbook they become the playbook admin of the new copy, which is why the Settings section is visible to them. The admin_only setting from the original is preserved on import (like other behavioral settings), but since the importer is the sole admin they always retain full edit access and can change the setting if they wish.

Thanks!

lindy65:
Thanks @catalintomai 

I've retested the scenarios above and everything is fixed/working as expected.

However, going through the process of retesting, I kept wondering whether "Attributes" tab should be "hidden" from ordinary users if "only admins can edit this playbook = true/enabled. Being able to see this tab doesn't make sense because I can't do anything on the screen but it says "No attributes yet - Add custom attributes to capture additional information about your playbook runs."
image

I also wondered about the "Configure metrics_" button in "Reports" tab - I can click on it and it takes me to the Retrospective section of the playbook in the backstage but I cannot edit anything (which is correct but I'm wasting time clicking on buttons that don't let me do anything as an ordinary member)
image

TO MAKE IT CLEAR TO ORDINARY USERS THAT THE PLAYBOOK IS EDITABLE ONLY BY A PLAYBOOK ADMIN, HOW ABOUT WE ADD SOME SORT OF TAG OR INDICATOR TO THE PLAYBOOK IN THE BACKSTAGE? I KNOW THIS WASN'T IN SCOPE OF THIS PR BUT... IT SOLVES A LOT OF THE PROBLEMS WHERE THE ORDINARY USER MIGHT NOT UNDERSTAND WHY THEY CAN CLICK ON THINGS BUT THEN DO NOTHING. I WAS THINKING OF A TAG TO THE LEFT OF THE PLAYBOOK NAME IN "VIEW ALL" SECTION OF BACKSTAGE OR PERHAPS JUST A "LOCK" ICON WITH A TOOLTIP "ONLY ADMINS CAN EDIT THIS PLAYBOOK". SO SOMEWHERE LIKE I'VE INDICATED BELOW ON THE SCREENSHOT. THOUGHTS?
image

As I mentioned yesterday, the previous issues were discovered *before* the playbook was run. I have seen a couple of issues with the playbook still not run by the owner but from the ordinary member's view in backstage > view all. "only admins can edit this playbook = true/enabled:

1. As an ordinary member, I "join" the playbook from the "View all" in backstage
    - From the 3-dot menu to the right of "Run" button, should "Edit" and "Archive" actions be visible? I can go through the flow of archiving the playbook but nothing happens. This might be confusing for the user as there's no feedback and I don't know that the playbook admin has set "only admins can edit this playbook" to true/enabled.
    - If I click on "Edit" then click on the playbook name drop-down menu, I see "Manage Access", "Rename", "Archive" - all actions I cannot successfully do but the flow exists and it *looks* like it's working but the update isn't effected. The only process that *is* actually blocked is renaming but I can still click on it, only to find out that I cannot rename.

image

image

image

2. With "only admins can edit this playbook = true/enabled, as an ordinary member I "**Run**" the playbook from the "View all" tab in backstage.
    - I *should* be able to *run* the playbook, correct?
    - When I ran the playbook, I changed the channel it is connected to - should I be able to?
    - In the channel, it still shows Lindy as the admin (correct on the original playbook) and sysadmin and user-1 as participants
       - What should the behavior be here? Should user-1 who "Ran" the playbook from the backstage not be the playbook owner now? (perhaps not because the playbook is set to "only admins can edit this playbook")
          - However, I can change/update Owner/Participants from within the run channel (which I shouldn't be able to do if the playbook is set to "only admins can edit this playbook", correct?)
            - I can also make a participant a run owner or remove them from the run altogether
          
image

image

image

image


I was thinking that if I (an ordinary member) run a playbook from the backstage that is set to "only admins can edit this playbook", I wouldn't be able to make any changes to anything in the playbook in the run channel but I can make all sorts of changes  and I'm not sure if I should or should *not* be able to. 
  - I can rename the playbook, assign people to tasks, delete tasks, finish the run, fill in and publish the retrospective, etc.


Maybe I am getting confused so I'll pause testing for now until you've had a chance to answer my questions above.

server/api/playbooks.go
edgarbellot, line 619: The [`Duplicate` handler](https://github.com/mattermost/mattermost-plugin-playbooks/blob/a26ba1771a7a02342aca5729d01c044a0d706c44/server/api/playbooks.go#L654) was gated with `PlaybookEdit` to prevent non-admin members from cloning an admin-locked playbook, but `exportPlaybook` here only requires `PlaybookViewWithPlaybook`. Combined with [`importPlaybook`](https://github.com/mattermost/mattermost-plugin-playbooks/blob/a26ba1771a7a02342aca5729d01c044a0d706c44/server/api/playbooks.go#L699-L704), which sets the importer as the sole admin of the new playbook, and the fact that `AdminOnlyEdit` carries the [`export:"admin_only_edit"` tag](https://github.com/mattermost/mattermost-plugin-playbooks/blob/a26ba1771a7a02342aca5729d01c044a0d706c44/server/app/playbook.go#L80), the export+import flow becomes functionally equivalent to `Duplicate` but without the edit gate.

Would it make sense to mirror the `Duplicate` check here when `AdminOnlyEdit` is true (or alternatively drop the `export:` tag on `AdminOnlyEdit` so the lock configuration doesn't travel with the export)? Otherwise a non-admin member of an admin-locked playbook can export and re-import to obtain a near-identical copy where they hold sole admin rights.

catalintomai:
Hi @lindy65

  Attributes tab / Configure metrics button — The attributes empty state now shows a read-only message instead of an action prompt, and the "Configure metrics" button is hidden for non-admins. Lmk if you're still seeing the old behavior.

  Lock icon suggestion — great idea, will need to address it in a separate PR, out of scope here.

  Edit/Archive in the dot-menu after joining — these items should appear disabled for non-admins. 

  Rename/Archive/Manage Access inside the editor — Rename and Archive should be disabled for non-admins. "Manage Access" is controlled by a separate permission unrelated to `admin_only_edit`, so it's intentionally not affected.

  Running a playbook and in-run actions — `admin_only_edit` only restricts editing the playbook template (its configuration). It does not restrict what happens inside a run (the run instance). So yes, ordinary members can start runs, change the run channel, assign owners/participants, complete tasks, publish the retrospective, etc.
  
  Thanks!

server/app/playbook.go
```diff
@@ -77,6 +77,8 @@ type Playbook struct {
 	// ChannelMode is the playbook>run>channel flow used
 	ChannelMode ChannelPlaybookMode `json:"channel_mode" export:"channel_mode"`
 
+	AdminOnlyEdit bool `json:"admin_only_edit" export:"admin_only_edit"`
```
edgarbellot: Since [`AdminOnlyEdit`](https://github.com/mattermost/mattermost-plugin-playbooks/blob/a26ba1771a7a02342aca5729d01c044a0d706c44/server/app/playbook.go#L80) is a plain `bool` without `omitempty`, and the [store `Update` writes it unconditionally](https://github.com/mattermost/mattermost-plugin-playbooks/blob/a26ba1771a7a02342aca5729d01c044a0d706c44/server/sqlstore/playbook.go#L712), any `PUT /playbooks/{id}` from a client that doesn't know about this new field will deserialize it as `false` and silently disable the lock. The [flip-gate in `PlaybookModifyWithFixes`](https://github.com/mattermost/mattermost-plugin-playbooks/blob/a26ba1771a7a02342aca5729d01c044a0d706c44/server/app/permissions_service.go#L233-L237) will pass as long as the requester is admin, even if disabling the lock was never their intent (e.g. a legacy script or integration using the pre-PR `app.Playbook` shape).

Would it be worth making this a `*bool` and only applying it when non-nil, or alternatively reading the old value when the request body doesn't explicitly include the field?

edgarbellot (changes requested):
@catalintomai great work! Left two suggestions :)

server/api/playbooks.go
catalintomai: @edgarbellot  - changed `AdminOnlyEdit` to `export:"-"` so the lock flag is excluded from exported JSON. Imported playbooks always start unlocked with the importer as sole admin. Export+import is a template-sharing mechanism, not a clone, so carrying the lock over doesn't make sense.

server/app/playbook.go
catalintomai: @edgarbellot  - `updatePlaybook` now reads the raw body first and checks whether `admin_only_edit` was present in the JSON keys. If the field is absent, the old value is preserved — so legacy clients or integrations that predate this field can't accidentally disable the lock. Added a test that issues a `PUT` without `admin_only_edit` and asserts the lock survives.

lindy65 (approved):
Thanks so much @catalintomai and sorry I got confused about normal user being able to run the playbook and change things in the actual run!

You've addressed all the issues I've found so I think it's good from my side 👍 

I've created https://mattermost.atlassian.net/browse/MM-68966 to address the "lock icon" suggestion

edgarbellot (approved):
LGTM! 🚀

asaadmahmood (approved):
LGTM

catalintomai:
@coderabbitai review

catalintomai:
@coderabbitai re: "Requested changes" - are there more changes? Or are all addressed?

catalintomai:
> @coderabbitai can you check if all the issues at https://github.com/mattermost/mattermost-plugin-playbooks/pull/2257#issuecomment-4547594218 have been addressed?

catalintomai:
@coderabbitai re: "No remaining open concerns from my previous review. The PR looks good to go from a review-comment perspective. 🎉" - then why don't you approve the PR?

PR #2258, opened 2026-05-04
CF-4: Restrict Playbook Finish to Run Owner

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68470

## Checklist

Summary

  Adds a new playbook setting — Restrict finish/restore/reassign to run owner — that, when enabled, limits who can finish, restore, or reassign ownership of a run to the run owner only. System admins can always bypass this restriction. Playbook admins additionally retain the ability to reassign run ownership even when the setting is on.

  ---
  Changes

  Backend

  - New OwnerGroupOnlyActions boolean field on the Playbook model, persisted via a schema migration.
  - New permission checks — RunFinish, RunRestore, RunChangeOwner — that enforce the owner-only gate when the flag is set. Bypass logic for system admins (all actions) and playbook admins (owner reassignment only).
  - The finish, restore, and change-owner API routes are moved out of the checkEditPermissions middleware and given their own explicit permission checks.
  - Toggling OwnerGroupOnlyActions is itself restricted to playbook admins and system admins.

  Frontend

  - New Settings section in the Playbook Editor with a toggle to enable/disable owner-only actions.
  - Finish and restore buttons in both the run view and the RHS are hidden or shown based on the new permission. Blocked users see a tooltip explaining the restriction.
  - System admins bypass the participant gate in the RHS so they see action controls directly.
  - New useIsBlockedByOwnerOnlyForFinishRestore hook encapsulates the client-side permission logic.

  ---
  How to Test

  1. Create a playbook. In the editor, open the Settings section and verify the Restrict finish/restore/reassign to run owner toggle is visible only to playbook admins and system admins.
  2. Enable the toggle and start a run. As a regular participant (not the owner), verify the Finish button is hidden in both the run view and the RHS. Hover over the disabled resume control on a finished run and confirm the tooltip says only the owner can restore.
  3. As the run owner, verify you can still finish and restore the run normally.
  4. As a system admin (not the run owner), verify you can finish, restore, and reassign ownership despite the restriction.
  5. As a playbook admin (not the run owner), verify you can reassign ownership but cannot finish or restore the run.
  6. Disable the toggle and verify all participants regain the ability to finish and restore the run immediately (no per-run grandfathering).
  7. Verify that regular participants (non-admins) cannot toggle the setting — the control should be read-only or hidden for them.

  ---
  Security

  - Run owner: can always finish, restore, and reassign ownership regardless of the OwnerGroupOnlyActions flag.
  - System admin: bypasses all owner-only restrictions for finish, restore, and reassign. A warning is logged when a system admin performs an owner-only action they do not own.
  - Playbook admin: can reassign run ownership even when OwnerGroupOnlyActions is enabled and they are not a run participant. Cannot finish or restore runs they do not own when the flag is set.
  - Regular participant: when OwnerGroupOnlyActions is enabled, cannot finish, restore, or reassign ownership unless they are the run owner.
  - Toggling the setting is restricted to playbook admins and system admins at the API layer — regular members with
  ManageProperties access cannot change it.
  - The flag is evaluated against the current playbook state, not a snapshot at run creation — enabling or disabling it takes effect immediately on all in-progress runs.


catalintomai:
Fixed — the `updateStatusDialog` handler now returns `403 Forbidden` for both `ErrNoPermissions` and `ErrNotFound` when `FinishRun` is true, so there is no longer a status-code difference an attacker could use to enumerate run IDs. See the updated block at `server/api/playbook_runs.go` lines 1088–1091.

server/app/permissions_service.go
```diff
@@ -668,3 +681,112 @@ func (p *PermissionsService) isChannelArchived(channelID string) bool {
...
+	if baseErr == nil {
+		return nil
+	}
+
+	if run == nil {
+		return baseErr
+	}
+	// Playbook admins can reassign ownership even if they are not a run participant —
+	// enabling legitimate handoffs when the original owner is unavailable.
+	if playbook != nil && p.IsPlaybookAdmin(userID, *playbook) {
```
JulienTant: It used to be that a playbook admin who is NOT a participant could not change it the owner. 

With the current code, it does not matter if `OwnerGroupOnlyActions` is true or false, they will be able to change the owner. If it was not intended we're probably missing a `p.OwnerGroupOnlyActions` - if it was intended the comment above the function is incorrect:

```
// When OwnerGroupOnlyActions is set, playbook admins may change ownership even without [...]
```

JulienTant (changes requested):

server/app/permissions_service.go
catalintomai: @JulienTant  - You are correct. `OwnerGroupOnlyActions` should be part of the condition (and it was) , removed by a coderabbit-induced regression change.  Updated the code and added a test.

JulienTant (approved):

jgheithcock (approved):
Nothing to add. LGTM

lindy65:
Thanks @catalintomai 

When a user who does not have permissions to finish a run opens a run in the RHS of a channel, "Time to wrap up?" and "Finish" flash briefly at the bottom of the screen before disappearing. Is there a way to prevent this from happening? (I'm just thinking of the case where this user might think something is wrong when they see something flash then disappear and not understanding why)
image

In this flow,
- Active run with "Restrict run management actions for non-owners" = true/enabled
- Normal member of the playbook is blocked / cannot see "Finish" in the RHS
- Playbook owner toggles "Restrict run management actions for non-owners" = false/disabled mid-run
- "Time to wrap up?" / "Finish" are not immediately visible to normal member of the playbook
   - They are visible after a refresh but should preferably be visible immediately?

In this flow (user-1 is playbook owner and leaves the channel),
- Observed by sysadmin in the playbook run channel: user-1 is still playbook owner so sysadmin changes ownership to herself (change is visible in the RHS)
- User-1 rejoins the playbook run channel and is shown as a playbook participant (correct) and should not be able to assign themself to the role of owner right? But they can.
   - In the playbook in the backstage, "Restrict run management actions for non-owners" is still true/enabled and Sysadmin is the owner (screenshot 1)
   - In user-1's playbook RHS, they've changed themselves to playbook owner and "Time to wrap up?" / "Finish" are visible (screenshot 2)

image

image

lindy65:
Thanks @catalintomai
 
I saw new commits on the PR so I retested the scenarios I mentioned [here](https://github.com/mattermost/mattermost-plugin-playbooks/pull/2258#issuecomment-4459252584) but they are not yet fixed.

Please let me know when I can retest :) Thanks!

catalintomai:
Hi @lindy65 
re: "Finish" flash briefly at the bottom of the screen" - do you still see it? Please do a hard browser refresh and try again and lmk.

re: "They are visible after a refresh but should preferably be visible immediately" - this is a known limitation - there is no `playbook_updated` WebSocket event only `playbook_run_update` exists - there should be other  examples (like renaming a key metric in the retrospective section on a playbook that has run with retrospective enabled). Solution would be to add WebSocket event that refreshes everyone at once, but that is beyond the scope of this PR. 

re: "and should not be able to assign themself to the role of owner right?" - if I understand the scenario correctly -  this should fall into the "playbook admins can reassign run ownership even when OwnerGroupOnlyActions is on" (see PR "Security" section). The rationale is a recovery path when the original owner is unavailable — but it also means a playbook admin who is also a run participant can hand ownership back to themselves.

Thanks!

server/app/permissions_service.go
```diff
@@ -669,3 +682,112 @@ func (p *PermissionsService) isChannelArchived(channelID string) bool {
...
+	if baseErr == nil {
+		return nil
+	}
+
+	if run == nil {
+		return baseErr
+	}
+	// When OwnerGroupOnlyActions is set, playbook admins can reassign ownership even if they
+	// are not a run participant — enabling legitimate handoffs when the original owner is unavailable.
+	if playbook != nil && playbook.OwnerGroupOnlyActions && p.IsPlaybookAdmin(userID, *playbook) {
```
edgarbellot, line 802: [`ChangeOwner`](https://github.com/mattermost/mattermost-plugin-playbooks/blob/af2300def5851181f700f1c7b0f73d47e48adc86/server/app/playbook_run_service.go#L1843-L1849) only checks that `ownerID` exists via `User.Get`, not that they are a member of the run's team. [`AddParticipants`](https://github.com/mattermost/mattermost-plugin-playbooks/blob/af2300def5851181f700f1c7b0f73d47e48adc86/server/app/playbook_run_service.go#L4083-L4104) silently drops non team members into `usersFailedToInvite` and returns nil, but `OwnerUserID` is then assigned regardless.

With this new bypass plus `OwnerGroupOnlyActions=true`, a playbook admin could hand off ownership to any Mattermost userID that isn't in the run's team. That user can't see the run's channel, and the flag blocks every other participant from finishing or restoring, so the run is effectively locked until a sysadmin intervenes.

Would it make sense to validate team membership for `ownerID` before updating the run?

server/app/permissions_service.go
```diff
@@ -669,3 +682,112 @@ func (p *PermissionsService) isChannelArchived(channelID string) bool {
...
+	if err := p.runManagePropertiesWithPlaybookRun(userID, run); err != nil {
+		return run, playbook, errors.Wrapf(err, "cannot %s run %s", actionName, runID)
+	}
+	if playbook == nil || !playbook.OwnerGroupOnlyActions { // nil == channel checklist or standalone run
+		return run, playbook, nil
+	}
+	if run.OwnerUserID == userID {
+		return run, playbook, nil
+	}
+	if IsSystemAdmin(userID, p.pluginAPI) {
```
edgarbellot, line 760: The new sysadmin and playbook admin bypasses are logged via `logrus.Warn` with structured fields, while the downstream action goes through `MakeAuditRecord` ([finishPlaybookRun](https://github.com/mattermost/mattermost-plugin-playbooks/blob/af2300def5851181f700f1c7b0f73d47e48adc86/server/app/playbook_run_service.go#L1219), [restorePlaybookRun](https://github.com/mattermost/mattermost-plugin-playbooks/blob/af2300def5851181f700f1c7b0f73d47e48adc86/server/app/playbook_run_service.go#L1446), [changePlaybookRunOwner](https://github.com/mattermost/mattermost-plugin-playbooks/blob/af2300def5851181f700f1c7b0f73d47e48adc86/server/app/playbook_run_service.go#L1802)). The override is traceable across both channels, but the audit record itself doesn't reflect that a security gate was overridden.

Would it be worth adding a `bypassReason` parameter to the audit records so the override shows up directly in the audit trail?

edgarbellot (changes requested):

server/app/permissions_service.go
catalintomai: @edgarbellot - should be addressed in the latest commit — ChangeOwner now validates team/channel membership before assigning. A test covering the scenario was added too.

server/app/permissions_service.go
catalintomai: @edgarbellot - I believe that threading a bypass reason from the permission layer down to MakeAuditRecord would touch the audit infrastructure across multiple operations and is out of scope here. As mentioned above, the logrus.Warn already fires at the exact moment the gate is bypassed, which provides the traceability.

server/app/permissions_service.go
edgarbellot: thank you!

server/app/permissions_service.go
edgarbellot: agreed. I saw you added `ownerUserID`, which allows some correlation. All good from my side 🚀

edgarbellot (approved):

lindy65 (approved):
> Hi @lindy65 re: "Finish" flash briefly at the bottom of the screen" - do you still see it? Please do a hard browser refresh and try again and lmk.

Thanks @catalintomai - this seems fixed now.

I've retested and didn't find any further issues

asaadmahmood (approved):
LGTM

PR #2259, opened 2026-05-04
CF-7: Add tasks to playbooks usage view

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68473

## Checklist

Summary

  Adds a Tasks column to the Playbooks run list (Runs view in Backstage). Each row now shows how many checklist tasks are complete out of total, along with a progress bar — giving users a quick at-a-glance view of run progress without opening the run.

  ---
  Changes

  Backend

  - PlaybookRun struct gains two computed (non-persisted) fields: task_total and task_completed.
  - A ComputeTaskProgress() method calculates these from the run's checklists at read time: hidden items are excluded, skipped and closed items count as completed.
  - Progress fields are populated on run creation, retrieval (GetPlaybookRun, GetPlaybookRuns), and timeline post addition.
  - Incremental WebSocket updates include these fields when they change, so the run list reflects task changes in real time.

  Frontend

  - New TaskProgress component renders the X/Y tasks label and a progress bar.
  - The component is displayed in a new Tasks column in the runs list, between the run metadata and the owner column.
  - Header and row grid columns are resized to accommodate the new column.
  - TypeScript types updated to include the optional task_total / task_completed fields on PlaybookRun.

  ---
  How to Test

  1. Navigate to Playbooks → Runs.
  2. Verify a new Tasks column appears in the run list header.
  3. Open a run that has checklist items; return to the Runs list and confirm the Tasks column shows the correct X/Y tasks count and a filled progress bar.
  4. For a run with no checklist items, confirm the Tasks cell is empty (no bar, no label).
  5. Check a checklist item as complete or skipped inside a run; return to the list and verify the Tasks counter and bar update in real time (without a page reload).
  6. For a run with hidden checklist items, confirm those items do not count toward the total.

  ---
  Security

  - Run participants and channel members: Can view task progress for any run visible to them through the standard run-list permissions — no change to existing access control.
  - Playbook members / admins: Can view all runs they already have access to; task progress follows the same visibility rules.
  - The two new fields are computed server-side from the run's checklist data, which is already fetched under existing authorization checks. No new data is exposed beyond what a user could see by opening the run itself.
  - Hidden checklist items (controlled by condition rules) are excluded from both the count and the total, so their existence is not leaked.


webapp/src/components/backstage/runs_list/task_progress.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+const TaskProgress = ({taskTotal, taskCompleted}: Props) => {
+    const {formatMessage} = useIntl();
+
+    if (taskTotal == null || taskTotal === 0) {
+        return null;
+    }
+
+    const completed = Math.max(0, taskCompleted ?? 0);
```
JulienTant: would `taskCompleted ?? 0` not be enough? Do we have a chance of having a negative number?

webapp/src/components/backstage/runs_list/task_progress.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const pct = Math.round(Math.max(0, Math.min((completed / taskTotal) * 100, 100)));
+    const label = formatMessage(
+        {id: 'pqR8tZ', defaultMessage: '{completed, number}/{total, number} tasks'},
```
JulienTant: We usually don't set IDs manually

JulienTant (changes requested):
Request change for the ID in the tsx, everything else is ok. I'm not a big fan of how many "manual" call we have to do in playbooks to compute data... but It's the pattern so can't really criticize it too much - I had to do it as well 😅

server/app/playbook_run_service.go
```diff
@@ -79,6 +79,8 @@ func (s *PlaybookRunServiceImpl) sendPlaybookRunObjectUpdatedWS(playbookRunID st
 			logger.WithError(err).Error("failed to get current state of playbook run")
 			return
 		}
+	} else {
+		currentRun.ComputeTaskProgress()
 	}
```
JulienTant, line 88: I would suggest as a safety measure to also compute task progress on "Previous" - I don't think we missed a callsite... but better be safe then sorry


```suggestion

  if previousRun != nil {
      previousRun.ComputeTaskProgress()
  }
```

JulienTant (approved):
LGTM

jgheithcock (approved):
Nothing to add. LGTM

lindy65:
Thanks @catalintomai 

I was wondering how "skipped" tasks are counted so I asked Claude and found out that they are counted as being checked off as "done" towards the total tasks. Is this correct?

Also, some inconsistencies Claude noticed in the unit & E2E tests - not sure whether you'd like to address them?
image

 Other than the above, looks good to me.

catalintomai:
Hi @lindy65 , re:  " I was wondering how "skipped" tasks are counted" - Skipping a task is interpreted here as "this doesn't apply / won't be done", so it's treated as resolved (not pending). I can see it going the other way as well, I agree - cc @DSchalla.

I believe I also addressed CC comments - see last commit.
Thanks.

lindy65:
Thanks @catalintomai - I'll go ahead an approve this PR but feel free to add me again for review if things change and need re-testng :)

lindy65 (approved):
Thanks Catalin, looks good to me 👍

DSchalla:
> Hi @lindy65 , re: " I was wondering how "skipped" tasks are counted" - Skipping a task is interpreted here as "this doesn't apply / won't be done", so it's treated as resolved (not pending). I can see it going the other way as well, I agree - cc @DSchalla.
> 
> I believe I also addressed CC comments - see last commit. Thanks.

I think either is totally fine, just should be documented. I could see value on mouse-over to have a legend of 15 completed, 3 skipped, 5 pending and indicate it with a striped bar for the skipped ones, but that's just future refinement. Thanks.

catalintomai:
@coderabbitai review

edgarbellot (approved):
LGTM!

asaadmahmood (approved):
LGTM

PR #2260, opened 2026-05-04
CF-10: Ability to fully disable Retrospectives

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68476

## Checklist

Ability to fully disable retrospectives per run

  Summary

  Adds the ability to enable or disable the retrospective section on a per-run basis, independent of the playbook-level
  retrospective_enabled setting. When disabled on a run, retrospective reminders are suppressed, the retrospective UI is hidden, and a timeline event is recorded. Re-enabling restores reminder scheduling for finished runs.

  Changes

  Backend

  - New ToggleRetrospectiveEnabled service method that persists the toggle, manages reminder scheduler side-effects, records timeline events, and broadcasts a WebSocket update.
  - New RunToggleRetrospective permission check: only the run owner or a system admin may change this setting.
  - New REST endpoint (PUT /api/v0/runs/{id}/retrospective-enabled) wired to the service method with the above permission guard.
  - Two new timeline event types: retrospective_enabled and retrospective_disabled.

  Frontend

  - New Enable/Disable Retrospective item in the run's context (dot) menu, visible only to users who pass the permission check (run owner or system admin).
  - Confirmation modal before applying the change (separate copy for enable vs disable).
  - useCanToggleRunRetrospective hook that mirrors the server-side permission rule.
  - Playbook editor's retrospective section accepts a disabled prop so input controls can be disabled independently of the playbook's retrospective_enabled field (used for the archived-playbook state). When retrospective_enabled is false on a playbook, the section renders a "no retrospective expected" placeholder instead of the editing controls.

  How to Test

  1. Open a run you own. Navigate to the Run Details Page (backstage). Click the dot menu (⋮) in the run header — confirm "Disable Retrospective" appears.
  2. Click Disable Retrospective, confirm the modal, and verify the retrospective section disappears from the run details page.
  3. Re-open the dot menu — confirm it now shows "Enable Retrospective". Click it and verify the retrospective section reappears.
  4. Log in as a non-owner participant. Open the same run's dot menu — confirm the toggle option is not present.
  5. Log in as a system admin on a run you do not own — confirm the toggle option is present and works.
  6. On an active run, disable its retrospective from the dot menu. Then finish the run. Verify that no retrospective reminder is posted to the channel.
  7. Re-enable the retrospective on that finished run (or on a finished run that still has retrospective enabled) — verify a reminder is posted immediately (if a reminder interval is configured on the playbook).

Security

  - Run owner: can toggle retrospective on/off for their own runs.
  - System admin: can toggle retrospective on any run, regardless of ownership.
  - Participants / channel members / other admins (non-system): no access — the toggle is not shown in the UI and the API returns 403.
  - Permission is enforced server-side in a dedicated permission check; the UI gate is additive only.


catalintomai:
@coderabbitai review

catalintomai:
@coderabbitai review

server/app/permissions_service.go
```diff
@@ -654,6 +654,19 @@ func GetRequesterInfo(userID string, pluginAPI *pluginapi.Client) (RequesterInfo
...
 }
 
+// RunToggleRetrospective checks whether userID can enable or disable the retrospective for runID.
+// Only the run owner or a system admin may change this setting.
+func (p *PermissionsService) RunToggleRetrospective(userID, runID string) error {
+	run, err := p.runService.GetPlaybookRun(runID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get run %s for permission check", runID)
+	}
+	if run.OwnerUserID == userID || IsSystemAdmin(userID, p.pluginAPI) {
```
JulienTant: Well that escalated quickly :P Is there nobody in between the run owner and sysadmin that should be able to flip the switch

JulienTant (approved):

server/app/permissions_service.go
catalintomai: @JulienTant - I went back and forth on this. Retrospectives are run-level outcomes, so it might make sense for the run owner (who knows the incident context) to decide. 
However, for compliance/governance reasons one could see playbook owner being added to the set. That said, playbook admins who want to toggle must become owner first, and owner-change is a legitimate admin action (they will need to add themselves as participants first though). That way we keep a simple permission model, while still opening the door to governance scenarios.
I can still see it going either way, if the change owner flow is considered to be too cumbersome. cc: @DSchalla for more context.

DSchalla: I think it's fine to start with that model given playbook admins can overwrite the owner, and that way its an intentional decision. We can always expand the permissions later while taking them away might be confusing.

jgheithcock (approved):
Nothing to add.

lindy65:
Thanks @catalintomai 

>How to Test
>1. Open a run you own. Open the run dot menu — confirm Disable Retrospective appears.

Which dot menu is this exactly Catalin? In the run in channel, I don't see a way to disable Retrospective directly
image

Should the instruction be to go to the run name drop-down menu or am I missing a dot menu it should appear on somewhere?
image

>6. Finish a run, then disable its retrospective. Verify no retrospective reminder is posted to the channel.

This is not working for me Catalin. The retrospective reminder *is* posted in the channel. (This might just be a "bad" test instruction because I would expect the retrospective post to be made in channel if Retrospective is still enabled when I "Finish" the playbook)
I ran a playbook with Retrospective = true/enabled. I finished the playbook and then disabled the retrospective. The post was made in the channel as soon as I clicked "Finish"
image

catalintomai:
@lindy65 re: On "which dot menu": The toggle lives on the Run Details Page (the backstage RDP view), not on channel posts. Navigate to the run's detail page, then click the run name dropdown at the top of the page — that's where "Disable Retrospective" appears. The channel-post three-dot menu is a separate component and does not include
  this option.
  
Screenshot 2026-05-19 at 15 12 40

catalintomai:
Hi @lindy65 , re: "I finished the playbook and then disabled the retrospective. " - one needs to disable first, then finish. Your sequence (finish → disable) is actually testing a different case: the reminder fires at finish-time because retrospective was still enabled when you clicked Finish, which is correct behavior.

That said, I made a change to remove the Enable/Disable Retrospective menu option for the channel's run RHS panel menu, that is because current master implementation does not have a retrospective section (unlike the one in the Run Details (backstage) page) - adding one would go beyond the scope of the current feature. 
Also, now the retrospective section should be present/absent from the RDS as the menu option is selected (before one would have had to refresh the page).
Thanks!

lindy65:
Thanks @catalintomai 

In this flow:
- Start a run with Retrospective = false / disabled
- When run is active, go to backstage and update Retrospective to = true / enabled
Observed: "Recent Activity" section shows a date/time stamp for the update but does not show what was done (see below)

image

In this flow:
- Start a run with Retrospective = false / disabled
- When run is active, go to backstage and update Retrospective to = true / enabled
- Finish run and click on "Yes, start retrospective" button on the post made in channel
Observed: Taken to the backstage but position is at the top of the playbook, not at the Retrospective section (previoiusly, in other playbooks, when clicking this button, I was taken to the Retrospective section in the backstage

image

After the flow above, I scroll down to the Retrospective area. Noticing that the field / input box is very narrow horizontally. It would be very nice if after clicking "Yes, start retrospective", I am taken to the Retrospective area in the backstage and my mouse is positioned in the field/input box (activated)
  - Is this possible?
     - If not, I think the field height should at least be increased

image

Once I've clicked in the narrow message input box, it does expand to a bigger box:
image

Bearing the above flow in mind from start of playbook, if I write the Retrospective and click on "Publish" there is no update in "Recent Activity" in the RHS and no post is made in channel. The message input field for the retrospective returns to it's narrow width and I get no feedback the retrospective was posted.

image

image

image


If I go back to the backstage and open the "finished" playbook then click on "Publish" retrospective AGAIN after the above flow, I get a modal "Are you sure you want to publish?"
  - I click on "Publish" and then I see it has been published and there is confirmation of this in the Recent Activity in the RHS
  - If I then go back to the channel I see the Retrospective published in channel

image

image

image

lindy65:
Thanks @catalintomai 

I've retested and have one remaining question...

I'm seeing that the message input box height for the Retrospective in the backstage is still very narrow. Were you not able to increase the height? Or do not think it's an issue?
  - It would also be very nice if focus could be in the message input box when I'm taken to the Retrospective section of the playbook backstage (after clicking "Yes, start retrospective" button in run channel - although this might be out of scope of this PR)

image

catalintomai:
Hi @lindy65 . The third item in your list above. https://github.com/mattermost/mattermost-plugin-playbooks/pull/2260#issuecomment-4496024977 - it is existing behavior, independent of the work in this PR - that includes the small box that can get expanded. With the latest commit I went beyond the scope of this PR and added a live feedback to the "Publish" action - now the button is greyed out in the run, you don't have to refresh or navigate away to a different screen. I suggest to open a bug on Asaad to maybe re-think the user experience for retrospectives, since it looks to me (and you) that it is lacking in some areas.

Thanks!

lindy65:
Thanks @catalintomai - looks good to me then.

I'll mention @asaadmahmood here on this PR so he can view all my comments about user experience for Retrospectives and decide whether an improvement ticket is necessary from a design perspective :)

lindy65 (approved):
Thanks @catalintomai - looks good to me now 👍

asaadmahmood (approved):
This LGTM, but I agree with @lindy65 that we need to improve the narrow textarea experience.

server/app/permissions_service.go
```diff
@@ -655,6 +655,19 @@ func GetRequesterInfo(userID string, pluginAPI *pluginapi.Client) (RequesterInfo
 	}, nil
 }
 
+// RunToggleRetrospective checks whether userID can enable or disable the retrospective for runID.
+// Only the run owner or a system admin may change this setting.
+func (p *PermissionsService) RunToggleRetrospective(userID, runID string) error {
```
edgarbellot, line 661: The new permission check only validates `OwnerUserID == userID || IsSystemAdmin`, but it does not verify that the user still has access to the run's channel. Since `OwnerUserID` is not cleared when a user is removed from the channel, an ex-member who is still the run owner can hit `PUT /runs/{id}/retrospective-enabled` and, on a finished+unpublished run, trigger [postRetrospectiveReminder](https://github.com/mattermost/mattermost-plugin-playbooks/blob/05207f3440e92bb37c7ef426133d34482e4e317b/server/app/playbook_run_service.go#L1727) which posts a `@channel` message into a channel they no longer belong to. Combined with toggling off/on repeatedly, this enables notification abuse from outside the channel.

Would it make sense to also require channel access here, similar to how other run mutators gate via `RunManageProperties`? Something like:

```go
if err := p.checkRunAccess(userID, run); err != nil {
    return err
}
if run.OwnerUserID == userID || IsSystemAdmin(userID, p.pluginAPI) {
    return nil
}

File: server/app/playbook_run_service.go [hunk 3]
```diff
@@ -1555,7 +1555,91 @@ func updateChecklistAndItemTimestamp(checklist *Checklist, item *ChecklistItem,
 	checklist.UpdateAt = timestamp
 }
 
-// GraphqlUpdate updates fields based on a setmap
+func (s *PlaybookRunServiceImpl) ToggleRetrospectiveEnabled(playbookRunID, userID string, enabled bool) error {
```
edgarbellot, line 1711: Just a suggestion: this new write path can mutate a run whose channel is archived (`DeleteAt > 0`), creating a fresh timeline event, a WS broadcast, and potentially a bot `@channel` post in a closed channel. The archived-channel guard in [runManagePropertiesWithPlaybookRun](https://github.com/mattermost/mattermost-plugin-playbooks/blob/05207f3440e92bb37c7ef426133d34482e4e317b/server/app/permissions_service.go#L475-L479) only applies to channelChecklist runs today, so this is consistent with the existing pattern for playbook runs rather than a regression introduced here.

However, the inmutability of artifacts tied to archived channels might be important for regulated environments. Would it make sense to short-circuit the toggle when `isChannelArchived(run.ChannelID)` returns true? It would also avoid the edge case where the re-enable branch posts a bot reminder into an archived channel.

edgarbellot (changes requested):
Thank you Catalin! Left a couple of suggestions.

server/app/permissions_service.go
catalintomai: @edgarbellot - fixed `RunToggleRetrospective` to require channel membership before evaluating the owner check — system admins bypass this. Added an E2E test that removes the owner from the channel and asserts the toggle returns `403` with state unchanged.

server/app/playbook_run_service.go
catalintomai: @edgarbellot - `ToggleRetrospectiveEnabled` now short-circuits early if the channel is archived, before any DB write, timeline event, WS broadcast, or bot post. Non-admin owners get a `400 Bad Request`; system admins bypass the guard (consistent with the channel-membership bypass in `RunToggleRetrospective`). The UI also hides the toggle for non-sysadmins when the channel is archived. Covered by unit and integration tests.

edgarbellot (approved):
Looking good! thank you :)

PR #2261, opened 2026-05-04
CF-11: Allow Playbooks only in New Channel

catalintomai:
## Summary

## Ticket Link
https://mattermost.atlassian.net/browse/MM-68477

## Checklist

Summary

  Adds a new playbook setting — "Require new channel for all runs" — that enforces channel creation on every run. When enabled, the "Link existing channel" option is hidden/disabled in the run start modal, and a hint informs the user that the playbook policy requires a new channel.

  ---
  Changes

  Backend

  - New NewChannelOnly boolean field on the Playbook model, persisted via a DB migration adding a column to IR_Playbook.
  - Validation at the API layer rejects a NewChannelOnly=true playbook whose ChannelMode is set to link an existing channel.
  - Run creation rejects any attempt to supply a channel_id when NewChannelOnly is enabled on the referenced playbook.
  - Field is included in playbook export/import.

  Frontend

  - New toggle in the playbook editor (Actions section): "Require new channel for all runs", with a confirmation dialog before enabling.
  - Run start modal respects new_channel_only: forces create_new_channel mode, disables the "Link to existing channel" radio button, and shows an explanatory hint below it.
  - New i18n strings added.

  ---
  How to Test

  1. Enable the setting: Open a playbook, go to the Actions section, toggle on "Require new channel for all runs" — confirm the dialog.
  2. Start a run: Click "Start a run" on the playbook. Verify the "Link to existing channel" radio button is disabled, "Create a run channel" is pre-selected, and the hint message is visible below the disabled radio.
  3. Try linking: Attempt to switch the channel mode to "Link existing channel" — it should be prevented.
  4. Start the run: Complete the modal and verify a new channel is created as expected.
  5. Disable the setting: Toggle off in the editor; start a run and verify both channel mode options are available again.
  6. Export/import round-trip: Export a playbook with new_channel_only=true, import it, and verify the setting is preserved.

  ---
  Security

  - Playbook editors/admins: Only members with edit access to the playbook can enable or disable the "Require new channel" setting. This matches the existing permission model for all playbook settings.
  - Run participants / non-editors: The policy is enforced server-side at run creation — even a direct API call supplying a channel_id will be rejected if the playbook has NewChannelOnly=true. The UI restriction is a UX convenience, not the sole enforcement boundary.
  - No new permission roles or privilege escalation paths are introduced. The feature gates behavior, not access.


catalintomai:
@coderabbitai review

catalintomai:
@coderabbitai review

webapp/src/components/modals/run_playbook_modal.tsx
```diff
@@ -300,30 +312,41 @@ type channelProps = {
...
-const ConfigChannelSection = ({teamId, channelMode, channelId, createPublicRun, onSetCreatePublicRun, onSetChannelMode, onSetChannelId}: channelProps) => {
+const ConfigChannelSection = ({teamId, channelMode, channelId, createPublicRun, newChannelOnly = false, onSetCreatePublicRun, onSetChannelMode, onSetChannelId}: channelProps) => {
     const {formatMessage} = useIntl();
     const createNewChannel = channelMode === 'create_new_channel';
-    const linkExistingChannel = channelMode === 'link_existing_channel';
+    const linkExistingChannel = channelMode === 'link_existing_channel' && !newChannelOnly;
     return (
         <ChannelContainer>
-            <ChannelBlock>
+            <ChannelBlock aria-disabled={newChannelOnly}>
```
JulienTant: if it's not an option, why bother showing all of this anyway?

JulienTant (approved):
Looks good, just left a small UX remark

catalintomai:
@coderabbitai review

webapp/src/components/modals/run_playbook_modal.tsx
catalintomai: @JulienTant  the prevalent pattern in the code seems to be greyed-out/disabled pattern (see public_private_selector.tsx, broadcast_channels_selector.tsx), not the "hide the section" one.

jgheithcock (approved):
I'm fine with this but I do want to leave a note about a GraphQL issue Cursor found:

`UpdatePlaybook` in server/api/graphql_root_playbook.go builds a `setmap` (including `ChannelMode` and `ChannelID`) and calls `playbookStore.GraphqlUpdate` directly. That path **does not call** `playbookService.Update`, so it **never** runs `ValidateNewChannelOnlyMode`.

So a client with permission to use GraphQL playbook mutations could leave `NewChannelOnly = true` (unchanged in that mutation) and set `ChannelMode `to link-existing (or otherwise move to an incompatible mode), producing a playbook row that REST/playbookService.Update would refuse to save.

I am okay with this as I'm hoping we can continue our work deprecating GraphQL as the only plugin using this API.

lindy65:
Thanks @catalintomai 

I can select "Link to an existing channel" after I have toggled "Require new channel for all runs"
image

If I click "Run" from the above screen, the next modal *does* show that a new channel needs to be created
image

A run is created in the previously set-up new channel "require new channel for all runs"
image

In a second test, I selected "require new channel for all runs" as well as "link to existing channel" = Town Square then ran the playbook and a new channel is still created (even although I didn't set it up / give it a name). So it looks like the flow is correct when running a playbook (even although "link to existing channel" is selected in backstage).

*Perhaps the best fix is to block/disable any input into "link to existing channel" toggle/drop-down if "require new channel for all runs" is selected? Not sure whether a tooltip is also a good idea to be displayed on "link to existing channel" toggle if "require new channel for all runs" is already selected?*

image

image

lindy65 (approved):
Thanks @catalintomai - the issues I saw before are fixed/working as expected.

catalintomai:
@coderabbitai review

asaadmahmood:
@catalintomai I'm facing an issue where this setting is not working.
https://www.loom.com/share/0a189a0c10314435b827b6ee1ed98b2a

asaadmahmood (changes requested):
See my issue above.

catalintomai:
> See my issue above.

Hi @asaadmahmood . I added an integration test that covers your refresh scenario, just to make sure - tried it locally and could not repro. I see that you are running locally, so it is likely that you hit some server error (my suspicion is that the DB schema does not include the newly added column that stores the setting) and the toggling is done in the UI but the setting is not saved at the server, so when you refresh the browser- it picks up the original setting (false). I recommend trying this in the cloud - that should run all the DB migrations and be closer to a prod. scenario.

asaadmahmood (approved):
LGTM

qsq8t65pxs:
@coderabbitai autofix stacked pr

server/app/playbook.go
edgarbellot, line 807: Since `NewChannelOnly` is a server-side security policy toggle, would it make sense to also expose it in [Playbook.Auditable()](https://github.com/mattermost/mattermost-plugin-playbooks/blob/b3054bd8164e0d2914a5cfd295f9eef3f7712041/server/app/playbook.go#L805-L815)? Right now the audit records emitted from Create / Update / Import (and from [CreatePlaybookRun](https://github.com/mattermost/mattermost-plugin-playbooks/blob/b3054bd8164e0d2914a5cfd295f9eef3f7712041/server/app/playbook_run_service.go#L391) for the attached playbook) go through this method, so the before/after value of the new toggle is not captured anywhere in the audit stream, only in the DB row itself.

edgarbellot (approved):
LGTM 👍  left one non-blocking suggestion.

PR #2263, opened 2026-05-12
MM-69624: Run/Playbook reports — Markdown · HTML · PDF export

calebroseland:
## Summary
Adds first-class export for Playbook Runs and Playbook templates in three formats — Markdown, HTML, and PDF — directly from the existing export modal. Markdown and HTML are always available on every deployment; PDF renders server-side via a Gotenberg sidecar when configured, and otherwise falls back transparently to the browser's print dialog (Save-as-PDF) using the same styled HTML. The result is a real audit-grade run report — KPI tiles, status pills, threaded transcript, typed timeline events, sanitized inline markdown — without requiring infrastructure changes from deployments that can't or won't run a sidecar.

Architecturally, `html_writer` is the single source of truth for visual fidelity (consumed both by the `.html` endpoint and by the Gotenberg adapter, with a byte-equality test enforcing it), `markdown_writer` is canonical for text portability with format rules pinned in code, and `HTMLPdfRenderer` is a thin pluggable interface so swapping the PDF engine doesn't ripple through the data layer. Security gating, audit logging, CSRF, permission scoping, and webhook redaction all carry through from the existing handler; the `.html` response ships a strict CSP and self-contained inline CSS so the browser-print path never reaches the network.

https://github.com/user-attachments/assets/74ba9104-4b44-491f-b849-3de069afb4da

[Example Run Report.pdf](https://github.com/user-attachments/files/27643584/Example.Run.Report.pdf)

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69624

## Prior
Supersedes the client-side POC in #2145 (`@react-pdf/renderer`), which was abandoned due to bundle-size, fidelity, performance, and data-exposure issues — see that PR's review thread for the original constraints this iteration addresses.


Willyfrog:
would it make sense to attach it to the channel? that way we can generate it and is already available for anyone involved

calebroseland:
> would it make sense to attach it to the channel? that way we can generate it and is already available for anyone involved

@Willyfrog if you mean posting the report to the channel (parallel to publishing the retrospective), I could definitely see the value in that—but because channel membership and run participation access are not mutually guaranteed, some UX or permissions refinement might be necessary.

Willyfrog:
> > would it make sense to attach it to the channel? that way we can generate it and is already available for anyone involved
> 
> @Willyfrog if you mean posting the report to the channel (parallel to publishing the retrospective), I could definitely see the value in that—but because channel membership and run participation access are not mutually guaranteed, some UX or permissions refinement might be necessary.

I thought about it, but if you are in the channel you already can see anything related to the run, right? Anyway, just an idea :)

calebroseland:
> > > would it make sense to attach it to the channel? that way we can generate it and is already available for anyone involved
> > 
> > 
> > @Willyfrog if you mean posting the report to the channel (parallel to publishing the retrospective), I could definitely see the value in that—but because channel membership and run participation access are not mutually guaranteed, some UX or permissions refinement might be necessary.
> 
> I thought about it, but if you are in the channel you already can see anything related to the run, right? Anyway, just an idea :)

Generally speaking, yes—but not always. If you start a run of a private playbook linked to a public channel, or if you invite someone to the private channel (of which theres's a linked private playbook run)—you would have access to the channel content, but not the (private playbook) run itself. 

I think it's a fantastic idea -- we could simply show the download link in the footer of the retrospective post if you have permission to access it :)

calebroseland:
@asaadmahmood a couple of questions:

1. Do we need markdown support? If only PDF is needed, I could trim a good bit of code to lighten the PR up.
2. Playbook export -- is this needed or useful? If not, I could trim or defer as a followup.
3. Export/download Dialog feedback -- should there be one, standard preset with hardcoded options, or are the optional sections/inclusions a good feature? Are custom presets needed? What sections are needed or wanted/unwanted?
4. Download flow and storage format -- is PDF/A a requirement? If so, could you provide specifics (e.g., PDF/A-1b, format or other requirements, etc.). Is direct-download a hard requirement? (The current flow uses a print&save-as-pdf flow with an optional opt-in deployment of https://gotenberg.dev/ for direct-download support)
5. PDF Layout design feedback -- fonts/colors, panel layouts, conversation UX, etc.

asaadmahmood:
@calebroseland 

> 1. Do we need markdown support? If only PDF is needed, I could trim a good bit of code to lighten the PR up.

Lets remove the markdown support for now.

> 2. Playbook export -- is this needed or useful? If not, I could trim or defer as a followup.

What do you mean by playbook export?

> 3. Export/download Dialog feedback -- should there be one, standard preset with hardcoded options, or are the optional sections/inclusions a good feature? Are custom presets needed? What sections are needed or wanted/unwanted?

We can have all of them on by default. Whatever you have in your demo is sufficient for now.

> 4. Download flow and storage format -- is PDF/A a requirement? If so, could you provide specifics (e.g., PDF/A-1b, format or other requirements, etc.). Is direct-download a hard requirement? (The current flow uses a print&save-as-pdf flow with an optional opt-in deployment of https://gotenberg.dev/ for direct-download support)

> 5. PDF Layout design feedback -- fonts/colors, panel layouts, conversation UX, etc.

We can have that separately, and we can add this feature in a feature flag on the playbooks plugin and have it off by default.

So we can try to get it in for the customer that needs it right now as soon as possible, and work on any improvements after.

PR #2269, opened 2026-05-15
ci: split lint into parallel go/web jobs and cache build artifacts

JulienTant:
## Summary
The CI `lint` job runs ~7m, with `make check-style` consuming ~6m of that (cold golangci-lint + tsc + tool installs every run). This PR splits the work into two parallel jobs and persists the caches that make repeat runs cheap.

- Split `make check-style` into `check-style-web` and `check-style-server` Makefile targets; the existing `check-style` still wraps both for local use.
- Replace the single `lint` GitHub Actions job with `lint-go` and `lint-web` running in parallel. Downstream jobs (`dist`, `dist-fips`, etc.) now wait on both. A no-op `lint` aggregator job is kept so existing branch protection rules requiring a check named `lint` keep working.
- Cache `bin/` (GOBIN), `~/.cache/go-build`, `~/.cache/golangci-lint`, `~/go/pkg/mod` for the go job, and `.eslintcache`, `.stylelintcache`, `.tsbuildinfo` for the web job.
- Pin `modd@v0.8` (was `@latest`) so the go-tools cache key is stable.
- Enable `incremental` + `tsBuildInfoFile` in `webapp/tsconfig.json` so the cached tsc build info is actually consumed.

`go vet ./...` is preserved. Tools are still installed via `go install` from the Makefile (now cached via `bin/`). No `paths-ignore` — lint still runs on every PR including doc-only ones.

### Measured impact

|                          | lint-go | lint-web | lint wall-clock |
| ------------------------ | ------- | -------- | --------------- |
| Before this PR           | —       | —        | **7m06s** (single `lint` job) |
| With this PR — cold      | 4m41s   | 3m51s    | **4m41s** (max of the two) |
| With this PR — warm      | 2m08s   | 3m39s    | **3m39s** (max of the two) |

`lint-go` shows the expected cold-vs-warm pattern (−55%); `lint-web` barely moved on this PR because no webapp source actually changed between runs — the eslint/stylelint/tsc caches will pay off on PRs that touch the webapp.

## Ticket Link
N/A — internal CI optimization.

## Checklist
- ~~Unit tests updated~~

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -150,6 +150,17 @@ jobs:
       - name: ci/check-diff-on-gomod
         run: git --no-pager diff --exit-code go.mod go.sum || (echo "Please run \"go mod tidy\" and commit the changes in go.mod and go.sum." && exit 1)
 
+  # Aggregator job so branch protection rules requiring a check named "lint"
+  # keep working after the split into lint-go / lint-web.
+  lint:
+    runs-on: ubuntu-22.04
+    needs:
+      - lint-go
+      - lint-web
+    steps:
+      - name: ci/lint-aggregator
+        run: echo "lint-go and lint-web both succeeded"
+
   dist:
     runs-on: ubuntu-22.04
     needs:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -151,15 +151,26 @@ jobs:
         run: git --no-pager diff --exit-code go.mod go.sum || (echo "Please run \"go mod tidy\" and commit the changes in go.mod and go.sum." && exit 1)
 
   # Aggregator job so branch protection rules requiring a check named "lint"
-  # keep working after the split into lint-go / lint-web.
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -98,6 +105,73 @@ jobs:
...
+        run: go mod tidy -v
+
+      - name: ci/check-diff-on-gomod
+        run: git --no-pager diff --exit-code go.mod go.sum || (echo "Please run \"go mod tidy\" and commit the changes in go.mod and go.sum." && exit 1)
+
+  # Aggregator job so branch protection rules requiring a check named "lint"
+  # keep working after the split into lint-go / lint-web. Runs even when an
+  # upstream lint job fails so the required check reports failure instead of
+  # being skipped.
+  lint:
```
JulienTant, line 157: As configured, the playbook CI requires a `lint` step to exist and pass - I don't have access to the settings to change this >_> So we're building it out of the result of the two other ones.

calebroseland (approved):

PR #2279, opened 2026-05-23
fix: rename "restart run" to "resume run" for clarity

### LINKED ISSUE #1837: Proposal: change "restart run" to "resume run"
When a playbook run is finished, it can be reopened with the "restart run" action (see PR #1448). I think the label is misleading. A restart would indicate _a new start from the beginning_, resetting existing progress.

Unfortunately, the confirm dialog does not clarify on this either:
> **Confirm restart run**
> Are you sure you want to restart the run?

Users might not proceed because of fear of data loss.

I therefore propose to **rename the action to "resume"**, indicating to pick-up where the run was stopped.

Renjini221:
## Summary
Renamed all instances of "restart run" to "resume run" to better 
reflect the action. Restarting implies starting over from the 
beginning, which may cause fear of data loss for users. "Resume run" 
clearly indicates picking up where the run was stopped.

## Ticket Link
Fixes: https://github.com/mattermost/mattermost-plugin-playbooks/issues/1837

## Checklist
- ~~Unit tests updated~~


Renjini221:
@calebroseland gentle ping! Would love to get this reviewed when you have a moment

calebroseland (changes requested):
Hi @Renjini221, Thanks for the PR! The intent is spot on — "resume" is much clearer than "restart" here.

However, the current diff edits locale JSON files directly, but those are generated output from `make i18n-extract`. The keys edited here (`Zg0obP`, `k5EChD`, `vqmRBs`) are stale — they don't exist in `en.json` and aren't used by the running application, so these changes won't take effect.

Here's what needs to change instead:

1. Change the `defaultMessage` values from `'Restart'`/`'restart'` to `'Resume'`/`'resume'` (and confirmation text equivalently) in the source files — e.g. `restore_run.tsx`, `controls.tsx`, `rhs_checklist_list_footer.tsx`.

2. Since the `isChannelChecklist` branches already use `'Resume'`, both sides of the ternary become identical — you can remove the conditionals entirely and simplify those spots.

3. Run `make i18n-extract` to regenerate `en.json` with the new keys.

4. Update the CSS class `restartRun` → `resumeRun` in `controls.tsx` and the matching `.restartRun` selector in the E2E test, along with the `.contains('Restart')` assertion.

5. Remove the manual locale JSON edits — those files are managed by the translation pipeline and cannot be edited directly.

Let me know if you have any questions!

Renjini221:
@calebroseland Thanks for the detailed review and guidance! I've now:

* Updated the actual source `.tsx` files instead of editing generated locale JSONs
* Simplified the ternary conditions where both branches became identical
* Renamed `restartRun` → `resumeRun`
* Updated the E2E test selectors/assertions
* Regenerated the i18n files using the extract workflow

Sry For the delay

Renjini221:
@calebroseland pls respond

calebroseland (changes requested):
Thanks for the further changes @Renjini221! There's still some changes to address -- looks like some conflicts to resolve and there are some extraction issues. 

1. See coderabbit's [`8af9f89` (this PR)](https://github.com/mattermost/mattermost-plugin-playbooks/pull/2279/changes/8af9f8917e053a6bb6b379c0238cfd2f7e19127b#r3324955909). To address, please undo all changes to `webapp/i18n/*` and then run `make i18n-extract` to regenerate `en.json` from source. There should be no changes to any other locales.

PR #2302, opened 2026-06-16
MM-69322: send single Content-Type header on /query

calebroseland:
## Summary
Apollo's `HttpLink` sets a lowercase `content-type`; `Client4.getOptions` adds a capitalized `Content-Type` and merges headers case-sensitively, so the GraphQL `/query` request went out as `Content-Type: application/json, application/json`. That trips WAF rules flagging multiple Content-Type headers (CVE-2023-38199 detection), forcing customers to whitelist us or weaken the rule.

This dedupes headers case-insensitively in the GraphQL fetch wrapper (`buildQueryFetchOptions`) so exactly one `Content-Type` is sent. Only the browser GraphQL path was affected; REST `doPost/doPut` pass no headers, so they were already fine.

_Note: the underlying cause is the case-sensitive header merge in `Client4.getOptions` (upstream `@mattermost/client`). This is a local unblock; an upstream fix would make the shared client correct for any future caller._

## Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-69322

## Test Plan
- Unit (`webapp/src/graphql/apollo.test.ts`): documents the raw duplicate from `Client4.getOptions`, asserts a single `content-type` after the fix, and verifies the dedup is general (not content-type-specific).
- E2E (`e2e-tests/tests/integration/playbooks/api/graphql_request_headers_spec.js`): intercepts a real browser-issued `/query` and asserts a single `Content-Type` value.

|  before  |  after  |
|----|----|
| CleanShot 2026-06-16 at 16 03 41@2x | CleanShot 2026-06-16 at 16 03 14 |

## Checklist
- [x] Unit tests updated (`webapp/src/graphql/apollo.test.ts`)
- [x] E2E test added (`e2e-tests/tests/integration/playbooks/api/graphql_request_headers_spec.js`)
- ~~Gated by experimental feature flag~~ (N/A)

### DIFF
--- a/e2e-tests/tests/integration/playbooks/api/graphql_request_headers_spec.js
+++ b/e2e-tests/tests/integration/playbooks/api/graphql_request_headers_spec.js
@@ -1,0 +1,66 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// ***************************************************************
+
+// Stage: @prod
+// Group: @playbooks
+
+describe('api > graphql_request_headers', {testIsolation: true}, () => {
+    let testTeam;
+    let testUser;
+    let testPublicPlaybook;
+
+    before(() => {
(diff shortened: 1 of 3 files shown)


jgheithcock (approved):
LGTM, thanks for adding the tests

PR #2310, opened 2026-06-19
MM-69385 - Restrict run management actions for non-members

catalintomai:
## Summary

On playbooks with **Restrict run management actions for non-owners** enabled, a non-owner could tick **Also mark the run as finished** when posting a status update. The update posted but the finish was silently rejected by the server — with no error or feedback.

This hides the *Also mark the run as finished* checkbox for users blocked by the owner-only restriction (reusing `useIsBlockedByOwnerOnlyForFinishRestore`), and enforces the gate in the submit logic so a blocked user can never reach the finish path. Owners and system admins are unaffected. Covered by new Jest and Cypress tests.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69385

### DIFF
--- a/e2e-tests/tests/integration/playbooks/runs/owner_only_finish_spec.js
+++ b/e2e-tests/tests/integration/playbooks/runs/owner_only_finish_spec.js
@@ -366,6 +366,41 @@ describe('runs > owner only finish', {testIsolation: true}, () => {
                 });
             });
         });
+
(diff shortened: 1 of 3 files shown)


lindy65 (approved):
Thanks @catalintomai - this is fixed 👍

jgheithcock (approved):
LGTM

PR #2311, opened 2026-06-19
MM-69384 - Disable toggles when "Only admins can edit this playbook" = true

catalintomai:
## Summary

With **Only admins can edit this playbook** enabled, the *Require new channel for all runs* and *Auto-archive channel* toggles were still editable for ordinary users — toggling them just flashed a `Failed to save setting` error. Every other setting was already read-only.

These two toggles keyed their `disabled` state off `archived` alone, ignoring `!canEdit`. They now use the same `disabled` (`archived || !canEdit`) as the surrounding settings, so they are read-only for non-admins. Covered by updated Cypress tests.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69384


lindy65 (approved):
Thanks @catalintomai - this is fixed 👍

jgheithcock (commented):
LGTM

jgheithcock (approved):

PR #2314, opened 2026-06-19
Gray out and explain read-only run name field in run playbook modal

JulienTant:
## Summary
When a playbook has a channel name template configured, the **Run name** field in the run playbook modal is read-only (the name is generated from the template). Previously the field still looked editable, which was confusing.

This PR:
- Styles the field with a disabled/grayed-out appearance when it's read-only (muted background and text, `not-allowed` cursor, no focus highlight).
- Adds a tooltip explaining *why* it can't be edited: the run name comes from the playbook's channel name template.

image

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69402

## Checklist
- [ ] Unit tests updated

### DIFF
--- a/webapp/i18n/en.json
+++ b/webapp/i18n/en.json
@@ -608,6 +608,7 @@
   "mm5vL8": "Only invited members",
   "mw9jVA": "Add a title",
   "n70CD4": "Are you sure you want to resume {name}?",
+  "nGnueQ": "The run name is set automatically from this playbook's channel name template and can't be edited here.",
   "nR9VfG": "Failed to save run number prefix. Please try again.",
   "nc8QpJ": "Recent Activity",
   "nkCCM2": "You will not be reminded again.",
--- a/webapp/src/components/assets/inputs.tsx
+++ b/webapp/src/components/assets/inputs.tsx
@@ -17,6 +17,16 @@ export const BaseInput = styled.input<{$invalid?: boolean; $readOnly?: boolean}>
     &:focus {
         box-shadow: ${(props) => (props.$invalid ? 'inset 0 0 0 2px var(--error-text)' : 'inset 0 0 0 2px var(--button-bg)')};
     }
+
+    ${(props) => props.$readOnly && `
+        background-color: rgba(var(--center-channel-color-rgb), 0.04);
+        color: rgba(var(--center-channel-color-rgb), 0.56);
+        cursor: not-allowed;
+
+        &:focus {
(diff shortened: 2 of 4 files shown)


catalintomai:
@JulienTant , fyi - getting the results below by running the "/review-code --full" on the PR using the https://github.com/catalintomai/mattermost-claude-agents agents. - might or might not (all) be valid but worth checking.

SHOULD_FIX 1 — Reuse the existing ConditionalTooltip instead of a hand-rolled ternary

  [simplicity-reviewer] · run_playbook_modal.tsx

  The PR extracts const input = (<BaseInput …/>) then renders readOnly ? <Tooltip>{input}</Tooltip> : input. The codebase already ships src/components/widgets/conditional_tooltip.tsx (verified) which does exactly this — wrap in Tooltip when show, else render the child bare. Using it removes the const input extraction, the ternary, and the new Tooltip import:

  <ConditionalTooltip
      show={Boolean(readOnly)}
      id='run-name-readonly-tooltip'
      content={formatMessage({defaultMessage: 'The run name is set automatically from this playbook\'s channel name template and can\'t be edited here.'})}
  >
      <BaseInput $invalid={Boolean(error)} $readOnly={readOnly} … />
  </ConditionalTooltip>

  Net simpler and aligns with the established pattern. Not blocking — the current code is correct.

  SHOULD_FIX 2 — Read-only reason and focus state aren't conveyed non-visually

  [accessibility-reviewer] + [ui-pattern-reviewer] (2-source convergence)

  Two related, in-diff a11y gaps — both non-blocking because the native readOnly attribute (present) already makes screen readers announce the field as read-only:
  - Reason not in the a11y tree: the explanation lives only inside the react-bootstrap tooltip popup (id on the popup, not referenced from the input). A screen-reader user hears "read only" but not why. Add aria-describedby on the input pointing at the explanation (e.g. an sr-only span, or wire the tooltip id) so the reason is statically available.
  - Focus ring muted (inputs.tsx): the new $readOnly &:focus block drops the ring to inset 0 0 0 1px …0.16 — identical to the rest state. The field is still Tab-reachable (readOnly ≠ disabled), so focus becomes visually indistinguishable (WCAG 2.4.7). Consider keeping a perceptible focus indicator.

JulienTant:
thanks @catalintomai I addressed the feedbacks :)

lindy65 (approved):
Thanks @JulienTant - this is working as expected now 👍

catalintomai (approved):

PR #2316, opened 2026-06-22
[MM-69404] Allow inter-plugin API calls on behalf of a user

hanzei:
## Summary

Other Mattermost plugins cannot call the Playbooks REST/GraphQL API today. When a plugin issues a request via `PluginHTTP` / `client.Plugin.HTTP`, the server routes it through `ServeInterPluginRequest`, which blanks `Mattermost-User-Id`. Our `MattermostAuthorizationRequired` middleware requires that header, so every inter-plugin call gets a 401. The request reaches us; it just has no user identity to authorize against.

This is needed for a custom customer integration (itself a plugin) that needs to read and act on Playbooks data as a specific user.

### What this changes

Extend `MattermostAuthorizationRequired` so that, when there is no authenticated user but the request is a trusted inter-plugin call, it promotes the asserted acting-user into `Mattermost-User-Id` and lets the request proceed. The calling plugin names the user via a custom `Mattermost-Plugin-Acting-User-Id` header.

Because every downstream handler already reads `Mattermost-User-Id` and runs per-user permission checks (`RunView`, etc.), **no handler changes are required** and the call is scoped to exactly what the asserted user can see. The change is ~2 files (`server/api/api.go` plus a middleware test). GraphQL is mounted under the same subrouter, so it is covered too.

### Trust model / security

- **Trust anchor:** the design relies on `Mattermost-Plugin-ID` being unforgeable by external clients. The server deletes it from the public request path and only sets it on genuine inter-plugin/internal paths (verified against the server version we depend on). If a future server change stops stripping it, the trust boundary moves, so this dependency is called out explicitly here.
- **Least privilege preserved:** every call is still authorized as the asserted user via the unchanged permission checks. The feature does not grant a plugin read access beyond what some user already has.
- **No anonymous access:** an inter-plugin request without an acting-user header is rejected.
- **Auditing:** the privileged acting-as-user grant is logged (caller plugin ID + acting user + method/url). `LogRequest` runs before promotion, so it would otherwise record a blank user and never the calling plugin.

### Example logs
```jsonl
{"timestamp":"2026-06-22 09:51:54.702 +02:00","level":"debug","msg":"Received HTTP request","caller":"app/plugin_api.go:1160","plugin_id":"playbooks","user_id":"","request_id":"4hbmhi8ybfrdubeirdiwoq4fsw","user_agent":"","method":"GET","url":"/api/v0/runs/re1x94mc1pnzbpphofgp5thqpe","plugin_caller":"github.com/mattermost/mattermost-plugin-playbooks/server/api/logger.go:48"}
{"timestamp":"2026-06-22 09:51:54.702 +02:00","level":"info","msg":"Authorized inter-plugin API request on behalf of user","caller":"app/plugin_api.go:1164","plugin_id":"playbooks","url":"/api/v0/runs/re1x94mc1pnzbpphofgp5thqpe","caller_plugin_id":"com.mattermost.plugin-starter-template","acting_user_id":"nswgu3thqtyq3fp6hkwn8xuk7a","method":"GET","plugin_caller":"github.com/mattermost/mattermost-plugin-playbooks/server/api/api.go:147"}
{"timestamp":"2026-06-22 09:51:54.725 +02:00","level":"debug","msg":"Handled HTTP request","caller":"app/plugin_api.go:1160","plugin_id":"playbooks","time":"22","status":"200","user_agent":"","method":"GET","url":"/api/v0/runs/re1x94mc1pnzbpphofgp5thqpe","user_id":"","request_id":"4hbmhi8ybfrdubeirdiwoq4fsw","plugin_caller":"github.com/mattermost/mattermost-plugin-playbooks/server/api/logger.go:61"}
```

## Ticket Link

https://mattermost.atlassian.net/browse/MM-69404

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -10,8 +10,10 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-plugin-playbooks/server/app"
 	"github.com/mattermost/mattermost-plugin-playbooks/server/config"
 
+	"github.com/mattermost/mattermost/server/public/model"
 	"github.com/mattermost/mattermost/server/public/pluginapi"
 )
 
@@ -33,20 +35,22 @@ type Handler struct {
 	APIRouter *mux.Router
 	root      *mux.Router
 	config    config.Service
+	auditor   app.Auditor
 }
 
 // NewHandler constructs a new handler.
-func NewHandler(pluginAPI *pluginapi.Client, config config.Service) *Handler {
+func NewHandler(pluginAPI *pluginapi.Client, config config.Service, auditor app.Auditor) *Handler {
 	handler := &Handler{
 		ErrorHandler: &ErrorHandler{},
 		pluginAPI:    pluginAPI,
 		config:       config,
+		auditor:      auditor,
 	}
 
 	root := mux.NewRouter()
 	root.Use(LogRequest)
 	api := root.PathPrefix("/api/v0").Subrouter()
-	api.Use(MattermostAuthorizationRequired)
+	api.Use(handler.MattermostAuthorizationRequired)
 
 	api.Handle("{anything:.*}", http.NotFoundHandler())
 	api.NotFoundHandler = http.NotFoundHandler()
@@ -126,7 +130,7 @@ const (
 // user it is acting on behalf of via the Mattermost-Plugin-Acting-User-Id header; that user is then
 // promoted into Mattermost-User-Id so every downstream per-user permission check applies to the
 // asserted user unchanged.
-func MattermostAuthorizationRequired(next http.Handler) http.Handler {
+func (h *Handler) MattermostAuthorizationRequired(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// Normal authenticated user request -- unchanged behavior.
 		if r.Header.Get(userIDHeader) != "" {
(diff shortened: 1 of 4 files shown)


server/api/api.go
```diff
@@ -104,15 +104,70 @@ func ReturnJSON(w http.ResponseWriter, pointerToObject interface{}, httpStatus i
...
+		if actingUserID := resolveInterPluginUser(r); actingUserID != "" {
+			// Record the privileged acting-as-user grant for auditing: the upstream LogRequest
+			// middleware runs before promotion, so it would otherwise log a blank user and never
+			// the calling plugin.
+			logrus.WithFields(logrus.Fields{
+				"caller_plugin_id": r.Header.Get(pluginIDHeader),
+				"acting_user_id":   actingUserID,
+				"method":           r.Method,
+				"url":              r.URL.String(),
+			}).Info("Authorized inter-plugin API request on behalf of user")
```
hanzei: Does the playbooks plugin export custom audit events? Should we add some for this case?

edgarbellot (approved):
LGTM!

JulienTant (approved):

jgheithcock (approved):
I don't have any issues from a code perspective but I do want to call out that this PR has a significant security impact (@edgarbellot fyi) - if everyone is fine with this, I'm fine also!

**Any** installed plugin can call any Playbook API as **any** user.

That seems a bit broad to me. I would have thought that the installed plugin would need to pass in the user it was operating in, or that there would need to be a defined user (that might have lower permissions) as the "Playbook API user" or even just the `EnableInterPluginAPI` flag that could be turned off. It seems odd that you can't even turn that off. DryRun calls this out as an Insecure Direct Object Reference (IDOR).

I'd like to make this PR and pattern part of the overall plugin architecture, we will need it for Docs and you can see it used across the board. That said, I would think we would want to make this more of a gate and less an open doorway.

@lieut-data - any thoughts on this?

lieut-data:
> **Any** installed plugin can call any Playbook API as **any** user.
> 
> That seems a bit broad to me.

@jgheithcock, plugins are root. This is an intentional design decision and not something we can change now, even if we wanted to.

I haven't looked at the implementation details -- would suggest we loop in @crspeller and @nickmisasi since I know they've been at the forefront of inter plugin API. But as far as allowing a plugin to leverage another plugin's API to effect changes on behalf of a user, I'm supportive of that flexibility. (Please ensure we have audit logs!)

jgheithcock:
> @jgheithcock, plugins are root. This is an intentional design decision and not something we can change now, even if we wanted to.

Great, thanks for the background context.

hanzei:
@jgheithcock One thing to mention: This PR does not expose a new security surface. Plugins can already create access tokens for _any_ user and call the playbooks API using that token. (In fact, that is what I did for a custom plugin)

server/api/api.go
hanzei: Moved the application logs to audit logs via a04dc720388e28488e41c14a59fd9a8739e52769.

hanzei:
Thanks for chiming in @lieut-data and also for the reminder about audit logs. I've added them via a04dc720388e28488e41c14a59fd9a8739e52769. Now the audit logs cover every request coming from another plugin, including the other plugin's ID and the user they are impersonating.

edgarbellot:
@jgheithcock Adding a bit of security context on top of Jesse's and Ben's points:

In our threat model, plugins are trusted components. They run inside the server's trust boundary with root-level access. As Ben noted, any plugin can already mint an access token for any user and call this same API with it, so this PR doesn't add a new security surface.

The real control here is installation: by default only sysadmins can install plugins, so trusting an installed plugin is equivalent to trusting a sysadmin.

On your suggestions (passing in the operating user, a lower-permission "Playbook API user", or an `EnableInterPluginAPI` flag): they're reasonable instincts, but they wouldn't add much security because plugins already run as root.

In this PR the boundary we do enforce is at the perimeter: the server strips these headers from any external request, and the new audit logs (thanks Jesse and Ben!) record which plugin made the call and which user it acted as.

hanzei:
Merging the PR per the above conversation

PR #2317, opened 2026-06-22
[MM-69405] Add inter-plugin client constructor for the Playbooks API

hanzei:
## Summary

Client-side companion to #2316. Lets other plugins call the Playbooks API over inter-plugin HTTP using the existing typed `client.Client`, instead of hand-rolling `*http.Request`s.

```go
pb, err := client.NewInterPluginClient(p.API.PluginHTTP, userID)
if err != nil {
    return err
}
run, err := pb.PlaybookRuns.Get(ctx, runID) // authorized as userID
```

`do` accepts anything with the `func(*http.Request) *http.Response` shape, so both `p.API.PluginHTTP` and `pluginClient.Plugin.HTTP` work without this package depending on the plugin API. The client is scoped to a single acting user for its lifetime (build one per user).

### What changed

Two adjustments make the existing client usable over inter-plugin HTTP:

- **API URL prefix is now a `Client` field.** Regular HTTP uses `plugins/playbooks/api/v0/`; inter-plugin requests use `playbooks/api/v0/`, because `PluginHTTP` routes by the first path segment (`/<pluginid>/*`) rather than the `/plugins/<id>/*` route the server uses for external HTTP.
- **A small `RoundTripper`** (`interPluginTransport`) injects the `Mattermost-Plugin-Acting-User-Id` header on every request and turns a nil `PluginHTTP` response (plugin not installed/enabled) into a clear error.

`NewInterPluginClient` fails closed: it errors if `do` is nil or `actingUserID` is empty, so a request can never go out without an asserted user.

### Dependency

This is inert until the server-side change (#2316) that promotes the acting-user header lands; until then these requests get a 401. No new module dependency: `client/go.mod` already requires `mattermost/server/public`, and the constructor takes a plain func rather than importing the plugin API.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-69405

### DIFF
--- a/client/interplugin.go
+++ b/client/interplugin.go
@@ -62,9 +62,20 @@ type interPluginTransport struct {
 }
 
 func (t *interPluginTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	req.Header.Set(actingUserIDHeader, t.actingUserID)
+	// RoundTrip must not modify the original request, so operate on a clone. The clone shares the
+	// original's Body, so closing it below also discharges RoundTrip's duty to close req.Body.
+	outReq := req.Clone(req.Context())
+	outReq.Header.Set(actingUserIDHeader, t.actingUserID)
 
-	resp := t.do(req)
+	// RoundTrip must always close the request body. PluginHTTP's buffered path closes and nils the
+	// body itself, but its streaming path does not, so close whatever remains once the call returns.
+	defer func() {
+		if outReq.Body != nil {
+			outReq.Body.Close()
+		}
+	}()
+
+	resp := t.do(outReq)
 	if resp == nil {
 		return nil, errors.Errorf("no response from the %s plugin; is it installed and enabled?", manifestID)
 	}
@@ -73,7 +84,7 @@ func (t *interPluginTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// error responses for non-2xx replies. Set it as net/http's Transport does for real network
 	// requests, so a non-2xx response yields an error rather than a nil-pointer panic.
 	if resp.Request == nil {
-		resp.Request = req
+		resp.Request = outReq
 	}
 
 	return resp, nil
--- a/client/interplugin.go
+++ b/client/interplugin.go
@@ -1,0 +1,73 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package client
+
+import (
+	"net/http"
+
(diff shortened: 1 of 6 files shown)


jgheithcock (approved):
LGTM

client/interplugin.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// PluginHTTP does not populate Response.Request, but the client dereferences it when building
+	// error responses for non-2xx replies. Set it as net/http's Transport does for real network
+	// requests, so a non-2xx response yields an error rather than a nil-pointer panic.
+	if resp.Request == nil {
+		resp.Request = req
+	}
+
+	return resp, nil
+}
```
JulienTant, line 91: Not blocking per se so I just leave that as a comment, but `RoundTrip` requires use to:

1. Not change the initial request ("RoundTrip should not modify the request, except for consuming and closing the Request's Body."): For this I propose a clone of the request.
2. Close the body no matter what ("RoundTrip must always close the body, including on errors"): we "simply" call close on the body if it's not nil.

(from `go doc net/http.RoundTripper`)

```suggestion
func (t *interPluginTransport) RoundTrip(req *http.Request) (*http.Response, error) {
      outReq := req.Clone(req.Context())
      outReq.Header.Set(actingUserIDHeader, t.actingUserID)

      resp := t.do(outReq)
      if resp == nil {
              if outReq.Body != nil {
                      outReq.Body.Close()
              }
              return nil, errors.Errorf("no response from the %s plugin; is it installed and enabled?", manifestID)
      }

      // PluginHTTP does not populate Response.Request, but the client dereferences it when building
      // error responses for non-2xx replies. Set it as net/http's Transport does for real network
      // requests, so a non-2xx response yields an error rather than a nil-pointer panic.
      if resp.Request == nil {
              resp.Request = outReq
      }

      return resp, nil
}
```

JulienTant (approved):
@hanzei LGTM - I left a remark to be more respectful of the original interface but it's not functionally blocking

client/interplugin.go
hanzei: Thanks for the suggestion @JulienTant!

1. Good call. The code now clones the context before sending it off.
2. https://pkg.go.dev/net/http#RoundTripper states

> 	// RoundTrip must always close the body, including on errors,
	// but depending on the implementation may do so in a separate
	// goroutine even after RoundTrip returns. This means that
	// callers wanting to reuse the body for subsequent requests
	// must arrange to wait for the Close call before doing so.
	
To honor that interface, `RoundTrip` now always closes the body, not just on error.

PR #2325, opened 2026-06-23
MM-69465: Fix non-deterministic plugin config keys from serialize() casing

JulienTant:
## Summary

Three Playbooks plugin settings (`EnableIncrementalUpdates`, `EnableTeamsTabApp`, `TeamsTabAppTenantIDs`) could silently fail to take effect because the plugin persisted them under a different key casing than the System Console writes.

The System Console saves a setting under its `plugin.json` `settings_schema` key, while `Configuration.serialize()` wrote some of them under CamelCase keys. Since the server's `LoadPluginConfiguration` lowercases every key before unmarshaling, the two variants collapsed into one struct field, and Go's randomized map iteration order made the surviving value **non-deterministic**. A stale value could clobber the admin's choice — e.g. `EnableIncrementalUpdates` set `true` in the console but read as `false` — which also produced repeated `cluster.SendBestEffort failed … message too long` errors and broke cross-device run sync.

**Fix:**
- `serialize()` now emits the exact `settings_schema` keys, so the console's copy and the plugin's copy always coincide and no duplicate is ever created.
- New `reconcileLegacyConfigKeys()` runs once at plugin activation: it reads the raw stored config (where the duplicate keys are still distinct), folds any pre-existing duplicate into the canonical key — keeping the console value — and deletes the stale key. Idempotent and safe under concurrent cluster activation.
- Added unit tests for the reconcile behavior plus a guard test that fails if `serialize()` keys ever collide case-insensitively or drift from `plugin.json`.

Verified end-to-end against a live local instance: recreated the duplicate (`enableincrementalupdates: true` + stale `EnableIncrementalUpdates: false`), re-activated the plugin, and confirmed the duplicate was removed and the console value (`true`) preserved.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-69465

## Checklist
- [x] Unit tests updated

### DIFF
--- a/server/config/service.go
+++ b/server/config/service.go
@@ -17,12 +17,12 @@ import (
 	"github.com/mattermost/mattermost/server/public/pluginapi"
 )
 
-// legacyConfigKeyAliases maps a canonical settings_schema key to the differently
-// cased key older plugin versions wrote via serialize(). See reconcileLegacyConfigKeys.
+// legacyConfigKeyAliases maps a legacy key older plugin versions wrote via
+// serialize() to its canonical settings_schema key. See reconcileLegacyConfigKeys.
 var legacyConfigKeyAliases = map[string]string{
-	"enableTeamsTabApp":        "EnableTeamsTabApp",
-	"teamsTabAppTenantIDs":     "TeamsTabAppTenantIDs",
-	"enableincrementalupdates": "EnableIncrementalUpdates",
+	"EnableTeamsTabApp":        "enableTeamsTabApp",
+	"TeamsTabAppTenantIDs":     "teamsTabAppTenantIDs",
+	"EnableIncrementalUpdates": "enableincrementalupdates",
 }
 
 // WebsocketPublisher defines interface for publishing websocket events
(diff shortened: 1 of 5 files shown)


hanzei:
Thanks for the PR, @JulienTant!

I've noticed that the "wrong" settings keys like `EnableIncrementalUpdates` stay in the config file. Should we clean them up? 
Screenshot from 2026-06-25 10-18-23

server/config/service.go
```diff
@@ -5,19 +5,29 @@ package config
...
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	"github.com/mattermost/mattermost/server/public/model"
 
 	"github.com/mattermost/mattermost/server/public/pluginapi"
 )
 
+// legacyConfigKeyAliases maps a canonical settings_schema key to the differently
+// cased key older plugin versions wrote via serialize(). See reconcileLegacyConfigKeys.
```
catalintomai: @JulienTant , 1/5 , but - the map is written new‑key‑first (newKey: oldKey), but the code copies data old‑key → new‑key. The declaration and the actual data movement run in opposite directions, which is not very readable imo.

catalintomai (approved):

JulienTant:
> Thanks for the PR, @JulienTant!
> 
> I've noticed that the "wrong" settings keys like `EnableIncrementalUpdates` stay in the config file. Should we clean them up? Screenshot from 2026-06-25 10-18-23

Uh I can't seem to be able to reproduce this - you have some steps to follow to get there? I tried with a fresh config with only the "wrong" key and one with both, in both case I'm left with only the "right" one

server/config/service.go
JulienTant: flipped the map to read `legacy -> canonical` in 5c453126.

PR #2345, opened 2026-07-01
[MM-69446] Fix run name precedence over literal channel template

jgheithcock:
## Summary
- Fix a v2.10.0 regression where a literal `channel_name_template` (e.g. `JIRA_ID: TITLE`) overrode a caller-supplied run `name` on `POST /api/v0/runs`
- Restore pre-CF-2 behavior: non-empty caller-supplied names take precedence over template resolution
- Add unit tests covering both the regression case and the no-name fallback

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69446

## Checklist
- [x] Unit tests updated

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -6219,6 +6219,13 @@ func (s *PlaybookRunServiceImpl) resolveRunName(playbookRun *PlaybookRun, pb *Pl
 	seqID := FormatSequentialID(pb.RunNumberPrefix, runNumber)
 	playbookRun.SequentialID = seqID
 
+	if userSuppliedName != "" {
+		resolvedRunName := strings.TrimSpace(truncateRunes(userSuppliedName, MaxRunNameLength))
+		playbookRun.Name = resolvedRunName
+		resolvedChannelName := strings.Join(strings.Fields(truncateRunes(resolvedRunName, model.ChannelDisplayNameMaxRunes)), " ")
+		return resolvedChannelName, nil
+	}
+
 	if formatFunc == nil {
 		formatFunc = s.makeRunNameFormatFunc()
 	}
@@ -6247,8 +6254,6 @@ func (s *PlaybookRunServiceImpl) resolveRunName(playbookRun *PlaybookRun, pb *Pl
 
 	if resolvedRunName == "" {
 		switch {
-		case userSuppliedName != "":
-			resolvedRunName = userSuppliedName
 		case seqID != "":
 			resolvedRunName = seqID + " - Untitled"
 		default:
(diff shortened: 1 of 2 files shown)


JulienTant:
@jgheithcock it seems to me like the jira ticket calls the new expected behavior introduced by CF-2 a bug... so I'm not quite sure what's right or wrong here. My understanding of CF-2 is that if a template exist, it takes precedence over the name field, and we don't provide a alternative right now.


I'll ask @catalintomai to review as he probably has more context on what is expected and why.

Augustin-FL:
@JulienTant @catalintomai It is sometimes desirable to allow end users to edit the playbook title template at launch time.
This is because not everything can be pre-build server side (see Mattermost support case 51883, for example). Some workflows rely on end-user input at launch time.

Perhaps an option like "allow title edit at launch time" should be added in the playbook outline?

JulienTant:
>  Some workflows rely on end-user input at launch time.

That's where the "attributes" come in, first you set the template with am attribute:

image

Then when you create your run, you can set the value:

image

But I agree that we might be missing a real "override/free form" format, so I[ just opened a PR](https://github.com/mattermost/mattermost-plugin-playbooks/pull/2346) to restore this function, I might have to loop in PM and design etc but feel free to take a look :)

Augustin-FL:
You're absolutely right. I didn't understand/grasp the concept of Attributes, and I was thinking Playbooks titles were now generated on the backend side only, without the possibility for a user to set up choices or free text.

I'd like to apologize for my previous message. This is a typical case of "user didn't read the documentation properly then said the product was not working". The user here being me.... Oops. Sorry about that. I should have held my comment.

Regarding the PR you opened: I let you decide if you want you pursue with it or not: I am not sure if it is necessary (It is also possible to set the title template to a single free text attribute)

catalintomai:
@Augustin-FL - couple of thoughts.

  1. Through the UI, if "Run name template" is set, the run creation modal does not allow changing the run name. The "Run name" field is read-only. This is by design. If no template is set, then the run name field is editable.

  2. I think your scenario is the API case where the template is set on the playbook, then a run is created via API with `name` also provided. Today, the template taking precedence matches the UI behavior: once a playbook has a run name template, that template controls the run name.

  3. I’m not convinced that the change in [66ec151](https://github.com/mattermost/mattermost-plugin-playbooks/pull/2345/commits/66ec151963408408f472c2e6a5e1b3211d558f8d), detecting `{` in the template, is the right boundary. In the Playbook Actions "Create a run channel" panel, "Run name template" can be either tokenized or literal text. A value without `{...}` is still a configured template. The PR is backend-only, so the UI still greys out "Run name" whenever a template is set, regardless of whether the template contains `{`. Making the UI consistent with this backend behavior would mean enabling/disabling the run name field based on the format of the template text, which seems confusing. A template is a template whether or not it has `{...}` placeholders.

  4. Does Julien’s attribute-based approach unblock your case for now? A user can still provide launch-time text by referencing an attribute in the template. I agree Julien’s PR sounds like the more user-friendly long-term solution.
  
  cc. @JulienTant , @jgheithcock , @asaadmahmood

Augustin-FL:
Yes, Julien's attribute-base approach unblock my case for now.

jgheithcock:
Great, thanks for all the feedback/pushback. Closing this one and will make a note in the ticket.

PR #2344, opened 2026-07-01
[MM-69462] Disable editing attributes and adding participants on finished runs

jgheithcock:
## Summary
- Disable click-to-edit affordances for run attributes when a playbook run is finished
- Hide "Add participant" and Manage controls in the RHS and participants view for finished runs
- Wire the existing backstage `editable` prop through to `PropertiesList` (was previously unused)

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69462

## Test plan
- [x] Open a finished playbook run in the RHS
- [x] Confirm run attributes are read-only (no hover/edit affordance, no "Failed to update property value" toast)
- [x] Confirm "Add participant" link/icon is hidden in the about section
- [x] Open the participants list for a finished run and confirm Manage/Add buttons are hidden
- [x] Open an in-progress run and confirm attributes and add-participant controls still work as before

## Checklist
- [ ] Gated by experimental feature flag
- [x] Unit tests updated

PR #2372, opened 2026-07-22
Cherry pick #2344 to release-2.10

JulienTant:
## Summary
Backport of #2344. On a finished playbook run, users could still edit attribute values and try to add participants, even though the server already blocks both actions — it looked like it worked, but nothing actually happened. This closes that gap by making attributes and the "add participant" option non-editable once a run is finished, so people aren't misled into thinking their change succeeded.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69462
(related: MM-68844, the server-side fix that already blocks this)

## Checklist
- [x] Unit tests updated


jgheithcock (approved):
LGTM

PR #2373, opened 2026-07-22
Cherry pick #2344 to release-2.4

JulienTant:
## Summary
Backport of #2344. On a finished playbook run, users could still try to add participants even though the server already blocks that action — it looked like it worked, but nothing actually happened. This closes that gap by hiding the "add participant" option once a run is finished, so people aren't misled into thinking it succeeded.

Only part of the original fix applies here, since this branch doesn't have the run-attributes feature that the other part of #2344 covered.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-69462
(related: MM-68844, the server-side fix that already blocks this)

## Checklist
- [x] Unit tests updated


jgheithcock (approved):
LGTM

PR #2381, opened 2026-07-27
MM-69627: Add beta task requirements for checklist items

asaadmahmood:
## Summary
- Adds task requirements so playbook authors can define required text fields on checklist items, collect values when checking off a task during a run, and review saved values in a collapsed accordion.
- Gates the feature behind a System Console **Beta Features** setting (`Task requirements`).
- Includes server validation on checkoff-to-closed, GraphQL/type updates, and unit tests for beta config, accordion, and fill modal.

Jira: https://mattermost.atlassian.net/browse/MM-69627

## Test plan
- [ ] Enable **System Console → Plugins → Playbooks → Beta Features → Task requirements**
- [ ] In a playbook editor, open a task’s More menu and add one or more requirements; confirm labels appear in the accordion
- [ ] Edit and remove requirements; confirm playbook save persists them
- [ ] Start a run, check off a task with requirements; confirm fill modal appears
- [ ] **Save** with partial values keeps the task open and shows values under the task
- [ ] **Save and mark complete** requires all fields and marks the task complete when filled
- [ ] With beta feature disabled, confirm Add/Edit requirements UI is hidden and checkoff does not require values
- [ ] Run `make check-style` and relevant unit tests (`go test` + webapp jest for requirements/beta)

Made with [Cursor](https://cursor.com)


webapp/src/components/backstage/beta_features_setting.tsx
```diff
@@ -0,0 +1,227 @@
...
+                            </FeatureCopy>
+                        </CheckboxLabel>
+                    </FeatureRow>
+                </AccordionBody>
+            )}
+        </Accordion>
+    );
+};
+
+const Accordion = styled.div`
```
Willyfrog, line 100: I really like this, but I wonder if this shouldn't be done by Mattermost's main repo and make it available for plugins

Willyfrog (commented):
not doing a full review, I was just curious on the accordion element. I left a comment on it

webapp/src/components/backstage/beta_features_setting.tsx
asaadmahmood: Ideally in the future we can just have it as a common component and export it. But I don't want this to pause this update.

server/app/playbook_run_service.go
```diff
@@ -2302,11 +2304,46 @@ func (s *PlaybookRunServiceImpl) ModifyCheckedState(playbookRunID, userID, newSt
...
+					itemToCheck.Requirements[i].Value = val
+					requirementsChanged = true
+				}
+			}
+		}
+	}
+
+	// Only require all fields when checking off while beta features are enabled.
+	// Saving values alone may leave some fields empty.
+	if newState == ChecklistItemStateClosed && !wasClosed && len(itemToCheck.Requirements) > 0 && s.configService.IsTaskRequirementsEnabled() {
```
JulienTant, line 2331: We should drop `!wasClosed` so we always validate that the requirement are met whenever the item checklist is "closed".

server/app/playbook.go
```diff
@@ -342,6 +342,9 @@ type ChecklistItem struct {
 	// TaskActions is an array of all the task actions associated with this task.
 	TaskActions []TaskAction `json:"task_actions" export:"-"`
```
JulienTant, line 343: Playbook templates can prefill requirement values into every run, it feels... "wrong". We should extend CleanUpChecklists to also blank `Requirements[].Value` on every checklist item, by adding a `ClearRequirementValues()` method to the `ChecklistItemCommon` interface so it's covered on both the REST and GraphQL playbook write paths!

server/api/playbook_runs.go
```diff
@@ -1476,7 +1477,12 @@ func (h *PlaybookRunHandler) itemSetState(c *Context, w http.ResponseWriter, r *
 		return
 	}
 
-	if err := h.playbookRunService.ModifyCheckedState(id, userID, params.NewState, checklistNum, itemNum); err != nil {
+	var opts []app.ModifyCheckedStateOptions
+	if params.RequirementValues != nil {
+		opts = append(opts, app.ModifyCheckedStateOptions{RequirementValues: params.RequirementValues})
+	}
+
+	if err := h.playbookRunService.ModifyCheckedState(id, userID, params.NewState, checklistNum, itemNum, opts...); err != nil {
```
JulienTant, line 1485: We should check if `errors.Is(err, app.ErrMalformedPlaybookRun)` and return a 40X instead of the default `HandleError` that returns 500.

webapp/src/components/checklist_item/fill_requirements_modal.tsx
```diff
@@ -0,0 +1,217 @@
...
+                nextErrors[req.id] = formatMessage({defaultMessage: 'This field is required to mark the task complete'});
+            }
+        }
+        setErrors(nextErrors);
+        return Object.keys(nextErrors).length === 0;
+    };
+
+    const handleSave = () => {
+        setErrors({});
+        onSave(getTrimmedValues());
```
JulienTant, line 64: We should make `onSave` and `onSaveAndComplete` async and only close the modal on success. On failure, keep the modal open and show a failure toast, reusing the exact toaster/ToastStyle.Failure pattern this file already uses for the assignee dropdown. Give the modal a saving state to disable its buttons while the request is in flight.

JulienTant (changes requested):
Thanks for the PR Asaad! Did a first pass and left some feedbacks, there's something that AI helped me discover that I don't have a place in the file to leave as a review so I leave it here: we need clear handling of having both task requirements + message trigger. If I set a task to be marked as done when a user types "resolved", how does it interact with requirements? Should they be mutually exclusive to simplify the process?

Willyfrog:
@asaadmahmood can you fix the CI?

asaadmahmood:
@Willyfrog Done. Sorry for the late response on this.

asaadmahmood:
@JulienTant Agreed on making task requirements and message triggers mutually exclusive.

Pushed in 4e3a53272:
- Playbook editor disables adding requirements while an enabled message trigger exists (and vice versa), with a short explanation tooltip
- Saving one clears the other if both somehow end up set
- Server rejects playbook writes (REST + GraphQL) that configure both on the same item
- Setting enabled mark-as-done task actions on a run item that already has requirements returns 400
- Keyword auto-complete skips items that have requirements

Happy to adjust the UX wording if you want something stricter (e.g. hide instead of disable).

server/api/graphql_playbook.go
```diff
@@ -228,20 +228,21 @@ func (c UpdateChecklist) GetItems() []app.ChecklistItemCommon {
...
+	AssigneeType            *string                `json:"assignee_type"`
+	AssigneePropertyFieldID *string                `json:"assignee_property_field_id"`
+	Command                 string                 `json:"command"`
+	CommandLastRun          float64                `json:"command_last_run"`
+	Description             string                 `json:"description"`
+	LastSkipped             float64                `json:"delete_at"`
+	DueDate                 float64                `json:"due_date"`
+	TaskActions             *[]app.TaskAction      `json:"task_actions"`
+	ConditionID             string                 `json:"condition_id"`
+	Requirements            *[]app.TaskRequirement `json:"requirements"`
```
Willyfrog, line 245: update api.yaml, so it stays up to date with the changes

Willyfrog (changes requested):
Something that it is out of scope for this PR: if the mobile app now tries to complete a task with requirements it will fail. I'm not sure when are we going to add support for this, but it should be documented somewhere to avoid surprises

server/app/playbook_run_service.go
asaadmahmood: Done — we now always validate requirements whenever the target state is closed (including already-closed items), not only on open→closed transitions.

server/app/playbook.go
asaadmahmood: Done — `ClearRequirementValues()` is on `ChecklistItemCommon` and `CleanUpChecklists` blanks requirement values on playbook write paths (REST + GraphQL).

server/api/playbook_runs.go
asaadmahmood: Done — `itemSetState` now returns 400 when `errors.Is(err, app.ErrMalformedPlaybookRun)`.

webapp/src/components/checklist_item/fill_requirements_modal.tsx
asaadmahmood: Done — fill modal saves are async with a saving state; modal stays open on failure and the parent shows a Failure toast.

server/api/graphql_playbook.go
asaadmahmood: Updated `server/api/api.yaml` in 0f2170ab0:
- `ChecklistItem.requirements` + `TaskRequirement` schema
- `requirement_values` on checklist item state updates
- notes on exclusivity with message triggers and that clients unable to collect values (e.g. mobile until supported) get 400 on close

webapp/src/components/backstage/beta_features_setting.tsx
asaadmahmood: Agreed this would be nicer as a shared Mattermost admin control eventually. For now we kept the beta accordion in Playbooks so we can ship the feature behind a plugin setting without blocking on a core change. Happy to revisit once there is a common plugin beta-features pattern in the main repo.

asaadmahmood:
@Willyfrog On the mobile gap: agreed it’s a real limitation while mobile can’t collect requirement values. I documented it in `server/api/api.yaml` on the checklist item state endpoint (clients that can’t supply `requirement_values` get 400 when closing items that have requirements and the beta flag is on). Happy to link that from a follow-up ticket / mobile work item if you want.

Willyfrog:
> @Willyfrog On the mobile gap: agreed it’s a real limitation while mobile can’t collect requirement values. I documented it in `server/api/api.yaml` on the checklist item state endpoint (clients that can’t supply `requirement_values` get 400 when closing items that have requirements and the beta flag is on). Happy to link that from a follow-up ticket / mobile work item if you want.

yeah, is mostly a "don't forget about that". 

The yaml is good, but it won't be looked at by most people turning the feature on. I was thinking more in the product docs

asaadmahmood:
@Willyfrog Created mobile PR too:
https://github.com/mattermost/mattermost-mobile/pull/10035

server/api/graphql_root_playbook.go
```diff
@@ -557,3 +560,22 @@ func validateUpdateTaskActions(checklists []UpdateChecklist) error {
 	}
 	return nil
 }
+
+func validateUpdateRequirementsExclusiveOfTaskActions(checklists []UpdateChecklist) error {
```
JulienTant, line 567: Should we also call this from `AddChecklistItem` and `AddChecklist`? Both potentially decode a Checklist (or checklistitem), we might want to gate them too.

Willyfrog (approved):

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -478,12 +495,17 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
             return null;
         }
 
+        const taskActionsDisabledReason = isPlaybookEditor && betaFeaturesEnabled && requirements.length > 0 ?
```
JulienTant, line 498: Should this apply to run as well?

JulienTant (approved):
Approved, just left one questions because it seems like we're not applying the same rules for runs?!

lindy65:
Hey @asaadmahmood - thanks - nice feature 😄 

🟥 Please add E2E tests

📄 Is documentation required?


1. Checking - should there be validations in the attached fields? I can type in random text for URLs and they are saved and the task is checked off as completed. Same for assignee.

image

2. I'm confused about this flow (although it is possible in playbooks right now - it doesn't make sense to me where all requirements have been completed). Perhaps there should just be a check like "Are you sure you want to uncheck this task?":
    - a task that has requirements that have been filled in and the task has been marked complete
    - I'm able to uncheck the task as having been completed - even although all requirements have been filled in
      - should this be allowed? 
      - why uncheck a completed task (unless you're removing the requirements which isn't possible from a run)

image

3. (Possible bug) Go to Playbooks backstage and click on "Tasks" (ensure you have a task with requirements assigned to you)

image

    - Tick off the task assigned to you and complete requirements

image

    - Task disappears because it's complete so click on "Filters" and select "Show checked tasks"

image


    - Task doesn't reappear even although you can see "2 assigned" at the top

image

    - On refresh, task reappears but is now **unchecked** and requirements are blank

image

4. Should the post made in channel - when the task can't be checked off via slash command - be more user friendly?

image

image

5. It doesn't seem like there is a character cap on a requirements field. Should there be?

image

6. Requirements modal breaks at 9 requirements. Separate question - what is the cap on requirements?

image

7. Complete requirements modal also breaks at 9 requirements

image

8. Should there be a recent / timeline activity trail for who filled in requirements?

image

asaadmahmood:
Thanks for the feedback @lindy65 

1. No validation, when we allow users to add different types of task requirements, like assignee and URL, or date, then we will add them.

2. Yes, you should be able to uncheck it, because maybe there's more requirements to be added, or it was inaccurately checked.

3. Fixed — completing from the Tasks inbox now sends `requirement_values`, and we only animate the row away after a successful save (so “Show checked tasks” / refresh no longer leaves it blank/unchecked).

4. Updated — `/playbook check` now opens an interactive dialog to fill requirements when there are ≤5 fields; otherwise it posts a clearer message with a link to the run.

5. Definitely, Fixed — labels capped at 128 chars, values at 1024 (enforced in UI + server). Max 20 requirements per task.

6. Fixed — edit requirements modal scrolls instead of overflowing.

7. Fixed — complete requirements modal scrolls instead of overflowing.

8. Fixed — saving requirement values (without a state change) now adds a timeline entry for who updated them.

Also noted on E2E/docs — will follow up separately.

asaadmahmood:
@lindy65 Also added Cypress e2e coverage in `e2e-tests/cypress/tests/integration/playbooks/playbooks/edit/task_requirements_spec.js`:

- Add a requirement from the task More menu and assert it persists
- Check off a run task → fill modal → save and mark complete
- Save requirement values as a draft without completing the task

lindy65:
Thanks @asaadmahmood 

Got it for 1. and 2. 👍 

All others retested and all fixed except for 6. and 7. which don't look fixed to me. My screen resolution is the recommended for my laptop (Lenovo ThinkPad) - 1920 x 1200

image

image

asaadmahmood:
Thanks @lindy65 — you're right on 6 and 7. The earlier fix lost to `GenericModal`'s `overflow: visible` styles.

Pushed a proper fix: the requirements field list itself scrolls (and the modal height is capped), so both the Add/Edit and Complete modals should stay within the viewport at 1920×1200 even with many requirements. Header/footer and the “Add another requirement” action stay visible while the fields scroll.

Please retest when SpinWick picks up the new artifact.

lindy65:
Hey @asaadmahmood - the spinwick did not update automatically and I'm getting an error trying to download the artifact [from here](https://github.com/mattermost/mattermost-plugin-playbooks/pull/2381#issuecomment-5374093733). Could this be because there are failing checks?

asaadmahmood:
@lindy65 Check now.

lindy65 (approved):
Thanks @asaadmahmood - everything looks good now 👍

PR #2386, opened 2026-07-29
Allow locking tasks so only the assignee can complete them

asaadmahmood:
## Summary
- Adds an `assignee_only_complete` setting on checklist items, with a lock toggle in the assignee dropdown (run details and channel RHS)
- When locked and assigned, only the assignee can check/uncheck the task; anyone can still change or clear the assignee
- Enforces the restriction on the server API and styles the assignee search input to use theme-aware Mattermost input colors

## Test plan
- [ ] Open a run checklist and assign a task to another user
- [ ] In the assignee dropdown, enable **Only the assignee can complete the task**
- [ ] Confirm a non-assignee (including run owner) sees a disabled complete checkbox and cannot toggle it via API
- [ ] Confirm the assignee can still complete the task
- [ ] Clear the assignee and confirm anyone can complete again
- [ ] Confirm the lock setting persists on playbook templates and is inherited by new runs
- [ ] In dark mode, confirm the assignee search input uses a dark background with a blue focus ring (not white)
- [ ] Confirm playbook editor pages load without GraphQL errors

Made with [Cursor](https://cursor.com)


JulienTant (approved):

server/app/playbook_run_service.go
```diff
@@ -2307,6 +2307,10 @@ func (s *PlaybookRunServiceImpl) ModifyCheckedState(playbookRunID, userID, newSt
 		return nil
 	}
 
+	if err := checkAssigneeOnlyComplete(itemToCheck, userID); err != nil {
```
calebroseland: (placeholder)
> **MUST_FIX — authz: `SkipChecklistItem` bypasses this guard**
>
> This guard runs on the `/state` path only. `SkipChecklistItem` (`playbook_run_service.go:3285`,
> unchanged by this PR) sets `State = ChecklistItemStateSkipped` with no `checkAssigneeOnlyComplete`
> call — and skipped items count as completed in `ComputeTaskProgress` (`playbook_run.go:271`).
>
> So a non-assignee can mark a locked task resolved, or wipe a locked task the assignee already
> closed. This is not an obscure API path: it is one click in the hover menu
> (`webapp/src/components/checklist_item/hover_menu.tsx:202`).
>
> That the same logical operation *is* guarded here makes the omission look unintentional rather
> than a deliberate carve-out.
>
> **Fix:** call `checkAssigneeOnlyComplete` after `checklistItemParamsVerify` in `SkipChecklistItem`,
> and map `ErrAssigneeOnlyComplete` to 403 in `itemSkip`.

calebroseland: (placeholder)
> **MUST_FIX — authz: `RestoreChecklistItem` bypasses this guard**
>
> Same root cause, opposite direction. `RestoreChecklistItem` (`playbook_run_service.go:3312`,
> unchanged by this PR) sets `State = ChecklistItemStateOpen` unguarded — effectively unchecking a
> locked, closed item — reachable via `PUT /runs/{id}/checklists/{n}/item/{m}/restore`
> (`api/playbook_runs.go:127`).
>
> Worth noting it also emits no `TimelineEvent`, so the bypass leaves less of an audit trail than
> the guarded path it circumvents.
>
> **Fix:** same guard, plus 403 mapping in `itemRestore`.

server/app/playbook_run_service.go
```diff
@@ -2405,6 +2409,60 @@ func (s *PlaybookRunServiceImpl) ToggleCheckedState(playbookRunID, userID string
 	return s.ModifyCheckedState(playbookRunID, userID, newState, checklistNumber, itemNumber)
 }
 
+// checkAssigneeOnlyComplete returns ErrAssigneeOnlyComplete when the item is locked to its
+// assignee and the acting user is not that assignee. Unassigned locked items are not restricted.
+func checkAssigneeOnlyComplete(item ChecklistItem, userID string) error {
```
calebroseland: (placeholder)
> **MUST_FIX — authz: bulk checklist endpoints clear every lock at once**
>
> This helper currently has one caller. `SkipChecklist` (`:3228`) and `RestoreChecklist` (`:3257`)
> — routes at `api/playbook_runs.go:119-120`, both unchanged by this PR — loop over every item in a
> checklist and force `State` with no per-item check.
>
> One request therefore overrides every lock in the checklist. A per-item control that a sibling
> bulk endpoint overwrites wholesale is not really a control.
>
> **Fix:** guard per item in both loops. Please make the failure mode an explicit decision rather
> than an accident, and state which you chose in the PR description:
> - **fail-closed** — reject the whole request if any item is locked to someone else, or
> - **partial-apply** — skip/restore everything except locked items, and report what was skipped.

webapp/src/components/checklist_item/checklist_item.tsx
```diff
@@ -620,7 +705,12 @@ export const ChecklistItem = (props: ChecklistItemProps): React.ReactElement =>
                     />
                     <CheckBoxButton
                         readOnly={props.readOnly}
-                        disabled={isSkipped() || props.playbookRunId === undefined || props.newItem}
+                        disabled={
+                            isSkipped() ||
+                            props.playbookRunId === undefined ||
+                            props.newItem ||
+                            (assigneeOnlyComplete && Boolean(assigneeID) && assigneeID !== currentUserId)
```
calebroseland: (placeholder)
> **MUST_FIX — a11y: the disabled checkbox is unexplained and unreachable (WCAG 2.1 AA 1.3.1, 3.3.2)**
>
> This new condition renders the checkbox `disabled` with no explanation anywhere in the row.
> `CheckBoxButton` (`inputs.tsx:48-56`) emits a bare `<input disabled>` — no `title`, no
> `aria-describedby`, no tooltip — and there is no lock glyph on the row.
>
> `disabled` also removes the control from the tab order, so keyboard and screen-reader users get a
> silently unreachable, unannounced control. The only affordance explaining *why* lives inside the
> assignee dropdown — a different surface the user has no reason to open.
>
> A sighted mouse user sees a greyed checkbox and guesses. A screen-reader user gets nothing at all.
>
> **Fix:**
> - use `aria-disabled` + swallow the change in `handleOnChange`, instead of `disabled`, so the
>   control stays focusable and announceable;
> - wrap in `WithTooltip` with a new string (e.g. *"Only {assignee} can complete this task"*) and
>   point `aria-describedby` at it;
> - render a lock glyph on the row when the item is locked and assigned.

server/api/playbook_runs.go
```diff
@@ -1477,6 +1478,10 @@ func (h *PlaybookRunHandler) itemSetState(c *Context, w http.ResponseWriter, r *
 	}
 
 	if err := h.playbookRunService.ModifyCheckedState(id, userID, params.NewState, checklistNum, itemNum); err != nil {
+		if errors.Is(err, app.ErrAssigneeOnlyComplete) {
```
calebroseland: (placeholder)
> **MUST_FIX — error handling: this 403 is silently swallowed by the client**
>
> This new 403 lands in a pre-existing swallow. `inputs.tsx:43-46` (not touched by this PR) does:
>
> ```ts
> const res = await props.onChange(newValue);
> if (res?.error) {
>     setIsChecked(isChecked);   // revert, no toast, no console
> }
> ```
>
> The optimistic tick appears and then silently vanishes, with no explanation.
>
> This is reachable in normal use, not just via a crafted request: the client-side gate reads local
> state that lags the server, so another user can lock or reassign the item between the WebSocket
> push and the click. The user then sees a checkbox that ticks and untick itself for no visible
> reason — the worst possible presentation of a permission denial.
>
> **Fix:** surface it — `toaster.add({toastStyle: ToastStyle.Failure})`, special-casing the 403 with
> the assignee-lock message rather than a generic failure string.

calebroseland (approved):
@asaadmahmood - I'm sorry for the delay.  Approving, but I see this as more of a "please don't" kind of guard  -- from that perspective, it works at the UI level, but there's potentially some gaps worth addressing, more details in the attachment.

One I wanted to highlight:
- The disabled checkbox of a assignee-only guarded task doesn't have any label indicating why it's disabled.

[pr-2386-blockers.md](https://github.com/user-attachments/files/31531145/pr-2386-blockers.md)

lindy65:
Hey @asaadmahmood 

1) If anyone who is a participant of a playbook can uncheck `Only the assignee can complete the task` what is the purpose of this feature?

2. I find this confusing:
    a) I assigned a task to `user-1` and ticked `Only the assignee can complete the task`
    b) If I click on "Assignee" icon on that task again I see the 2nd screenshot and, even although `user-1` is actually assigned to this task, this is what's confusing:
        - There a bold "No assignee" at the top
        - 2 users are highlighted `Run owner` and `user-1`
     c) Also confusing - if I click "No assignee", `user-1` is unassigned (the "No assignee" looked to me like there was no assignee - not that it removes the assignee)

image


image

3) There is no tooltip on the checkbox of a task when it's assigned to someone else (other than me) and `Only the assignee can complete the task` is checked. (As another user I might *think* I could complete the task and try to click but am blocked and I don't know why)

image

4) This flow doesn't make sense to me:
    - Task is assigned to `user-1` and `Only the assignee can complete the task` is checked/locked
    - `user-1` checks off the task as complete
    Observed: any other playbook participant can go and change assignee after the task has been completed (this may be how it works currently on playbooks but... for this particular feature where we're locking completion of a task to an assignee, it doesn't make sense that the assignee can be changed after the task is checked off as complete)

asaadmahmood:
@lindy65 Thanks for the review — we addressed the main behaviour gaps from your feedback. Could you take another pass when you have a moment?

### Behaviour changes
1. **Lock purpose / who can unlock or reassign** — When *Only the assignee can complete the task* is on, only the **assignee** or **run owner** can change the assignee or turn the lock off. Other participants are blocked in the UI and on the server (including after the task is completed).
2. **Disabled checkbox tooltip** — Non-assignees hovering a locked task’s checkbox now see: *Only the assignee can complete the task*.
3. **Clear assignee label** — Renamed **No Assignee** → **Remove assignee**, and styled it as a semibold link so it reads as an action rather than a status.

### Still noting from your review
- The dual highlight of **Run Owner** + a specific user in the assignee dropdown looks like a separate selection-state UX issue; we didn’t change that in this pass.
- Run owner / assignee can still reassign a locked completed task (others cannot). Happy to tighten that further if you want assignee frozen after completion.

Please re-test and let us know if anything still feels off. Thanks!

asaadmahmood:
@coderabbitai review

lindy65:
Hey @asaadmahmood - I'll be re-testing shortly but.... is there a Jira ticket for this PR?

asaadmahmood:
@lindy65 Yup, here you go:
https://mattermost.atlassian.net/browse/MM-69626

lindy65:
Thanks @asaadmahmood :)

1. This flow is a little broken... Perhaps we can put some kind of warning on the screen when a user clicks "Only the assignee can complete the task" that they need to add an assignee first (if one is not yet added)?
    - Click "Only the assignee can complete the task" on any task in a playbook RHS that does not have an assignee yet
    - Do not assign the task to anyone
    - As a playbook participant (not run owner), hover over assignee icon on that task
      Observed: Tooltip - "Only the assignee or run owner can change the assignee" (but the task is not yet assigned)
    - As a playbook participant (not run owner), check off that task as completed
      Observed: Able to check off the task as completed even although "Only the assignee can complete the task" is enabled

image


image


2. Another flow that perhaps needs some thought?:
    - as user 'lindy' (run participant) 
    - assign a task to 'user-1' (run participant)
    - as user 'lindy' check the box "Only the assignee can complete the task"
    - user 'lindy' realises she checked the box in error but now she cannot uncheck the box as only the assignee or run owner can
      - Does user 'lindy' just ask in the run channel for run owner to uncheck the "Only the assignee can complete the task" box?


3. There is no tooltip ("Only the assignee or run owner can change the assignee") on a task assignee's name where "Only the assignee can complete the task" box has been checked if I access the task as a run participant from the edit icon of that task in the RHS

image

asaadmahmood:
@lindy65 I will update 1 to only work when you select the assignee.
2. Yes she probably should just ask the run owner. It would be too cumbersome to track or even allow people to revert their decision, would go back to what I was doing previously and keep it open.
3. I'll add that tooltip there too.

lindy65:
> @lindy65 I will update 1 to only work when you select the assignee. 2. Yes she probably should just ask the run owner. It would be too cumbersome to track or even allow people to revert their decision, would go back to what I was doing previously and keep it open. 3. I'll add that tooltip there too.

OK, thanks @asaadmahmood - let me know when ready to re-test

asaadmahmood:
@lindy65 Should be good to retest.

lindy65:
Thanks @asaadmahmood 

1. Something weird is happening when I now try to click in the box for "Only the assignee can complete the task" BEFORE I select an assignee. If I keep clicking on the box (still without having selected an assignee) it eventually ticks the box. It's intermittent - sometimes I only click a few times until the box is ticked. At others I have to click for quite a while until the box is ticked.
  - It also takes quite a few ticks to un-tick the box once it's ticked. Sometimes it doesn't un-tick and I have to exit the modal and go back into assignee modal to see the box un-ticked

image


2. Following on from 1) I think there should be a tooltip on the box of "Only the assignee can complete the task" that instructs the user to assign the task before clicking the box.

3. Actually, I'm going to stop testing because I'd like to clarify first how exactly you implemented 1) in your comment [here](https://github.com/mattermost/mattermost-plugin-playbooks/pull/2386#issuecomment-5490334452). I was thinking that the box would be un-tickable UNTIL an assignee is chosen but that doesn't seem to be what cursor did? I think the box should be un-tickable with a tooltip saying "choose an assignee first". The way it's working now is it's only sometimes clickable (see my first comment) and there's no tooltip saying why for the user

image

asaadmahmood:
@lindy65 
Sent a video to you in the DMs.


https://github.com/user-attachments/assets/3d28efe9-cfa2-4910-9ee7-9a3316b8a8f8

lindy65:
> @lindy65 Sent a video to you in the DMs.
> 
>  CleanShot.2026-09-02.at.4.12.09.PM.2.mp4

OK, thanks @asaadmahmood - I get where you're coming from. I'll retest with your intended fix in mind

lindy65 (approved):
Thanks @asaadmahmood - retested after your fixes and video explanation and everything looks good.

PR #2388, opened 2026-07-30
MM-70041: Enforce membership permission checks consistently on playbook creation

edgarbellot:
## Ticket Link

Fixes [MM-70044](https://mattermost.atlassian.net/browse/MM-70044).

```release-note
Fixed an issue where playbook membership permission checks were not applied consistently between creating and updating a playbook.
```

[MM-70044]: https://mattermost.atlassian.net/browse/MM-70044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e-tests/cypress/tests/integration/playbooks/channels/broadcast_spec.js
+++ b/e2e-tests/cypress/tests/integration/playbooks/channels/broadcast_spec.js
@@ -111,7 +111,6 @@ describe('channels > broadcast', {testIsolation: true}, () => {
                                     userId: testUser.id,
                                     broadcastChannelIds: [testPublicChannel1.id, testPrivateChannel1.id],
                                     broadcastEnabled: true,
-                                    otherMembers: [testAdmin.id],
                                     invitedUserIds: [testAdmin.id],
                                 }).then((playbook) => {
                                     rootDeletePlaybook = playbook;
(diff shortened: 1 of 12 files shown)


edgarbellot:
Thanks for the review, @coderabbitai.

Regarding the two outside-diff suggestions on `noAddedMembersWithoutPermission` in `server/app/permissions_service.go`:

1. `reflect.DeepEqual` treating `nil` vs empty `[]PlaybookMember{}` as different.
2. Role comparison via `strings.Join` being order-sensitive.

Both behaviors predate this PR — `noAddedMembersWithoutPermission` is a straight extraction of the existing comparison logic that previously lived inline in `PlaybookModifyWithFixes`, with no change to how it compares members or roles. Since these are pre-existing characteristics of the modification path and not something introduced or altered by this change, they're out of scope here. Suggest tracking them in a separate follow-up ticket/PR instead.

JulienTant (changes requested):
@edgarbellot if I'm not mistaken, this fix can be bypassed (and the e2e helper added already implements the bypass as its workaround!)

1. `POST /playbooks` with no members, the server grants you `playbook_admin`
2. `PUT /playbooks/{id}` with the members array you wanted. This is allowed because `playbook_admin` still holds manage_members/manage_roles by default. Roles are "merged", so stripping them from `playbook_member` doesn't help...

edgarbellot:
@coderabbitai review

edgarbellot:
@JulienTant added a commit addressing your feedback, great catch! thank you

JulienTant (approved):
LGTM

PR #2389, opened 2026-07-31
[MM-70073] Widen task-assignee auto-add-participant check to any authorized actor

JulienTant:
## Summary

A user reported that assigning a task to someone who isn't yet a run participant doesn't actually add them as a participant, grant them channel access, or notify them — the assignment silently "succeeds" with none of those side effects.

The root cause: `addAssigneeParticipantAndDM` (the helper responsible for this auto-add) only triggers if the actor is the run owner or a system admin. But every entry point that reaches this helper (`SetAssignee`, `SetRoleAssignee`, `SetPropertyUserAssignee`, `SetRunPropertyValue`) is already gated by `RunManageProperties`, which is far more permissive — it allows any existing run participant, not just the owner. Worse, the explicit "invite a participant" GraphQL mutation (`AddRunParticipants`) already lets any participant add an arbitrary new participant with the identical outcome. So the owner/admin-only restriction wasn't preventing any actual privilege escalation — it was just a second, narrower, inconsistent gate that produced a silent partial failure whenever a non-owner participant tried to assign a task to someone new.

This PR:
- Widens the actor check inside `addAssigneeParticipantAndDM` to delegate to `RunManageProperties`, matching the permission model already enforced everywhere else (no signature change — the helper already receives the run ID).
- Adds a warning log on the (now effectively unreachable via HTTP) deny path instead of a silent no-op, in case a future internal caller bypasses the HTTP layer.
- Fixes a related, pre-existing bug this change would otherwise expose more often: the assignment DM was sent even when `AddParticipants` silently dropped a target for failing team/DM-channel membership validation, leaking the run name and URL to someone who was never actually added. The DM is now gated on the target actually landing in the participant list — only for that specific case; the owner-assignment and already-a-participant paths (which never call `AddParticipants`) still send the DM exactly as before.
- Inverts a stale test that asserted the old (buggy) restriction, and adds a new integration test covering the fixed behavior across `SetAssignee`, `SetRoleAssignee`'s creator-role branch, the membership-validation-failure DM guard, and a regression check that the DM still fires for owner/already-participant assignment.
- Adds a Cypress e2e test that drives the real UI as a non-owner participant assigning a task to a non-participant, and asserts they show up in the run's Participants panel.

No API endpoint, request/response shape, or OpenAPI change — this only changes an internal authorization check on existing endpoints.

## QA Steps

**Setup**: a playbook run with at least two non-admin users — User A (a run participant, but not the owner) and User B (a team member not yet a run participant).

1. Log in as **User A** (non-owner participant) and open the run.
2. In the run's checklist, assign a task to **User B**.
3. Verify:
   - The task now shows User B as the assignee.
   - User B appears in the run's **Participants** list (RHS Participants panel).
   - User B receives a bot DM notifying them of the assignment, with a link to the run.
   - If the run/playbook has "Add participants to channel automatically" enabled and User A has permission to manage channel members, User B is also added to the run's channel.
4. **Regression** — repeat steps 2-3 as the **run owner** assigning a different non-participant user: behavior should be identical (unaffected by this change).
5. **Edge case** — as User A, assign a task to a valid Mattermost user who is **not a member of the run's team**: the assignee field should update, but that user should **not** be added as a participant and should **not** receive a DM (this closes a separate notification leak).
6. **Regression** — as User A, assign a task to the run owner, or to a user who is already a run participant: the assignment DM should still be sent as before.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-70073

## Checklist
- [x] Unit tests updated
- ~~Planned cherry-picks: `release-X.Y`~~

### DIFF
--- a/e2e-tests/cypress/tests/integration/playbooks/runs/assignee_auto_add_participant_spec.js
+++ b/e2e-tests/cypress/tests/integration/playbooks/runs/assignee_auto_add_participant_spec.js
@@ -77,9 +77,14 @@ describe('runs > assignee auto-add participant (MM-70073)', {testIsolation: true
             // # Visit the run details page as the non-owner participant
             cy.visit(`/playbooks/runs/${playbookRun.id}`);
 
-            // # Assign the task to the non-participant target user via the real UI flow
-            getChecklistTasks().eq(0).findByTestId('hover-menu-edit-button').click();
-            cy.findByTestId('assignee-profile-selector').click();
+            // # Assign the task to the non-participant target user via the real UI flow.
+            // The assignee-profile-selector testid must be scoped to this checklist item:
+            // the RHS Info panel's Owner row renders an identically-testid'd selector that's
+            // on screen by default, so an unscoped findByTestId would match two elements.
+            getChecklistTasks().eq(0).within(() => {
+                cy.findByTestId('hover-menu-edit-button').click();
+                cy.findByTestId('assignee-profile-selector').click();
+            });
             cy.contains('.playbook-react-select__option', '@' + testTarget.username).click();
 
             cy.wait('@setAssignee');
--- a/e2e-tests/cypress/tests/integration/playbooks/runs/assignee_auto_add_participant_spec.js
+++ b/e2e-tests/cypress/tests/integration/playbooks/runs/assignee_auto_add_participant_spec.js
@@ -1,0 +1,99 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// ***************************************************************
+// - [#] indicates a test step (e.g. # Go to a page)
+// - [*] indicates an assertion (e.g. * Check the title)
+// ***************************************************************
+
+// Stage: @prod
+// Group: @playbooks
+
+describe('runs > assignee auto-add participant (MM-70073)', {testIsolation: true}, () => {
+    let testTeam;
+    let testOwner;
+    let testParticipant;
+    let testTarget;
+    let testPlaybook;
+
+    const getChecklistTasks = () => cy.findByTestId('run-checklist-section').findAllByTestId('checkbox-item-container');
+
+    before(() => {
+        cy.apiInitSetup().then(({team, user}) => {
+            testTeam = team;
+            testOwner = user;
+
+            // # Create a non-owner run participant
+            cy.apiCreateUser().then(({user: participant}) => {
+                testParticipant = participant;
+                cy.apiAddUserToTeam(testTeam.id, testParticipant.id);
+            });
+
+            // # Create a team member who will be the assignment target — a team member, but
+            // not yet a run participant
+            cy.apiCreateUser().then(({user: target}) => {
+                testTarget = target;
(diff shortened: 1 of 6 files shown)


catalintomai:
@julientant -looks good, might want to check with DanielS re: requirements, the previous limiting CF-5 behavior was explicitly defined as such (as the comments indicate).  Thanks.

catalintomai (approved):

lindy65 (approved):
Thanks @JulienTant - tested as per test steps and everything's working as expected. 👍

e2e-tests/cypress/tests/integration/playbooks/runs/assignee_auto_add_participant_spec.js
```diff
@@ -0,0 +1,104 @@
...
+            cy.intercept('PUT', '**/api/v0/runs/*/checklists/*/item/*/assignee').as('setAssignee');
+
+            // # Visit the run details page as the non-owner participant
+            cy.visit(`/playbooks/runs/${playbookRun.id}`);
+
+            // # Assign the task to the non-participant target user via the real UI flow.
+            // The assignee-profile-selector testid must be scoped to this checklist item:
+            // the RHS Info panel's Owner row renders an identically-testid'd selector that's
+            // on screen by default, so an unscoped findByTestId would match two elements.
+            getChecklistTasks().eq(0).within(() => {
```
jgheithcock, line 85: Suggest calling `getChecklistTasks().eq(0).trigger('mouseover');` before the `.within(...` block to invoke the `.hover()` so that the opacity is not zero.

server/app/playbook_run_service.go
```diff
@@ -5842,23 +5842,44 @@ func (s *PlaybookRunServiceImpl) addAssigneeParticipantAndDM(playbookRunID, acto
...
+					logrus.WithError(err).WithFields(logFields).Warn("failed to determine actor's permission to add assignee as run participant")
 				}
-			} else {
-				// Actor lacks permission to add resolvedUserID; skip DM to avoid leaking run details.
+				return
+			}
+			addedUserIDs, err := s.AddParticipants(playbookRunID, []string{resolvedUserID}, actorUserID, false, false)
+			if err != nil {
+				logrus.WithError(err).WithFields(logFields).Warn("failed to add assignee as participant")
+				return
```
jgheithcock: AI says that `AddParticipants` can return non-error but with a non-empty list, so delete the return and fall through to the `!slices.Contains()` code.

jgheithcock (changes requested):
Two little issues

jgheithcock (approved):
LGTM! Thanks!

DSchalla:
Thanks Julien for the brief conversation, makes sense to me - Wondering if it's worth to callout that it allows members to add someone to a channel even if the team/system permission schema blocks that otherwise via assigning a task? 

The behavior makes sense, just good to document and clarify.

JulienTant:
@DSchalla - it is not the case tho, we do a permission check before inviting users to a channel on behalf of the playbooks member. The member can invite someone to the playbook but not to the channel - those are separate capabilities. Unless I missed a path? 😅

DSchalla:
Sorry, I've misread the PR description, my bad. :-) QA Steps make it very clear now, thanks!

PR #2391, opened 2026-08-05
MM-70119: fix blank admin DM for checklist due-date trial nudge

JulienTant:
## Summary

When a non-admin user without the required license tries to set a due date on a checklist item, Playbooks sends a DM to all System Admins nudging them to start a trial. That DM rendered completely blank — no title, no body text, just a divider line and a "Start 30-day trial" button — because `NotifyAdmins` never had a `case` for the `start_trial_to_set_checklist_item_due_date` message type (defined in `webapp/src/constants.ts` and used in `webapp/src/components/checklist_item/duedate.tsx`). `title` and `text` stayed as empty strings when the switch fell through.

This was the only one of the 9 `AdminNotificationType` values missing its corresponding backend case, and it's been broken since the checklist due-date feature was introduced in 2022 (PR #1115).

This PR adds the missing case, using the same copy already present in the frontend modal (`upgrade_modal_data.tsx`) for consistency, plus a table-driven regression test asserting every `AdminNotificationType` produces a non-blank DM — so a future addition to the frontend enum that forgets the matching backend case fails a unit test instead of shipping silently.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-70119

## QA Test Steps

Precondition: server has no active Professional/Enterprise trial or license (Team Edition, or a licensed edition with the trial not started/expired), and cloud licensing is off.

1. Log in as a non-admin user who is a member of a playbook run's channel.
2. Open the run's checklist (RHS) and try to set a due date on any checklist item (click the due-date icon/button on a task).
3. Confirm the upgrade nudge modal that appears (title "Work more effectively" / text about assigning due dates) by clicking its confirm button as a non-admin.
4. Log in as a System Admin (or open a second session).
5. Open the DM from the "Playbooks" bot to the admin account.

**Expected:** the DM shows a title ("Work more effectively"), body text ("Assign due dates to tasks so assignees can prioritize and get things done."), a learn-more/upgrade footer, and the "Start 30-day trial" button.

**Before the fix:** the DM showed only a blank divider line and the "Start 30-day trial" button, with no title or text.

**Regression check:** repeat steps 1-5 for a couple of the other upgrade-nudge entry points (e.g. RHS timeline "add to timeline", playbook run channel export, key metrics) to confirm those DMs still render their title/text correctly.

## Checklist
- [x] Unit tests updated
- [ ] Planned cherry-picks: `release-2.11`, `release-2.10`, `release-2.9`, `release-2.4`

### DIFF
--- a/server/bot/poster.go
+++ b/server/bot/poster.go
@@ -236,6 +236,10 @@ func (b *Bot) NotifyAdmins(messageType, authorUserID string, isTeamEdition bool)
 		message = fmt.Sprintf("@%s requested access to ask for status updates in playbook runs", author.Username)
 		title = "Try request update with a free trial"
 		text = "Request updates for playbook runs in a single click and get notified directly when an update is posted. Start a free, 30-day trial to try it out.\n" + footer
+	case "start_trial_to_set_checklist_item_due_date":
+		message = fmt.Sprintf("@%s requested access to set due dates on checklist items.", author.Username)
+		title = "Work more effectively"
+		text = "Assign due dates to tasks so assignees can prioritize and get things done.\n" + footer
 	}
 
 	actions := []*model.PostAction{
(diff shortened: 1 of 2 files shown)


lindy65:
Hey @JulienTant - I'm not seeing the fixed DM message to the system admin - for me it is still broken.

I have set up a team edition `/cloud` server then uploaded playbooks with this PR's tar.gz (I had to add a Professional licence to install playbooks but then removed the licence so it's back to a team edition now)

image


Let me know if you need credentials to my `/cloud` server to take a look

lindy65:
Hi @JulienTant - did we determine whether this is actually a bug or not?

PR #2398, opened 2026-08-17
[MM-70231] record skip/restore task state changes

vish9812:
## Summary

Skipping or restoring a task left `StateModified` untouched and created no timeline event, so clients could not tell when a task was skipped or by whom — they fell back to the timestamp of an earlier check/uncheck.

**Server**

- Set `StateModified` on both the skip and restore paths, and emit a `task_state_modified` timeline event carrying `action`, `task`, and `item_id`.
- Pass a `nil` current run to `sendPlaybookRunObjectUpdatedWS` so the broadcast is rebuilt from the store and ships the newly created event, matching what `ModifyCheckedState` already did.
- Extract the event construction shared by all four transitions into `createTaskStateModifiedEvent`, name the actions as constants, and export the payload as `TaskStateModifiedDetails`.
- Add audit records to `SkipChecklistItem` and `RestoreChecklistItem`, which previously had none, plus idempotency guards so a repeated call cannot write a second event or move the timestamp forward.
- `ModifyCheckedState` now sets `LastSkipped` when skipping, making it a strict superset of `SkipChecklistItem`. `LastSkipped` is deliberately left alone on restore: it records when a task was last skipped, not whether it still is.

**Webapp**

- The hover menu's skip/restore item now goes through `onChange`, so a click is a single write to the item state endpoint. Removed `clientSkipChecklistItem` and `clientRestoreChecklistItem`.

This leaves one code path for a task state change instead of two with different side effects. The dedicated `/skip` and `/restore` endpoints remain for API consumers and now behave consistently with `/state`.

## Ticket Link

Server/Webapp: https://mattermost.atlassian.net/browse/MM-70231
Mobile: https://mattermost.atlassian.net/browse/MM-69381

## Checklist

- [x] Unit tests updated — new `server/app/skip_restore_checklist_item_test.go` covers both endpoints, the no-op guards, timeline event contents and timestamps, the WS broadcast, and `ModifyCheckedState`'s `LastSkipped` handling for skip/restore/check
- [ ] ~~Planned cherry-picks: `release-X.Y`~~

### DIFF
--- a/server/app/playbook_run_service.go
+++ b/server/app/playbook_run_service.go
@@ -3303,8 +3303,6 @@ func (s *PlaybookRunServiceImpl) SkipChecklistItem(playbookRunID, userID string,
 	model.AddEventParameterToAuditRec(auditRec, "taskTitle", itemToSkip.Title)
 	model.AddEventParameterToAuditRec(auditRec, "currentState", itemToSkip.State)
 
-	// Without this guard a repeated skip writes a second timeline event and moves the timestamp
-	// forward, which clients render as a task that was skipped twice.
 	if itemToSkip.State == ChecklistItemStateSkipped {
 		auditRec.Success()
 		return nil
@@ -3398,8 +3396,7 @@ func (s *PlaybookRunServiceImpl) RestoreChecklistItem(playbookRunID, userID stri
 	model.AddEventParameterToAuditRec(auditRec, "taskTitle", itemToRestore.Title)
 	model.AddEventParameterToAuditRec(auditRec, "currentState", itemToRestore.State)
 
-	// A repeat call would otherwise duplicate the timeline event.
-	if itemToRestore.State == ChecklistItemStateOpen {
+	if itemToRestore.State != ChecklistItemStateSkipped {
 		auditRec.Success()
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


server/app/playbook_run_service.go
```diff
@@ -3283,52 +3283,155 @@ func (s *PlaybookRunServiceImpl) RestoreChecklist(playbookRunID, userID string,
...
 	}
 
+	itemToRestore := &playbookRunToModify.Checklists[checklistNumber].Items[itemNumber]
+
+	// Add current context to audit
+	model.AddEventParameterToAuditRec(auditRec, "taskTitle", itemToRestore.Title)
+	model.AddEventParameterToAuditRec(auditRec, "currentState", itemToRestore.State)
+
+	// A repeat call would otherwise duplicate the timeline event.
+	if itemToRestore.State == ChecklistItemStateOpen {
```
calebroseland: Should the guard be `itemToRestore.State != ChecklistItemStateSkipped` instead?

> RestoreChecklistItem guards on `State == Open`, so a checked item falls through — unchecked, but recorded as `action: "restore"`. ModifyCheckedState calls that same Closed→Open transition `uncheck`, so the two paths still disagree on the transition this PR unifies. `!= Skipped` fixes it. The unchecking isn't new (old code had no guard); the wrong event label is. Webapp can't hit it after your change, but `/restore` is still registered.

calebroseland (approved):
Thanks for straightening this out, @vish9812! Approving, but might be worth looking at the one finding.

vish9812:
Nice catch @calebroseland. Fixed now.

vish9812:
@lindy65 this is the server-side fix that's needed for the related mobile PR to show the "skipped/restored" correctly. There is no new UI feature in this, so you can mainly test that there are no regressions on the web-side for this PR.

lindy65 (approved):
Thanks @vish9812 - regression-tested the front-end side and didn't find any issues 👍

vish9812:
Thanks @lindy65 
Once it's merged, I'll update and test the mobile ticket once more and will let you know.

lindy65:
Sounds good, thanks @vish9812

PR #2399, opened 2026-08-18
MM-69943: Require channel post permission to run checklist slash commands

jgheithcock:
## Summary

Executing a checklist slash command was authorized only by run-level permissions, which did not
account for whether the caller can post in the run's channel. `RunChecklistItemSlashCommand` now
requires `PermissionCreatePost` on the run's channel, checked before the checklist item is
resolved so the response does not vary with the item's contents.

The check is applied in the app service rather than the API handler, so it covers every caller of
this operation. It is a no-op for channel checklists and DM/GM runs, where `RunManageProperties`
already requires posting permission on the channel; it only tightens playbook-based runs.
`itemRun` maps `ErrNoPermissions` to a 403, matching the `checkEditPermissions` middleware and the
other handlers in this file that surface this sentinel from a service call.

## Ticket Link

Fixes [MM-69943](https://mattermost.atlassian.net/browse/MM-69943).

## Checklist

- [x] Unit tests updated
- [ ] ~~Planned cherry-picks~~ handled by backend automation

[MM-69943]: https://mattermost.atlassian.net/browse/MM-69943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/api/playbook_runs.go
+++ b/server/api/playbook_runs.go
@@ -1653,7 +1653,11 @@ func (h *PlaybookRunHandler) itemRun(c *Context, w http.ResponseWriter, r *http.
 
 	triggerID, err := h.playbookRunService.RunChecklistItemSlashCommand(playbookRunID, userID, checklistNum, itemNum)
 	if err != nil {
-		h.HandleError(w, c.logger, err)
+		if errors.Is(err, app.ErrNoPermissions) {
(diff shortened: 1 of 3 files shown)


JulienTant (approved):

hmohammed-prodsec (approved):
LGTM 👍

PR #2402, opened 2026-08-25
MM-70245: Bound the records a GraphQL request may hydrate

jgheithcock:
## Summary
The GraphQL `runs` query defaulted to a 10,000 record page when a client omitted
pagination, and nothing bounded how many records a single request could hydrate
in total across all of the fields it resolves. Every run is hydrated with its
full checklists regardless of which fields the client selected, so a request can
cost considerably more memory than its size suggests.

This bounds a single run listing at 1,000 records — the largest page any current
client asks for — and gives each request a shared budget of 5,000 records across
the run and playbook resolvers. Page sizes above the bound are clamped rather
than rejected, so existing clients keep working, and the budget is charged by the
records actually fetched, so a request that asks for a large page but matches few
records is unaffected.

**Worst case after this change:** the budget is checked before a listing is
fetched and charged after it returns, so up to `MaxParallelism` (5) listings can
be in flight when the budget is nearly exhausted. The practical ceiling is
therefore roughly 10,000 runs per request (4,999 + 5 × 1,000) rather than a hard
5,000. This is a deliberate trade-off: reserving the full page size up front
would give a hard ceiling, but it would also require capping the `playbooks` page
size, which would truncate results for teams with more than 1,000 playbooks.
`playbooks` keeps its existing page size and is bounded by the shared budget alone.

No client changes are needed: the webapp's largest `runs` request is `first: 1000`,
and the remaining call sites use `first: 1`, `first: 8`, or no pagination.

Added unit tests for the page-size bound and the request budget (including
concurrent field resolution under `-race`), plus integration tests covering both
a query that stays within the budget and one that exceeds it. Regression tests
added; verified they fail when the fix is reverted.

## Ticket Link
https://mattermost.atlassian.net/browse/MM-70245

## Checklist
- [x] Unit tests updated
- [ ] Planned cherry-picks: TBD, no fix versions set on the ticket yet

### DIFF
--- a/server/api/graphql.go
+++ b/server/api/graphql.go
@@ -141,6 +141,7 @@ type GraphQLContext struct {
 	config               config.Service
 	permissions          *app.PermissionsService
 	licenceChecker       app.LicenseChecker
+	recordBudget         *recordBudget
 	favoritesLoader      *dataloader.Loader[favoriteInfo, bool]
 	playbooksLoader      *dataloader.Loader[playbookInfo, *app.Playbook]
 	statusPostsLoader    *dataloader.Loader[string, []app.StatusPost]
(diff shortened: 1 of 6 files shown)


jgheithcock:
@coderabbitai resolve

hmohammed-prodsec (approved):

JulienTant (approved):

PR #2409, opened 2026-09-14
Replace unavailable minio image in CI with mirrored copy

edgarbellot:
## Summary

`minio/minio:RELEASE.2019-10-11T00-38-09Z` was removed from Docker Hub along
with the rest of the `minio/minio` repository, so the `minio` service
container used in the e2e CI jobs (`e2e-cypress-tests`,
`e2e-cypress-tests-fips`, `e2e-playwright-tests`) can no longer be pulled,
failing all three jobs.

This swaps it for `mattermostdevelopment/mirrored-minio:RELEASE.2026-07-17T12-07-51Z`,
the mirrored copy of the Chainguard minio image that the monorepo already
migrated to in mattermost/mattermost#38495.

No behavior change: the minio service container starts without an explicit
command in both the old and new image, so in either case it exits after
printing its help text rather than actually running the server. There is no
healthcheck on this service, so CI was never blocked on it actually being up.
This PR only restores the ability to pull the image; it does not change or
fix that pre-existing behavior.

## Release Note
```release-note
NONE
```


JulienTant:
Thanks, i'll update #2411  when this is merged!

JulienTant (approved):

PR #2410, opened 2026-09-14
[DO NOT MERGE] Diagnostic: revert #2402 to isolate e2e regression

edgarbellot:
Diagnostic only, not for merging.

Includes the minio image fix from #2409 plus a revert of #2402, to check
whether #2402 is the cause of the widespread, deterministic e2e failures
seen in #2409 (lhs-navigation, rhs-runs-list, property-field-row, playbook
editor, etc. all failing identically across every parallel job/retry).

Will be closed once the diagnosis is done.


edgarbellot:
Diagnosis complete: #2402 ruled out as the cause of the e2e failures (same failures reproduce with it reverted). Closing this diagnostic PR.

PR #2418, opened 2026-09-17
MM-70770: Prevent todo digest DM from silently failing for users with many items

JulienTant:
## Summary

The daily/weekly todo digest DM silently fails for users who own a large number of playbook runs or have many assigned tasks. They simply never receive the message, with no indication anything went wrong. The root cause is that the digest has no bound on how many items it includes — on a busy instance the assembled post can exceed Mattermost's post size limit, and when it does the server rejects it quietly.

The fix caps runs sections at 20 items in all cases, and tasks at 20 items for the scheduled digest. The slash command (`/playbook todo`) leaves tasks uncapped because no dedicated paginated task page exists yet (tracked separately). When a runs section is truncated, a "…and N more [View all runs]" footer links to `/playbooks/runs`. A secondary size check on the fully assembled post remains as a last-resort backstop.

| Section | Scheduled digest | `/playbook todo` |
|---|---|---|
| Overdue runs | 20 + link to `/playbooks/runs` | 20 + link to `/playbooks/runs` |
| In-progress runs | 20 + link to `/playbooks/runs` | 20 + link to `/playbooks/runs` |
| Assigned tasks | 20 + "use `/playbook todo`" | uncapped |

## Ticket Link

MM-70770

## Checklist
- [x] Unit tests updated
- [ ] Planned cherry-picks: `release-2.12`


server/app/playbook_run_service.go
```diff
@@ -3980,7 +3988,17 @@ func (s *PlaybookRunServiceImpl) buildTodoDigestMessage(userID string, force boo
 		message += part3
 	}
 
-	return &model.Post{Message: message}, nil
+	T := i18n.GetUserTranslations(user.Locale)
+	footer := "\n" + T("app.user.digest.message_truncated")
+	capped := capDigestMessage(message, footer, model.PostMessageMaxRunesV2)
```
hanzei, line 3995: v12 will contain https://github.com/mattermost/mattermost/pull/37300, making post basically arbitrary long. What is the impact for this PR? I expect a recap to never hit that limit. Does it still make sense from a UX perspective to print the whole digest when force=true?

hanzei (changes requested):
Thanks, Julien! One questing regarding the impact of https://github.com/mattermost/mattermost/pull/37300 on this PR below.

server/app/playbook_run_service.go
JulienTant: Thanks @hanzei I had no clue this PR existed, and here I was thinking that having 16k characters worth of tasks was a lot 😬 

Everything run related has a link to the run list, but one of our issue is we don't have such page for "todos", we do have a some fake RHS in playbooks page but nothing to trigger it through a URL or a dedicated page (which would be nice).

I propose this for now:

- For the 2 runs category: always cap to 20 and show a link to see the run list
- For the task list, if the message is:
  - automatic (force=false) limit to 20 tasks and inform the user they  can use the slash command
  - manual (force=true / slash command) don't set the limit until we have a better solution.

Associated with a ticket to improve UX:

> The daily/weekly digest shows assigned tasks capped at 20 with a footer pointing users to `/playbook todo` to see the rest. The slash command currently shows all tasks uncapped as a stopgap — but a proper paginated page for tasks doesn't exist yet. The runs sections already have this: they cap at 20 and link to /playbooks/runs. Tasks need the same treatment.

WDYT?

hanzei: Yeah, a dedicated "todo" page would be nice. the slash command is a bit clunky.

I'm even fine not filling out the follow-up ticket. I would much rather spend that time on a dedicated page.

hanzei (approved):
Thanks!

hanzei (approved):
Thanks!

